PR #3, opened 2019-01-07
Use pre-commit.js instead of git-pre-commit => removed dependency on yarn

pepakriz:
https://github.com/observing/pre-commit requires more universal syntax for pre-commit definition. So you can use `npm` instead of `yarn` if you want

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,11 +44,13 @@
   },
   "scripts": {
     "build": "tsc",
-    "precommit": "yarn lint",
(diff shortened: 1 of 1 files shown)


B4nan:
TY, I did close #4 with this, because you mentioned it there, but now I am not sure if it really fixes this? Should I keep it opened?

pepakriz:
I opened #4 because this PR is not checked by Travis. Maybe integration is not enabled. I don't see any pipeline.

PR #5, opened 2019-01-08
WIP: BaseEntity: remove magic fields `createdAt` and `updatedAt`

pepakriz:
It can be done in a different way - by hooks

### DIFF
--- a/lib/Utils.ts
+++ b/lib/Utils.ts
@@ -7,7 +7,7 @@ import { getMetadataStorage } from './MikroORM';
 
 export class Utils {
 
-  private static readonly DIFF_IGNORED_KEYS = ['_id', '_initialized', 'createdAt', 'updatedAt'];
+  private static readonly DIFF_IGNORED_KEYS = ['_id', '_initialized'];
 
   static isObject(o: any): boolean {
     return typeof o === 'object' && o !== null;
--- a/tests/Utils.test.ts
+++ b/tests/Utils.test.ts
@@ -65,7 +65,7 @@ describe('Utils', () => {
     expect(Utils.diff({a: 'a'}, {a: 'b', b: ['c']})).toEqual({a: 'b', b: ['c']});
     expect(Utils.diff({a: 'a', b: ['c']}, {b: []})).toEqual({b: []});
     expect(Utils.diff({a: 'a', b: ['c']}, {a: 'b'})).toEqual({a: 'b'});
-    expect(Utils.diff({_id: 'a', createdAt: 1, updatedAt: 1}, {_id: 'b', createdAt: 2, updatedAt: 2})).toEqual({}); // ignored fields
+    expect(Utils.diff({_id: 'a'}, {_id: 'b'})).toEqual({}); // ignored fields
     expect(Utils.diff({a: new Date()}, {a: new Date('2018-01-01')})).toEqual({a: new Date('2018-01-01')});
   });
 
@@ -84,7 +84,7 @@ describe('Utils', () => {
     author2.favouriteBook = book;
     author2.version = 123;
     await orm.em.persist([author1, author2, book]);
-    expect(Utils.diffEntities(author1, author2)).toEqual({ name: 'Name 2', favouriteBook: book._id });
+    expect(Utils.diffEntities(author1, author2)).toEqual({ name: 'Name 2', favouriteBook: book._id, updatedAt: author2.updatedAt });
   });
 
   test('copy', () => {
--- a/tests/Utils.test.ts
+++ b/tests/Utils.test.ts
@@ -79,12 +79,11 @@ describe('Utils', () => {
 
   test('prepareEntity changes entity to string id', async () => {
     const author1 = new Author('Name 1', 'e-mail');
-    const book = new Book('test', author1);
     const author2 = new Author('Name 2', 'e-mail');
-    author2.favouriteBook = book;
-    author2.version = 123;
-    await orm.em.persist([author1, author2, book]);
-    expect(Utils.diffEntities(author1, author2)).toEqual({ name: 'Name 2', favouriteBook: book._id, updatedAt: author2.updatedAt });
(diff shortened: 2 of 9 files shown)


B4nan:
I guess you are right that we should remove this magic and let users design their own BaseEntity to support such behaviour. Similar problem is there with `id` and `_id` magic that was baked for mongo only (`id` is string value of mongo's `ObjectID` which is under `_id`). 

But I would like to keep this possible via BaseEntity, so you do not have to pollute your entities like this (lot of boiler plate code in every single entity). Never tried, but I am afraid entity inheritance will not work out of the box. If it would be hard, I was thinking about adding simple configuration switch to disable those fields -- that would be pretty simple and straightforward. 

Anyway, this is gonna be BC, adding `v2` milestone.

lib/Utils.ts
```diff
@@ -7,7 +7,7 @@ import { getMetadataStorage } from './MikroORM';
 
 export class Utils {
 
-  private static readonly DIFF_IGNORED_KEYS = ['_id', '_initialized', 'createdAt', 'updatedAt'];
+  private static readonly DIFF_IGNORED_KEYS = ['_id', '_initialized'];
```
B4nan: I would be careful here, not 100% sure, but this might break something.

tests/entities/Test.ts
```diff
@@ -4,7 +4,7 @@ import { BaseEntity, Entity, Property } from '../../lib';
 export class Test extends BaseEntity {
 
   @Property()
-  name: string;
+  name: string = '';
```
B4nan: Do you need the default value here?

tests/Utils.test.ts
```diff
@@ -79,12 +79,11 @@ describe('Utils', () => {
 
   test('prepareEntity changes entity to string id', async () => {
     const author1 = new Author('Name 1', 'e-mail');
-    const book = new Book('test', author1);
     const author2 = new Author('Name 2', 'e-mail');
-    author2.favouriteBook = book;
```
B4nan: This is wrong, the test is about case with single book by one author, which is also favourite book of another author. And it is there to be sure we do not store the whole object in database, but only the referenced id. 

If this test fails randomly, then you had to break something in the code. Please do not fix stuff like that by simplifying tests.

tests/entities/Test.ts
pepakriz: Yes, because of un unrelated bug - https://github.com/B4nan/mikro-orm/blob/master/lib/UnitOfWork.ts#L116 - `itemChangeSet` can be null. I'll fix it in separated PR

lib/Utils.ts
pepakriz: Really? My idea is to drop all magic properties from entities. And after this commit, fields `createdAt` and `updatedAt` will be standard fields like other. Do you agree?

tests/Utils.test.ts
pepakriz: I understand the idea of the test, but after my changes, it's doing the same job. Isn't it?

The reason why I replaced Author for Book is the `diffEntities` method could return random data. Sometimes it contains `updatedAt` field, sometimes not. It depends on a time interval between field updating.

Previously `updatedAt`, `createdAt` were being excluded from the result, so the easiest way to fix it was to change tested entity which doesn't contain dynamic fields like `Date`.

lib/Utils.ts
B4nan: I definitely want to have this feature without the need to use hooks. I want that for all entities, so its just a boiler plate for me. As I said, I am ok with putting this under config toggle or moving the feature to some decorator, etc. 

One way to do this would be with custom decorators like typeorm does. I am currently working on that drivers support and of course, all this BaseEntity magic is the hardest part to manage (rest is almost ready). I am trying to handle that id: string vs _id: ObjectID thing to custom decorator too, that might be the correct way imho (I am totally fine with having to declare those fields, but not with the need for hooks). 

Back to this specific line, I am not 100% sure that the update hook is fired only when we write to database (it should be, just not sure). It might result in updating the `updatedAt` field even if we do not change other fields. But maybe I am just too scared, I will try just to remove those two from `DIFF_IGNORED_KEYS` without other changes to see how it works in the real world :]

Anyway I was thinking about setting up a slack channel so we can discuss those architectural changes first.

tests/Utils.test.ts
B4nan: How is this connected to your PR? The test was not about created/updated fields, but as I said, about writing reference id instead of the whole object (which is valid case in mongo).

Btw did you try running tests without your changes? It should be stable (at least on my macbook it is, as well as on CI afaik). The issue you are talking about sounds strange.

B4nan: Hmm I think I understand finally - you removed those fields from ignored and it fails randomly. Anyway, the test is still valid.

This can be solved easily by using different expectation (`toMatchObject` instead of `toEqual`), or by removing those created/updated fields from the returned difference manually before asserting.

B4nan: Or we could simply assert that `name === 'Name 2'` and `favouriteBook === book._id` with two simple asserts.

pepakriz: > Or we could simply assert that `name === 'Name 2'` and `favouriteBook === book._id` with two simple asserts.

The best and the simplest :+1:  I'll change it

B4nan: Actually we could assert only that `diff.favouriteBook === book._id` here, as that is the real thing the test is about. There are many others that cover basic scalar property updates.

B4nan:
I did try to rebase your branch and finish the job there, but it was full of conflicts after I merged the drivers branch. 

Anyway now you need to define those fields manually like this to have it working:

```
  @Property()
  createdAt = new Date();

  @Property({ onUpdate: () => new Date() })
  updatedAt = new Date();
```

I also fixed hooks - your approach was not working because hooks were run agains entities too late (after we compute the change set) so changes to entity done inside hook was never persisted.

PR #6, opened 2019-01-11
Run lint on CI

pepakriz:


.travis.yml
```diff
@@ -13,6 +13,8 @@ cache:
   directories:
     - 'node_modules'
 
-script: ORM_PARALLEL=1 yarn test
+script:
+  - yarn run lint
```
B4nan: `yarn lint` and `yarn test` is enough -- this is how `yarn` works, why not to use it?

pepakriz: It prevents collision with the native yarn scripts. But I'm OK with aliases.

Thank you!

PR #7, opened 2019-01-11
Improve options typing

pepakriz:


lib/EntityFactory.ts
```diff
@@ -127,14 +127,12 @@ export class EntityFactory {
...
 
-    if (this.options.debug) {
-      this.logger(`ORM entity discovery started`);
-    }
+    this.logger(`ORM entity discovery started`);
 
     const project = new Project();
     const sources: SourceFile[] = [];
 
-    if (!this.em.options.entitiesDirsTs) {
```
B4nan: what is wrong with this?

lib/EntityFactory.ts
```diff
@@ -145,17 +143,13 @@ export class EntityFactory {
     this.options.entitiesDirs.forEach(dir => this.discover(sources, dir));
     const diff = Date.now() - startTime;
 
-    if (this.options.debug) {
```
B4nan: the purpose of debug flag was basically to reduce log messages (make them less verbose). 

as a user i want to be able to see query logger stuff, but i want to be able to disable that, not sure if toggling the `logger` option is better. 

originally this was introduced just to be able to disable those discovery messages. correct solution would be probably to introduce more robust logger interface with various log levels, and let users choose the level they want.

lib/MikroORM.ts
```diff
@@ -31,27 +40,12 @@ export class MikroORM {
     return orm;
   }
 
-  constructor(public options: Options) {
-    if (!this.options.dbName) {
```
B4nan: i would like to keep this human readable error messages. if i understand your solution, it will be type-checked only by TS compiler? i dont think that is enough (you can bypass those TS errors, right?)

the trick with merging default options is fine, i would just keep the validation here.

lib/MikroORM.ts
```diff
@@ -31,27 +40,12 @@ export class MikroORM {
...
-      throw new Error('No database specified, please fill in `dbName` option');
-    }
-
-    if (!this.options.entitiesDirs || this.options.entitiesDirs.length === 0) {
```
B4nan: also this particular validation is about the need to have at least one entity directory, while your change allows empty array which is invalid.

lib/EntityFactory.ts
pepakriz: Agree, so I'll revert it

pepakriz:
@B4nan it's ready to merge. A lot of thinks was removed, so PR is only about better Options typing. Next time I'll try to make smaller and more indipendent PRs.

BTW because the code coverage has been decreased (-0.01% :smile:), PR is marked as failing. Is it necessary?

B4nan:
I guess that is just because you removed some tested code, don't worry about that :]

PR #8, opened 2019-01-11
Add support for MySQL [WIP]

B4nan:
Progress:
- [x] refactor db manipulation to drivers
- [x] introduce @PrimaryKey decorator (remove it from BaseEntity)
- [x] add support for M:N pivot tables
- [x] introduce mongodb and mysql drivers
- [x] improve support for joins (populate N:1 and 1:N)
- [x] fix rest of mysql tests
- [x] handle mysql driver configuration properly (hardcoded now)
- [x] add basic transactions api to EM
- [x] update readme
- [x] add support for naming strategy so we can have underscored lowercase table and field names in MySQL
- [x] remove metadata cache (implement first test run caching somehow?) used in QB tests


lib/EntityManager.ts
```diff
@@ -50,9 +50,8 @@ export class EntityManager {
     return em.identityMap;
   }
 
-  getCollection(entityName: string): MongoCollection {
-    const col = this.metadata[entityName] ? this.metadata[entityName].collection : entityName;
-    return this.db.collection(col);
+  getDriver<D extends IDatabaseDriver = IDatabaseDriver>(): D {
```
pepakriz: You can use generics on the class instead.

tests/entities-mysql/Author2.ts
```diff
@@ -0,0 +1,80 @@
...
+
+import { Book2 } from './Book2';
+
+@Entity()
+export class Author2 extends BaseEntity {
+
+  static beforeDestroyCalled = 0;
+  static afterDestroyCalled = 0;
+
+  @PrimaryKey()
```
pepakriz: I vote for `@PrimaryProperty`

pepakriz: Or duo `@PrimaryColumn` and `@Column`

lib/decorators/PrimaryKey.ts
```diff
@@ -0,0 +1,41 @@
...
+  };
+}
+
+export type PrimaryKeyOptions = {
+  name?: string;
+  type?: any;
+  [prop: string]: any;
+}
+
+export type IPrimaryKey = number | string | ObjectID;
```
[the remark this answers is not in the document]
B4nan: Name conflict in the very same file. Will think about renaming the decorator.

lib/EntityManager.ts
B4nan: That was my first approach but it was throwing strange errors on very strange places when I provided default type (and I do not want to require the type on `EntityManager`). Will keep it this way for now, feel free to send PR once this one is merged :]

tests/entities-mysql/Author2.ts
B4nan: We could also go for similar approach doctrine uses -- use multiple decorators like `@Property()` and `@Primary()` or `@Id()`. 

Basically the only purpose of the decorator is now to provide magic `_id`/`id` getter and setter for mongo (plus I think we should add validation so there is always exactly one PK). It would be imho more DRY/SRP approach.

PR #13, opened 2019-02-26
Add support for usage with vanilla JS

B4nan:
- [x] allow users to provide custom metadata provider
- [x] provide TS and JS metadata providers


B4nan:
WDYT @stalniy?

stalniy:
At the first sight looks cool! :)

I need to spend a bit more time to understand MetadataProvider API. Will give feedback tomorrow in the morning.

stalniy:
Sorry for the delay with feedback a bit busy... My wife has birthday on this Saturday, so can't spend enough time on Open Source :) 
I'll look into this till the end of the week

B4nan:
I am merging this now, feel free to add comments anyway, any feed back appreciated :]

stalniy:
Looks good. 

As far as I see currently MetadataStorsge is restricted to load metadata from files/directories (its `discover`) do something like this. 

What would be awesome for my project is an ability to implement custom logic for discovery process. I want to load metadata from database (this would allow to implement metadata part of EAV design on top of existing ORM). 

To do so `discover` needs to be an async function which delegates everything to `MetadataProvider`. 

WDYT?

B4nan:
Load metadata from another ORM? :] Sounds strange, but whatever :] 

So what you are proposing, if I understand, is to be able to use vanilla JS entities and make them behave based on schema stored in database? Why do you want this? What about having common base entity instead? 

> As far as I see currently MetadataStorsge is restricted to load metadata from files/directories (its discover).

`MetadataStorage` is not an extension point, it is the wrapper around the whole metadata thing. It reads `entities` (with #18) or `entitiesDirs` options to find existing entities (which are always files). 

The extension point is `MetadataProvider` with basically only one method: 
```
abstract discoverEntity(meta: EntityMetadata, name: string): void
```

`name` is the name of discovered entity, and `meta` is the place where to store the metadata. So it is not really like this, it is not restricted to files/directories. So just by changing the signature to async method, you could provide something like this:

```
import { Cascade, EntityMetadata, EntityProperty, MetadataProvider, ReferenceType } from 'mikro-orm';

export class DatabaseMetadataProvider extends MetadataProvider {

  async discoverEntity(meta: EntityMetadata, name: string): void {
    const db = await this.getMetadataDatabase();
    const properties = await db.find({ entity: name });
    Object.entries(properties).forEach(prop => {
      if (typeof prop === 'string') {
        schema.properties[name] = { type: prop };
      }
    });

    Utils.merge(meta.properties, properties.map(prop => this.initProperty(prop)));
  }

  private async getMetadataDatabase(): MongoClient { ... }
  private async initProperty(data: any): EntityProperty { ... }

}
```

I should probably also export `Utils` helpers from `index.ts` so `Utils.merge` can be used when extending internal parts of ORM like this.

B4nan:
Thinking about this now, I will probably rename the method to `MetadataProvider.loadEntityMetadata()` as the entity is already discovered (ORM knows about it), and the responsibility of this method is to load the metadata (schema), not to discover it.

B4nan:
Btw there is metadata caching in place, you would need to disable that or provide custom cache adapter that would handle cache invalidation somehow. Now it checks modified time of entity file, which wouldn't work for this case.

B4nan:
I did change it `MetadataProvider.loadEntityMetadata()` and made it async.

stalniy:
Cool! Thank you very much!

So, the reasons behind `DatabaseMetadataProvider` is to be able to make ORM work with EAV (Entity Attribute Value) in the similar way as with regular database structure. 

In EAV, you need to store metadata of all entities in database (so the user can change them later).

**Probably you heard that EAV is an anti-pattern. Why do I want to use it?**

Because this is a very good way to design data-driven applications, EAV allows fast write and easy schema migrations. 

**What else will I need to customize?**

`UnitOfWork` and `EntityMapper` (I think you name it `EntityLoader`). But I didn't have time to dive deeper into your source code :) Probably will start investigate it more in the begining of April

PR #61, opened 2019-06-09
feat(core): add support for mapping raw DB results via EM.map()

### LINKED ISSUE #56: Use 3rd party query builder
It would be awesome to be able to use custom QueryBuilder like https://knexjs.org/. It has a lot of cool features, migration support, seeding, etc.

B4nan:
Closes #56


stalniy:
Awesome! Thanks!

PR #80, opened 2019-07-25
MikroORM v3

B4nan:
**Progress**
- [x] Change `autoFlush` default value to `false` #63
- [x] Use knex as query runner #64
- [x] Enable connection pooling #64
- [x] Rework schema generator to use knex #81
- [x] Remove deprecated 1:m `fk` parameter #87
- [x] Move `MetadataStorage.getClassName()` to `NamingStrategy` #15
- [x] Use instance of metadata instead of global storage #91
- [x] Add onInit lifecycle hook #83
- [x] Support virtual property getters #82
- [x] Support multiple conditions in JOINs #70
- [x] Add entity generator based on current db schema #78, #98 
- [x] Implement schema updates #97
- [x] Add basic CLI tool #101, #102
- [x] Improve logging - add namespaces, colors and highlighting #108, #109
- [x] Support for read replica connections #77, #116 
- [x] Add `Reference<T>` wrapper to allow better type checking of m:1 and 1:1 references #107, #117 
- [x] Add support for MariaDB #110, #120
- [x] Add `findOneOrFail` method to entity manager and repository #133, #142 
- [x] Add debug CLI command to ease debugging of ORM's configuration #136
- [x] Allow logging full query including params #155
- [x] Add type-safe way to define relationships #146
- [x] Validate one to one relationship metadata #149 
- [x] Auto-wire missing references from owner to inverse side #149 
- [x] Allow populating all relationships via passing `populate: true` #160 
- [x] Add `findAndCount()` to EM and repository interface #123 
- [x] Allow querying by relationships #157 
- [x] Add support for eager loading #168
- [x] Improve typing of FilterQuery #124, #193
- [x] Rework IEntity interface and entity definition #171, #193
- [x] Allow mapping results of query builder to entities directly
- [x] Add support for bundling via Webpack #196, #200
- [x] Allow using composite PK in M:N pivot tables #121, #204
- [x] Do not require `entity` attribute in collection decorators #207
- [x] Add support for migrations via umzug #209
- [x] Add migrations support to CLI #209
- [x] Add support for enums via `@Enum()` #215, #232 
- [x] Add support for create/drop database #237
- [x] Add `ReflectMetadataProvider` #240
- [ ] Add support for user defined indexes via `@Index()` #226
- [ ] Support searching by regexp in query builder

**Documentation**
- [x] autoFlush
- [x] new schema generator API
- [x] entity generator
- [x] CLI
- [x] virtual properties
- [x] new naming strategy api (`getClassName()`)
[...]

PR #84, opened 2019-08-06
WIP: Fix for using globs

NickBolles:
This fixes issues with using globs for entity dirs

<details> 
  <summary><h3>Click here for all of my investigation / rambling</h3></summary>
   <span>

```
0|server  | Error saving recipe TypeError: Cannot read property 'name' of undefined
0|server  |     at Function.className (/home/coder/project/packages/server/node_modules/mikro-orm/dist/utils/Utils.js:156:28)
0|server  |     at MongoConnection.getCollectionName (/home/coder/project/packages/server/node_modules/mikro-orm/dist/connections/MongoConnection.js:153:26)
0|server  |     at MongoConnection.getCollection (/home/coder/project/packages/server/node_modules/mikro-orm/dist/connections/MongoConnection.js:19:36)
0|server  |     at MongoConnection.insertOne (/home/coder/project/packages/server/node_modules/mikro-orm/dist/connections/MongoConnection.js:84:28)
0|server  |     at MongoDriver.nativeInsert (/home/coder/project/packages/server/node_modules/mikro-orm/dist/drivers/MongoDriver.js:33:32)
0|server  |     at ChangeSetPersister.persistEntity (/home/coder/project/packages/server/node_modules/mikro-orm/dist/unit-of-work/ChangeSetPersister.js:37:37)
0|server  |     at ChangeSetPersister.persistToDatabase (/home/coder/project/packages/server/node_modules/mikro-orm/dist/unit-of-work/ChangeSetPersister.js:20:20)
0|server  |     at UnitOfWork.commitChangeSet (/home/coder/project/packages/server/node_modules/mikro-orm/dist/unit-of-work/UnitOfWork.js:200:35)
0|server  |     at processTicksAndRejections (internal/process/task_queues.js:86:5)
```

I added a log for 

https://github.com/B4nan/mikro-orm/blob/master/lib/connections/MongoConnection.ts#L181-L184
```
  private getCollectionName(name: EntityName<IEntity>): string {
    name = Utils.className(name);
    console.log(this.metadata)
    return this.metadata[name] ? this.metadata[name].collection : name;
  }
```
```
0|server  |   Recipe:
0|server  |    { properties:
0|server  |       { _id: [Object],
0|server  |         createdAt: [Object],
0|server  |         updatedAt: [Object],
0|server  |         name: [Object],
0|server  |         description: [Object] },
0|server  |      primaryKey: '_id',
0|server  |      path:
0|server  |       '/home/coder/project/packages/server/src/Recipe/entities/recipe.entity.ts',
0|server  |      name: 'Recipe',
0|server  |      constructorParams: [],
0|server  |      extends: undefined },
```

This fails because there is an entry in metadata for the entity, but the collection property isn't set. 

### Workaround? Maybe?

If I specify the collection on the entity
```typescript
@Entity({ collection: "Recipe" })
export class Recipe {
//...
```
Here's the metadata now
```
0|server  |   Recipe:
0|server  |    { properties:
0|server  |       { _id: [Object],
0|server  |         createdAt: [Object],
0|server  |         updatedAt: [Object],
0|server  |         name: [Object],
0|server  |         description: [Object] },
0|server  |      primaryKey: '_id',
0|server  |      path:
0|server  |       '/home/coder/project/packages/server/src/Recipe/entities/recipe.entity.ts',
0|server  |      collection: 'Recipe',
0|server  |      name: 'Recipe',
0|server  |      constructorParams: [],
0|server  |      extends: undefined },
```

But then I get another error:
```
Error saving recipe TypeError: Cannot read property 'name' of undefined
0|server  |     at Function.fromMergeWithoutPK (/home/coder/project/packages/server/node_modules/mikro-orm/dist/utils/ValidationError.js:49:68)
0|server  |     at EntityValidator.validatePrimaryKey (/home/coder/project/packages/server/node_modules/mikro-orm/dist/entity/EntityValidator.js:55:43)
0|server  |     at EntityManager.merge (/home/coder/project/packages/server/node_modules/mikro-orm/dist/EntityManager.js:154:20)
0|server  |     at UnitOfWork.commitChangeSet (/home/coder/project/packages/server/node_modules/mikro-orm/dist/unit-of-work/UnitOfWork.js:203:17)
0|server  |     at processTicksAndRejections (internal/process/task_queues.js:86:5
```

https://github.com/B4nan/mikro-orm/blob/master/lib/entity/EntityValidator.ts#L66-L70
```typescript
  validatePrimaryKey<T extends IEntityType<T>>(entity: EntityData<T>, meta: EntityMetadata): void {
    if (!entity || (!entity[meta.primaryKey] && !entity[meta.serializedPrimaryKey])) {
      throw ValidationError.fromMergeWithoutPK(meta);
    }
  }
```

Here's my entity (console.logged from EntityRepository#GetRepository) It doesn't have an _id, so it fails this check.

```
0|server  |   Recipe {
0|server  |   createdAt: 2019-08-01T17:09:13.160Z,
0|server  |   updatedAt: 2019-08-01T17:09:13.160Z,
0|server  |   name: 'My Recipe',
0|server  |   description: 'Some cool recipe Created: 1564679353160' }
```

## More Logs!!!

```
0|server  | Getting Unit of work: 
0|server  | Insert one into :  Recipe string
0|server  | Get Collection Name for:  Recipe string
0|server  | Get Collection Name for classname:  Recipe string
0|server  | metdata:  true { BaseEntity:
0|server  |    { properties: { _id: [Object], createdAt: [Object], updatedAt: [Object] },
0|server  |      primaryKey: '_id',
0|server  |      path:
0|server  |       '/home/coder/project/packages/server/src/shared/entities/BaseEntity.ts' },
0|server  |   Recipe:
0|server  |    { properties:
0|server  |       { _id: [Object],
0|server  |         createdAt: [Object],
0|server  |         updatedAt: [Object],
0|server  |         name: [Object],
0|server  |         description: [Object] },
0|server  |      primaryKey: '_id',
0|server  |      path:
0|server  |       '/home/coder/project/packages/server/src/Recipe/entities/recipe.entity.ts',
0|server  |      collection: 'Recipe',
0|server  |      name: 'Recipe',
0|server  |      constructorParams: [],
0|server  |      extends: undefined },
0|server  |   User:
0|server  |    { properties:
0|server  |       { username: [Object],
0|server  |         allServices: [Object],
0|server  |         emails: [Object],
0|server  |         deactivated: [Object],
0|server  |         recipes: [Object] },
0|server  |      path:
0|server  |       '/home/coder/project/packages/server/src/User/entities/user.entity.ts',
0|server  |      name: 'User',
0|server  |      constructorParams: [],
0|server  |      extends: 'BaseEntity' } }
0|server  | CollectionName:  Recipe string
0|server  | Get Collection Name for:  Recipe string
0|server  | Get Collection Name for classname:  Recipe string
0|server  | metdata:  true { BaseEntity:
0|server  |    { properties: { _id: [Object], createdAt: [Object], updatedAt: [Object] },
0|server  |      primaryKey: '_id',
0|server  |      path:
0|server  |       '/home/coder/project/packages/server/src/shared/entities/BaseEntity.ts' },
0|server  |   Recipe:
0|server  |    { properties:
0|server  |       { _id: [Object],
0|server  |         createdAt: [Object],
0|server  |         updatedAt: [Object],
0|server  |         name: [Object],
0|server  |         description: [Object] },
0|server  |      primaryKey: '_id',
0|server  |      path:
0|server  |       '/home/coder/project/packages/server/src/Recipe/entities/recipe.entity.ts',
0|server  |      collection: 'Recipe',
0|server  |      name: 'Recipe',
0|server  |      constructorParams: [],
0|server  |      extends: undefined },
0|server  |   User:
0|server  |    { properties:
0|server  |       { username: [Object],
0|server  |         allServices: [Object],
0|server  |         emails: [Object],
0|server  |         deactivated: [Object],
0|server  |         recipes: [Object] },
0|server  |      path:
0|server  |       '/home/coder/project/packages/server/src/User/entities/user.entity.ts',
0|server  |      name: 'User',
0|server  |      constructorParams: [],
0|server  |      extends: 'BaseEntity' } }
0|server  | Merging Entity Recipe {
0|server  |   createdAt: 2019-08-01T17:27:29.960Z,
0|server  |   updatedAt: 2019-08-01T17:27:29.963Z,
0|server  |   name: 'My Recipe',
0|server  |   description: 'Some cool recipe Created: 1564680449959',
0|server  |   __primaryKey: 5d4321011bb0124e0584efed }  With data undefined
0|server  | Entity Classname is  My Recipe string metadata:  false undefined
0|server  | Throwing merge without PK for entity undefined and Meta undefined
0|server  | Error saving recipe TypeError: Cannot read property 'name' of undefined
0|server  |     at Function.fromMergeWithoutPK (/home/coder/project/packages/server/node_modules/mikro-orm/dist/utils/ValidationError.js:49:68)
0|server  |     at EntityValidator.validatePrimaryKey (/home/coder/project/packages/server/node_modules/mikro-orm/dist/entity/EntityValidator.js:60:37)
0|server  |     at EntityManager.merge (/home/coder/project/packages/server/node_modules/mikro-orm/dist/EntityManager.js:164:20)
0|server  |     at UnitOfWork.commitChangeSet (/home/coder/project/packages/server/node_modules/mikro-orm/dist/unit-of-work/UnitOfWork.js:203:17)
0|server  |     at processTicksAndRejections (internal/process/task_queues.js:86:5)
```

### Stack
It looks like that got us a little further (modified it in node_modules)

but now we get this error (again)
```
Error saving recipe TypeError: Cannot read property 'name' of undefined
    at Function.fromMergeWithoutPK (/home/coder/project/packages/server/node_modules/mikro-orm/dist/utils/ValidationError.js:49:68)
    at EntityValidator.validatePrimaryKey (/home/coder/project/packages/server/node_modules/mikro-orm/dist/entity/EntityValidator.js:55:43)
    at EntityManager.merge (/home/coder/project/packages/server/node_modules/mikro-orm/dist/EntityManager.js:139:24)
    at UnitOfWork.commitChangeSet (/home/coder/project/packages/server/node_modules/mikro-orm/dist/unit-of-work/UnitOfWork.js:192:25)
    at processTicksAndRejections (internal/process/task_queues.js:86:5)
```

This is called from
https://github.com/B4nan/mikro-orm/blob/master/lib/unit-of-work/UnitOfWork.ts#L240-L244

### Conclusion 1

**EntityManager#merge - When creating an entity with data only `const recipe = await this.recipeRepository.create({ ...recipeInput, description });` it will be called with the data, which won't be an entity, which won't have an ID and will fail merge**

# Problem two
If I use new Recipe, Object.assign and then persistAndFlush, it correctly creates an ID in the __primaryKey property. But still doesn't work
```
Merging Entity Recipe {
0|server  |   createdAt: 2019-08-01T17:41:46.322Z,
0|server  |   updatedAt: 2019-08-01T17:41:46.325Z,
0|server  |   name: 'My Recipe',
0|server  |   description: 'Some cool recipe Created: 1564681306322',
0|server  |   __primaryKey: 5d43245aec82af4f3c8a569c }  With data undefined
0|server  | Entity Classname is  My Recipe string metadata:  false undefined
0|server  | Throwing merge without PK for entity undefined and Meta undefined
```

```typescript
const recipe = new Recipe();
    const description = recipeInput.description + " Created: " + Date.now();
    Object.assign(recipe, recipeInput, { description });
    // const recipe = await this.recipeRepository.create({ ...recipeInput, description });
    console.log("before persist and flush recipe", recipe, "typeof", typeof recipe);
    await this.recipeRepository.persistAndFlush(recipe);
    console.log("after persist and flush recipe", recipe);
    recipe.description = "goo";
    console.log("after set description recipe", recipe);
    try {
      await this.recipeRepository.persistAndFlush(recipe);
    } catch (e) {
      console.log("Error saving recipe", e);
    }
    console.log("Recipe Saved", recipe);
    return recipe;
```

#### Stack
persistAndFlush -> EntityManager#flush -> UnitOfWork#commit -> UnitOfWork$commitChangeSet -> ChangeSetPersister#persistToDatabase -< persistEntity#ChangeSetPersister -> MongoDriver#nativeInsert -> MongoConnection#insertOne -> MongoConnection#getCollectionName
image

image

Now we have `undefined` as the `collection` and we call two more lines, then we call `this.getCollection` with undefined

image

Which calls collectionName again, on undefined

image

which calls className on undefined `name = utils_1.Utils.className(name);`

### Conclusion 2
**getCollectionName fails if the metadata doesn't have a collection**

#### Fixes?
I think the failure is correct. If you pass in undefined `collection` to `getCollectionName` it should probably fail.

But returning undefined from getCollectionName in the first place seems wrong. We should probably change this line from:
```typescript
return this.metadata[name] ? this.metadata[name].collection : name;
```

to
```typescript
return (this.metadata[name] && this.metadata[name].collection) ? this.metadata[name].collection : name;
```
or, IMO a little bit cleaner
```typescript
return (this.metadata[name] && this.metadata[name].collection) || name;
```

But, the root cause might be, why is collection null in the first place??

# More problems

Ugh.... Okay lets run the example in nestjs-mikro-orm. That seems to work for both SQLLite and mongodb. 

What's the difference??
Version:
2.7.1
my version
2.7.4

okay, so let's try using 2.7.1....
We get an error because `./**/entities/` is not a directory... interesting, so glob patterns was added since 2.7.1?? 

So let's try an entities directory. If the only change I make is move recipe.entity.ts to an `./entities/` folder, it works perfectly....... UGH! So there's an issue with loading the files with globby.

  </span>
</details>

### DIFF
--- a/tests/MikroORM.test.ts
+++ b/tests/MikroORM.test.ts
@@ -27,6 +27,14 @@ describe('MikroORM', () => {
     await expect(MikroORM.init({ dbName: 'test', baseDir: BASE_DIR, entities: [FooBaz2], cache: { enabled: false }, entitiesDirsTs: ['entities'] })).rejects.toThrowError(error);
   });
 
+  test('should work with globs', async () => {
+    // todo: confirm that entities are loaded correctly
+    await expect(MikroORM.init({ dbName: 'test', baseDir: BASE_DIR, entitiesDirs: ['./entities-nested/**/entities'], cache: { enabled: false }, entitiesDirsTs: ['./entities-nested/**/entities'] })).not.toThrowError();
+    await expect(MikroORM.init({ dbName: 'test', baseDir: BASE_DIR, entitiesDirs: ['entities-nested/**/entities'], cache: { enabled: false }, entitiesDirsTs: ['entities-nested/**/entities'] })).not.toThrowError();
+    await expect(MikroORM.init({ dbName: 'test', baseDir: BASE_DIR, entitiesDirs: ['entities-nested/**/entities/'], cache: { enabled: false }, entitiesDirsTs: ['entities-nested/**/entities/'] })).not.toThrowError();
+    await expect(MikroORM.init({ dbName: 'test', baseDir: BASE_DIR, entitiesDirs: ['./entities-nested/**/entities'], cache: { enabled: false } })).not.toThrowError();
+  });
+
   test('should init itself with entity manager', async () => {
     const orm = await MikroORM.init({
       entitiesDirs: ['entities'],
--- a/lib/metadata/MetadataDiscovery.ts
+++ b/lib/metadata/MetadataDiscovery.ts
@@ -1,15 +1,14 @@
+import { extname, parse } from 'path';
 import * as globby from 'globby';
-import { extname } from 'path';
 
 import { EntityClass, EntityClassGroup, EntityMetadata, EntityProperty, IEntityType } from '../decorators';
 import { EntityManager } from '../EntityManager';
 import { Configuration, Logger, Utils, ValidationError } from '../utils';
+import { Cascade, EntityHelper, ReferenceType } from '../entity';
 import { MetadataValidator } from './MetadataValidator';
 import { MetadataStorage } from './MetadataStorage';
-import { Cascade, EntityHelper, ReferenceType } from '../entity';
 
 export class MetadataDiscovery {
-
   private readonly metadata = MetadataStorage.getMetadata();
   private readonly namingStrategy = this.config.getNamingStrategy();
   private readonly metadataProvider = this.config.getMetadataProvider();
@@ -18,13 +17,15 @@ export class MetadataDiscovery {
   private readonly validator = new MetadataValidator();
   private readonly discovered: EntityMetadata[] = [];
 
-  constructor(private readonly em: EntityManager,
-              private readonly config: Configuration,
-              private readonly logger: Logger) { }
+  constructor(
+    private readonly em: EntityManager,
+    private readonly config: Configuration,
+    private readonly logger: Logger
+  ) {}
 
   async discover(): Promise<Record<string, EntityMetadata>> {
     const startTime = Date.now();
-    this.logger.debug(`ORM entity discovery started`);
+    this.logger.debug('ORM entity discovery started');
     this.discovered.length = 0;
     const tsNode = process.argv[0].endsWith('ts-node') || process.argv.slice(1).some(arg => arg.includes('ts-node'));
 
@@ -39,7 +40,9 @@ export class MetadataDiscovery {
     // validate base entities
     this.discovered
       .filter(meta => meta.extends && !this.discovered.find(m => m.prototype.constructor.name === meta.extends))
-      .forEach(meta => { throw ValidationError.fromUnknownBaseEntity(meta); });
+      .forEach(meta => {
+        throw ValidationError.fromUnknownBaseEntity(meta);
+      });
 
     // ignore base entities (not annotated with @Entity)
     const filtered = this.discovered.filter(meta => meta.name);
@@ -50,37 +53,30 @@ export class MetadataDiscovery {
     this.logger.debug(`- entity discovery finished after ${diff} ms`);
 
     const discovered: Record<string, EntityMetadata> = {};
-    this.discovered
-      .filter(meta => meta.name)
-      .forEach(meta => discovered[meta.name] = meta );
+    this.discovered.filter(meta => meta.name).forEach(meta => (discovered[meta.name] = meta));
 
     return discovered;
   }
 
   private async discoverDirectory(basePath: string): Promise<void> {
-    const files = await globby('*', { cwd: Utils.normalizePath(this.config.get('baseDir'), basePath) });
+    const globsToIgnore = ['*.js.map', '*.d.ts', '.*', 'index.[jt]s'];
+    const glob = [`${basePath}/(*.[jt]s)`, `**/!{${globsToIgnore.join(',')}}`];
+    const files = await globby(glob, { cwd: this.config.get('baseDir') });
+
     this.logger.debug(`- processing ${files.length} files from directory ${basePath}`);
 
     for (const file of files) {
-      if (
-        !file.match(/\.[jt]s$/) ||
-        file.endsWith('.js.map') ||
-        file.endsWith('.d.ts') ||
-        file.startsWith('.') ||
-        file.match(/index\.[jt]s$/)
-      ) {
-        this.logger.debug(`- ignoring file ${file}`);
-        continue;
-      }
-
       const name = this.getClassName(file);
-      const path = Utils.normalizePath(this.config.get('baseDir'), basePath, file);
+      const path = Utils.normalizePath(this.config.get('baseDir'), file);
       const target = require(path)[name]; // include the file to trigger loading of metadata
       await this.discoverEntity(target, path);
     }
   }
 
-  private async discoverEntity<T extends IEntityType<T>>(entity: EntityClass<T> | EntityClassGroup<T>, path?: string): Promise<void> {
+  private async discoverEntity<T extends IEntityType<T>>(
+    entity: EntityClass<T> | EntityClassGroup<T>,
+    path?: string
+  ): Promise<void> {
     entity = this.metadataProvider.prepare(entity);
     this.logger.debug(`- processing entity ${entity.name}`);
 
@@ -88,7 +84,7 @@ export class MetadataDiscovery {
     meta.prototype = entity.prototype;
     meta.path = path || meta.path;
     meta.toJsonParams = Utils.getParamNames(entity.prototype.toJSON || '').filter(p => p !== '...args');
-    const cache = meta.path && await this.cache.get(entity.name + extname(meta.path));
+    const cache = meta.path && (await this.cache.get(entity.name + extname(meta.path)));
 
     if (cache) {
       this.logger.debug(`- using cached metadata for entity ${entity.name}`);
@@ -227,19 +223,24 @@ export class MetadataDiscovery {
   private definePivotTableEntity(meta: EntityMetadata, prop: EntityProperty): EntityMetadata | undefined {
     if (prop.reference === ReferenceType.MANY_TO_MANY && prop.owner && prop.pivotTable) {
       const pk = this.namingStrategy.referenceColumnName();
-      const primaryProp = { name: pk, type: 'number', reference: ReferenceType.SCALAR, primary: true } as EntityProperty;
+      const primaryProp = {
+        name: pk,
+        type: 'number',
+        reference: ReferenceType.SCALAR,
+        primary: true
+      } as EntityProperty;
       this.initFieldName(primaryProp);
 
-      return this.metadata[prop.pivotTable] = {
+      return (this.metadata[prop.pivotTable] = {
         name: prop.pivotTable,
         collection: prop.pivotTable,
         primaryKey: pk,
         properties: {
           [pk]: primaryProp,
           [meta.name]: this.definePivotProperty(prop, meta.name, prop.type),
-          [prop.type]: this.definePivotProperty(prop, prop.type, meta.name),
-        },
-      } as EntityMetadata;
+          [prop.type]: this.definePivotProperty(prop, prop.type, meta.name)
+        }
+      } as EntityMetadata);
     }
   }
 
@@ -271,7 +272,8 @@ export class MetadataDiscovery {
   }
 
   private getClassName(file: string): string {
-    const name = file.split('.')[0];
+    const filename = parse(file).name;
+    const name = filename.split('.')[0];
     const ret = name.replace(/-(\w)/, m => m[1].toUpperCase());
 
     return ret.charAt(0).toUpperCase() + ret.slice(1);
@@ -304,5 +306,4 @@ export class MetadataDiscovery {
 
     return '1';
   }
-
 }
--- a/lib/metadata/TypeScriptMetadataProvider.ts
+++ b/lib/metadata/TypeScriptMetadataProvider.ts
@@ -1,12 +1,13 @@
+import { isAbsolute } from 'path';
 import { Project, SourceFile } from 'ts-morph';
 import { pathExists } from 'fs-extra';
 
-import { MetadataProvider } from './MetadataProvider';
+import * as globby from 'globby';
 import { EntityMetadata, EntityProperty } from '../decorators';
 import { Utils } from '../utils';
+import { MetadataProvider } from './MetadataProvider';
 
 export class TypeScriptMetadataProvider extends MetadataProvider {
-
   private readonly project = new Project();
   private sources: SourceFile[];
 
@@ -54,29 +55,49 @@ export class TypeScriptMetadataProvider extends MetadataProvider {
 
   private async initSourceFiles(): Promise<void> {
     const tsDirs = this.config.get('entitiesDirsTs');
+    const baseDir = this.config.get('baseDir');
 
     if (tsDirs.length > 0) {
-      const dirs = await this.validateDirectories(tsDirs);
+      const dirs = await this.validateDirectories(tsDirs, baseDir);
       this.sources = this.project.addExistingSourceFiles(dirs);
     } else {
       this.sources = this.project.addSourceFilesFromTsConfig(this.config.get('tsConfigPath'));
     }
   }
 
-  private async validateDirectories(dirs: string[]): Promise<string[]> {
+  private async validateDirectories(dirs: string[], baseDir: string): Promise<string[]> {
     const ret: string[] = [];
 
     for (const dir of dirs) {
-      const path = Utils.normalizePath(this.config.get('baseDir'), dir);
-
-      if (!await pathExists(path)) {
-        throw new Error(`Path ${path} does not exist`);
-      }
-
-      ret.push(Utils.normalizePath(path, '**', '*.ts'));
+      ret.push(...(await this.validateDirectory(dir, baseDir)));
     }
 
     return ret;
   }
 
+  private async validateDirectory(dir: string, baseDir: string): Promise<string[]> {
+    const ret: string[] = [];
+    let path;
+    if (isAbsolute(dir)) {
+      path = dir;
+    } else {
+      path = Utils.normalizePath(baseDir, dir);
+    }
+
+    if (globby.hasMagic(path)) {
+      const dirs = await globby(path, {
+        cwd: this.config.get('baseDir'),
+        onlyDirectories: true
+      });
+      for (const file of dirs) {
+        ret.push(...(await this.validateDirectory(file, baseDir)));
+      }
+    } else {
+      if (!(await pathExists(path))) {
+        throw new Error(`Path ${path} does not exist`);
+      }
+      ret.push(Utils.normalizePath(path, '**', '*.ts'));
+    }
+    return ret;
+  }
 }
--- a/lib/metadata/TypeScriptMetadataProvider.ts
+++ b/lib/metadata/TypeScriptMetadataProvider.ts
@@ -1,4 +1,3 @@
-import { isAbsolute } from 'path';
 import { Project, SourceFile } from 'ts-morph';
 import { pathExists } from 'fs-extra';
 
--- a/lib/metadata/TypeScriptMetadataProvider.ts
+++ b/lib/metadata/TypeScriptMetadataProvider.ts
@@ -77,26 +77,29 @@ export class TypeScriptMetadataProvider extends MetadataProvider {
 
   private async validateDirectory(dir: string, baseDir: string): Promise<string[]> {
     const ret: string[] = [];
-    let path;
-    if (isAbsolute(dir)) {
-      path = dir;
-    } else {
-      path = Utils.normalizePath(baseDir, dir);
-    }
+    const path = Utils.toAbsolutePath(dir, baseDir);
 
     if (globby.hasMagic(path)) {
-      const dirs = await globby(path, {
-        cwd: this.config.get('baseDir'),
-        onlyDirectories: true
-      });
-      for (const file of dirs) {
-        ret.push(...(await this.validateDirectory(file, baseDir)));
-      }
-    } else {
-      if (!(await pathExists(path))) {
-        throw new Error(`Path ${path} does not exist`);
-      }
-      ret.push(Utils.normalizePath(path, '**', '*.ts'));
+      return this.validateDirGlob(path, baseDir)
+    }
+
+    if (!(await pathExists(path))) {
+      throw new Error(`Path ${path} does not exist`);
+    }
+    ret.push(Utils.normalizePath(path, '**', '*.ts'));
+
+    return ret;
+  }
+
+  private async validateDirGlob(dirGlob: string, baseDir: string): Promise<string[]> {
+    const ret: string[] = [];
+    const dirs = await globby(dirGlob, {
+      cwd: baseDir,
+      onlyDirectories: true
+    });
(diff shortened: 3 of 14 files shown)


B4nan:
First of all, please remove all code style changes so the diff is more readable (it looks like you applied some automatic formatter, right?). Also please stick to current project's code style (empty lines inside class definition, different style for constructors, do not add unnecessary brackets...). 

There was some outage in travis, I just restarted the build, now it ended but a lot of tests are failing, so obviously please fix them all (and aim to keep the 100% code coverage).

tests/MikroORM.test.ts
```diff
@@ -27,6 +27,14 @@ describe('MikroORM', () => {
     await expect(MikroORM.init({ dbName: 'test', baseDir: BASE_DIR, entities: [FooBaz2], cache: { enabled: false }, entitiesDirsTs: ['entities'] })).rejects.toThrowError(error);
   });
 
+  test('should work with globs', async () => {
+    // todo: confirm that entities are loaded correctly
+    await expect(MikroORM.init({ dbName: 'test', baseDir: BASE_DIR, entitiesDirs: ['./entities-nested/**/entities'], cache: { enabled: false }, entitiesDirsTs: ['./entities-nested/**/entities'] })).not.toThrowError();
```
B4nan: afaik this will never throw, you need to use `await expect(promise).resolves.toBeInstanceOf(MikroORM)` or something similar (or check other possibilities here https://jestjs.io/docs/en/tutorial-async).

B4nan:
Also please simplify those new entities. If you use them only for testing the entity discovery, then 2-3 simple entities should be enough, and there is no reason to have hooks and complex associations inside them.

B4nan:
Btw the reason why `collection` property was not set on given entity metadata is that is was not discovered properly. The metadata collection is first build based on decorators present in entities, and then all entities needs to be discovered based on `entities/entitiesDirs` options. In this process all default values will be handled (not just `collection`, but also stuff from naming strategy, etc). 

This is the method that needs to be fired on all entities: https://github.com/B4nan/mikro-orm/blob/a41d1ec50c824dbc92478a63b35caa6d63fb0737/lib/metadata/MetadataDiscovery.ts#L83

If you enable debug mode and set logger (https://b4nan.github.io/mikro-orm/debugging/) you should probably see there are no entities discovered.

B4nan:
About your conclusions:

> Conclusion 1
> EntityManager#merge - When creating an entity with data only const recipe = await this.recipeRepository.create({ ...recipeInput, description }); it will be called with the data, which won't be an entity, which won't have an ID and will fail merge

You are mixing two things here. `em.create()` is there so you can programatically create entity instance. 

```typescript
// these 2 are equivalent
const recipe = await this.recipeRepository.create(...);
const recipe = new Recipe(...);
```

If you want to merge the `recipe` entity, you need to be sure it has ID. Not sure what exactly are you trying to achieve. Normally you construct the entity and persist it (then it becomes managed ~ merged). Using `em.create()` and `em.merge()` like this is usually only for restoring cached entity (so it has ID, and you need to make it managed ~ part of the identity map). 

> Conclusion 2
> getCollectionName fails if the metadata doesn't have a collection

This is again because of the discovery process not passed thru the given entity. The method will always return a string if discovery happened, as `collection` property of metadata is always set. One exception is base entity I think, but that is invalid here (the method is never called on base entity). 

> More problems

globby was afaik added in v2.0.0, but its true I didn't have much time for maintaining the examples repository, will need to update all of them soon, but currently I am focusing on finishing v3 (#80) first.

B4nan:
What I am missing here is snippets where you call `MikroORM.init()`, as there is the problem - in the configuration. 

Few more notes about the discovery process and the reason why it works like this:
- we support also vanilla JS entities, where no decorators are used
- by defining a decorator in the entity, it will automatically create a record in metadata dict
- but JS entities won't have records there as they do not use decorators
- TS entities could also not be initialized if they were not loaded into the memory (so decorators where never executed)

NickBolles:
Thanks for all of the quick responses. Sorry that this is kinda a mess. Mostly this is a stab at fixing it for my use case. I'm sure it's not a great fix, and might break stuff for other use cases. 

> First of all, please remove all code style changes so the diff is more readable (it looks like you applied some automatic formatter, right?). Also please stick to current project's code style (empty lines inside class definition, different style for constructors, do not add unnecessary brackets...).
> 
> There was some outage in travis, I just restarted the build, now it ended but a lot of tests are failing, so obviously please fix them all (and aim to keep the 100% code coverage).

Yea, prettier... It's great in your own project, but when you work on an open source project it blows everything up. I should have put a WIP tag on this (added now), cuz style and tests that def needs to be fixed. I couldn't get the tests setup locally, so I was kinda leaning on Travis there.




> Also please simplify those new entities. If you use them only for testing the entity discovery, then 2-3 simple entities should be enough, and there is no reason to have hooks and complex associations inside them.

For sure, again, I didn't get tests running locally, so I was stabbing in the dark a bit. I have mariadb (mysql compatible... right?), mongodb and postgres setup on my NAS, how can I run the tests against those DBs?


> If you enable debug mode and set logger (https://b4nan.github.io/mikro-orm/debugging/) you should probably see there are no entities discovered.

Yea... I found this after a way, and it showed the `Source file for entity ${file} not found, check your 'entitiesDirsTs' option` message right away, which led me to the globs not loading correctly. Part of the problem too is that my Nest.js setup was suppressing the error message/stack, which made it harder. But even with either entitiesDirsTs set (to a glob, which it didn't support I guess) or not set, which I think should load the files from the tsConfig, it wouldn't boot up for me.



> About your conclusions:

First off, like I said in my issue response, take this with a grain of salt. `Conclusions` is probably a strong word. I was mostly just trying to organize my thoughts.

> If you want to merge the recipe entity, you need to be sure it has ID. Not sure what exactly are you trying to achieve. Normally you construct the entity and persist it (then it becomes managed ~ merged). Using em.create() and em.merge() like this is usually only for restoring cached entity (so it has ID, and you need to make it managed ~ part of the identity map).

Gotcha, I kept messing with stuff, so this was one iteration of that. But it sounds like it's meant to be used as:
1. Create instance of entity - `new Recipe()`
2. Modify it
3. Persist it `recipeRepository.persist(recipe)` (or entityManager)

> What I am missing here is snippets where you call `MikroORM.init()`, as there is the problem - in the configuration.

I'm using nestjs-mikro-orm
`app.module.ts`
```typescript
    MikroOrmModule.forRootAsync({
      useClass: AppMikroOrmOptions
    }),
```

`MikroORMOptions.ts`
```typescript
import { MikroOrmOptionsFactory, MikroOrmModuleOptions } from "nestjs-mikro-orm";

export class AppMikroOrmOptions implements MikroOrmOptionsFactory {
  createMikroOrmOptions(): Promise<MikroOrmModuleOptions> | MikroOrmModuleOptions {
    return {
      type: "mongo",
      dbName: "goalsapp-dev",
      clientUrl: "mongodb://127.0.0.1",
      entitiesDirs: ["./**/entities"],
      entitiesDirsTs: ["./**/entities"],
      baseDir: __dirname,
      cache: {
        enabled: true, // This is required?
        options: {
          cacheDir: ".build/type-cache"
        }
      },
      debug: true
    };
  }
}
```

I'll try to get an example repo setup, but might be a few days.

tests/MikroORM.test.ts
NickBolles: Yea, this test should really test that the entities are loaded correctly, so this really needs a more full featured test.

B4nan:
Ok understood :] In the meantime I spent some time with making the linter more strict (already merged to master), so that should help you with the code style a bit once you rebase. 

About tests, I would definitely go for docker, once you have it installed, then its just about booting up via `docker-compose up -d` (drop the `-d` during initial setup so you know it works). Its pretty easy to setup, even on windows it took me less than hour... And yes, it should be mariadb compatible, as the mysql driver states that too. 

About your init snippet: 

- `baseDir: __dirname,` this part might be the problem, as you are using the current directory where the `MikroORMOptions.ts` file is stored as the base directory. If that is not the root of your project ~ if your entities are not stored in its subdirectories ~ then it can't work. By dropping this option, `process.cwd()` will be used as a default. 
- caching is enabled by default and you should be able to just change the `cacheDir` as options are deeply merged with defaults. 

Strange that you did not get the source file missing error without the debug mode enabled. Would be good to get to the bottom of this as well. 

BTW what I do like in your PR is that support for both relative and absolute paths 👍 Maybe I would suggest to split this into multiple PRs, one for support of absolute paths, another for using globs (+ possibly more). Try to be granular as possible, as the PR will get squashed and only 1 commit message will be there. That commit message will be part of the changelog, so it should be informative as possible.

NickBolles:
> Ok understood :] In the meantime I spent some time with making the linter more strict (already merged to master), so that should help you with the code style a bit once you rebase.

Nice!

> About tests, I would definitely go for docker, once you have it installed, then its just about booting up via `docker-compose up -d` (drop the `-d` during initial setup so you know it works). Its pretty easy to setup, even on windows it took me less than hour... And yes, it should be mariadb compatible, as the mysql driver states that too.

The only issue with this is that I do most of my dev inside of a coder/code-server docker container on my home server (docker rocks! lol). I can probably just fire it up on my laptop or on the host of my home server to test it though. That's awesome that you have the docker-compose ready to go!

> Speaking of tests, I would like to see this problem covered with unit tests instead of booting up the whole ORM. It can be pretty slow (seconds), so testing multiple configurations might end up being +1 min on tests easily, which is bad. This should be easy to extract and test only the lookup mechanism. Another problematic part will be with entity names. Currently it is required that all entities names are unique as the metadata storage is global (it needs to be because of how decorators work). I would love to find a way around this but that is another story...

I agree. If I can get the tests running locally this would all be a lot easier to do with tests (TDD).
Have you considered splitting the tests between unit tests and E2E/integration, or whatever you want to call it?

> About your init snippet:
> 
> * `baseDir: __dirname,` this part might be the problem, as you are using the current directory where the `MikroORMOptions.ts` file is stored as the base directory. If that is not the root of your project ~ if your entities are not stored in its subdirectories ~ then it can't work. By dropping this option, `process.cwd()` will be used as a default.

I think this breaks down if I run from the base of my lerna repo though. I think there's a way around that though. I'll look into that.

> * caching is enabled by default and you should be able to just change the `cacheDir` as options are deeply merged with defaults.

...This works fine now if I comment out enabled. Why wouldn't it before? Ugh...

> Strange that you did not get the source file missing error without the debug mode enabled. Would be good to get to the bottom of this as well.

Again, maybe I did, but nest was hiding it. I can test this by falling back to the released version.

> BTW what I do like in your PR is that support for both relative and absolute paths 👍 Maybe I would suggest to split this into multiple PRs, one for support of absolute paths, another for using globs (+ possibly more). 

It seems easiest if we, right at the beginning convert everything to absolute, it makes things a lot easier IMHO.

>Try to be granular as possible, as the PR will get squashed and only 1 commit message will be there. That commit message will be part of the changelog, so it should be informative as possible.

For sure. Also conventional commits are awesome. I had just started using them before starting this PR. 

More of a high level, does this change make sense? I think I need to create a minimum representation of the issue (with and without nest maybe?) to test against and verify that it's an issue that you actually want to fix.

B4nan:
> More of a high level, does this change make sense? I think I need to create a minimum representation of the issue (with and without nest maybe?) to test against and verify that it's an issue that you actually want to fix.

Yeah I think it does make sense, if I understood it correctly. Still it might be much easier to use the approach with doing lookup yourself and passing the array of entities instead. 

Example project with the issue would be great. Nest is fine I think, I am familiar with it as I do use it as well in my projects.

NickBolles:
> Example project with the issue would be great. Nest is fine I think, I am familiar with it as I do use it as well in my projects.

Here's a reproduction repo. https://github.com/NickBolles/mikro-orm-globby-issue
If you have time, I'd appreciate you making sure I'm no doing anything dumb lol.

B4nan:
Thanks for the repo, first thing that is not 100% correct (although it should be working) is your setup of entitiesDirs. As you mentioned, thanks to globby, it should work fine, but generally you want to have it more specific. In your example, your source files will always be under `src`, and compiled files under `dist`, so it should be something like:

```
    MikroOrmModule.forRoot({
      entitiesDirs: ["./dist/**/entities"],
      entitiesDirsTs: ["./src/**/entities"],
```

Also note that by default, `entitiesDirsTs` is optional and if not provided `entitiesDirs` is used, so having those two with equal value is not needed.

I also believe that the problem here is the `baseDir` option, as you are putting there current directory, which should be `src` if run via `ts-node`, and probably `dist` if compiled and run via `node`. Try to comment that out, then the default `process.cwd()` will be used from where above settings (with `dist` and `src`) should work. 

Also I saw you are using `@Property({ default: callback })`, that is afaik not supported, it is meant only for schema generator and it should be scalar value that will be used as sql default for given column.

NickBolles:
> Thanks for the repo, first thing that is not 100% correct (although it should be working) is your setup of entitiesDirs. As you mentioned, thanks to globby, it should work fine, but generally you want to have it more specific. In your example, your source files will always be under `src`, and compiled files under `dist`, so it should be something like:
> 
> ```
>     MikroOrmModule.forRoot({
>       entitiesDirs: ["./dist/**/entities"],
>       entitiesDirsTs: ["./src/**/entities"],
> ```

> Also note that by default, `entitiesDirsTs` is optional and if not provided `entitiesDirs` is used, so having those two with equal value is not needed.
> 
> I also believe that the problem here is the `baseDir` option, as you are putting there current directory, which should be `src` if run via `ts-node`, and probably `dist` if compiled and run via `node`. Try to comment that out, then the default `process.cwd()` will be used from where above settings (with `dist` and `src`) should work.

I guess I'm confused by this. If we are running from ts-node, the src/app.module.ts file will be running, and __dirname should be the `src/` directory. If we have compiled and running `dist/app.module.js` the __dirname should be the `dist/` directory. Either way, you are starting from the same point and doing a glob from src or dist, which IMO is how it should be. I preferrably want my compiled app to be self contained (so basically src folder down can compile to js and then just use package.json, npm inistall and `node app.js` to run) Although for node apps it's a bit of a futile effort.

Updated the repo with these changes though.

> Also I saw you are using `@Property({ default: callback })`, that is afaik not supported, it is meant only for schema generator and it should be scalar value that will be used as sql default for given column.

Yea should have left that out, was messing with it. Oops. removed

B4nan:
The thing is that (if not run via ts-node), ORM needs to know both paths, as it has to be able to locate the TS source files (for sniffing all the types correctly). So if compiled and run via node, then you will get the `baseDir` set to `dist`, from where you can't find those TS source files as they live one step above. The fallback is to use `tsconfig.json`, and again, if you setup baseDir this way, even the `tsconfig.json` won't be found.

NickBolles:
> The thing is that (if not run via ts-node), ORM needs to know both paths, as it has to be able to locate the TS source files (for sniffing all the types correctly). So if compiled and run via node, then you will get the `baseDir` set to `dist`, from where you can't find those TS source files as they live one step above. The fallback is to use `tsconfig.json`, and again, if you setup baseDir this way, even the `tsconfig.json` won't be found.

Gotcha, so Typescript is required in production for Mikro-ORM? Is there a way we can compile the types? Basically it would be dumping the cache into the dist folder, right? sot hat we don't need typescript in production? 
This also seems like a tangent. Don't want to get stuck on this. I'll update the PR tonight with styling and test changes.

B4nan:
It's not required, you can still specify types manually in decorators, only those that are missing will be sniffed via reflection. Also you do not need "TS in production", you just need TS source files deployed so they can be analysed (via `ts-morph`). I personally do deploy TS files, as it's server side thing, it does not matter much so I am fine with that approach. But if you specify all types then this whole process will be skipped. 

You can also use MikroORM with vanilla JS, and specify the schema manually https://b4nan.github.io/mikro-orm/usage-with-js/.

B4nan:
Where are we with this one? I am not going to merge new features in v2 (master branch), so I guess we can close this one and if you want to implement this, please base it on current dev branch. 

Still, I would rather see discussion in an issue prior to implementation, so we can agree what it should do.

PR #126, opened 2019-09-19
[docs] add postgres to toc

lookfirst:
(sorry i missed this, maybe there is some way to re-use the toc and the sidebar?)


B4nan:
I was looking into that but I didn't find an easy way. One possibility would be to drop the homepage toc and use the sidebar there as well, but sidebar has to be defined as html so links are not working before actual build... 

I was also looking into different ways of creating the website, ideally with some tool that would allow versioning of docs, but for now there are more important things to work on :]

PR #128, opened 2019-09-19
[fix] validateOptions not considering entitiesDirsTs

lookfirst:
also updated the code to wrap and error message to be a bit less confusing and friendly.

### DIFF
--- a/lib/utils/Configuration.ts
+++ b/lib/utils/Configuration.ts
@@ -149,8 +149,11 @@ export class Configuration {
       throw new Error('No database specified, please fill in `dbName` option');
     }
 
-    if (this.options.entities.length === 0 && this.options.entitiesDirs.length === 0 && this.options.warnWhenNoEntities) {
-      throw new Error('No entities found, please use `entities` or `entitiesDirs` option');
+    if (this.options.warnWhenNoEntities && 
+        this.options.entities.length === 0 && 
+        this.options.entitiesDirs.length === 0 && 
+        this.options.entitiesDirsTs.length === 0) {
+      throw new Error('All of the `entities`, `entitiesDirs` and `entitiesDirsTs` options are empty. Maybe you forget to define this in the configuration?');
     }
   }
 


lookfirst:
Maybe also consider adding options.isValid() or some such...

B4nan:
`entitiesDirsTs` is optional, you always have to define `entitiesDirs` or `entities`, if you omit `entitiesDirsTs`, then all your project source files will be searched (based on your `tsconfig.json`).

maybe we could highlight that in docs somewhere

B4nan:
> Maybe also consider adding options.isValid() or some such...

Well options is just a POJO, and this code is inside a validation method already. Although its not public, but the validation happens automatically from the constructor (with possibility to disable it).

lookfirst:
I'm so confused.

B4nan:
`entitiesDirsTs` is currently just a way to make entity discovery faster. during this process, TS metadata provider tries to read missing types, e.g. when you define m:1 like this:

```typescript
@ManyToOne()
prop: PropType;
```

Then there is no information about the type being `PropType` in the runtime (which is needed). What TS metadata provider does is that it tries to find the TS source file for given entity, reads its properties and their types via reflection (`ts-morph`) and fills what's missing. 

When you whitelist folders with your entity source files via `entitiesDirsTs`, then `ts-morph` will search only thru those, otherwise it reads your `tsconfig.json` and searches all possible source files defined there. 

Another way is to simply fill the type manually, either via `type` or via `entity` callback:

```typescript
@ManyToOne({ type: 'PropType' })
prop: PropType;

@ManyToOne({ entity: () => PropType })
prop: PropType;
```

Then the whole TS metadata provider just validates that there is no type missing and `ts-morph` is not involved, so there is no actual need for `entitiesDirsTs`.

There are some words about this here: https://mikro-orm.io/metadata-cache/. Although it could be definitely also somewhere else, some 'internals explained' section maybe.

lookfirst:
Maybe a different name for that variable would make sense.

PR #129, opened 2019-09-19
add example with sqlite

lookfirst:


B4nan:
is this really true? i thought the filename can be pretty much any filename, regardless its extension (i would expect it to allow also files without extensions)

lookfirst:
when i had it without the suffix...

[Error: SQLITE_CANTOPEN: unable to open database file] {
  errno: 14,
  code: 'SQLITE_CANTOPEN'
}

lookfirst:
try it in the little test project i just uploaded

B4nan:
that sounds more like a permissions issue, will give it a try but it sounds really strange, especially when i was not able to find that in sqlite driver docs.

lookfirst:
what permissions? i'm on an osx mac and running in my account. adding .db fixed the issue.

B4nan:
I don't know what your error is about, but all tests are passing on my machine if I omit the extension or change it to something else (tried `.ext`).

B4nan:
Tried your example project, and it fails with that error for database `test`, but `test123` works... Guess why - you already have something called `test` in your root directory - its the `test` directory :]

lookfirst:
HAHAHAHAHAHAHA

PR #154, opened 2019-09-22
[fix] pg defaultClientURL is missing an s

lookfirst:
I can't tell which one it should be, it just seems like postgre seems wrong. =)

https://node-postgres.com/features/connecting
https://github.com/iceddev/pg-connection-string#usage


B4nan:
Next time please try to add some meaningful commit/PR message, it took me some time to realise what you changed (without looking at the diff). 

Also would be nice if you follow contributing guidelines, especially this line:

> Commit your changes using a descriptive commit message that follows defined commit message conventions. Adherence to these conventions is necessary because release notes are automatically generated from these messages.

I am using squash&merge so I will still have final control over the commit message, but I want to see something I can just reformat, instead of coming up with a meaningful sentence of what you did :]

Note about tests - this project has 100% code coverage (all lines and branches are covered). That pretty much means every change you do will cause tests to fail. Would be good if you set up the project and run tests locally before pushing changes (but feel free to use CI for that, I don't care as long as its just small change like this, but once you see tens on broken tests, that is really something you need to fix locally first).

lookfirst:
100% code coverage doesn't mean the code is correct. ;-)

I'll work to do better for you, but I will say that the first thing I do is look at the code since that is the source of truth for anything. 

After doing this open source thing solid for 20+ years now, I'm less of a commit nazi. Better things to worry about.

B4nan:
> 100% code coverage doesn't mean the code is correct. ;-)

That is not what I said. All I said is that is usually means you will need to change tests too, not just the code :]

> Better things to worry about.

Well sure, but without good commit messages it's hard to collaborate :] Reading code is slower/harder that reading good commit message.

lookfirst:
It is literally a one line change and actually I read code pretty fast.

B4nan:
> It is literally a one line change and actually I read code pretty fast.

Sure, I am not talking about this particular PR. Still commit message is not just for me or you, but for every user. It will be also used in changelog. 

(again that does not really applies to this particular PR, as I will make it `chore` or `refactor` commit anyway that will not get into changelog)

lookfirst:
Also, your tests require mongodb to be running. Not going to run that, sorry.

B4nan:
> Also, your tests require mongodb to be running. Not going to run that, sorry.

Not a fan of docker?

lookfirst:
Not on OSX. I only run it if I absolutely have to. Also seems like a lot of overhead to just run something that brew runs just fine.

lookfirst:
I'm less of a fan of mongo, despite one of my best friends working for monogo from almost day one. =) Aggregation queries and lack of proper transactions, sucks. Might as well use postgres which supports that, jsonb and more.

lookfirst:
`PostgreSqlDriver`

missing an s in a lot of places.

B4nan:
> `PostgreSqlDriver`
> 
> missing an s in a lot of places.

Where exactly? It is PostgreSQL, so the driver is called `PostgreSqlDriver`. Or what do you mean?

lookfirst:
Oh, you're right! always thought it was with an s! My apologies and bad. =)

B4nan:
Can you please fix that failing test so we can merge this?

`tests/EntityManager.postgre.test.ts:94:40`

j:
@lookfirst ps, mongo does have transaction support for a while now as well as the aggregation framework and some cool aggregation framework rollup functionality in the next version :P

We process millions of pre-aggregated reports and it's super snappy.

lookfirst:
@j Cool man. Glad it is working for you. For me, I wouldn't go back to it. Postgres does everything mongo does, and more.

PR #155, opened 2019-09-22
feat(logging): allow logging full query including params

B4nan:
Added new LoggerNamespace 'query-params'. When setting `debug: true`, it will be also included.

To log only statements instead of full queries, whitelist only 'query' namespace: `debug: ['debug']`.


B4nan:
Here you go @lookfirst :] Will be part of next alpha.

PR #173, opened 2019-09-26
docs: declare correct type of codeblock

lookfirst:
should probably find/replace the whole docs for `bash`....


B4nan:
Last PR changed to sh, now bash in one place. Not sure which one it should be, but we should unify that definitely.

lookfirst:
yes, i matched what was in this file, which is generally the good thing to practice and that is why i noted it in the comment above. ;-)

PR #188, opened 2019-10-05
feat(drivers): add native UUID postgres type

pepakriz:


B4nan:
Thanks! Could you target dev branch instead? I don't wanna publish more features in v2.

B4nan:
Also it would be good to define some existing table with the uuid type and test how entity generator behaves in that case.

pepakriz:
@B4nan can you help me, please? Why CI reports different results than local run? Some strange inconsistencies. I'm not able to debug it.

B4nan:
Do you test it in dev branch? Looks like you don’t. You created new entity which will require adjusting snapshots of schema generator tests. Also I would remove the label mongo entity, as mongo entities requires objectid PK now I guess. Will add more comments in a minute once I got home.

B4nan:
I'd suggest adding the new entity to postgres schema directly, that way you can also test entity generator. Those failing tests are just about outdated snapshots, those things that are reported as missing should be simply added if you wanna have it this way, but I would rather define the table in schema instead.

pepakriz:
Ok, but why tests run successfully on my PC?

B4nan:
Maybe I got it wrong, but this is what I think happened:

- you created a fork based on master
- you added changes (on top of master)
- I asked to changing the base to dev branch
- you changed it just here in github, so the CI is running your changes on top of dev branch (which is correct as that is what will get merged)
- you are running locally still on fork based on master

pepakriz:
1) I've changed target branch to `dev`
2) locally rebased to dev branch (`git rebase --onto upstream/dev HEAD~1`)
3) force pushed PR

I think the problem is somewhere else. Can you try my branch on your machine please?

B4nan:
Ok will try. Btw why `HEAD~1`? That does skip the latest commit, or not? (not that it should matter here)

B4nan:
Cloned your fork, switched to `pg-uuid` branch and tests are failing just like in CI. This has to be just your local copy not being up to date. 

Try to get fresh clone I guess...

pepakriz:
Uff, I've deleted docker volumes, created volumes from scratch and finally I got the same errors as on the CI. I apologize...

tests/entities/Label.ts
```diff
@@ -0,0 +1,19 @@
+import { Entity, IEntity, PrimaryKey, Property } from '../../lib';
+import { v4 } from 'uuid';
+
+@Entity()
+export class Label {
```
B4nan: Let's remove this mongo entity as discussed before. Otherwise LGTM.

tests/__snapshots__/EntityGenerator.test.ts.snap
```diff
@@ -370,6 +370,19 @@ export class FooBaz2 {
 ",
   "import { Entity, PrimaryKey, Property } from 'mikro-orm';
 
+@Entity()
+export class Label2 {
+
+    @PrimaryKey({ type: 'uuid' })
+    uuid: string;
```
B4nan: would be great to have the default value set here as in the original entity

    @PrimaryKey({ type: 'uuid' })
    uuid: string = v4();

(and adding the import as well - there is automatic imports clean up happening, so it should be enough to simply put the import there and it should be cleaned up if not used)

or at least via `default` attribute as a database function:

    @PrimaryKey({ type: 'uuid', default: 'uuid_generate_v4()' })
    uuid: string;

but that should happen only if you have this default used in the schema, if there is no default, the first approach would be better.

tests/entities/Label.ts
pepakriz: Done

tests/__snapshots__/EntityGenerator.test.ts.snap
pepakriz: Good idea! Can we solve it in separated PR, please? It leads to a bigger refactoring

PR #191, opened 2019-10-08
fix(generator): fixed default values and types for nullable properties

pepakriz:


B4nan:
Thanks!

I think it would be better to go for `undefined` rather than `null` though. Both approaches are supported in ORM, but usually one would design the entity like this instead:

```typescript
  @Property()
  age?: number;
```

Maybe it could be configurable, but I would be ok with just the `undefined` version. WDYT?

pepakriz:
I agree with the `undefined` version.

B4nan:
Talking about nullable props, I think it should be possible to get the information about property being possibly `undefined` or `null` via `ts-morph`, so the `nullable: true` inside decorator could be even optional (I would still keep it there when generating entities though)...

pepakriz:
@B4nan yeah, It would be great if can be possible to define nullability by TS type. Because current behavior can lead to inconsistency between decorator definition and TS type. So :+1:

lib/schema/EntityGenerator.ts
```diff
@@ -61,21 +61,21 @@ export class EntityGenerator {
...
+    const definition = this.getPropertyDefinition(column, prop, type, defaultValue);
 
     writer.blankLineIfLastNot();
     writer.writeLine(decorator);
     writer.writeLine(definition);
     writer.blankLine();
   }
 
-  private getPropertyDefinition(prop: string, type: string, defaultValue: any): string {
+  private getPropertyDefinition(column: Column, prop: string, type: string, defaultValue: any): string {
```
B4nan: can you please refactor this a bit so we maintain the 100% branch coverage? this should be enough:

```typescript
const ret = `${prop}${column.nullable ? '?' : ''}: ${type}`;

// string defaults are usually things like SQL functions
if (!defaultValue || typeof defaultValue === 'string') {
  return ret + ';';
}

return `${ret} = ${defaultValue};`;
```

B4nan:
> @B4nan yeah, It would be great if can be possible to define nullability by TS type. Because current behavior can lead to inconsistency between decorator definition and TS type. So 👍

So it was possible indeed, implemented in https://github.com/mikro-orm/mikro-orm/commit/cff0dd4c117db0efa1edcaa4a9403a6a8490a8ab.

Issue #171, opened 2019-09-26
Rework IEntity interface and entity definition


thynson:
I'm invesgating this package, and very satisfied with the promising features.
However, it make me a little uncomfortable that entity class need to extends `IEntity` interface, as it just pollute the namespace.
```typescript

@Entity()
class MyEntity {

}

// isInitialized added to MyEntity as it's a member function of IEntity
// while it's a very common name it may be overrided accidently
interface MyEntity extends IEntity {}

// isInitialized is typed to be always presents, but actually it's not
new MyEntity().isInitialized(); // TypeError: (intermediate value).isInitialized is not a function
``` 
I guess that entity manager need need to track instance through the help of `IEntity`, but this could be in a better way, e.g. using symbol instead.

B4nan:
Yeah I am not very happy with that. How would using symbol help here? The main reason for this is to add that extra public API to all entities (like `init` or `isInitialized`). 

> new MyEntity().isInitialized(); // TypeError: (intermediate value).isInitialized is not a function

This code should work if you started the ORM first (aka the discovery process went thru this entity).

thynson:
I suppose fields prefixed with `__` like `__uuid` is not expected to be used out use code. 
```typescript
// Entity.ts
const _UUIDKey = Symbol(); 

export interface MyInterface {
    [_UUIDKey]: string; // this field can only be access in this file as _UUIDKey is not exported
}

// SomeEntity.ts

export class SomeEntity {

}
export interface MyEntity extends MyInterface {}
```

For some extra public API like `isInitialized()`, I think they shoud be repository method to make entity class clean and  persistence ignorance.

B4nan:
I was also thinking about making all the methods defined as optional, obviously then to use them you would have to add `!`:

```typescript
if (await entity.isInitialized!()) {
  await entity.init!();
}
```

With that approach it should be possible to simply implement the interface (as it would contain only optional fields and methods):

```typescript
export class MyEntity implements IEntity { ... }
```

> I suppose fields prefixed with __ like __uuid is not expected to be used out use code.

Yeah I was going to wrap those into `__internal`, we can have it as a symbol, but as it is prefixed with `__` I do not think we need to care much about collisions... Also I would have to export the symbol as there are multiple places where it is used.

> For some extra public API like isInitialized(), I think they shoud be repository method to make entity class clean and persistence ignorance.

[...]

PR #193, opened 2019-10-09
feat(core): simplify entity definition and rework typings of FilterQuery

### LINKED ISSUE #124: Improve typing of FilterQuery
**Describe the bug**
You can put to ```em::find``` anything as $where condition that completely breaks any compile time checking of queries. It would be nice to keep typing at least for simple cases as for example TypeORM provides.

B4nan:
Now it is no longer needed to merge entities with `IEntity` interface, that was polluting entity's interface with internal methods.
New interfaces `IdentifiedEntity<T>`, `UuidEntity<T>` and `MongoEntity<T>` are introduced, that should be implemented by entities. They
are not adding any new properties or methods, keeping the entity's interface clean.

This also introduces new strictly typed `FilterQuery<T>` implementation based on information provided by those new interfaces (type of primary key
in particular). It is not used in query builder, as there string keys needs to be supported for aliasing to work (`{ 'a.books': 1 }`). 

**BREAKING CHANGES:**
`IEntity` interface no longer has public methods like `toJSON()`, `toObject()` or `init()`. One can use `wrap()` method provided by ORM that
will enhance property type when needed with those methods (`await wrap(book.author).init()`). To keep all methods available on the
entity, you can still use interface merging with `WrappedEntity<T, PK>` that both extends `IEntity<T, PK>` and defines all those methods. 
FilterQuery now does not allow using smart query operators. You can either cast your condition as any or use object syntax instead
(instead of `{ 'age:gte': 18 }` use `{ age: { $gte: 18 } }`).

Closes #124, #171

### DIFF
--- a/docs/custom-driver.md
+++ b/docs/custom-driver.md
@@ -98,12 +98,12 @@ export class MyCustomSchemaHelper extends DatabaseDriver {
   protected readonly platform = new MyCustomPlatform;
 
   // and implement abstract methods
-  find<T extends IEntity>(entityName: string, where: FilterQuery<T>, populate?: string[], orderBy?: Record<string, QueryOrder>, limit?: number, offset?: number): Promise<T[]>;
-  findOne<T extends IEntity>(entityName: string, where: FilterQuery<T> | string, populate: string[]): Promise<T | null>;
-  nativeInsert<T extends IEntityType<T>>(entityName: string, data: EntityData<T>): Promise<QueryResult>;
-  nativeUpdate<T extends IEntity>(entityName: string, where: FilterQuery<IEntity> | IPrimaryKey, data: EntityData<T>): Promise<QueryResult>;
-  nativeDelete<T extends IEntity>(entityName: string, where: FilterQuery<IEntity> | IPrimaryKey): Promise<QueryResult>;
-  count<T extends IEntity>(entityName: string, where: FilterQuery<T>): Promise<number>;
+  find<T extends AnyEntity>(entityName: string, where: FilterQuery<T>, populate?: string[], orderBy?: Record<string, QueryOrder>, limit?: number, offset?: number): Promise<T[]>;
+  findOne<T extends AnyEntity>(entityName: string, where: FilterQuery<T> | string, populate: string[]): Promise<T | null>;
+  nativeInsert<T extends AnyEntityType<T>>(entityName: string, data: EntityData<T>): Promise<QueryResult>;
+  nativeUpdate<T extends AnyEntity>(entityName: string, where: FilterQuery<T> | IPrimaryKey, data: EntityData<T>): Promise<QueryResult>;
+  nativeDelete<T extends AnyEntity>(entityName: string, where: FilterQuery<T> | IPrimaryKey): Promise<QueryResult>;
+  count<T extends AnyEntity>(entityName: string, where: FilterQuery<T>): Promise<number>;
 
 }
 ```
--- a/docs/defining-entities.md
+++ b/docs/defining-entities.md
@@ -42,13 +42,13 @@ export class Book {
 
 }
 
-export interface Book extends IEntity<string> { }
+export interface Book extends AnyEntity<string> { }
 ```
 
-You will need to extend Book's interface with `IEntity`. The interface represents internal 
+You will need to extend Book's interface with `AnyEntity`. The interface represents internal 
 methods added to your entity's prototype via `@Entity` decorator.
 
-> `IEntity` is generic interface, its type parameter depends on data type of normalized primary
+> `AnyEntity` is generic interface, its type parameter depends on data type of normalized primary
 > key produced by used driver. SQL drivers usually use `number` and Mongo driver uses `string`.
 > This type default to union type `number | string`. Keep in mind that you have to worry about 
 > this only when you define your primary key as `_id` instead of `id`.
@@ -107,7 +107,7 @@ export class Author {
 
 }
 
-export interface Author extends IEntity { }
+export interface Author extends AnyEntity { }
 ```
 
 More information about modelling relationships can be found on [modelling relationships page](relationships.md).
--- a/docs/entity-constructors.md
+++ b/docs/entity-constructors.md
@@ -36,7 +36,7 @@ export class Book {
 
 }
 
-export interface Book extends IEntity { }
+export interface Book extends AnyEntity { }
 ```
 
 [&larr; Back to table of contents](index.md#table-of-contents)
--- a/docs/entity-helper.md
+++ b/docs/entity-helper.md
@@ -3,11 +3,11 @@
 
 # EntityHelper and Decorated Entities
 
-## Updating Entity Values with IEntity.assign()
+## Updating Entity Values with Entity.assign()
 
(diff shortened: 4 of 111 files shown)


B4nan:
Example of entity definition in SQL drivers (integer AI primary key):

```typescript
@Entity()
export class FooBar implements IdentifiedEntity<FooBar> {

  @PrimaryKey()
  id: number;

  @Property()
  name: string;

}
```

For mongo it is a bit more complicated as now both `_id` and `id` will need to be defined:

```typescript
@Entity()
export class FooBar implements MongoEntity<FooBar> {

  @PrimaryKey()
  _id: ObjectId;

  @SerializedPrimaryKey()
  id: string;

  @Property()
  name: string;

}
```

B4nan:
Not yet fully polished, but would love to hear your feedback @lookfirst, @vimmerru, @pepakriz.

Available for testing via `dev` tag (`3.0.0-dev.1`).

lookfirst:
Damn, no wonder you've been so quiet recently. Good work!

pepakriz:
I have just a little experience with the orm code, but it looks very impressive!

thynson:
Wow, you made it so quickly. Excellent work!

lookfirst:
What if I want an `id: string`?

B4nan:
> What if I want an id: string?

There is still `IEntity` interface for that, where you specify what property is PK and its type will be used. There is also `PrimaryKeyType` symbol that can be used to make sure the type is correctly set in `FilterQuery`, but it should not be needed (at least with TS 3.6 it looks like that, I added that symbol when developing it on TS 3.5 but it might work there as well). 

```typescript
@Entity()
export class FooBar implements IEntity<Test2, 'id'> {

  @PrimaryKey()
  id: string;

}
```

Btw just found out there are still some issues with `FilterQuery`, it allows to put anything into property on the top level (nested conditions are correctly checked though):

```typescript
const ok: FilterQuery<Author2> = { books: { author: { born: new Date() } } };
const fail: FilterQuery<Author2> = { books: { undef: { born: new Date() } } };
const shouldFail: FilterQuery<Author2> = { undef: { author: { born: new Date() } } };
```

lookfirst:
Yea, don't worry about supporting older versions of TS. Happy to stay current with that. Running 3.6.2 and may even switch to the 3.7 beta soon to get all the awesome `?` improvements.

Re: IEntity, I must admit that it feels awkward to have to switch between two different interfaces and formats just because I'm using a string or number primary key.

As always, excellent work. Super appreciate.

B4nan:
> Re: IEntity, I must admit that it feels awkward to have to switch between two different interfaces and formats just because I'm using a string or number primary key.

You can use `IEntity` for all entities if you want. Those extra interfaces are there just to simplify things, usually one will use the same type of PKs in all entities imho. 

```typescript
export type IEntity<T = any, PK extends keyof T = keyof T> = { [K in PK]: T[K] };
export type WrappedEntity<T, PK extends keyof T> = IWrappedEntity<T, PK> & IEntity<T, PK>;
export type IdentifiedEntity<T extends { id: number }> = IEntity<T, 'id'>;
export type UuidEntity<T extends { uuid: string }> = IEntity<T, 'uuid'>;
export type MongoEntity<T extends { _id: IPrimaryKey; id: string }> = IEntity<T, 'id' | '_id'>;
```

lookfirst:
Maybe `IEntity` should be `AnyEntity` and `IdentifiedEntity` should be `NumericEntity`

lookfirst:
Or NumberEntity

B4nan:
I picked `IdentifiedEntity` because of the requirement for `id` property... I think we could support both number and string there, important is to know the property name, its type can be inferred. 

Not sure about `AnyEntity`, lets wait for other opinions, I am not against renaming it, I know the name is too generic (but - generic as the interface itself). It is used to define `any` entity, so maybe it could be good name.

thynson:
Why not 
```typescript
type IdentifiedEntity<T extends { id: R}, R=number> = IEntity<T, 'id'>;
```
so user has the ability to specify the type of primary key

B4nan:
That's pretty much what I just said in last comment :] Will make this happen, was thinking about it anyway... 

I even think we could just have (other types of PK than string and number are not supported, mongo requires _id property so no need to take ObjectId into account here):

```typescript
type IdentifiedEntity<T extends { id: number | string }> = IEntity<T, 'id'>;
```

lookfirst:
I am not a fan of prefixing interfaces with `I`, it is a cheapsauce naming technique. interfaces are the things you use, not just look at.

B4nan:
> I am not a fan of prefixing interfaces with I, it is a cheapsauce naming technique. interfaces are the things you use, not just look at.

Me neither, but... naming is hard, you know :D 

Let's call it `AnyEntity` unless someone proposes something better.

lookfirst:
Change...

```
export type IEntity<T = any, PK extends keyof T = keyof T> = { [K in PK]: T[K] };
export type IdentifiedEntity<T extends { id: number }> = IEntity<T, 'id'>;
```

into...

```
export type AnyEntity<T = any, PK extends keyof T = keyof T> = { [K in PK]: T[K] };
export type IdEntity<T extends { id: number | string }> = AnyEntity<T, 'id'>;
```

Seems sane and now is a good time given the breaking changes...

B4nan:
Ok done with the `FilterQuery` issue hopefully, also renamed `IEntity` and `IdentifiedEntity` to `AnyEntity` and `IdEntity` plus `IdEntity` now allows using both number and string PK. 

Published as `3.0.0-dev.3`.

B4nan:
Found one more small issue with `FilterQuery`: using operators with properties of `Date` type are failing (e.g. `{ born: { $gte: new Date('...') } }`), but that should be hopefully easy to fix.

pepakriz:
I'm not sure about `FilterQuery` typing. The current behavior is unfriendly with private properties for example. I've been thinking about typing and I'd like to have cli command which generates TS types. This principle is used by prisma and it has one great advantage - you're not limited by TS.  Generated types are simple as possible without TS magic which can break something every TS version. So I'd completely prefer this way.

B4nan:
Well that was first approach I was thinking about, but I don't like the idea of generating them only via CLI, it would have to be somehow automated (which could work with enabled caching probably). Still current approach looks good to me, so I am not very keen to rewrite that... Will test it for couple more days in the wild, but so far it works good. 

> The current behavior is unfriendly with private properties for example.

Not sure if that is important, as private props are not really supported by the ORM (there is no mechanism that would call your setters instead of just assigning the value internally). Any other issue with current approach? 

> Generated types are simple as possible without TS magic

Still they would be pretty complex as well I think, although definitely less generic/more specific. I did not pay much attention to prisma yet, but from what I saw, they have more strict rules on the query type, e.g. they do not allow query by PK (`{ author: { id: 1 } }` instead of `{ author: 1 }` ). 

Let's see what others are thinking.

lookfirst:
Example project updated, all tests pass. No complaints. =) After midnight here, so I'll do my main project tomorrow morning.

B4nan:
Merging this now and closing both connected issues, but feel free to continue with discussion. For bugs please open new issues.

PR #197, opened 2019-10-12
refactor: em `transactional` method return value & improve types

pepakriz:


B4nan:
So this change is just about returning what was returned from transactional callback, preserving the type? 

Sounds good, I just dont like this being called a fix :) lets call it refactoring

pepakriz:
@B4nan how you wish. changed

PR #198, opened 2019-10-12
refactor: em and connection generic types

pepakriz:


lib/EntityManager.ts
```diff
@@ -5,28 +5,28 @@ import { AbstractSqlDriver, IDatabaseDriver } from './drivers';
...
-  getDriver<D extends IDatabaseDriver = IDatabaseDriver>(): D {
-    return this.driver as D;
+  getDriver(): D {
+    return this.driver;
   }
 
-  getConnection<C extends Connection = Connection>(type?: 'read' | 'write'): C {
-    return this.driver.getConnection(type) as C;
+  getConnection(type?: 'read' | 'write'): ReturnType<D['getConnection']> {
+    return this.driver.getConnection(type) as any;
```
pepakriz: Probably a bug in TS. I'll check it again later. But outer usage works correctly.

lib/EntityManager.ts
```diff
@@ -48,7 +48,7 @@ export class EntityManager {
 
   createQueryBuilder(entityName: EntityName<AnyEntity>, alias?: string, type?: 'read' | 'write'): QueryBuilder {
     entityName = Utils.className(entityName);
-    return new QueryBuilder(entityName, this.metadata, this.driver as AbstractSqlDriver, this.transactionContext, alias, type);
+    return new QueryBuilder(entityName, this.metadata, this.driver as any as AbstractSqlDriver, this.transactionContext, alias, type);
```
pepakriz: I'll try to find a better solution.

lib/MikroORM.ts
```diff
@@ -66,11 +66,11 @@ export class MikroORM {
   }
 
   getSchemaGenerator(): SchemaGenerator {
-    return new SchemaGenerator(this.driver as AbstractSqlDriver, this.metadata);
+    return new SchemaGenerator(this.driver as any as AbstractSqlDriver, this.metadata);
```
pepakriz: I'll try to find a better solution.

lib/MikroORM.ts
```diff
@@ -66,11 +66,11 @@ export class MikroORM {
...
   }
 
   getEntityGenerator(): EntityGenerator {
-    return new EntityGenerator(this.driver as AbstractSqlDriver, this.config);
+    return new EntityGenerator(this.driver as any as AbstractSqlDriver, this.config);
```
pepakriz: I'll try to find a better solution.

lib/utils/Configuration.ts
```diff
@@ -163,13 +163,17 @@ export class Configuration {
...
 
-  private initDriver(): IDatabaseDriver {
+  private initDriver(): D {
     if (!this.options.driver) {
       const driver = Configuration.PLATFORMS[this.options.type];
       this.options.driver = require('../drivers/' + driver)[driver];
     }
 
-    return new this.options.driver!(this);
+    if (!this.options.driver) {
```
pepakriz: Because TS don't understand of line 169

pepakriz: Or we can preserve the original solution.

pepakriz: reverted

lib/EntityManager.ts
```diff
@@ -5,28 +5,29 @@ import { AbstractSqlDriver, IDatabaseDriver } from './drivers';
...
+import { Transaction } from './connections';
+import { MySqlDriver } from './drivers/MySqlDriver';
 
-export class EntityManager {
+export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
 
   private readonly validator = new EntityValidator(this.config.get('strict'));
   private readonly repositoryMap: Record<string, EntityRepository<AnyEntity>> = {};
-  private readonly entityLoader = new EntityLoader(this);
+  private readonly entityLoader: EntityLoader = new EntityLoader(this);
```
B4nan: why is this needed?

lib/EntityManager.ts
```diff
@@ -5,28 +5,29 @@ import { AbstractSqlDriver, IDatabaseDriver } from './drivers';
 import { EntityData, EntityMetadata, EntityName, AnyEntity, IPrimaryKey, FilterQuery, Primary, Dictionary } from './types';
 import { QueryBuilder, QueryOrderMap, SmartQueryHelper } from './query';
 import { MetadataStorage } from './metadata';
-import { Connection, Transaction } from './connections';
+import { Transaction } from './connections';
+import { MySqlDriver } from './drivers/MySqlDriver';
```
B4nan: forgotten unused import

lib/EntityManager.ts
B4nan: can we cast it to `ReturnType<D['getConnection']>` instead?

lib/EntityManager.ts
```diff
@@ -48,7 +49,13 @@ export class EntityManager {
 
   createQueryBuilder(entityName: EntityName<AnyEntity>, alias?: string, type?: 'read' | 'write'): QueryBuilder {
     entityName = Utils.className(entityName);
-    return new QueryBuilder(entityName, this.metadata, this.driver as AbstractSqlDriver, this.transactionContext, alias, type);
+
+    const driver = this.driver as IDatabaseDriver<any>;
```
B4nan: move this line above (so the blank line is before the `if` block). 

also why `IDatabaseDriver<any>`? if its just about TS complaining about the instanceof, you can cast to unknown or object.

lib/MikroORM.ts
```diff
@@ -7,22 +7,22 @@ import { Configuration, Logger, Options } from './utils';
...
-    const orm = new MikroORM(options);
+  static async init<D extends IDatabaseDriver = IDatabaseDriver>(options: Options<D> | Configuration<D>): Promise<MikroORM<D>> {
+    const orm = new MikroORM<D>(options);
     const driver = await orm.connect();
 
     try {
       const discovery = new MetadataDiscovery(MetadataStorage.init(), orm.driver.getPlatform(), orm.config, orm.logger);
       orm.metadata = await discovery.discover();
-      orm.em = new EntityManager(orm.config, driver, orm.metadata);
+      orm.em = new EntityManager<D>(orm.config, driver, orm.metadata);
```
B4nan: here the `<D>` would be inferred from `driver` variable type, right?

lib/MikroORM.ts
```diff
@@ -66,11 +66,21 @@ export class MikroORM {
   }
 
   getSchemaGenerator(): SchemaGenerator {
-    return new SchemaGenerator(this.driver as AbstractSqlDriver, this.metadata);
+    const driver = this.driver as IDatabaseDriver<any>;
```
B4nan: again, please maintain blank line before block statements. 

but here I would rather drop `driver` variable and do this instead:

    if (!(this.driver as unknown instanceof AbstractSqlDriver)) {
      throw new Error('Not supported by given driver');
    }

    return new SchemaGenerator(this.driver, this.metadata);

lib/MikroORM.ts
```diff
@@ -66,11 +66,21 @@ export class MikroORM {
...
+    if (!(driver instanceof AbstractSqlDriver)) {
+      throw new Error('Not supported by given driver');
+    }
+
+    return new SchemaGenerator(driver, this.metadata);
   }
 
   getEntityGenerator(): EntityGenerator {
-    return new EntityGenerator(this.driver as AbstractSqlDriver, this.config);
+    const driver = this.driver as IDatabaseDriver<any>;
```
B4nan: same here

tests/EntityGenerator.test.ts
```diff
@@ -40,4 +40,12 @@ describe('EntityGenerator', () => {
     await orm.close(true);
   });
 
+  test('not supported [mongodb]', async () => {
+    const orm = await initORMMongo();
+
```
B4nan: i would remove those blank lines here

tests/EntityGenerator.test.ts
```diff
@@ -40,4 +40,12 @@ describe('EntityGenerator', () => {
...
+    expect(() => orm.getEntityGenerator()).toThrowError('Not supported by given driver');
+
```
B4nan: and here

tests/QueryBuilder.test.ts
```diff
@@ -1015,6 +1015,14 @@ describe('QueryBuilder', () => {
     expect(qb.getAliasForEntity(Author2.name, node)).toBeUndefined();
   });
 
+  test('not supported [mongodb]', async () => {
+    const orm = await initORMMongo();
+
```
B4nan: again, please remove those blank lines here

tests/QueryBuilder.test.ts
```diff
@@ -1015,6 +1015,14 @@ describe('QueryBuilder', () => {
...
+    expect(() => orm.em.createQueryBuilder(Author2, 'a')).toThrowError('Not supported by given driver');
+
```
B4nan: and here

tests/SchemaGenerator.test.ts
```diff
@@ -255,4 +255,12 @@ describe('SchemaGenerator', () => {
     await orm.close(true);
   });
 
+  test('not supported [mongodb]', async () => {
+    const orm = await initORMMongo();
+
```
B4nan: and again here

tests/SchemaGenerator.test.ts
```diff
@@ -255,4 +255,12 @@ describe('SchemaGenerator', () => {
...
+    expect(() => orm.getSchemaGenerator()).toThrowError('Not supported by given driver');
+
```
B4nan: and here

lib/EntityManager.ts
pepakriz: `error TS7022: 'entityLoader' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.`

lib/EntityManager.ts
pepakriz: Ok, I've replaced it by `object` type

lib/MikroORM.ts
pepakriz: right

lib/EntityManager.ts
pepakriz: Wow, it works! Truly magic! :smiley:

tests/SchemaGenerator.test.ts
pepakriz: I've preserved this new line to be consistent with other tests.

tests/QueryBuilder.test.ts
pepakriz: I've preserved this new line to be consistent with other tests.

PR #200, opened 2019-10-14
Add support for bundling with Webpack

jsprw:
A few patches to allow to bundle all dependencies with webpack with the following setup:

**mikro orm config** (entitiesDirs** will be ignored)
```
await MikroORM.init({
    ...
    entities: [Entity1, Entity2, ...],
    ...
})
```

**webpack.config.js**
```
const optionalModules = [
    ...Object.keys(require("knex/package.json").browser),
    ...Object.keys(require("mikro-orm/package.json").peerDependencies)
];
...
module.exports = {
    ...
    target: "node",
    ...
    plugins: [
        new webpack.EnvironmentPlugin({
            WEBPACK: true //this will replace all process.env.WEBPACK with the boolean true
        }),
        new webpack.IgnorePlugin({
            checkResource: rsrc => {
                if (optionalModules.indexOf(rsrc.split("/")[0]) !== -1) {
                    try {
                        require.resolve(rsrc);
                        return false;
                    } catch (e) {
                        return true;
                    }
                }
                return false;
            }
        }),
        ...
    ],
    ...
}   
```

Documentation will follow once merged. (How could this be tested?)


lib/entity/EntityFactory.ts
```diff
@@ -54,8 +54,13 @@ export class EntityFactory {
   }
 
   private createEntity<T extends AnyEntity<T>>(data: EntityData<T>, meta: EntityMetadata<T>): T {
-    const path = Utils.absolutePath(meta.path, this.config.get('baseDir'));
-    const Entity = require(path)[meta.name];
+    let Entity;
+    if (process.env.WEBPACK) {
```
B4nan: please add blank line above if statement

lib/metadata/MetadataDiscovery.ts
```diff
@@ -27,14 +27,22 @@ export class MetadataDiscovery {
     const startTime = Date.now();
     this.logger.log('discovery', `ORM entity discovery started`);
     this.discovered.length = 0;
-    const tsNode = this.config.get('tsNode') || process.argv[0].endsWith('ts-node') || process.argv.slice(1).some(arg => arg.includes('ts-node')) || !!require.extensions['.ts'];
 
-    if (this.config.get('entities').length > 0) {
+    if (process.env.WEBPACK) {
```
B4nan: this can be simplified. add if check for `process.env.WEBPACK && this.config.get('entities').length === 0` and throw. 
then keep the rest as it was - `entities` array is always used if provided. 

also would be good to add at least basic docs (1-2 sentences and example) and link to the docs from the error.

lib/metadata/TypeScriptMetadataProvider.ts
```diff
@@ -26,6 +26,9 @@ export class TypeScriptMetadataProvider extends MetadataProvider {
       } else if (prop.entity) {
         prop.type = Utils.className(prop.entity());
       } else if (!prop.type) {
+        if (process.env.WEBPACK) {
+          throw new Error(`Webpack bundles require the type property to be set in Column decorators. (${meta.className}.${name})`);
```
B4nan: `type` or `entity`, both are fine. also it is not `Column` decorator, but `@Property` decorator (keep the `@` sign too).

tests/Utils.test.ts
```diff
@@ -181,11 +181,11 @@ describe('Utils', () => {
   });
 
   test('absolutePath', () => {
-    expect(Utils.absolutePath('./test')).toBe(process.cwd() + '/test');
-    expect(Utils.absolutePath('test')).toBe(process.cwd() + '/test');
-    expect(Utils.absolutePath(process.cwd() + '/tests/')).toBe(process.cwd() + '/tests');
-    expect(Utils.absolutePath('./tests/cli')).toBe(process.cwd() + '/tests/cli');
-    expect(Utils.absolutePath('')).toBe(process.cwd());
+    expect(Utils.absolutePath('./test')).toBe(Utils.normalizePath(process.cwd() + '/test'));
```
B4nan: would be good to rebase to current dev, those changes are already merged and they should not be part of the diff at all.

B4nan:
Also please in the docs example keep the same codestyle: use single quotes, use `includes` instead of `indexOf`, keep blank lines around block statements (e.g. if), as well as blank line before return statement (if the block is longer than 2 lines).

lib/metadata/MetadataDiscovery.ts
jsprw: ` !!require.extensions['.ts'];`

This part is a problem in Webpack, added it to the elseif part.

lib/metadata/TypeScriptMetadataProvider.ts
```diff
@@ -26,6 +26,9 @@ export class TypeScriptMetadataProvider extends MetadataProvider {
       } else if (prop.entity) {
         prop.type = Utils.className(prop.entity());
       } else if (!prop.type) {
+        if (process.env.WEBPACK) {
+          throw new Error(`Webpack bundles requires the any of the properties type or entity to be set in @Property decorators. (${meta.className}.${name})`);
```
B4nan:
```suggestion
          throw new Error(`Webpack bundling requires either 'type' or 'entity' attributes to be set in @Property decorators. (${meta.className}.${name})`);
```

lib/metadata/MetadataDiscovery.ts
```diff
@@ -27,11 +27,14 @@ export class MetadataDiscovery {
...
-    const tsNode = this.config.get('tsNode') || process.argv[0].endsWith('ts-node') || process.argv.slice(1).some(arg => arg.includes('ts-node')) || !!require.extensions['.ts'];
+
+    if (process.env.WEBPACK && this.config.get('entities').length === 0) {
+      throw new Error("Webpack bundles only supports pre-defined entities. Please use the 'entities' option. See the documentation for more information.");
+    }
 
     if (this.config.get('entities').length > 0) {
       await Utils.runSerial(this.config.get('entities'), entity => this.discoverEntity(entity));
-    } else if (tsNode) {
+    } else if (this.config.get('tsNode') || process.argv[0].endsWith('ts-node') || process.argv.slice(1).some(arg => arg.includes('ts-node')) || !!require.extensions['.ts']) {
```
B4nan: let's move this to `Utils.detectTsNode()`

lib/entity/EntityFactory.ts
```diff
@@ -54,8 +54,14 @@ export class EntityFactory {
   }
 
   private createEntity<T extends AnyEntity<T>>(data: EntityData<T>, meta: EntityMetadata<T>): T {
-    const path = Utils.absolutePath(meta.path, this.config.get('baseDir'));
-    const Entity = require(path)[meta.name];
+    let Entity;
+
+    if (process.env.WEBPACK) {
+      Entity = this.config.get('entities').find((f) => (f as Function).name === meta.name);
```
B4nan:
```suggestion
      Entity = this.config.get('entities').find((f: Function) => f.name === meta.name);
```

lib/entity/EntityFactory.ts
```diff
@@ -54,8 +54,14 @@ export class EntityFactory {
   }
 
   private createEntity<T extends AnyEntity<T>>(data: EntityData<T>, meta: EntityMetadata<T>): T {
-    const path = Utils.absolutePath(meta.path, this.config.get('baseDir'));
-    const Entity = require(path)[meta.name];
+    let Entity;
```
B4nan: this could be typed to `Constructor<T>` i think

lib/metadata/MetadataValidator.ts
```diff
@@ -41,7 +41,6 @@ export class MetadataValidator {
       throw ValidationError.duplicateEntityDiscovered(duplicates);
     }
 
-    // validate base entities
```
B4nan: why did you remove it?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
B4nan: drop this file, and add `.vscode` to `.gitignore`

tests/Webpack.test.ts
```diff
@@ -0,0 +1,61 @@
+import { BookWp, AuthorWp } from './entities-webpack';
+import { BookWpI, AuthorWpI } from './entities-webpack-invalid';
+import { MikroORM } from '../lib';
+import { wipeDatabase, BASE_DIR } from './bootstrap';
+import { MetadataDiscovery } from '../lib/metadata';
+
+describe('EntityFactory', () => {
```
B4nan: fix the test case name

.gitignore
```diff
@@ -8,3 +8,4 @@
 /.coveralls.yml
 !.gitkeep
 /mikro_orm_test.db
+/.vscode
```
B4nan: please keep the new line at the end of file

tests/Webpack.test.ts
```diff
@@ -32,16 +32,49 @@ describe('Webpack', () => {
...
+
+    await new MetadataDiscovery(
+      orm.getMetadata(),
+      orm.em.getDriver().getPlatform(),
+      orm.config,
+      orm.config.getLogger(),
+    ).discover();
+    const authors = await orm.em.getRepository(AuthorWp).findAll({ populate: true });
+    expect(authors.length).toBe(1);
+    expect(authors[0].books.isInitialized).toBeTruthy();
```
B4nan: `isInitialized` is method, it will be always truthy.

tests/Webpack.test.ts
```diff
@@ -32,16 +32,49 @@ describe('Webpack', () => {
...
+    await orm.close(true);
+    done();
```
B4nan: there is no need to use `done` function (generally I am not very happy with the code style of your tests, but I can refactor that myself later, important thing for me is the coverage...)

jsprw: Yeah understandable, having some trouble with properly factoring the tests and properly covering each file/line as I don't have much experience with writing unit tests.. 

Sorry for taking so long to properly write 4(!) tests..

The commits should now restore coverage to 100%.

tests/Webpack.test.ts
```diff
@@ -121,7 +119,6 @@ describe('Webpack', () => {
       expect(e.message).toBe(
         "Webpack bundles only supports pre-defined entities. Please use the 'entities' option. See the documentation for more information.",
       );
-      done();
```
B4nan: actually here it is needed, ideally you should rewrite this to try/catch with awaiting the init(), then it will run fine without the `done` callback. btw line 60 of EntityFactory is still not covered.

PR #202, opened 2019-10-15
Webpack docs

jsprw:
Added webpack section to deployment documentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 /temp
 /tests/generated-entities
 /.idea
+/.vscode
 /yarn-error.log
 /.coveralls.yml
 !.gitkeep
--- a/README.md
+++ b/README.md
@@ -126,16 +126,20 @@ app.use((req, res, next) => {
 
 More info about `RequestContext` is described [here](https://mikro-orm.io/identity-map/#request-context).
 
-Now you can start defining your entities (in one of the `entitiesDirs` folders):
+Now you can start defining your entities (in one of the `entitiesDirs` folders). This is how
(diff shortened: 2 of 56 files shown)


B4nan:
Looking good, I just don't like this sentence very much `If you want to deploy the least amount of files possible, you should use Webpack`. Personally I see this webpack solution viable only for angular universal or electron apps where the result is actually ran in browser based environment. Generally I would not suggest to use webpack (or any other build tool made for frontend) when its needed. 

Also please use `MikroORM` instead of `Mikro-ORM`. `mikro-orm` is the package name, `MikroORM` is the library name. I don't want to promote that mixture of both :]

jsprw:
> Personally I see this webpack solution viable only for angular universal or electron apps where the result is actually ran in browser based environment.

[Angular universal](https://angular.io/guide/universal) is not run in browser based environments. It's run on a node server and produces data to be sent to the browser environment. 

> Generally I would not suggest to use webpack (or any other build tool made for frontend) when its not needed.

In my opinion, it works well as you only need to install node on the server and no other node modules. Furthermore the plugins are nice to have. Anyways, I have removed the sentence.

> Also please use MikroORM instead of Mikro-ORM. mikro-orm is the package name, MikroORM is the library name. I don't want to promote that mixture of both :]

Fixed.

B4nan:
> Angular universal is not run in browser based environments. It's run on a node server and produces data to be sent to the browser environment.

Yeah I know how that works, I am using that myself for a long time already. But my point is that it is forcing you to use webpack, or not? 

Never mind, LGTM, thanks!

PR #209, opened 2019-10-19
feat(core): add support for migrations via umzug

B4nan:
Basic support for migrations is now provided via umzug. `Migrator` class is available that can be used to generate new migration file with current schema diff from `SchemaGenerator`.

```typescript
const migrator = orm.getMigrator();
await migrator.createMigration(); // creates file Migration20191019195930.ts
await migrator.up(); // runs migrations up to the latest
await migrator.up('up-to-name'); // runs migrations up to given version
await migrator.down('down-to-name'); // runs migrations down to given version
```

Migrations are classes that extend `Migration` abstract class:

```typescript
export class Migration20191019195930 extends Migration {

  async up(): Promise<void> {
    this.addSql('select 1 + 1');
  }

  isTransactional(): boolean {
    return false;
  }

}
```

`down` method by default throws. also `isTransactional(): boolean` can be implemented to disable transactions for given migration. `Configuration` object and driver instance are available.

Available for testing in `3.0.0-alpha.30`.


B4nan:
WDYT @lookfirst @pepakriz?

lookfirst:
Nice option for sure. I looked but didn't see integration with the cli?

Also... allOrNothing is todo? If it is what I think it is, I'd default true as you don't want a migration to fail and leave things in a bad state.

B4nan:
CLI will follow probably today. All or nothing is inspired by doctrine and there they have it defaults to false. It will wrap all migrations inside one parent transaction.

https://www.doctrine-project.org/projects/doctrine-migrations/en/latest/reference/configuration.html#configuration

lookfirst:
Funny how the default is false, but the example is true.

image

lookfirst:
Looks like they default to false because not all databases have transactions, not because it is the right thing to do.

B4nan:
CLI support added in alpha-31

B4nan:
Removed the `allOrNothing` for now as I am not sure how to implement that on top of umzug. Maybe I could just manually run `begin/commit/rollback` and force knex to use single connection (no pooling), but I think this could be a good start and we can improve on the go.

lookfirst:
hmmm... yes, one transaction please. no need for pooling for schema changes. i don't want to do schema changes without a transaction, that is suicide.

B4nan:
I must have been blind, it was actually pretty easily :] Already got it implemented, just need to polish some tests. Will enable it by default as suggested. When transactions are disabled via the `transactional` config toggle, it will be ignored.

lookfirst:
All those years of working in PHP has ruined your eyesight. ;-)

B4nan:
Implemented in https://github.com/mikro-orm/mikro-orm/commit/acb1d24f6fa4429e543dd0e7159bb61694b5f42f, also improved success messages and other small things.

PR #211, opened 2019-10-22
test: 'fix' memory leak in graceful-fs

B4nan:
`graceful-fs` and `jest` does not like each other, so this removes the `graceful-fs` implementation before running tests and adds it back afterwards.

https://github.com/facebook/jest/issues/6399
https://github.com/facebook/jest/issues/6814
https://github.com/isaacs/node-graceful-fs/issues/102

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,8 +36,10 @@
   "homepage": "https://mikro-orm.io",
   "scripts": {
     "build": "rimraf dist && tsc",
-    "test": "jest --runInBand",
-    "coverage": "rimraf temp tests/generated-entities && jest --runInBand --coverage",
+    "pretest": "node tests/pre-test",
+    "test": "node --expose-gc ./node_modules/.bin/jest --runInBand --logHeapUsage",
+    "posttest": "node tests/post-test",
+    "coverage": "rimraf temp tests/generated-entities && yarn test --coverage",
     "coveralls": "cat ./coverage/lcov.info | coveralls",
     "lint": "tslint -p ."
   },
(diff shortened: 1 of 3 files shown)


lookfirst:
wow, ugly.

B4nan:
Yeah, ugly. Hopefully this can be removed soon, but currently CI builds started to fail randomly because I had too much test cases. And it reached 1g heap size, without this hack...

Its sad that but I will rather do this than manually rerun failed builds and hope it will fit the limit. Also adding more test files would make this more and more often...

lookfirst:
Ha! I guess it is possible to do too much unit testing... =)

Yea... shrug... just a test hack, no biggie.

package.json
```diff
@@ -36,8 +36,10 @@
   "homepage": "https://mikro-orm.io",
   "scripts": {
     "build": "rimraf dist && tsc",
-    "test": "jest --runInBand",
-    "coverage": "rimraf temp tests/generated-entities && jest --runInBand --coverage",
+    "pretest": "node tests/pre-test",
```
seyfer: is any performance test before and after available? I have added this to my project - see no difference in heap size usage

B4nan: travis builds were randomly reaching 1g memory limit, with this change the max heap size is quite stable around 300m. you can take a look at the travis builds before and after:

before: https://travis-ci.com/mikro-orm/mikro-orm/jobs/248182459
after: https://travis-ci.com/mikro-orm/mikro-orm/jobs/248297817

unfortunately i do not have the logging of heap size there before this actual change.

B4nan: maybe you have actual leak in your app, or it is caused by another dependency than graceful-fs...

PR #233, opened 2019-11-11
[docs] tune the changeset docs...

lookfirst:
simplify the english a lot...

(note: this is wip, help me add an example of the constructors)

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,9 +57,36 @@ await em.flush();
 
 MikroORM allows you to implement your domain/business logic directly in the entities. 
 To maintain always valid entities, you can use constructors to mark required properties. 
+Let's define the `User` entity used in previous example:
 
 ```typescript
-Example using constructors that were just mentioned...
+@Entity()
+export class User implements IdEntity<User> {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property()
+  name!: string;
+
+  @OneToOne()
+  address?: Address;
+
+  @ManyToMany()
+  cars = new Collection<Car>(this);
+
+  constructor(name: string) {
+    this.name = name;
+  }
+
+}
+```
+
+Now to create new instance of the `User` entity, we are forced to provide the `name`:
+
+```typescript
+const user = new User('John Doe'); // name is required to create new user instance
+user.address = new Address('10 Downing Street'); // address is optional
 ```
 
 Once your entities are loaded, make a number of synchronous actions on your entities,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,18 +53,19 @@ user.cars.add(car);
...
+MikroORM allows you to implement your domain/business logic directly in the entities. 
+To maintain always valid entities, you can use constructors to mark required properties. 
 
-Once your entities are loaded, you can just work with them and forget about persistence. 
-As you do not have to care about persistence, most of entity interactions can be synchronous. 
-When you have done all changes, you call `em.flush()`. It will trigger computing of change 
-sets. Only entities that were changed will generate database queries, if there are no changes, 
-no transaction will be started. 
+```typescript
+Example using constructors that were just mentioned...
```
B4nan: Not sure how to make good example for this, as it requires the entity definition to be clear, which can be a bit too verbose for this section... 

https://mikro-orm.github.io/docs/v3/entity-constructors/

(looking at that page, I should add `?` and `!` to props to examples too, especially in that particular one)

lookfirst: Then maybe remove the part about the constructors. I don't like talking about something, especially on the main readme, without at least showing a basic example.

B4nan: Well I think it's one of the big differences when it comes to comparison with other ORMs so I'd say it deserves the space for example rather than removing it... :] 

Let's define `User` entity used in those examples, it can be as simple as:

```typescript
@Entity()
export class User implements IdEntity<User> {

  @PrimaryKey()
  id!: number;

  @Property()
  name!: string;

  @OneToOne()
  address?: Address;

  @ManyToMany()
  cars = new Collection<Car>(this);

  constructor(name: string) {
    this.name = name;
  }

}

const user = new User('John Doe'); // name is required to create new user instance
user.address = new Address('10 Downing Street'); // address is optional
```

Thinking about this, I guess I should also add validation for missing ctor params when creating the entity via `em.create()`...

B4nan: Simple entity definition (almost no place where you have to repeat yourself) is also one of core features I'd like to promote, so example like this should be definitely at the top too...

B4nan: I guess we should also change the street so it is different from the following example that causes the update address query (changing the street number could be enough).

lookfirst: looks good. feel free to edit this PR and add that. =)

B4nan: done, i guess we can merge it now?

lookfirst:
Looks good to me!

PR #281, opened 2019-12-22
Update web pack documentation

jsprw:
This option must be disabled to prevent the metadata analysers from running else web pack builds will fail on start.

### DIFF
--- a/docs/docs/deployment.md
+++ b/docs/docs/deployment.md
@@ -119,6 +119,7 @@ await MikroORM.init({
     // ...
     entities: await getEntities(),
     cache: { enabled: false },
+    discovery: { alwaysAnalyseProperties: false } //since rc-4
     // ...
 });
 


B4nan:
If you set `WEBPACK = true`, then this will become default. 

See: https://github.com/mikro-orm/mikro-orm/blob/master/lib/MikroORM.ts#L45

jsprw:
> If you set `WEBPACK = true`, then this will become default.


Hi, that will break the local development without Webpack: https://github.com/mikro-orm/mikro-orm/blob/1093d88b989e5dfec2b8cee1f835194db76c6eb7/docs/docs/deployment.md#dynamically-loading-dependencies 

require.context() is an alias in Webpack to replace that part with a dynamic import but that won't work if you don't build it with Webpack.

B4nan:
Ok but still, we should use the same configuration for development as well, otherwise you might end up with errors because of missing types when building the production version via webpack. 

The `ReflectMetadataProvider` specifically should be used here, it will allow you to not specify all types (primitives should be supported at least). What about adding `webpack: boolean` configuration option, so the code you are changing would be:

```typescript
await MikroORM.init({
  entities: await getEntities(),
  webpack: true,
});
```

Then in the `MikroORM` class the condition I linked would check for this config flag instead of the environment variable. 

WDYT?

jsprw:
That would be even better, great idea! You could also name this config variable disableDynamicFileAccess or disableFileScanning to remove the mention of Webpack.

B4nan:
`disableDynamicFileAccess` sounds good... about that additional validation, I would keep that, it's still valid to use TsMorphMetadatProvider with webpack, let's just change the wording so it does not mention webpack explicitly.

B4nan:
`discovery.disableDynamicFileAccess` will be available in next RC. Please check the changes in docs if it's ok.

PR #292, opened 2020-01-07
Update configuration.md

shynome:
if you want find entity in entities, `discovery.requireEntitiesArray` config is required.
I spent an hour to find out why I have set the entites but the mikro-rom still requires the file, hope this can save time for others

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -10,6 +10,9 @@ entities in selected folders.
 ```typescript
 MikroORM.init({
   entities: [Author, Book, Publisher, BookTag],
+  discovery: {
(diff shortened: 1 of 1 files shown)


B4nan:
what do you mean? there are 2 ways you can define entities, by folders (`entitiesDirs` option) or by reference (`entities` option). `requireEntitiesArray` is just about forcing you to use `entities` variant (which is in some cases required, like when you bundle via webpack).

B4nan:
ouch, that is not a docs issue, its a bug. closing this now, will create new issue for this so you can subscribe to it.

B4nan:
here it is: https://github.com/mikro-orm/mikro-orm/issues/293

shynome:
Thanks for your quickly reply, I have subscribed that issue

PR #323, opened 2020-02-11
fix: add TS non-null assertion operator for non-nullable columns

soroushjp:
Please let me know if there is a reason for this behaviour, and I can close or update PR.

### DIFF
--- a/docs/docs/entity-generator.md
+++ b/docs/docs/entity-generator.md
@@ -43,4 +43,3 @@ $ ts-node generate-entities
 
 - many to many relations are not supported, pivot table will be represented as separate entity
 - in mysql, tinyint columns will be defined as boolean properties
-- in postgresql, `NOT NULL` columns won't be generated with the `!` TypeScript non-null assertion operator
--- a/docs/docs/entity-generator.md
+++ b/docs/docs/entity-generator.md
@@ -43,3 +43,4 @@ $ ts-node generate-entities
 
 - many to many relations are not supported, pivot table will be represented as separate entity
 - in mysql, tinyint columns will be defined as boolean properties
+- in postgresql, `NOT NULL` columns won't be generated with the `!` TypeScript non-null assertion operator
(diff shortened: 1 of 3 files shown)


soroushjp:
@B4nan Do you know off the top of your head which database drivers this applies to? Is it all applicable drivers, beyond `postgresql`?

soroushjp:
Removed second part of original comment, since it was inaccurate (amended commit to avoid multiple commits for this PR).

B4nan:
I think this is just not yet implemented. It should apply to all drivers, the code for writing the entity code is same for all of them. 

Will be happy to merge this, but it would be better to fix the behaviour instead :] 

> Removed second part of original comment, since it was inaccurate

Yeah I was just about to write that it should work with the nullable cols: https://github.com/mikro-orm/mikro-orm/blob/master/lib/schema/EntityGenerator.ts#L73

I think the only necessary change to support this is on the very same line.

soroushjp:
Thanks for pointing me to the relevant line @B4nan. How does this new fix look to you?

B4nan:
This will definitely require changes to tests, in the entity generator snapshot at least:

https://github.com/mikro-orm/mikro-orm/blob/master/tests/__snapshots__/EntityGenerator.test.ts.snap

soroushjp:
Sounds good @B4nan, makes sense. I will need to finish this during the weekend unfortunately, due to work commitments today and during the week.

B4nan:
Looking at the snapshot, there is one more small thing missing - the entity interface are not there (`IdEntity<T>` and others). But that will be a bit more complicated (one need to pick the right interface based on the actual PK). 

> Sounds good @B4nan, makes sense. I will need to finish this during the weekend unfortunately, due to work commitments today and during the week.

It should be enough to remove the snapshot and run the tests locally to regenerate it. I don't think there are more tests affected by this. I can do that for you if you don't want to set tests up locally (should be easy though, firing up `docker-compose` should be enough).

soroushjp:
I can work on regenerating those snapshots when I'm off work @B4nan . If you'd rather get this merged ASAP, feel free to re-run yourself as well.

Issue #342, opened 2020-02-18
populate() should accept a single entity


darkbasic:
**Is your feature request related to a problem? Please describe.**
In GraphQL you have to write a resolver for each field which has not been returned by the parent. The easiest way to achieve so with mikro-orm is to call `populate` on that field. Unfortunately `populate` only accepts arrays of entities and not a single entity, so you end up with the following, which is pretty ugly:

```ts
async getMatchCreator(match: Match) {
  return (await this.repo.populate([match], ['creator']))[0].creator;
}
```

`populate` signature:
```ts
populate(entities: T[], populate: string | string[] | boolean, where?: FilterQuery<T>, orderBy?: QueryOrderMap, refresh?: boolean, validate?: boolean): Promise<T[]>;
```

**Describe the solution you'd like**
`populate` should also accept a single entity. The previous code would be way cleaner:

```ts
getMatchCreator(match: Match) {
  return this.repo.populate(match, ['creator']).creator;
}
```

B4nan:
This is already supported, just not on the repository level (and all repository does is to forward the call to entity manager, so in fact it is just a typing issue).

PR #343, opened 2020-02-19
fix(typing): align EntityRepository populate signature to EntityManager

darkbasic:
Closes https://github.com/mikro-orm/mikro-orm/issues/342


B4nan:
Thanks!

Issue #346, opened 2020-02-20
Collection: add a new initAndGetItems() method


darkbasic:
**Is your feature request related to a problem? Please describe.**
In GraphQL you have to write a resolver for each field which has not been returned by the parent. When dealing with collections the easiest way to achieve so with mikro-orm is to `await` for `init()` on the collection and then call `getItems()`. This works well, but unfortunately it's a little bit ugly:
```
async getMatchPartecipants(match: Match): Promise<User[]> {
  return (await match.partecipants.init()).getItems();
}
```

**Describe the solution you'd like**
I would like to have some sugar in the form of an `initAndGetItems()` method, which would allow me to get rid of the `(await ...)` and in turn get more readable code:

```
getMatchPartecipants(match: Match): Promise<User[]> {
  return await match.partecipants.initAndGetItems();
}
```

If you think this would be ok for you I could submit a PR to implement it.

B4nan:
`Collection` is iterator, so you can do `[...await collection.init()]`. `getItems()` also creates a shallow copy of the array so the behaviour is equal. 

But sure we can have a method for that. Maybe `loadItems()` would be better name, feel free to send a PR.

stephenh:
+1 for `loadItems()`. If starting from scratch, I'd go for `load()` but given the existing `getItems()`, then `loadItems()` is more consistent.

B4nan:
Also, calling `Collection.init` will always query the database and reload the collection. The new `loadItems()` method should not do that.

PR #347, opened 2020-02-20
feat(core): add loadItems() method to Collection

darkbasic:
Sugar for await init() and getItems()

Closes https://github.com/mikro-orm/mikro-orm/issues/346


B4nan:
Thanks

PR #351, opened 2020-02-21
feat(core): rework commit logic of UoW to ensure right query order

B4nan:
- uses topological sorting to find correct order of entities when persisting to ensure FK availability
- postpones m:n collection updates (removes them from entity change sets)
- changes the logic when extra updates are used


B4nan:
Lol, so Array.sort() was really not stable before node 10 :D

Issue #266, opened 2019-12-11
Integrate dataloader for ref/collection lazy loading


stephenh:
I'm using mikro to implement a GraphQL server. In GraphQL, queries are not known ahead of time, and so the "endpoints"/resolvers are implemented very piecemeal, i.e. "get this project's clients", "okay now get the 1st client's name", "okay now get the 2nd client's name".

However, this can easily lead to N+1s where if a project has 100 clients, and the `Client` resolver's `name` function is called 100 times, on each call it does a `(await clientRef.load()).name`, we'll end up with 100 individual SQL calls.

The standard way of solving this in JS GraphQL implementations is to use dataloader (https://github.com/graphql/dataloader), which lets the programmer still implement the resolvers in the ^ fashion (I'm loading client 1, then client 2, then client 3) but behind the scenes it uses the JS event loop to turn each of the "load client 1", "load client 2", "load client 3" promises into a single meta-promise of "load clients 1, 2, 3" that then, once that promise is fulfilled with a batch-friendly "load all three" implementation (typically hand-written by the programmer), dataloader breaks up the list of "client 1, client 2, client3" back out into the individual promises, and gives it back to each call site, who know thinks "hey I just got a single client" but really in an N+1 safe way.

Granted, the scheme requires the GraphQL implementation to trigger the "load name for all 100 clients" in a single tick of the event loop, but this happens almost defacto by just calling the "load name" method from a regular for loop.

Anyway, b/c I need this "batch load refs" behavior, I've hand-written a dataloader that looks like:

```typescript
export class EntityDataLoader {
  private loader: DataLoader<Reference<any>, AnyEntity<any, any>>;

  constructor(em: EntityManager) {
    this.loader = new DataLoader<Reference<any>, AnyEntity<any, any>>(async refs => {
      // Group the refs per entity and mapValues to get their primary keys
      const idsPerEntity: Record<string, any[]> = {};
      asSequence(refs)
        .groupBy(ref => ref.__meta.className)
        .forEach((refs, key) => {
          idsPerEntity[key] = refs.map(ref => ref.__primaryKey);
        });
      // Load all the rows for each entity
      const promises = Object.entries(idsPerEntity).map(([entity, ids]) => {
        return em.getRepository(entity).find({ id: ids } as any);
      });
      await Promise.all(promises);
      // Now once the entities are already in our Unit Of Work, we can load the refs w/o triggering N SQL calls.
      return refs.map(async ref => await ref.load());
    });
  }

[...]

PR #364, opened 2020-02-26
[RFC] Dataloader

darkbasic:
**NOTE**: this is not meant to be merged, it's just a small GraphQL project based on Mikro-ORM to showcase the `EntityDataLoader` implementation. If the approach is deemed to be sane I'll integrate it into Mikro-ORM and rebase this PR.

In order to run it be sure to remove the parent `node_modules` directory and to run `docker-compose up` and `yarn codegen`:
```
cd dataloader
rm -rf ../node_modules
yarn
yarn codegen
docker-compose up -d
yarn test
```

```
$ yarn test
yarn run v1.22.0
$ jest
  console.log tests/postgre.test.ts:161
    W/o dataloader: 13 ms

  console.log tests/postgre.test.ts:162
    W/ dataloader: 7 ms

  console.log tests/postgre.test.ts:202
    W/o dataloader: 14 ms

  console.log tests/postgre.test.ts:203
    W/ dataloader: 8 ms

  console.log tests/postgre.test.ts:243
    W/o dataloader: 147 ms

  console.log tests/postgre.test.ts:244
    W/ dataloader: 43 ms

 PASS  tests/postgre.test.ts (5.949s)
  Backend
    ✓ isConnected() (290ms)
    ✓ should fetch all users (300ms)
    ✓ should fetch all sports and their specialties (One To Many) (264ms)
    ✓ should return the same entities with EntityDataLoader (280ms)
    ✓ should be faster with EntityDataLoader (305ms)
    ✓ should fetch all specialties and their sport (Many To One) (254ms)
    ✓ should return the same entities with EntityDataLoader (277ms)
    ✓ should be faster with EntityDataLoader (287ms)
    ✓ should fetch a BIG query (431ms)
    ✓ should return the same entities with EntityDataLoader (442ms)
    ✓ should be faster with EntityDataLoader (415ms)

Tests:       11 passed, 11 total
Snapshots:   4 passed, 4 total
Time:        6.608s, estimated 10s
Ran all test suites.
Done in 7.97s.
```

Comment out `debug: ['query']` in bootstrap.ts in order to see which queries get executed.

The important file to look at is `dataloader/src/schema/datasources/utils/index.ts`: https://github.com/mikro-orm/mikro-orm/blob/fa646506149528d72cbdeda17462c2fa39339fef/dataloader/src/schema/datasources/utils/index.ts#L162

See also https://github.com/mikro-orm/mikro-orm/issues/266

Thanks @stephenh for the idea.


darkbasic:
P.S.
In the project you can also see a function called `getItemNoPopulate`: this is a stub for an even deeper optimization where I traverse the whole `info` object in order to generate a single big query for the whole tree. This is obviously GraphQL specific and so cannot be integrated into Mikro-ORM. Anyway the current implementation is way simpler and just takes the already available IDs from the Refs if no other fields gets queried. It's not used at all in the tests, I simply left it because the approach could be interesting enough to make its own library.

darkbasic:
I expanded the dataloader to also dataload find and findOne queries. So right now it can automatically dataload references, collections, find queries and findOne queries (limitations apply, because it only supports one level of depth and it doesn't yet support the full set of operators).
I also renamed the existing tests as `graphql.test.ts` and I added new ones which don't use GraphQL at all: `reference.test.ts`, `collection.test.ts`, `find.test.ts` and `find-one.test.ts`.
The codebase is still not performance optimized, but I added performance tests and it's way faster than non-dataloaded code-paths nonetheless.
I apologize for the hard-to-read code, please consider it as a proof of concept.

darkbasic:
```
$ yarn test
yarn run v1.22.4
$ jest --runInBand
 PASS  tests/graphql.test.ts (12.636s)
  ● Console

    console.log tests/graphql.test.ts:206
      W/o dataloader: 15 ms
    console.log tests/graphql.test.ts:207
      W/ dataloader: 9 ms
    console.log tests/graphql.test.ts:247
      W/o dataloader: 22 ms
    console.log tests/graphql.test.ts:248
      W/ dataloader: 8 ms
    console.log tests/graphql.test.ts:288
      W/o dataloader: 275 ms
    console.log tests/graphql.test.ts:289
      W/ dataloader: 64 ms
    console.log tests/graphql.test.ts:329
      W/o dataloader: 36 ms
    console.log tests/graphql.test.ts:330
      W/ dataloader: 25 ms
    console.log tests/graphql.test.ts:370
      W/o dataloader: 56 ms
    console.log tests/graphql.test.ts:371
      W/ dataloader: 23 ms
    console.log tests/graphql.test.ts:411
      W/o dataloader: 83 ms
    console.log tests/graphql.test.ts:412
      W/ dataloader: 26 ms

 PASS  tests/find.test.ts
  ● Console

    console.log tests/find.test.ts:83
      W/o dataloader: 18 ms
    console.log tests/find.test.ts:84
      W/ dataloader: 4 ms
    console.log tests/find.test.ts:149
      W/o dataloader: 22 ms
    console.log tests/find.test.ts:150
      W/ dataloader: 14 ms
    console.log tests/find.test.ts:204
      W/o dataloader: 19 ms
    console.log tests/find.test.ts:205
      W/ dataloader: 11 ms

 PASS  tests/find-one.test.ts
  ● Console

    console.log tests/find-one.test.ts:73
      W/o dataloader: 25 ms
    console.log tests/find-one.test.ts:74
      W/ dataloader: 3 ms
    console.log tests/find-one.test.ts:137
      W/o dataloader: 19 ms
    console.log tests/find-one.test.ts:138
      W/ dataloader: 11 ms
    console.log tests/find-one.test.ts:195
      W/o dataloader: 19 ms
    console.log tests/find-one.test.ts:196
      W/ dataloader: 10 ms

 PASS  tests/collection.test.ts
  ● Console

    console.log tests/collection.test.ts:80
      W/o dataloader: 17 ms
    console.log tests/collection.test.ts:81
      W/ dataloader: 10 ms

 PASS  tests/reference.test.ts
  ● Console

    console.log tests/reference.test.ts:68
      W/o dataloader: 10 ms
    console.log tests/reference.test.ts:69
      W/ dataloader: 3 ms


Test Suites: 5 passed, 5 total
Tests:       48 passed, 48 total
Snapshots:   15 passed, 15 total
Time:        23.302s, estimated 26s
Ran all test suites.
Done in 24.26s.
```

darkbasic:
I added some more tests to cover usage of find and findOne dataloader when mixing different entities or same entities with different filters.

Also please note that while the find/findOne dataloader currently uses the Repository API it will probably move to the Entity Manager API.

thekevinbrown:
@darkbasic, we're not using GraphQL, but the approach of doing joins to load the whole entity graph is something I've used in other ORMs that is really powerful. Why couldn't we incorporate that behaviour as an extension to the `populate` functionality that already exists?

darkbasic:
@thekevinbrown I've probably made some confusion for those who do not use GraphQL, but while this approach can indeed by used with GraphQL it can also be used by anything else. Moreover, this approach is sub-optimal for GraphQL, because with GraphQL you could traverse the `info` object to accurately predict the shape of the query in advance. So the approach I used in this PR is better suited as a general purpose solution rather than a GraphQL specific one.
That said, I *do* plan to incorporate this functionality in existing load/loadMany/find/findOne methods, it's just not part of this PR. The reasons are the following:
1) This is meant as a RFC rather than a real PR
2) I'm planning to use this in production in a few weeks/months, so it needs to be standalone until it's ready to be merged
3) @B4nan is already working on v4, so it makes little sense to try integrating it in v3

darkbasic:
I ported the dataloader to v5 and I removed anything related to graphql. I also wrote 29 tests to make sure it works in all circumstances and it's faster. Tomorrow I'll add more details about what it's able to do and what it isn't. In the meantime could you please re-open this? I force pushed and the PR got closed @B4nan 
To try it just run:
`yarn`
`yarn docker`
`DOCKER=true yarn test`

B4nan:
I guess you should open new PR, there is something wrong with the rebase you made...

image

darkbasic:
@B4nan sorry I messed up with the rebase, it should be possible to reopen it now.
In the next revision I will also strip out the pg database and switch to in-memory sqlite.

Instead of integrating it inside mikro-orm I implemented the dataloader as a standalone piece of software which consumes both the public and (sometimes) the internal API. All the relevant code is in `src/dataloader/EntityDataloader.ts`. Internally I use three different dataloaders: one for References, one for Collections and another one for Queries (find/findOne). A single instance of the EntityDataloader class can deal with all kind of entities, meaning that you can mix different entities in the same event loop. In addition to this it can also deal with the same Entities with different query conditions and it will automatically take care of separating the different entities, extrapolating the greatest common divisor between the different conditions in order to create a single query for each entity and finally filter the result to assign it to the different queries.

The **Reference loader** is the simplest and IMO could be easily integrated in the upcoming mikro-orm v5. This is especially simple because I don't have to filter the results to reassign them, instead after firing the batched query I can simply call `Reference.load()` on each reference knowing that thanks to the Identity Map pattern it won't fire any more queries.

The **Collection loader** is a bit more complex, mainly because I have to manually filter the results to reassign them and also because the public API doesn't expose enough metadata so I have to resort to the internal one sometimes. I'm fairly confident that after cleaning it up a little bit it could find its way to the upcoming mikro-orm v5.

The reason why we desperately need this in GraphQL applications resides in the way GraphQL works: if I want to query a list of Events and their creators it first queries the events, then **for each of them** it will also **separately** query the creator (loads the reference). It has to do so because there are no endpoints and the frontend could query whatever it wants, so if you ask the backend to query events it cannot know in advance if you are also going to ask for their creators or not. The traditional way to deal with this is to dataload each single query separately, but this especially annoying and I think that the ORM could do this instead.

The Reference dataloader and the Collection dataloader would already solve 90% of my use cases, with the following exceptions:

1. Entities with composed keys. For example my application shows a list of Matches (for different Sports) and for each of them I can load a collection of its participants. For each of them I want to know their level in that sport (how strong they are). Since each Level depends not only on the User (obviously) but also on the Sport it cannot be a simple reference in the User entity and it requires a composite key on user+sport. Since I cannot use a reference I cannot obviously dataload it (another example is friendship relationships, which are composite as well: user1+user2).
2. Querying subsets of collections. Whenever I don't need the full collection but just a subset of it I will have to make another query and thus I cannot dataload the collection. A workaround would be to use loadItems anyway and simply filter the results, which may be applicable if you won't have to filter out most of the results.

For the previous reasons I also created the **Query dataloader**, which dataloads generic queries. This is the most complex one because it needs to first analyze the query conditions in order to compute the minimum amount of queries necessary to pull all the entities you need, then it needs to filter out the results and analyze each condition to reassign them. I currently support only a small subset of the operators which allows me to always reduce the conditions to a single query. I also support just one syntax, you can see some usage examples in the tests. Cutting on features allowed to keep it simple, but because of that I don't think we would be able to integrate it. Also, while the Reference and Collection dataloader always save time, the Query dataloader is a double edged weapon: if the queries you're making are completely unrelated it will take longer because of the additional processing. This is not a problem in GraphQL applications because usually in the same event loop you will find batches of similar queries with similar conditions, but this is something to consider if it wants to be general purpose. Also it could be difficult to keep it performant while broadening the supported features. For this reason I don't think this is ready to be included in its current form, but for sure it would be a really nice challenge to tackle this.

B4nan:
Thanks for the write up, will try to review this soon. Unfortunately still can't reopen this (also the diff is not available).

image

jgiunta1:
@darkbasic is there any update on this? Currently running into this issue in my TypeGraphql application

darkbasic:
@jgiunta1 I forgot to reopen the PR, but I'm basically waiting for feedback from @B4nan

amaranth:
I'm a bit confused on what this PR is supposed to be. It seems the old one was adding an example project showing how to build a GraphQL project and included your `EntityDataLoader` (which seems to be the key addition here) to make that project more efficient. Then after you tried to update/rebase it you seem to have accidentally deleted everything from the project except your example which is why github is very confused about what to do with it.

Was this meant to be an example on steroids that included how you thought dataloader and mikro-orm could be used together or was it supposed to add `EntityDataLoader` to the API? I would think that latter is what we'd want to aim for, in something like a `@mikro-orm/dataloader` package.

darkbasic:
@amaranth this was supposed to be just a RFC, not a real PR. In fact I didn't implement the dataloading logic inside mikro-orm because it was too early and would have been harder to rebase for no reason. Instead I mirrored a subset of the API into a separate class just to give a taste of how it would look like once integrated. You can start having a look from there: https://github.com/darkbasic/mikro-orm/blob/dataloader/src/dataloader/EntityDataLoader.ts

> something like a @mikro-orm/dataloader package

The idea was to get some thoughts from @B4nan about the possibility of integrating a similar approach inside MikroORM itself, but as time passes it looks like he might not be interested in such a feature. If that's the case maybe I might re-evaluate mirroring a sub-portion of the API into a separate package like you suggested.

corbt:
> Instead I mirrored a subset of the API into a separate class just to give a taste of how it would look like once integrated.

Hey @darkbasic, I'm looking for a DataLoader solution for Mikro-ORM. I'm trying to get a sense for the current status here; is the `EntityDataLoader` class a good starting point for use in production in my own app?

Thanks!

darkbasic:
Hi @corbt I wouldn't call it production ready yet, but the idea behind it was to aim at something production capable down the road thus it's backed by tons of tests including performance ones. I'm using it into another project with even more tests but it's not in production yet.

I am considering to turn this into its own self-contained package because of the lack of upstream interest, so if you're willing to battle test this in production I would encourage you to do so.

@B4nan would you prefer `@mikro-orm/dataloader` as suggested by @amaranth or would you rather prefer it to stay under my name?

corbt:
> if you're willing to battle test this in production I would encourage you to do so.

Awesome, would love to test this! Would you mind putting it in a repository with a small README? That'll help me figure out what the intended public API is. 🙂

B4nan:
> @B4nan would you prefer @mikro-orm/dataloader as suggested by @amaranth or would you rather prefer it to stay under my name?

Let's first have something released and tested outside, then we can move it under the `@mikro-orm` organization after some initial testing and feedback. I am open for adding such package, I was just not sure about changing internals to use dataloaders, but having helpers in separate package totally makes sense.

corbt:
@darkbasic, would you mind giving a simple example of how to make this work with Mikro-ORM 5? I'm trying to adapt it but running into type errors that aren't immediately obvious to me.

darkbasic:
@corbt I'm sorry but I previously missed your last comment. You can find plenty of examples in the tests. My previous RFC iteration had graphql specific examples while this one comes with more general purpose ones.

darkbasic:
New PR for v6: https://github.com/mikro-orm/mikro-orm/pull/4321

PR #373, opened 2020-02-29
refactor: use Promise.all to await for independent promises

pepakriz:


tests/issues/GH228.test.ts
```diff
@@ -69,7 +69,7 @@ describe('GH issue 228', () => {
       populate: true,
     });
     expect(mock.mock.calls[0][0]).toMatch('select `e0`.* from `a` as `e0` order by `e0`.`type_id` asc');
-    expect(mock.mock.calls[1][0]).toMatch('select `e0`.* from `b` as `e0` where `e0`.`id` in (?)');
+    expect(mock.mock.calls[2][0]).toMatch('select `e0`.* from `b` as `e0` where `e0`.`id` in (?)');
```
pepakriz: This test depends on the execution order :(

B4nan: I guess we could reorder the `calls` array then? Or check its length (to be sure we do not fire more queries than expected) and check for particular queries via sth like `includes`. 

```typescript
const queries = mock.mock.calls.map(c => c[0]);
expect(queries).toHaveLength(3);
expect(queries).toContain('... sql1 ...');
expect(queries).toContain('... sql2 ...');
```

pepakriz: Thanks for suggestions. I've used sort fn.

B4nan:
Hmm looks like its stuck again, what ever, thanks!

pepakriz:
@B4nan It was stuck because of this issue: https://github.com/mikro-orm/mikro-orm/pull/387
You should wait for the "green" pipeline :)

PR #374, opened 2020-02-29
feat(query-builder): improved typing of where(), andWhere(), orWhere()

CyriacBr:
As discussed in the Slack channel, this improves typings and provides intelliSense for the `QueryBuilder` `where` methods, without breaking change.


B4nan:
I guess we could use that type also on `having`, `join`, `leftJoin` and `delete` methods.

PR #377, opened 2020-03-01
fix: fixed github actions coverage

pepakriz:


pepakriz:
@B4nan it failed again. So maybe wrong configuration on the coveralls.io side? I'd like to fix it because every PR is failing due to this issue.

B4nan:
The problem is with gh actions not providing secrets when running against forks. Tried to fix this already by not running the coveralls request when the token is missing, but apparently its not working.

pepakriz:
Wow, I've solved it. Take a look...

B4nan:
Perfect, thanks!

Issue #118, opened 2019-09-14
Error when handle decorator @PrimaryKey


chaoyangnz:
`TypeError: Cannot read property '1' of null
    at Function.lookupPathFromDecorator (/Users/chao.yang/Work/stuff-content/node_modules/mikro-orm/dist/utils/Utils.js:177:78)
    at /Users/chao.yang/Work/stuff-content/node_modules/mikro-orm/dist/decorators/PrimaryKey.js:12:23
    at DecorateProperty (/Users/chao.yang/Work/stuff-content/node_modules/reflect-metadata/Reflect.js:553:33)
`

OS: Mac OSX
Node: 10.x.x
Framework: NestJS
DB: Postgres

B4nan:
Could you please add `console.log(stack, line);` to `node_modules/mikro-orm/dist/utils/Utils.js` file right before this error occurred (line 176 I guess) and pass the output here?

chaoyangnz:
could you give me a link about the bug of `tslib` you mentioned?

B4nan:
issue #100, it was fixed in v2.7.6

B4nan:
Any updates on this one?

B4nan:
I will close this one as resolved, let me know if you still have problems and I will reopen it, but be sure to include the debug output I was asking for as well as your ORM version.

CyriacBr:
@B4nan I was working on a lib based on mikro-orm just now and the cli couldn't generate the cache. I had the same issue as @chaoyangnz .  
It appears that the `lookupFromDecorator` expect each stack trace line to contain the file path inside parentheses. However, on my repo that wasn't the case, no idea why.  
Check https://github.com/mikro-orm/mikro-orm/pull/378. This is working for me.

PR #378, opened 2020-03-02
fix(utils): lookupPathFromDecorator() incorrect processing of stack t…

CyriacBr:
…race

Related to https://github.com/mikro-orm/mikro-orm/issues/118.
It appears that the `lookupFromDecorator` method expects each stack trace line to contain the file path inside parentheses. However it seems not to be always the case. On one of my project, when I log `stack[line]` this is what it outputs: `     at /long/path/to/entity.file.ts:x:y`.  
Naturally this would cause the lookup method to crash.

### DIFF
--- a/lib/utils/Utils.ts
+++ b/lib/utils/Utils.ts
@@ -338,7 +338,15 @@ export class Utils {
       line++;
     }
 
-    meta.path = Utils.normalizePath(stack[line].match(/\((.*):\d+:\d+\)/)![1]);
+    if (stack[line].match(/\(.+\)/i)) {
+      meta.path = Utils.normalizePath(
+        stack[line].match(/\((.*):\d+:\d+\)/)![1],
+      );
+    } else {
+      meta.path = Utils.normalizePath(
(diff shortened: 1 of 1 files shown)


B4nan:
please provide this description in the PR, now it says only “race...”.

Also please add a test to keep 100% coverage, you introduced new branch that is not covered. This way it will be also easily readable how is your usecase different.

CyriacBr:
> please provide some description in the PR, now it says only “race...”.

I actually replied to the linked issue to explain the problem:

> @B4nan I was working on a lib based on mikro-orm just now and the cli couldn't generate the cache. I had the same issue as @chaoyangnz .
> It appears that the lookupFromDecorator expects each stack trace line to contain the file path inside parentheses. However, on my repo that wasn't the case, no idea why.
> Check #378. This is working for me.

> Also please add a test to keep 100% coverage, you introduced new branch that is not covered. This way it will be also easily readable how is your usecase different.  

Sure.

B4nan:
I know you replied there, but here is the right place to have the information. All PRs should have a meaningful description.

PR #387, opened 2020-03-06
fix: default property value

pepakriz:
It fixes failing master branch. See https://github.com/mikro-orm/mikro-orm/pull/373/checks?check_run_id=490535503


B4nan:
lol thanks, somehow i forgot to add this change, but i had it locally :D

PR #392, opened 2020-03-08
feat(mongo): add support for transactions in mongodb

### LINKED ISSUE #34: Implement transactions in mongo driver
MongoDB 4 supports transactions, so we should add support for that. 

Maybe via new driver keeping the current `MongoDriver` for MongoDB 3, or via some configuration option that would be disabled by default (or automatically enabled via feature detection/version comparison).

B4nan:
Mongo driver now supports transactions, although implicit transactions are by default disabled.
We can either use explicit transaction demarcation via `em.trasactional()`, or enable them globally
via `implicitTransactions: true`.

Closes #34


B4nan:
Available in v3.4.0

Issue #66, opened 2019-06-18
Add support for Map collections or composite key.


Darpholgshon:
I have a table definition which has a composite primary key and was mapped in Java as a CollectionOfElements.

Example:
Supplier:   { id, name, created_at, updated_at }
SupplierConfiguration: {supplier_id, property_key, property_value}

The composite key is supplier_id and property_key, and in Hibernate this is managed as a Map<String, String> using the decorators below.
```
@CollectionOfElements(fetch = FetchType.EAGER)
@JoinTable(name = "supplier_configuration", joinColumns = @JoinColumn(name = "supplier_id"))
@MapKey(columns = @Column(name = "property_key"))
@Column(name = "property_value")
public Map<String, String> getConfiguration() {
  return configuration;
}
```

B4nan:
Hey there! I am definitely planning to support composite PKs. What do you think about the way Nextras ORM does it?

https://nextras.org/orm/docs/2.3/entity#toc-primary-and-primary-proxy

In your example, you are not defining the child entity (`SupplierConfiguration`), and instead you are defining the collection of config options in the Supplier entity directly, right? I would rather define the child entity, and move the composite PK definition there. Then the config collection definition would be pretty much classic `@OneToMany`:

```typescript
@Entity()
export class Supplier {

  @OneToMany({ fieldName: 'property_value' })
  config = new Collection<SupplierConfiguration>(this);

  getConfiguration(): Record<string, string> {
    return this.config.getItems().map(option => option.value);
  }

}

@Entity()
export class SupplierConfiguration {

  @PrimaryKey({ composite: ['supplier', 'key'] })
  id: string;

  @ManyToOne()
  supplier: Supplier;

  @Property({ fieldName: 'property_key' })
  key: string;

  @Property({ fieldName: 'property_value' })
  value: string;

}
```

The composite PK here would be only virtual property, or we could move this to `@Entity` decorator. WDYT?

```typescript
@Entity({ compositePK: ['supplier', 'key'] })
export class SupplierConfiguration {

  // ...

}
```

Unfortunately there is no eager loading ATM, I am planning to add that as well, but not sure when. But when I am thinking about it now, it could be pretty easy to implement.

B4nan:
Mmmmm I forgot about the mapping part to dictionary, but that can be easily done in the `getConfiguration()` method:

```typescript
getConfiguration(): Record<string, string> {
[...]

PR #395, opened 2020-03-10
feat(core): add support for composite keys

B4nan:
Related #66

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,7 @@ There is also auto-generated [CHANGELOG.md](CHANGELOG.md) file based on commit m
 - [Unit of Work](https://mikro-orm.io/unit-of-work/)
 - [Transactions](https://mikro-orm.io/transactions/)
 - [Cascading persist and remove](https://mikro-orm.io/cascading/)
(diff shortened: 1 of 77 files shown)


darkbasic:
I'll give it a try, I'm pretty sure I will have to refactor my dataloading PR to support this.

B4nan:
This is still WIP. Will send more changes and tests later today, with support for FK as PK, and basically for all what doctrine supports (as I am designing the tests based on their docs):

https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/tutorials/composite-primary-keys.html

Not sure if I will also add support for having composite key entities as owners in relations, that might be a bit tricky... But I will give it a try too :]

B4nan:
Quick progress report, support for composite entities in relationships is almost ready, I am aiming to release 3.5 later this week (it will take few more days for sure, but the hard parts are implemented already).

B4nan:
Still need to do some code clean up, but pretty much all of the functionalities are now implemented. Will release this probably tomorrow.

PR #398, opened 2020-03-12
Fix undefined variables in migration

jsprw:
Column names are capitals in MySQL 8.0.19 (MySQL Community Server - GPL). Without the aliases, the columns will be returned like COLUMN_NAME, DATA_TYPE, COLUMN_DEFAULT etc. and causing undefined column properties (for example line 95) and breaking the migration if the table structure is existent (creating the structure from scratch works fine, it is just the updating of existing databases that is broken).


B4nan:
Interesting, so can this be considered as "fixing support for mysql 8"? As that sounds more understandable from the changelog perspective. 

Also please fix the tests, should be just about fixing those snapshots. Otherwise we are good to merge.

jsprw:
I don't know if it is only MySQL 8, I just started using the migrator and got loads of undefined errors and after looking further into it, it was because of these queries returning the data in capitals.

B4nan:
Well the fix is only in the mysql helper, so I think we can call it this way. Other drivers use different approach/queries to get the column list, so they should be fine. 

Anyway, thanks, will release it as part of 3.5 probably later this week.

PR #412, opened 2020-03-18
feat(core): allow for the number zero as a primary key

pbohnenkamp:
There are some cases where an entity will have a primary key of 0.
Our use case is a table of static statuses that are pervasive
throughout our application. We add the rows when we generate the
schema, not through the ORM. We only want to map relationships
to those statuses with the ORM. Some of these statuses start with
the id of 0 versus 1. Previous to this fix, there were generic
falsy checks that would take a primary key with a value of 0 as
if the primary key was not set. This fix makes those checks
specific for both null and undefined values.

BREAKING CHANGE: entities that were using a primary key of 0 to
signify that the primary key was not defined will now interpret
the zero to be the actual primary key.

To migrate, no longer default the primary key to 0. Use null
and undefined to signify a primary key that is not defined.

### DIFF
--- a/lib/entity/EntityAssigner.ts
+++ b/lib/entity/EntityAssigner.ts
@@ -31,7 +31,8 @@ export class EntityAssigner {
         value = props[prop].customType.convertToJSValue(value, platform);
       }
 
-      if (props[prop] && [ReferenceType.MANY_TO_ONE, ReferenceType.ONE_TO_ONE].includes(props[prop].reference) && value && EntityAssigner.validateEM(em)) {
+      // tslint:disable-next-line:triple-equals - Really want the double equals here to pass for both null and undefined and allow for keys of numeric 0
+      if (props[prop] && [ReferenceType.MANY_TO_ONE, ReferenceType.ONE_TO_ONE].includes(props[prop].reference) && value != null && EntityAssigner.validateEM(em)) {
         return EntityAssigner.assignReference<T>(entity, value, props[prop], em!);
       }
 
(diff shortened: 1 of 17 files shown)


B4nan:
I am quite fine with the proposal, but what I don't like is that `==` comparison. There is already `Utils.isDefined()` method, if you think we need to check for `null` too, then let's add a bool param to that method that will also consider `null` as falsy value. 

Also, do we really need those new entities in tests? Looks like not, so I would reuse existing ones (to have better test speed), also I would add that test to one of existing test files so we do not have to initialize the ORM again (that is the slow part).

pbohnenkamp:
Hi, just read your comments, thanks. I will make the adjustments and re-push. Probably won't be until this weekend though

B4nan:
Yeah no rush, btw I do not consider this as a breaking change, but if you think it really is, we can have it as part of v4 (will start to work on v4 right after I am done with composite keys #395).

B4nan:
What about this? I am planning to release one more feature release (3.6) and then will do some hard breaking changes, namely splitting the project into multiple packages so it will be really challenging to rebase it (v4 development is done in `dev` branch, so will keep releasing bug fixes if needed). 

I would suggest to get this out in 3.6 (which I would like to release later this week, once I deal with #416 and #421).

pbohnenkamp:
Apologies for disappearing for a bit, had a work deadline. I was on the
fence whether this would be a breaking change, as I figured no one should
have used a 0 to indicate an undefined key. But then I thought, why would
someone choose a primary key of 0 :)

I just rebased the changes in for 3.5 and I think I can have everything
updated by the end of the day.

Thanks!

PR #423, opened 2020-03-25
Check for undefined before calling get metadata

darkbasic:


lib/metadata/MetadataDiscovery.ts
```diff
@@ -453,7 +453,7 @@ export class MetadataDiscovery {
   }
 
   private defineBaseEntityProperties(meta: EntityMetadata): void {
-    const base = this.metadata.get(meta.extends);
+    const base = meta.extends && this.metadata.get(meta.extends);
 
     if (!meta.extends || !base) {
```
B4nan: we can now simplify this 

```suggestion
    if (!base) {
```

PR #426, opened 2020-03-27
feat(core): allow for the number zero as a primary key

pbohnenkamp:
There are some cases where an entity will have a primary key of 0.
Our use case is a table of static statuses that are pervasive
throughout our application. We add the rows when we generate the
schema, not through the ORM. We only want to map relationships
to those statuses with the ORM. Some of these statuses start with
the id of 0 versus 1. Previous to this fix, there were generic
falsy checks that would take a primary key with a value of 0 as
if the primary key was not set. This fix makes those checks
specific for both null and undefined values.

### DIFF
--- a/lib/entity/EntityAssigner.ts
+++ b/lib/entity/EntityAssigner.ts
@@ -31,7 +31,7 @@ export class EntityAssigner {
         value = props[prop].customType.convertToJSValue(value, platform);
       }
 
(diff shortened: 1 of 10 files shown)


pbohnenkamp:
Apologies for the totally new pull request, after merging all the commits from upstream, I couldn't figure out how to get the change to stay at one commit for the previous pull request.

B4nan:
Thanks!

pbohnenkamp:
My pleasure. I like what you have put together! If you want help, I could
give at least a few hours of time a week. Let me know if there are any
features/bugs you want addressed that would help get me more familiar with
design/architecture and work my way up to larger stuff.

PR #429, opened 2020-03-28
Add migration pattern to the documentation

Dschoordsch:
This needs to be adapted to find .js files, thus should also be
documented.


B4nan:
Thanks

PR #475, opened 2020-04-12
feat(core): split project into multiple packages

B4nan:
This allows to have lighter production builds (e.g. do not ship ts-morph or knex when not used).

Also removes the need to import drivers from `mikro-orm/dist`, now we have dedicated driver packages that already have the underlying db connector as a dependency.

Available packages:
- `@mikro-orm/core`
- `@mikro-orm/reflection` - `TsMorphMetadataProvider`
- `@mikro-orm/cli` - CLI support, requires entity-generator, migrator and knex
- `@mikro-orm/knex` - SQL support
- `@mikro-orm/entity-generator`
- `@mikro-orm/migrations`
- `@mikro-orm/mysql`
- `@mikro-orm/mariadb`
- `@mikro-orm/mysql-base` - Common implementation for mysql and mariadb
- `@mikro-orm/sqlite`
- `@mikro-orm/postgresql`
- `@mikro-orm/mongodb`

For easier transition, meta package `mikro-orm` is still present, reexporting `core`, `reflection` and `cli` packages.

**BREAKING CHANGE:**
Some parts of the ORM now resides under their own packages.
Default metadata provider has been changed to `ReflectMetadataProvider`.


B4nan:
@lookfirst so what do you say about this breaking change? :D

lookfirst:
It is what it is... not the end of the world. 

I have to do that on one of my react projects, I'm just too lazy to set it all up.

PR #485, opened 2020-04-16
feat(mongo): add support for nullable unique indexes

### LINKED ISSUE #484: MongoDB Optional One-to-One
**Describe the bug**
One-to-one relations add unique indexes in MongoDB, but sparse option should be also added for optional attributes.

**To Reproduce**
1. Create a one-to-one relation.
2. Add multiple documents keeping optional attribute undefined.
3. Unique index will raise exception.

**Expected behavior**
Multiple documents with undefined optional attribute should be ok.

osayfun:
Adds sparse option to nullable unique indexes

Closes #484


B4nan:
Thanks!

Issue #486, opened 2020-04-17
Initializing a relation can result in the loss of pending changes


followben:
**Describe the bug**
Lazily populating a bidirectional relation inadvertently writes any pending changes for the parent (and any onward relations) to the EntityManager's `originalEntityData` map. When the resulting UnitOfWork is committed, those changes are omitted from the changeSet and never written to the DB.

**To Reproduce**
1. Fetch an entity instance but don't populate any bidirectional relations
2. Change a property on the fetched instance (e.g. a scalar value)
3. Initialize a bidirectional relation and change it (e.g. add a new instance to a Collection)
4. Call `em.persistAndFlush()`

**Expected behavior**
Both changes are persisted to the database.

**Actual behavior**
Only the 2nd change, after the relation `init()`, is persisted.

**Versions**

| Dependency | Version |
| - | - |
| node | 13.10.1 |
| typescript | 3.8.3 |
| mikro-orm | 3.6.7 |
| pg | 7.18.2 |

B4nan:
Thanks for the repro, appreciated!

PR #487, opened 2020-04-17
Test changes not lost when initing a bidirectional relation

followben:
Demonistrates #486.


B4nan:
Fixed, test case included in the test suite. Thanks again

PR #503, opened 2020-04-21
feat(core): add support for single table inheritance

### LINKED ISSUE #33: Add support for single table inheritance
Similar to what doctrine implements: https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/inheritance-mapping.html#single-table-inheritance

But less verbose (one decorator should be enough).

B4nan:
```typescript
@Entity({
  discriminatorColumn: 'discr',
  discriminatorMap: { person: 'Person', employee: 'Employee' },
})
export class Person {
  // ...
}

@Entity()
export class Employee extends Person {
  // ...
}
```

Closes #33

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -19,11 +19,12 @@ discuss specifics.
 
 ## Planned changes for v4
 
+- [x] Single table inheritance ([#33](https://github.com/mikro-orm/mikro-orm/issues/33))
+- [x] Allow adding items to not initialized collections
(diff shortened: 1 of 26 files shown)


totalard:
What is the need of discriminator here? both can connect to single table, and person only get what the entity property has, and employee will get what it has, employee is extending person, so it has both person and employee.

i cant understand the use of discriminator column.

B4nan:
You need to be able to tell what the type of the row is. This is nothing new, every ORM with STI support does it this way.

PR #509, opened 2020-04-23
fix(docs): fix link to MySQL EntityManager test

lukasfrana:


B4nan:
Thanks. Looks like GH is having outage so the CI fails...

B4nan:
> I also noticed that in the header EntityRepository<T> on https://mikro-orm.io/docs/repositories is rendered backslash but here on GitHub is not rendered.

Afaik this is a markdown problem, docusaurus (used for docs) uses some different flavour of it I think. Originally I had that as `code snippet` which solves it, but the GH pages template was rendering it with wrong font size or something like that. I guess we could use code snippets there now.

lukasfrana:
Yeah, I see 500 a lot.

Also I noticed backslash in heading in docs [here](https://mikro-orm.io/docs/repositories/#entityrepositoryt-api), but I don't know how to fix it. Markdown on GitHub is rendered correctly.

PR #510, opened 2020-04-24
docs: small grammar fix

lookfirst:


B4nan:
Would be good to edit also the next docs, as you just changed 3.6 and when releasing next version this change would be forgotten.

PR #514, opened 2020-04-25
feat(mapping): add support for embeddables

B4nan:
Embeddables are classes which are not entities themselves, but are embedded in entities.

```typescript
@Embeddable()
class Address {

  @Property()
  street!: string;

  @Property()
  city!: string;

  @Property()
  country!: string;

}

@entity()
class User {

  @PrimaryKey()
  id!: number;

  @Embedded(() => Address)
  address!: Address;

}
```


murbanowicz:
Does it work with all dB? I mean pg and mongodb as well?

B4nan:
Yes it does, it works for all drivers: https://github.com/mikro-orm/mikro-orm/pull/514/files#diff-75e1159357f882865c05f8e43f1506a7

Issue #214, opened 2019-10-24
Static checking of population


vimmerru:
**Is your feature request related to a problem? Please describe.**

For the moment ```em::find(...)``` and ```em::find(..., { populate: ['something'] })``` return exactly the same entity types and compiler don't know was references populated or not.

We have the solution to explicitly define references https://github.com/mikro-orm/mikro-orm/blob/dev/docs/entity-references.md#better-type-safety-with-referencet-wrapper but this solution is partial as you don't know were the references populated or not. You can await all populated references that will not perform database querying, but if you just remember population you may release the code that will produce a lot of unexpected queries to production. Also it causes significant boilerplate.

It would be nice if ``em::find(..., { populate: ['something'] })``` will return mapped type with all populated fields de-referenced.

**Describe the solution you'd like**

Here is a very rough idea how to define types:

```typescript
class Reference<T> {
    unwrap<T>(): T {
        // implementation isn't important
        return undefined as any;
    }
}

type UnwrappedReference<T> = T extends Reference<infer U> ? U : T;

type PopulatedEntity<E, R extends keyof E> = {
    [k in keyof E]: k extends R ? UnwrappedReference<E[k]> : E[k] 
}

class Credentials {
}

class Organization {
}

class User {
    name!: string;
    credentials!: Reference<Credentials>;
    organization!: Reference<Organization>;
}

function populate<E, R extends keyof E>(entities: E[], ...fields: R[]): PopulatedEntity<E, R>[] {
    // implementation isn't important
    return undefined as any;
}

const users = [new User()];

const populated1 = populate(users, 'credentials');
const credentials1: Credentials = populated1[0].credentials;
const organization: Reference<Organization> = populated1[0].organization;

const populated2 = populate(users, 'credentials', 'organization');
const credentials2: Credentials = populated2[0].credentials;
const organization2: Organization = populated2[0].organization;
```

**Additional context**

Types i provided allows to populate only one level of references. There is no generic ts solution to populate at any level, but if we limit deepness of population to some meaningful value like 3-5 correct typing is possible.

vimmerru:
> how to design the populate api to support nesting and maintain type safety?

I would prefer some monad-style population builder

```
const populated =await  new PopulateHelper([new Employee()])
    .populate('company', 'alliance')
    .populate('university')
    .exec();
```

> how will the solution for 3 levels depth look like?

Here is a proof-of-concept PopulateHelper typings for 2-levels. Any additional level is just one, but long line of code

[...]

Issue #160, opened 2019-09-24
[feature] Populate all


lookfirst:
I'd love an option to just load a top level entity and have it fully populated without having to do anything. Maybe just `{populate: true}`.

B4nan:
Implemented in `3.0.0-alpha.15`, as always, feel free to reopen.

lookfirst:
sweet! totally makes my day. loading `createdBy` all the time sucks. ;-)

i know you have your own tests, but this also works:

https://github.com/lookfirst/mikro-example/blob/master/test/db/populate.spec.ts

lookfirst:
One other idea, what about using an annotation for this, like @Load. This way, I could just specify fields that I always wanted populated...

B4nan:
Yeah, eager loading, that is also on the roadmap. Will probably do in as part of v3, as it should be pretty easy thanks to how populate support works now.

lookfirst:
Oh interesting, populate `string | true` doesn't allow me to pass in a variable. it should be string | boolean i think.

lookfirst:
Apologies, I don't see a way to reopen this issue.

B4nan:
Its `string[] | true`, the array part is your problem I think. I could also support plain string.

lookfirst:
No, it is a problem passing a boolean variable.

```
	async getRig(id: string, populate: boolean = true): Promise<Rig | null> {
		console.log('in get rig!');
		return this.getOrm().em.findOne(Rig, id, {populate});
	}
```

```
Error:(20, 10) TS2769: No overload matches this call.
  Overload 1 of 2, '(entityName: EntityName<Rig>, where: string | number | Record<string, any> | { toString?(): string; toHexString?(): string; } | Partial<Rig>, options?: FindOneOptions | undefined): Promise<...>', gave the following error.
    Argument of type '{ populate: boolean; }' is not assignable to parameter of type 'FindOneOptions'.
      Types of property 'populate' are incompatible.
        Type 'boolean' is not assignable to type 'true | string[] | undefined'.
  Overload 2 of 2, '(entityName: EntityName<Rig>, where: string | number | Record<string, any> | { toString?(): string; toHexString?(): string; } | Partial<Rig>, populate?: true | string[] | undefined, orderBy?: Record<...> | undefined): Promise<...>', gave the following error.
    Argument of type '{ populate: boolean; }' is not assignable to parameter of type 'true | string[] | undefined'.
[...]

PR #517, opened 2020-04-27
Add support for alternative loading strategies

### LINKED ISSUE #440: Support for alternative loading strategies
_**Apologies if this is documented somewhere, but I have not been able to find anything about it.**_

Not really a problem, but more of an inconvenience. It would be nice to be able to specify the [Loading technique](https://docs.sqlalchemy.org/en/13/orm/loading_relationships.html) without having to drop to using the query builder.

I would like to be able to specify the loading technique at the time I choose to load relationships.

Looking at the docs on [Collections](https://mikro-orm.io/docs/collections) and seeing how relationships are currently loaded, one potential API for this could be something like:

```ts
type Strategy = 'lazy' | 'joined' | 'subquery' | 'selectin';
```

```ts
const author = orm.em.findOne(Author, '...', ['books', 'publishers']); // current

const author = orm.em.findOne(Author, '...', {
  books: {
    strategy: 'joined',
  },
  publishers: {
    strategy: 'subquery',
  },
}); // suggested

await author.books.init({
  strategy: 'joined',
});
```

Similar to SQLAlchemy, you could even specify the load type at the relationship level:

```ts
@Entity()
export class Book {
  @ManyToOne({
    entity: () => Author,
    strategy: 'joined',
  })
  author!: Author;
}
```

As mentioned above, the alternative right now seems to be to use the query builder.

**Additional context**

This feature seems fairly standard to ORMs. Some examples include [SQLAlchemy](https://docs.sqlalchemy.org/en/13/orm/loading_relationships.html) and [Objection.js](https://vincit.github.io/objection.js/api/query-builder/eager-methods.html#withgraphfetched)

---

Thanks for the great project btw!

willsoto:
This adds support for alternate load strategies. As of now, only one additional strategy is supported: `joined`. Currently, the only way to specify the strategy is on on the decorator for the relationship:

```ts
@Entity()
export class Author2 {
  @OneToMany({
    entity: () => Book2,
    mappedBy: 'author',
    orderBy: { title: QueryOrder.ASC },
    strategy: LoadStrategy.JOINED, // new property
  })
  books!: Collection<Book2>;
}
```
---

- [ ] Tests
- [ ] Docs
- [ ] Cleanup

Closes #440

### DIFF
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -3,7 +3,7 @@ import { v4 as uuid } from 'uuid';
 import { Configuration, RequestContext, Utils, ValidationError, SmartQueryHelper } from './utils';
 import { EntityAssigner, EntityFactory, EntityLoader, EntityRepository, EntityValidator, IdentifiedReference, Reference, ReferenceType, wrap } from './entity';
 import { LockMode, UnitOfWork } from './unit-of-work';
-import { IDatabaseDriver, FindOneOptions, FindOptions, EntityManagerType, PopulateOptions, Populate } from './drivers';
+import { IDatabaseDriver, FindOneOptions, FindOptions, EntityManagerType, Populate, PopulateOptions } from './drivers';
 import { EntityData, EntityMetadata, EntityName, AnyEntity, IPrimaryKey, FilterQuery, Primary, Dictionary } from './typings';
 import { QueryOrderMap } from './enums';
 import { MetadataStorage } from './metadata';
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -598,20 +598,8 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
     if (Array.isArray(populate)) {
       return populate.map(field => {
         if (Utils.isString(field)) {
-          const relation = meta.properties[field];
-
-          // If no relation is found, we assume that the relation is invalid and don't attempt to get the strategy.
-          // By returning a field/relation that doesn't exist, validation code will handle that error upstream.
-          if (!relation) {
-            return {
-              field,
-            };
-          }
-
-          return {
-            field,
-            strategy: relation.strategy,
-          };
+          const strategy = meta.properties[field]?.strategy;
+          return { field, strategy };
         }
 
         return field;
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -3,7 +3,7 @@ import { v4 as uuid } from 'uuid';
 import { Configuration, RequestContext, Utils, ValidationError, SmartQueryHelper } from './utils';
 import { EntityAssigner, EntityFactory, EntityLoader, EntityRepository, EntityValidator, IdentifiedReference, Reference, ReferenceType, wrap } from './entity';
 import { LockMode, UnitOfWork } from './unit-of-work';
-import { IDatabaseDriver, FindOneOptions, FindOptions, EntityManagerType } from './drivers';
+import { IDatabaseDriver, FindOneOptions, FindOptions, EntityManagerType, PopulateOptions, Populate } from './drivers';
 import { EntityData, EntityMetadata, EntityName, AnyEntity, IPrimaryKey, FilterQuery, Primary, Dictionary } from './typings';
 import { QueryOrderMap } from './enums';
 import { MetadataStorage } from './metadata';
@@ -83,7 +83,9 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
     this.validator.validateParams(where);
     const options = Utils.isObject<FindOptions>(populate) ? populate : { populate, orderBy, limit, offset };
     options.orderBy = options.orderBy || {};
-    const results = await this.driver.find<T>(entityName, where, { ...options, populate: this.preparePopulate(options.populate) }, this.transactionContext);
+    const preparedPopulate = this.preparePopulate(entityName, options.populate);
+    const opts = { ...options, populate: preparedPopulate };
+    const results = await this.driver.find<T>(entityName, where, opts, this.transactionContext);
 
     if (results.length === 0) {
       return [];
@@ -97,7 +99,7 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
     }
 
     const unique = Utils.unique(ret);
-    await this.entityLoader.populate<T>(entityName, unique, options.populate || [], where, options.orderBy, options.refresh);
+    await this.entityLoader.populate<T>(entityName, unique, preparedPopulate, where, options.orderBy, options.refresh);
 
     return unique;
   }
@@ -155,7 +157,10 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
     }
 
     this.validator.validateParams(where);
-    const data = await this.driver.findOne(entityName, where, { ...options, populate: this.preparePopulate(options.populate) }, this.transactionContext);
+    const preparedPopulate = this.preparePopulate(entityName, options.populate);
+    options.populate = preparedPopulate;
+
+    const data = await this.driver.findOne(entityName, where, { ...options, populate: preparedPopulate }, this.transactionContext);
 
     if (!data) {
       return null;
@@ -494,7 +499,7 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
     return ret;
   }
 
-  async populate<T extends AnyEntity<T>, K extends T | T[]>(entities: K, populate: string | string[] | boolean, where: FilterQuery<T> = {}, orderBy: QueryOrderMap = {}, refresh = false, validate = true): Promise<K> {
+  async populate<T extends AnyEntity<T>, K extends T | T[]>(entities: K, populate: string | Populate, where: FilterQuery<T> = {}, orderBy: QueryOrderMap = {}, refresh = false, validate = true): Promise<K> {
     const entitiesArray = Utils.asArray(entities);
 
     if (entitiesArray.length === 0) {
@@ -502,7 +507,8 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
     }
 
     const entityName = entitiesArray[0].constructor.name;
-    await this.entityLoader.populate(entityName, entitiesArray, populate, where, orderBy, refresh, validate);
+    const preparedPopulate = this.preparePopulate(entityName, populate);
+    await this.entityLoader.populate(entityName, entitiesArray, preparedPopulate, where, orderBy, refresh, validate);
 
     return entities;
   }
@@ -579,13 +585,49 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
       await this.lock(entity, options.lockMode, options.lockVersion);
     }
 
-    await this.entityLoader.populate(entityName, [entity], options.populate || [], where, options.orderBy || {}, options.refresh);
+    const preparedPopulate = this.preparePopulate(entityName, options.populate);
+
+    await this.entityLoader.populate(entityName, [entity], preparedPopulate, where, options.orderBy || {}, options.refresh);
 
     return entity;
   }
 
-  private preparePopulate(populate?: string[] | boolean) {
-    return Array.isArray(populate) ? populate : [];
+  private preparePopulate(entityName: string, populate?: string | Populate): PopulateOptions[] {
+    const meta = this.metadata.get(entityName);
+
+    if (Array.isArray(populate)) {
+      return populate.map(field => {
+        if (Utils.isString(field)) {
+          const relation = meta.properties[field];
+
+          // If no relation is found, we assume that the relation is invalid and don't attempt to get the strategy.
+          // By returning a field/relation that doesn't exist, validation code will handle that error upstream.
+          if (!relation) {
+            return {
+              field,
+            };
+          }
+
+          return {
+            field,
+            strategy: relation.strategy,
+          };
+        }
+
+        return field;
+      });
+    }
+
+    if (Utils.isString(populate)) {
+      const relation = meta.properties[populate];
+
+      return [{
+        field: populate,
+        strategy: relation.strategy,
+      }];
+    }
+
+    return [];
   }
 
 }
--- a/packages/core/src/drivers/IDatabaseDriver.ts
+++ b/packages/core/src/drivers/IDatabaseDriver.ts
@@ -4,7 +4,7 @@ import { QueryOrderMap, QueryFlag } from '../enums';
 import { Platform } from '../platforms';
 import { MetadataStorage } from '../metadata';
 import { LockMode } from '../unit-of-work';
-import { Collection } from '../entity';
+import { Collection, LoadStrategy } from '../entity';
 import { EntityManager } from '../index';
 import { DriverException } from '../exceptions';
 
@@ -46,7 +46,7 @@ export interface IDatabaseDriver<C extends Connection = Connection> {
 
   aggregate(entityName: string, pipeline: any[]): Promise<any[]>;
 
-  mapResult<T extends AnyEntity<T>>(result: EntityData<T>, meta: EntityMetadata): T | null;
+  mapResult<T extends AnyEntity<T>>(result: EntityData<T>, meta: EntityMetadata, populate?: PopulateOptions[]): T | null;
 
   /**
    * When driver uses pivot tables for M:N, this method will load identifiers for given collections from them
@@ -75,7 +75,7 @@ export interface IDatabaseDriver<C extends Connection = Connection> {
 }
 
 export interface FindOptions {
-  populate?: string[] | boolean;
+  populate?: Populate;
   orderBy?: QueryOrderMap;
   limit?: number;
   offset?: number;
@@ -87,7 +87,7 @@ export interface FindOptions {
 }
 
 export interface FindOneOptions {
-  populate?: string[] | boolean;
+  populate?: Populate;
   orderBy?: QueryOrderMap;
   groupBy?: string | string[];
   lockMode?: LockMode;
@@ -97,3 +97,10 @@ export interface FindOneOptions {
   schema?: string;
   flags?: QueryFlag[];
 }
+
+export type Populate = (string | PopulateOptions)[] | boolean;
+
+export type PopulateOptions = {
+  field: string;
+  strategy?: LoadStrategy;
+};
--- a/packages/core/src/entity/EntityLoader.ts
+++ b/packages/core/src/entity/EntityLoader.ts
@@ -1,11 +1,12 @@
 import { AnyEntity, EntityProperty, FilterQuery } from '../typings';
-import { EntityManager } from '../index';
-import { ReferenceType } from './enums';
+import { EntityManager, Populate } from '../index';
+import { ReferenceType, LoadStrategy } from './enums';
 import { Utils, ValidationError } from '../utils';
 import { Collection } from './Collection';
 import { QueryOrder, QueryOrderMap } from '../enums';
 import { Reference } from './Reference';
 import { wrap } from './wrap';
+import { PopulateOptions } from '../drivers';
 
 export class EntityLoader {
 
@@ -14,25 +15,25 @@ export class EntityLoader {
 
   constructor(private readonly em: EntityManager) { }
 
-  async populate<T extends AnyEntity<T>>(entityName: string, entities: T[], populate: string | string[] | boolean, where: FilterQuery<T> = {}, orderBy: QueryOrderMap = {}, refresh = false, validate = true, lookup = true): Promise<void> {
-    if (entities.length === 0 || populate === false) {
+  async populate<T extends AnyEntity<T>>(entityName: string, entities: T[], populate: PopulateOptions[] | boolean, where: FilterQuery<T> = {}, orderBy: QueryOrderMap = {}, refresh = false, validate = true, lookup = true): Promise<void> {
+    if (entities.length === 0 || populate === false || (Array.isArray(populate) && populate.length > 0)) {
       return;
     }
 
     populate = this.normalizePopulate(entityName, populate, lookup);
 
-    const invalid = populate.find(field => !this.em.canPopulate(entityName, field));
+    const invalid = populate.find(({ field }) => !this.em.canPopulate(entityName, field));
 
     if (validate && invalid) {
-      throw ValidationError.invalidPropertyName(entityName, invalid);
+      throw ValidationError.invalidPropertyName(entityName, invalid.field);
     }
 
-    for (const field of populate) {
-      await this.populateField<T>(entityName, entities, field, where, orderBy, refresh);
+    for (const pop of populate) {
+      await this.populateField<T>(entityName, entities, pop.field, where, orderBy, refresh);
     }
   }
 
-  private normalizePopulate(entityName: string, populate: string | string[] | true, lookup: boolean): string[] {
+  private normalizePopulate(entityName: string, populate: PopulateOptions[] | true, lookup: boolean): PopulateOptions[] {
     if (populate === true) {
       populate = this.lookupAllRelationships(entityName);
     } else {
@@ -144,7 +145,20 @@ export class EntityLoader {
 
     const filtered = Utils.unique(children);
     const prop = this.metadata.get(entityName).properties[f];
-    await this.populate<T>(prop.type, filtered, [parts.join('.')], where[prop.name], orderBy[prop.name] as QueryOrderMap, refresh, false, false);
+
+    await this.populate<T>(
+      prop.type,
+      filtered,
+      [{
+        field: parts.join('.'),
+        strategy: LoadStrategy.SELECT_IN,
+      }],
+      where[prop.name],
+      orderBy[prop.name] as QueryOrderMap,
+      refresh,
+      false,
+      false
+    );
   }
 
   private async findChildrenFromPivotTable<T extends AnyEntity<T>>(filtered: T[], prop: EntityProperty, field: keyof T, refresh: boolean, where?: FilterQuery<T>, orderBy?: QueryOrderMap): Promise<AnyEntity[]> {
@@ -195,13 +209,13 @@ export class EntityLoader {
     return children.filter(e => !wrap(e[field], true).isInitialized()).map(e => Utils.unwrapReference(e[field]));
   }
 
-  private lookupAllRelationships(entityName: string, prefix = '', visited: string[] = []): string[] {
+  private lookupAllRelationships(entityName: string, prefix = '', visited: string[] = []): PopulateOptions[] {
     if (visited.includes(entityName)) {
       return [];
     }
 
     visited.push(entityName);
-    const ret: string[] = [];
+    const ret: PopulateOptions[] = [];
     const meta = this.metadata.get(entityName);
 
     Object.values(meta.properties)
@@ -213,14 +227,17 @@ export class EntityLoader {
         if (nested.length > 0) {
           ret.push(...nested);
         } else {
-          ret.push(prefixed);
+          ret.push({
+            field: prefixed,
+            strategy: LoadStrategy.SELECT_IN,
+          });
         }
       });
 
     return ret;
   }
 
-  private lookupEagerLoadedRelationships(entityName: string, populate: string[], prefix = '', visited: string[] = []): string[] {
+  private lookupEagerLoadedRelationships(entityName: string, populate: PopulateOptions[], prefix = '', visited: string[] = []): PopulateOptions[] {
     if (visited.includes(entityName)) {
       return [];
     }
@@ -237,7 +254,10 @@ export class EntityLoader {
         if (nested.length > 0) {
           populate.push(...nested);
         } else {
-          populate.push(prefixed);
+          populate.push({
+            field: prefixed,
+            strategy: LoadStrategy.SELECT_IN,
+          });
         }
       });
 
--- a/packages/core/src/entity/EntityLoader.ts
+++ b/packages/core/src/entity/EntityLoader.ts
@@ -16,7 +16,7 @@ export class EntityLoader {
   constructor(private readonly em: EntityManager) { }
 
   async populate<T extends AnyEntity<T>>(entityName: string, entities: T[], populate: PopulateOptions[] | boolean, where: FilterQuery<T> = {}, orderBy: QueryOrderMap = {}, refresh = false, validate = true, lookup = true): Promise<void> {
-    if (entities.length === 0 || populate === false || (Array.isArray(populate) && populate.length > 0)) {
+    if (entities.length === 0 || populate === false) {
       return;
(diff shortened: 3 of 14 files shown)


packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -63,15 +63,60 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
...
+      .select(selects)
       .populate(options.populate)
       .where(where as Dictionary)
       .orderBy(options.orderBy!)
-      .limit(1)
+      // .limit(1)
       .setLockMode(options.lockMode)
       .withSchema(options.schema)
-      .execute('get');
+      .execute('all');
```
willsoto, line 119: When joining records (like in a one-to-many) you necessarily get back >1 row, so that is why these changes were made.

packages/core/src/EntityManager.ts
```diff
@@ -155,13 +155,19 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
...
+    }, this.transactionContext);
 
     if (!data) {
       return null;
     }
 
-    entity = this.merge(entityName, data as EntityData<T>, options.refresh) as T;
+    const entityData = !Array.isArray(data) ? [data] : data;
+
+    entity = this.merge(entityName, entityData[0] as EntityData<T>, options.refresh) as T;
```
willsoto, line 170: This is the point at which the raw query results actually become instances of things, so I'd imagine this is where I should also map any additional properties that come through?

packages/core/src/EntityManager.ts
```diff
@@ -155,13 +155,19 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
...
+      ...options,
+      populate: preparedPopulate,
+    }, this.transactionContext);
 
     if (!data) {
       return null;
     }
 
-    entity = this.merge(entityName, data as EntityData<T>, options.refresh) as T;
+    const entityData = !Array.isArray(data) ? [data] : data;
```
B4nan, line 168:
```suggestion
    const entityData = Utils.asArray(data);
```

packages/core/src/EntityManager.ts
B4nan: as noted above, `em.merge()` should not serve this purpose, its method for merging entities = making them managed. 

`DatabaseDriver.mapResult()` is the right place I think, probably by overriding the implementation in `AbstractSqlDriver`. that is what gets called when you use QB.getResult() - in fact it uses `em.map()` which first uses driver's `mapResult` followed by `em.merge()`. Then we could probably rewrite the `AbstractSqlDriver.findOne` so it uses `QB.getResult()` (or actually `QB.getSingleResult()` for `findOne`).

or maybe it should happen in the `Hydrator` (that is used in `EntityFactory`).

packages/knex/src/AbstractSqlDriver.ts
B4nan: although this makes sense, i would rather do the mapping via QB as suggested above, as I dont like the idea of driver returning multiple rows for findOne. currently driver returns POJOs that are converted to entities via the `em.merge()` call, but we could return entities directly I hope (at least for `AbstractSqlDriver` - we can keep the `em.merge()` call sitting on the EM level to maintain compatibility with mongo driver. 

or we can narrow it for all drivers, as we are aiming v4 with this, we can afford some breaking changes.

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -63,15 +63,60 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
       options.fields.unshift(pk);
     }
 
-    return this.createQueryBuilder(entityName, ctx, !!ctx)
-      .select(options.fields || '*')
+    const joinedLoads: string[] = options.populate.filter((field) => {
+      const relation = meta.properties[field];
+
+      return relation.strategy === 'joined';
+    });
```
willsoto, line 70: Related to this, I am wondering where the best place would to convert `populate` from `string[]` to `PopulateOptions[]` so that it can be passed around to all subsequent call sites who might need it.

My thinking is that we will get `populate`, `find`, `findOne` etc for free if we can standardize the data structure early on.

```ts
// all the various signatures for populate
populate = "books";
populate = ["books", "tags"];
populate = [
  {
    field: "books",
    strategy: "joined",
  },
  "tags",
];
```

into

```ts
populate = [
  {
    field: "books",
    strategy: "joined",
  },
  {
    field: "tags",
    strategy: "selectin",
  },
];
```

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -63,15 +63,60 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
...
+
+        Object.values(properties)
+          .filter(({ reference }) => reference === 'scalar')
+          .forEach((descriptor) => {
+            const [fieldName] = descriptor.fieldNames;
+            const aliasedProp = `"${alias}"."${fieldName}" as "${alias}_${fieldName}"`;
+
+            selects.push(aliasedProp);
+            qb.select(aliasedProp);
+            qb.join(relationName, alias);
```
willsoto, line 103: I am not sure if I was doing something wrong, but if I don't call `select` prior to calling `join`, I get an error from the `join` call since there is nothing in the `_aliasMap` yet.

Is this expected or am I doing something? I guess this isn't necessarily a problem, I'm mostly curious.

packages/knex/src/AbstractSqlDriver.ts
willsoto: > i would rather do the mapping via QB as suggested above, as I dont like the idea of driver returning multiple rows for findOne.

To be clear, are you suggesting that we convert the POJOs into actually entities in the QB? Or convert the POJO into the expected structure. Eg we take the flat object we get (from op) and turn it into:

```js
const author = {
  createdAt: "2020-04-26T22:32:53.482Z",
  updatedAt: "2020-04-26T22:32:53.490Z",
  name: "Albert Camus",
  email: "albert.camus@email.com",
  age: null,
  termsAccepted: false,
  optional: null,
  identities: null,
  born: null,
  bornTime: null,
  id: 1,
  books: [
    {
      uuid_pk: "1ce80133-0379-440d-b92e-e18e93428434",
      created_at: "2020-04-26T22:32:53.482Z",
      title: "The Stranger",
      perex: null,
      price: null,
      double: null,
      meta: null,
    },
  ],
};
```

> but we could return entities directly I hope (at least for AbstractSqlDriver - we can keep the em.merge() call sitting on the EM level to maintain compatibility with mongo driver.

If we return the above data structure, the `AbstractSqlDriver` could then map over the new structure, instantiate entities and return those back to the EM.

Does that sound right?

packages/knex/src/AbstractSqlDriver.ts
B4nan: Kinda expected, we could work that around by making the join evaluation lazy (in the finalize method), but I don't think its an actual issue for us.

packages/knex/src/AbstractSqlDriver.ts
B4nan: Agreed that it should happen early as possible, probably right after the `canPopulate` call, or maybe even inside it (once we validate it is ok). So probably at the EM level, I guess the `preparePopulate` method can be a good fit here.

packages/knex/src/AbstractSqlDriver.ts
B4nan: First of all, you can already see that your current payload is something in the middle, as the root entity properties are already mapped to the entity prop names (camel case), while the book entity props are intact (snake case as from the db). This is what happens in the `mapResult` method of the driver, and that is where I think we should handle the child entities too so we do not have this kind of mixed state (some props mapped, some not).

About the QB, it already supports mapping to entities, that is what the `getResult()` method is for, which calls `em.map()` (which calls `driver.mapResult()` followed by `em.merge()`). So we might leverage that probably, that was my suggestion. 

So to wrap up, I think the `driver.mapResult()` should be where we convert the output to a real graph (not entities, but POJOs), so for findOne we still have single object in the return value (btw `mapResult()` is also used when doing `qb.execute()`)

packages/core/src/EntityManager.ts
```diff
@@ -583,13 +591,41 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
...
+    const meta = this.metadata.get(entityName);
+
+    if (Array.isArray(populate)) {
+      return populate.map((option) => {
+        if (typeof option === 'string') {
+          const relation = meta.properties[option];
+
+          return {
+            field: option,
+            // TODO: Ideally we specify the default strategy somewhere else (maybe when the decorator is defined?)
```
B4nan, line 611: yes, the defaults should be set during the discovery (not in decorator, as there are other ways to define entities than just decorators). 

this should ideally happen in `EntitySchema` helper, that is also used internally in all cases (with all metadata providers).

packages/core/src/drivers/IDatabaseDriver.ts
```diff
@@ -89,6 +89,8 @@ export interface FindOneOptions {
   schema?: string;
 }
 
+export type Populate = (string | PopulateOptions)[] | boolean;
```
B4nan, line 92: there is actually one thing i forgot to mention, as part of v4 i would like to deal with the #214 somehow, which will also need different format for this. we will have to be able to have strong typing support there, so maybe something like an object or tuples. 

so the `PopulateOptions` should be fine hopefully for that usecase, but we will need to extend it to support nesting (currently it is possible to have `populate: ['author.books.tags']`)

this is just a heads up, i dont think you need to care much about it now

willsoto:
@B4nan - Ok, we are getting closer now...I've gotten it to the point where the joined pojos are being transformed correctly as suggested via `mapResult`. By doing that, we get the conversion to entities for free (as far as I can tell) and there are no additional queries fired since those entities are loaded already.

packages/core/src/entity/enums.ts
```diff
@@ -13,3 +13,8 @@ export enum Cascade {
   REMOVE = 'remove',
   ALL = 'all',
 }
+
+export enum LoadStrategy {
+  SELECTIN = 'selectin',
+  JOINED = 'joined'
+}
```
willsoto, line 20: I converted this to an `enum` for consistency since you use enums to represent so many other options on the decorators.

B4nan: good. what about renaming `SELECTIN = 'selectin'` to `SELECT_IN = 'select-in'`?

tests/JoinedLoads.test.ts
```diff
@@ -0,0 +1,34 @@
+import { MikroORM } from '@mikro-orm/core';
+import { PostgreSqlDriver } from '@mikro-orm/postgresql';
+import { initORMPostgreSql, wipeDatabasePostgreSql } from './bootstrap';
+import { Author2, Book2 } from './entities-sql';
+
+describe('Joined loading', () => {
```
willsoto, line 6: I'd additionally like to test that the correct number of queries are fired, so any suggestions for that?

eg when joined loading one relation, we should see 1 query, not 2

packages/core/src/entity/enums.ts
willsoto: no problem 👍

tests/JoinedLoads.test.ts
B4nan: mock logger and assert the actual queries. i do it quite often, search for `mock.calls` in the tests

willsoto:
Also, this code is really gross...I am basically inlining everything and plan to back through and refactor (I hope you understand as I am adding a feature while learning the code base 😅  )

B4nan:
Yeah sure, no problem at all :] 

Btw I am rebasing the `dev` branch quite often with bug fixes from master, so you might need to create new branch for this and copy the changes via stash or something. But it can happen later, once we will be getting close to merge.

willsoto:
I feel like the code is _mostly_ in a good place for a more in-depth review. I still have 17 failing tests locally and I am currently looking into why they are failing with my changes. Some of them don't seem related and others seem to be non-deterministic, but `dev` is currently passing so I am looking into them.

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -187,18 +249,92 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
...
+    // alias all fields in the primary table
+    Object.values(meta.properties)
+      .filter(({ reference, persist }) => {
+        return reference === 'scalar' && persist !== false;
+      })
+      .forEach((descriptor) => {
+        const [fieldName] = descriptor.fieldNames;
+
+        selects.push(fieldName);
+        queryBuilder.select(fieldName);
```
B4nan, line 317: `qb.select()` does override, maybe you want `qb.addSelect()`? or do we even need that line?

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -187,18 +249,92 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
...
+      });
+
+    joinedLoads.forEach((relationName) => {
+      const prop = meta.properties[relationName];
+      const entity = this.metadata.get(prop.type);
```
B4nan, line 322: i dont like the name `entity` here. usually i use `meta2` to point to the other side, not great, but better than calling it `entity` when its not entity but its metadata.

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -187,18 +249,92 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
...
+      const { properties } = entity;
```
B4nan, line 323: and as this is the only place where you use the `entity` variable, i would suggest to inline it so we dont have to care about naming :] 

```typescript
const properties = this.metadata.get(prop.type).properties;
```

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -187,18 +249,92 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
...
+
+      Object.values(properties)
+        .filter(({ reference, persist }) => reference === 'scalar' && persist !== false)
+        .forEach((descriptor) => {
+          const [fieldName] = descriptor.fieldNames;
```
B4nan, line 328: i see you really love destructing :D 

but here i dont see the point, this is too verbose, when all we want is to access the first array element.

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -187,18 +249,92 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
...
+      });
+
+    joinedLoads.forEach((relationName) => {
+      const prop = meta.properties[relationName];
+      const entity = this.metadata.get(prop.type);
+      const { properties } = entity;
+
+      Object.values(properties)
+        .filter(({ reference, persist }) => reference === 'scalar' && persist !== false)
+        .forEach((descriptor) => {
```
B4nan, line 327: we use `prop` everywhere in the code base so please stick to that.
```suggestion
        .forEach(prop => {
```

packages/knex/src/query/QueryBuilder.ts
```diff
@@ -85,6 +85,12 @@ export class QueryBuilder<T extends AnyEntity<T> = AnyEntity> {
     return this.init(QueryType.COUNT);
   }
 
+  selectAliasedField(field: string, schema: string, index?: number) {
```
B4nan, line 88: this should probably live elsewhere (abstract sql driver i guess?), also the method does not select anything.

and it looks like you never use the `index` param - what should that be about?

packages/knex/src/query/QueryBuilder.ts
```diff
@@ -45,7 +45,7 @@ export class QueryBuilder<T extends AnyEntity<T> = AnyEntity> {
               private readonly connectionType?: 'read' | 'write',
               private readonly em?: SqlEntityManager) { }
 
-  select(fields: string | KnexQueryBuilder | (string | KnexQueryBuilder)[], distinct = false): this {
+  select(fields: string | Field[], distinct = false): this {
```
B4nan, line 48:
```suggestion
  select(fields: Field | Field[], distinct = false): this {
```

packages/knex/src/query/QueryBuilder.ts
```diff
@@ -491,6 +498,12 @@ export class QueryBuilder<T extends AnyEntity<T> = AnyEntity> {
 
 }
 
+type KnexStringRef = Knex.Ref<string, {
```
B4nan, line 501:
```suggestion
type KnexStringRef = Ref<string, {
```

(and import `Ref` from `knex` if its not yet there)

packages/knex/src/query/QueryBuilder.ts
```diff
@@ -491,6 +498,12 @@ export class QueryBuilder<T extends AnyEntity<T> = AnyEntity> {
...
+  [alias: string]: string;
+}>;
+
+export type Field = (KnexStringRef | string | KnexQueryBuilder);
```
B4nan, line 505: no need to wrap in parens, also i would move the `string` to the first place (looks weird between those knex types :P)

B4nan:
Sorry for the delay, the code looks good, most of the comments are just code style suggestions so its aligned with the codebase. 

Btw I think we will need to explicitly list all the fields when selecting anyway, as I am planning to introduce lazy scalar properties in v4, so we will need to be able to blacklist something from the select clause instead of using `*`.

tests/JoinedLoads.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    Object.assign(orm.em.config, { logger });
+
+    (await orm.em.findOne(Author2, {
+      id: author2.id,
+    }, {
+      populate: ['books'],
+    }));
+
+    expect(mock.mock.calls.length).toBe(1);
+    expect(mock.mock.calls[0][0]).toMatch('select "e0"."id", "e0"."created_at", "e0"."updated_at", "e0"."name", "e0"."email", "e0"."age", "e0"."terms_accepted", "e0"."optional", "e0"."identities", "e0"."born", "e0"."born_time", "books"."uuid_pk" as "books_uuid_pk", "books"."created_at" as "books_created_at", "books"."title" as "books_title", "books"."perex" as "books_perex", "books"."price" as "books_price", "books"."double" as "books_double", "books"."meta" as "books_meta" from "author2" as "e0" inner join "book2" as "books" on "e0"."id" = "books"."author_id" where "e0"."id" = $1');
```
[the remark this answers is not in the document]
willsoto: Agreed. In one of my first passes I added a numeric suffix, removed it and then forgot to add it back.

I think `b1`, `c2`, `b3`, etc should be sufficient.

tests/QueryBuilder.test.ts
```diff
@@ -457,7 +457,7 @@ describe('QueryBuilder', () => {
 
   test('select by m:n inverse side (that is not defined as property) via populate', async () => {
     const qb = orm.em.createQueryBuilder(Test2);
-    qb.select('*').populate(['publisher2_tests']).where({ 'publisher2_tests.Publisher2_owner': { $in: [ 1, 2 ] } }).orderBy({ 'publisher2_tests.id': QueryOrder.ASC });
+    qb.select('*').populate([{ field: 'publisher2_to_test2' }]).where({ 'publisher2_to_test2.Publisher2_owner': { $in: [ 1, 2 ] } }).orderBy({ 'publisher2_to_test2.id': QueryOrder.ASC });
```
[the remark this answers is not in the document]
willsoto: Ah yes, I remember having merge conflicts here and I didn't notice the actual name changing. 👍

packages/core/src/entity/EntityLoader.ts
```diff
@@ -14,25 +15,25 @@ export class EntityLoader {
 
   constructor(private readonly em: EntityManager) { }
 
-  async populate<T extends AnyEntity<T>>(entityName: string, entities: T[], populate: string | string[] | boolean, where: FilterQuery<T> = {}, orderBy: QueryOrderMap = {}, refresh = false, validate = true, lookup = true): Promise<void> {
-    if (entities.length === 0 || populate === false) {
+  async populate<T extends AnyEntity<T>>(entityName: string, entities: T[], populate: PopulateOptions[], where: FilterQuery<T> = {}, orderBy: QueryOrderMap = {}, refresh = false, validate = true, lookup = true): Promise<void> {
```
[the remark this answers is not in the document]
willsoto: Unintentional on my part. But I am curious what `true` implies?

packages/knex/src/AbstractSqlDriver.ts
willsoto: Let me check, `addSelect` might do what I want here..

packages/knex/src/AbstractSqlDriver.ts
willsoto: `meta2` seems reasonable enough

packages/knex/src/AbstractSqlDriver.ts
willsoto: Your preference here would be?

```
const fieldName = descriptor.fieldNames[0];
```

packages/knex/src/query/QueryBuilder.ts
willsoto: I actually think I was already thinking about [this](https://github.com/mikro-orm/mikro-orm/pull/517#discussion_r419172881) when I added that.

Any naming suggestions? `generateAliasForField`?

packages/core/src/entity/EntityLoader.ts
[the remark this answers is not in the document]
B4nan: https://github.com/mikro-orm/mikro-orm/issues/160

packages/knex/src/AbstractSqlDriver.ts
B4nan: my preference is to not introduce new variable when not needed (especially for vars that are used only once), i would just use the `prop.fieldNames[0]` where needed. 

sometimes it makes sense to improve readability (like when you compose complex conditions), here i dont think so.

packages/knex/src/query/QueryBuilder.ts
B4nan: yeah sounds good. i would be even less verbose (e.g. `getAliasedField()`), but it might not be that readable, idk, naming is hard :]

packages/core/src/EntityManager.ts
```diff
@@ -579,13 +593,47 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
...
+          // By returning a field/relation that doesn't exist, validation code will handle that error upstream.
+          if (!relation) {
+            return {
+              field: option,
+            };
+          }
+
+          return {
+            field: option,
+            strategy: relation.strategy,
```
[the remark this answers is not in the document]
willsoto: Not totally clear on the question here 😬

packages/knex/src/query/QueryBuilder.ts
willsoto: So looking at this code more, the reason why I put it in the `QueryBuilder` is because it needs access to the `knex` instance and that is currently not public. 

I do agree with the rename, so I did that. Also, the `index` param is used:

```
const as = index ? `${schema}${index}_${field}` : `${schema}_${field}`;
```

packages/knex/src/AbstractSqlDriver.ts
willsoto: `addSelect` did the trick here 👍

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -187,18 +249,92 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
...
 
+  protected joinedLoads(meta: EntityMetadata, populate: PopulateOptions[]): string[] {
+    return populate.filter(({ field }) => {
+      const relation = meta.properties[field];
+
+      if (!relation) {
+        return false;
+      }
+
+      return relation.strategy === LoadStrategy.JOINED;
```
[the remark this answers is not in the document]
willsoto: Not clear on what you mean here 🙃

packages/knex/src/query/QueryBuilder.ts
```diff
@@ -321,6 +322,10 @@ export class QueryBuilder<T extends AnyEntity<T> = AnyEntity> {
     return qb;
   }
 
+  getRefForField(field: string, schema: string, as: string) {
```
B4nan, line 325: i can't find the thread, but this can definitely live in abstract sql driver as i suggested. you have the knex instance there accessible from `this.connection.getKnex()`. its stateless/singleton, same instance as in `QB.knex`.

B4nan: also please rename `as` to `alias`

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -69,19 +74,66 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
...
+    joinedLoads.forEach((relationName, index) => {
+      const prop = meta.properties[relationName];
+      const properties = this.metadata.get(prop.type).properties;
 
-    return this.rethrow(qb.execute('get'));
+      ret[relationName] = ret[relationName] || [];
+      const relationPojo = {};
+
+      Object.values(properties)
+        .filter(({ reference }) => reference === 'scalar')
```
B4nan, line 123: use enum

tests/JoinedLoads.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    Object.assign(orm.em.config, { logger });
+
+    await orm.em.findOneOrFail(Author2, {
+      id: author2.id,
+    }, {
+      populate: ['books'],
+    });
+
+    expect(mock.mock.calls.length).toBe(1);
+    expect(mock.mock.calls[0][0]).toMatch('select "e0"."id", "e0"."created_at", "e0"."updated_at", "e0"."name", "e0"."email", "e0"."age", "e0"."terms_accepted", "e0"."optional", "e0"."identities", "e0"."born", "e0"."born_time", "books"."uuid_pk" as "b0_uuid_pk", "books"."created_at" as "b0_created_at", "books"."title" as "b0_title", "books"."perex" as "b0_perex", "books"."price" as "b0_price", "books"."double" as "b0_double", "books"."meta" as "b0_meta" from "author2" as "e0" inner join "book2" as "books" on "e0"."id" = "books"."author_id" where "e0"."id" = $1');
```
B4nan, line 57: this still looks a bit wrong, you are aliasing `book2` table as `books`, which can still conflict if you populate such property twice (e.g. on different entities, but its still called `books`).

it should be aligned with the prefix, so `b0`.

personally i would even use the same counter, so we have `a0` and `b1` instead of `a0` and `b0`. 
1. i would expect that to be easier to maintain/code
2. i feel like it helps understanding what is the root entity there.

i would also change the root alias from static `e0` to be dependent on actual entity name (so really `a0` here for author entity), but we can do that later (i can imagine that change to require lots of test changes...), will be happy to do it myself

B4nan:
Later today I will be merging support for computed fields via `@Formula()` decorator to `dev`, I believe there will be some changes necessary to support this too. It's basically about being able to define custom sql snippets that will be automatically added to the select clause:

https://blog.eyallupu.com/2009/07/hibernate-derived-properties.html

Would be great if you could then rebase this and make sure it works with joined strategy too. Will ping you here and point you to what probably needs to be addressed.

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -32,15 +32,17 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
   async find<T extends AnyEntity<T>>(entityName: string, where: FilterQuery<T>, options?: FindOptions, ctx?: Transaction<KnexTransaction>): Promise<T[]> {
     const meta = this.metadata.get(entityName);
     options = { populate: [], orderBy: {}, ...(options || {}) };
-    options.populate = this.autoJoinOneToOneOwner(meta, options.populate as string[]);
+
+    const populate = options.populate as PopulateOptions[];
+    options.populate = this.autoJoinOneToOneOwner(meta, populate);
```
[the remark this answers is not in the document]
willsoto: `populate` is used on line 45 as well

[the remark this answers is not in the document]
willsoto: I need to type assert in additional places since `options.populate` is a much wider type

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -208,18 +263,87 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
...
     return [...populate, ...toPopulate];
   }
 
+  protected joinedLoads(meta: EntityMetadata, populate: PopulateOptions[]): string[] {
+    return populate
+      .filter(({ field }) => meta.properties[field]?.strategy === LoadStrategy.JOINED)
+      .map(({ field }) => field);
+  }
+
+  protected processJoinedLoads<T extends AnyEntity<T>>(results: object[], joinedLoads: string[]): T {
```
B4nan, line 288: looking at the failed tests, and one of your problems is that here you always return object, even when no entity is found

so we should first check for empty array and return `null` in such case

B4nan:
Support for formulas is merged: https://github.com/mikro-orm/mikro-orm/pull/553

Currently it works only with the root entity, as that is the only one that gets actually mapped. With your changes, we will need to load those formulas for joined (populated) entities too.

here is the part that does the job, implemented at the QB level:

https://github.com/mikro-orm/mikro-orm/pull/553/files#diff-e32c03fc87a03e74dd84d662768fbd4eR461-R469

To support this with joined strategy, we will need to adjust the aliased name (e.g. `price_taxed -> b1_price_taxed`) and we will need to make sure we add it to selects also for joined entities.

Should be quite easy, I can help with that once we have all tests passing.

willsoto:
So I am down to 7 test failures. I have some ordering issues and then I have a feeling I am not handling `populate: true` correctly somewhere, but I can't see it. I will continue to work on this but let me know if you see anything...

B4nan:
You still have wrongly rebased changes there, mainly the schema dumps are wrong, as well as pivot table names in the entities, you are missing some tables there and some have bad names. Looks like you just omitted many of the changes I made, but kept some parts of it. For me ~200 tests are failing. 

This will be really hard to fix, as you made changes to fixtures so you are actually getting false positives, your tests might be passing but the changes I made are not present in the code. It would be better if you checkout fresh brach from `dev` and use git stash to move the changes from this brach, so you clearly see all things that changed. 

If you are lost there I can try to do this myself (via new branch/PR, as fixing this would take me days probably, as its basically about checking all the changes I made in dev branch).

willsoto:
I’ll go ahead and just close this PR and cherry pick my changes into a new branch. Are you going to continue rebasing dev?

willsoto:
Closing in favor of #556

B4nan:
Thanks, will look into this when I have some spare time. Will try to not rebase dev till this one is merged (or the new one actually).

PR #521, opened 2020-04-28
build: add Node.js 14

Yaty:
Node.js 14 is now available, I think it should be tested :)

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,2 @@
+github: [B4nan]
+custom: 'https://paypal.me/mikroorm'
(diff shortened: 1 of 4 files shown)


B4nan:
i would expect the CI to fail, let's see. a lot of dependencies are out of date because of node 8 support - in `dev` branch is v4 with up to date dependencies, so we could try that instead

B4nan:
yes, failing as expected. please target the `dev` branch.

B4nan:
You will need to base it on dev branch too (might be easier to create new MR).

Yaty:
Yes I will do that

PR #522, opened 2020-04-28
feat(core): add support for Node.js 14

Yaty:
here we go again :)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 13.x]
+        node-version: [10.x, 12.x, 13.x, 14.x]
 
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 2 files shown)


B4nan:
Should be caused by node-postgres, it is supposed to be fixed in the latest patch version, can you try to upgrade it?

https://github.com/brianc/node-postgres/issues/2180

B4nan:
You will need to update the lock file too I guess (by actually installing it via `yarn`)

B4nan:
Already did the update in dev branch, so its enough to rebase this now.

B4nan:
apparently there is one more issue with FileCacheAdapter:

```
TypeError: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received type number (123)TypeError [ERR_INVALID_ARG_TYPE]: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received type number (123)
```

Yaty:
alright it seems ok now, thanks for guidance :)

B4nan:
Good, thank you! Let's make this feature commit so it gets to the changelog :]

Issue #535, opened 2020-05-02
Utils.shouldIgnoreProperty() crashes when trying to evaluate a getter


followben:
**Describe the bug**
When using the `@Property()` decorator with `persist: false` on a getter to serialize calculated property in an entity's json, mikro crashes when that entity is fetched in a `find()` method because `shouldIgnoreProperty()` tries to evaluate it.

**Stack trace**
```
at Reference.getEntity (lib/entity/Reference.ts:68:13)
      at MyEntity.get someCalculatedProperty [as someCalculatedProperty] (tests/issues/GHWIP.test.ts:20:19)
      at Function.shouldIgnoreProperty (lib/utils/Utils.ts:156:30)
      at lib/utils/Utils.ts:127:17
          at Array.forEach (<anonymous>)
      at Function.prepareEntity (lib/utils/Utils.ts:126:55)
      at UnitOfWork.merge (lib/unit-of-work/UnitOfWork.ts:45:55)
      at EntityManager.merge (lib/EntityManager.ts:326:26)
      at EntityManager.findOne (lib/EntityManager.ts:178:19)
      at EntityManager.findOneOrFail (lib/EntityManager.ts:204:20)
```

**To Reproduce**
1. Create an entity with getter that uses a relationship and decorate it with `@Property({persist: false})`
2. Attempt to load the entity from the db

**Expected behavior**
The entity should load without a crash as it should not evaluate the getter until it's accessed during serialisation.

**Additional context**

**Versions**

| Dependency | Version |
| - | - |
| node | 13.10.1 |
| typescript | 3.8.3 |
| mikro-orm | 3.6.11 |
| pg | 7.18.2 |

B4nan:
Mmmm and what you have in the getter? I think we need to evaluate it, why is it crashing when its accessed? Maybe you should just add some condition to the getter implementation?

B4nan:
Yeah but agreed we can do what you suggested, we can return early for virtual properties.

followben:
Cheers @B4nan - early return is all that's needed I think. Fix is up at that PR if you want it.

PR #536, opened 2020-05-02
fix(core): avoid evaluating virtual properties on find

followben:
As per #535


followben:
It does actually. `prop.name in entity` is ok, but `entity[prop.name]` is not

B4nan:
Thanks

Issue #527, opened 2020-04-30
MikroORM v4


B4nan:
## Progress

- [x] Single table inheritance (#33, #503)
- [x] Allow adding items to not initialized collections (#489)
- [x] Use absolute path to entity file in static MetadataStorage keys (fix for 'Multiple property decorators' validation issues) (#488)
- [x] Embedded entities (allow in-lining child entity into parent one with prefixed keys) (#514)
- [x] Add support for Node.js 14 (#522)
- [x] Support subqueries in QB (#525)
- [x] Nested conditions in `qb.update()` queries via subqueries (#319, #537)
- [x] Nested conditions in `em.remove()` via subqueries (#492, #538)
- [x] Use custom errors for specific cases (unique constraint violation, db not accessible, ...) (#539, #611)
- [x] Paginator helper or something similar ([doctrine docs](https://www.doctrine-project.org/projects/doctrine-orm/en/latest/tutorials/pagination.html)) (#544)
- [x] Add `groupBy` and `distinct` to `FindOptions` and `FindOneOptions`
- [x] Support computed properties via `@Formula()` decorator (#553)
- [x] Allow the use of TS path mapping (#554)
- [x] Add custom types for blob, array, json (#476)
- [x] Pass entity instance as parameter in `onCreate` and `onUpdate` (#564)
- [x] Cache metadata only with ts-morph provider (#569)
- [x] Allow using different primary key types than ObjectId in mongo driver (#349, #568)
- [x] Lazy scalar properties (allow having props that won't be loaded by default, but can be populated) (#427, #585)
- [x] Support for alternative loading strategies (#440, #556, #600)
- [x] Support external hooks when using EntitySchema (hooks outside of entity) (#516, #614)
- [x] Execute hooks via `EventManager` (adds `EventArgs` to hook params) (#622, #623)
- [x] Add support for flush events (allows to modify change sets) (#637, #642)
- [x] Use `.d.ts` files in `TsMorphMetadataProvider` (#616)
- [x] Ensure correct order of tables in schema generator (#617)
- [x] Ability to specify entities glob instead of entitiesDir glob (#605, #618) 
- [x] Refactor internal dependencies to support Yarn PnP (#645)
- [x] Association scopes/filters ([hibernate docs](https://docs.jboss.org/hibernate/orm/3.6/reference/en-US/html/filters.html)) (#385, #663)
- [x] Add support for entity and property comments (#668)
[...]

PR #554, opened 2020-05-07
feat(cli): allow the use of TS path mapping

AlexAegis:
This enables the use of path aliases in the config file.

Example `tsconfig.json`

```json
{
	"compilerOptions": {
		"paths": {
			"@foo": ["src/foo/index.ts"]
		}
	},
}
```

Example `mikro-orm.config.ts`

```ts
import * as foo from '@foo';

const config = foo.myFancyFunctionThatReturnsAFancyConfig();

export default config;
```

### DIFF
--- a/packages/core/src/utils/ConfigurationLoader.ts
+++ b/packages/core/src/utils/ConfigurationLoader.ts
@@ -56,6 +56,27 @@ export class ConfigurationLoader {
     return paths;
   }
 
+  static async registerTsNode(tsConfigPath = process.cwd() + '/tsconfig.json') {
+    require('ts-node').register({
+      project: tsConfigPath,
+    });
+
+    if (await pathExists(tsConfigPath)) {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const tsConfig = require(tsConfigPath);
+
+      /* istanbul ignore next */
+      const paths = tsConfig?.compilerOptions?.paths;
+
+      if (paths) {
+        require('tsconfig-paths').register({
+          baseUrl: tsConfig.compilerOptions.baseUrl,
+          paths: tsConfig.compilerOptions.paths,
+        });
+      }
+    }
+  }
+
 }
 
 export interface Settings {
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -53,6 +53,7 @@
     "cli-highlight": "^2.1.4",
     "cli-table3": "^0.6.0",
     "fs-extra": "^9.0.0",
+    "tsconfig-paths": "^3.9.0",
     "yargonaut": "^1.1.4",
     "yargs": "^15.3.1"
   }
--- a/packages/cli/src/CLIHelper.ts
+++ b/packages/cli/src/CLIHelper.ts
@@ -39,9 +39,7 @@ export class CLIHelper {
     const settings = await ConfigurationLoader.getSettings();
 
     if (settings.useTsNode) {
-      require('ts-node').register({
-        project: settings.tsConfigPath,
-      });
+      await ConfigurationLoader.registerTsNode(settings.tsConfigPath);
     }
 
     // noinspection HtmlDeprecatedTag
--- a/tests/cli/CLIHelper.test.ts
+++ b/tests/cli/CLIHelper.test.ts
@@ -4,6 +4,8 @@ jest.mock('../../tests/mikro-orm.config.js', () => ({ type: 'mongo', dbName: 'fo
 jest.mock('../../tests/mikro-orm.config.ts', () => ({ type: 'mongo', dbName: 'foo_bar', entitiesDirs: ['.'] }), { virtual: true });
 const pkg = { 'mikro-orm': {} } as any;
 jest.mock('../../tests/package.json', () => pkg, { virtual: true });
+const tsc = { compilerOptions: {} } as any;
+jest.mock('../../tests/tsconfig.json', () => tsc, { virtual: true });
 const cwd = process.cwd;
 (global as any).process.cwd = () => '../../../../tests';
 const log = jest.fn();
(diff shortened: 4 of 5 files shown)


packages/core/src/utils/ConfigurationLoader.ts
```diff
@@ -32,6 +32,14 @@ export class ConfigurationLoader {
     return {};
   }
 
+  static async getTsConfig(): Promise<Dictionary> {
+    if (await pathExists(process.cwd() + '/tsconfig.json')) {
```
B4nan, line 36: we should use the path from the `package.json` settings (the one from `getSettings` method, `tsConfigPath`)

AlexAegis: I see.. Then I think I've found a bug (Assuming the default settings should be applied everywhere). In `CLIHelper.ts` the `configure` function directly calls the `ConfigurationLoader::getSettings` method. But the defaults are set in the constructor of `Configuration`.

`CLIHelper::configure` is also used by the `cli.ts` itself. Shouldn't the defaults supposed to be applied there too?


If not then then it's fine. But then I need to apply the default myself.

Another thing is, that if I want to only `require` the Settings once then I'd have to require the tsconfig there. But it would look nicer in its own method, and anyway, eslint doesn't allow me to require into variables. But I've seen that `await ConfigurationLoader.getSettings()` is used a few times anyway. I suggest memoizing this function, then it would be 100% correct.

B4nan: Not really a bug, you are mixing two things, package.json settings (`CLIHelper.getSettings()`) and ORM configuration (`Configuration` class). They are for different things. I don't want to adjust the ORM config anyhow - overriding stuff based on something from package.json seems just wrong. The whole point of this package.json settings is to allow to configure CLI correctly - so to let user say where the right config is and whether we want to run via ts-node or not. And to do so, one can specify where the tsconfig.json file is. 

All I was asking for is to use the `settings.getTsConfig` as a path for the require, so as a parameter to this function. Usign the default from `Configuration` seems wrong, although thee default is same, the class has nothing to do with CLI itself.

package.json
```diff
@@ -125,6 +125,7 @@
     "semantic-release": "^17.0.4",
     "ts-jest": "^25.4.0",
     "ts-node": "^8.8.2",
+    "tsconfig-paths": "^3.9.0",
```
B4nan, line 128: why do you need it here?

packages/core/src/utils/ConfigurationLoader.ts
```diff
@@ -32,6 +32,16 @@ export class ConfigurationLoader {
     return {};
   }
 
+  static async getTsConfig(): Promise<Dictionary> {
+    const settings = await ConfigurationLoader.getSettings();
```
B4nan, line 36: as said, this should not be here and it should not depend on `Configuration`

what you had there before was ok, i just wanted you to prefer the path from `settings.tsConfigPath` if provided

package.json
AlexAegis: For the tests `jest.spyOn(require('tsconfig-paths'), 'register')`

packages/core/src/utils/ConfigurationLoader.ts
AlexAegis: Whats the problem with using the default value defined there as a fallback? The same one was there, I just didn't want to duplicate it after I saw it defined there already. Or this whole function shouldn't be there?

package.json
B4nan: it is not needed, as it will be hoisted by yarn. by your logic we would have to have everything here, all the db connectors and basically all dependencies of every package.

packages/core/src/utils/ConfigurationLoader.ts
B4nan: This whole function shouldn't be here, if you need it, move it back to CLIHelper, but without that extra call to getSettings. I said you should pass that value via parameter, not call it again. 

Again, what you had there in the first commit was ok, I just wanted to have 1 line change (using the path from already loaded settings object), instead you rewrote it and made it much more complex :]

AlexAegis: I do noticed that it doesn't really make sense to have a method like that at all. And it would make much more sense to just have this included at the config, and I did just that. But I need help because I can't make the test work. `tsconfig.json` should be mocked in there, but it only works if it's required two directory levels more inside, in `ConfigurationLoader.ts`, but if I put it in the `configure` method I can't get it to work.

Also, the require is in an IFEE because eslint doesn't let me require into var. And I don't want to ignore or turn off rules.

package.json
AlexAegis: I've took it out

packages/core/src/utils/ConfigurationLoader.ts
B4nan: Just add eslint ignore comment there, the rule is irrelevant here, by using IIFE (its IIFE, not IFEE :]) you are just getting around it, without fixing the root cause (you are still using `require()`).

B4nan:
I am sorry but I already spend more time giving you reviews than I would spend with implementing this myself. If you are stuck, I can finish this myself, no probs. 

Generally thanks for this, I was planning to do it myself for few months already :]

AlexAegis:
I don't have enough experience with Jest to fix the test, so yes it would be nice to have some help.

> I already spend more time giving you reviews than I would spend with implementing this myself.

This is a little discouraging. I'd like to use this library but there is a lot to implement and I don't want to demand every little change like this, but to contribute myself.

packages/core/src/utils/ConfigurationLoader.ts
AlexAegis: Oh yeah, that was a typo. Before top level async was introduced I've used it a lot.

That's why I asked you what would you prefer, ignore it explicitly or placing it in a function.

B4nan:
> This is a little discouraging. I'd like to use this library but there is a lot to implement and I don't want to demand every little change like this, but to contribute myself.

Please understand that I am spending my own free time with this project. And I already spend 2+ hours with this small change, that I would have done in less then hour. I am spending dozens of hours every week like this. 

So I simply said that I will rather finish this myself than spend another 2+ hours on reviews, because my time is very limited and I have [many](https://github.com/mikro-orm/mikro-orm/issues/527) things to work on. 

I am more than happy to see more contributions, but:
1. we should first agree on what and how (and why) should be changed. Here you send the PR right ahead, and when I asked for one simple change, you started elaborating on the why. I simply can't explain to you every decision I made.
2. you should be able to test things yourself and provide tests for your changes. testing things against the CI is not very welcome - every push you do is sending me email notification. 
3. this project has high standards in terms of code style and test coverage, if you want to contribute, you must adhere to that. and again, I did offer to you I will handle the testing issue myself. 

So I am sorry if you feel discouraged, but please try to understand my point of view as well.

AlexAegis:
This last push is not "testing against the CI" I can run the tests myself. I just incorporated the last requested change you mentioned what I can solve. If you have the time to look at the mock issue I had please take a look at it.

B4nan:
Ok so this was quite tricky, the problem was that `package.json` was required from `ConfigurationLoader` and `tsconfig.json` was required in `CLIHelper`, but both were using the same relative (mocked) `process.cwd()`. So the solution was to actually move the code to `ConfigurationLoader`.

Once again sorry if I offended or discourage you, and thanks for your contribution.

B4nan:
Ok so this was quite tricky, the problem was that `package.json` was required from `ConfigurationLoader` and `tsconfig.json` was required in `CLIHelper`, but both were using the same relative (mocked) `process.cwd()`. So the solution was to actually move the code to `ConfigurationLoader`.

Once again sorry if I offended or discourage you, and thanks for your contribution.

PR #566, opened 2020-05-12
docs: improve code snippets in entity definition section

thekevinbrown:
Currently on https://mikro-orm.io/docs/defining-entities these two blocks aren't styled correctly as the markdown parser doesn't recognise the language hint. Fixing the spacing to match the rest of the file should fix this.


B4nan:
Hmm but this will break the numbering of the ordered list, right?

thekevinbrown:
Not sure to be honest. Is there a place where I can see the rendering on the PR to check?

B4nan:
Nope but you can run the docs on localhost. Just `cd docs && yarn && yarn start`

thekevinbrown:
Good news, looks good to me locally.

Screen Shot 2020-05-12 at 5 57 30 pm

B4nan:
Ok, not intended properly but we can live with that :] Thanks!

PR #568, opened 2020-05-12
feat(mongo): allow using different primary key types than ObjectId

### LINKED ISSUE #349: Allow using different PK type than ObjectId in mongo driver
**Describe the bug**
I want to be able to have `_id` as `string` instead of `ObjectId` for Mongo collections. 

I believe the issue is with `denormalizePrimaryKey` which by default converts data into ObjectId at
https://github.com/mikro-orm/mikro-orm/blob/master/lib/platforms/MongoPlatform.ts#L32

I tried searching, however, I couldn't find any workaround. Maybe it's not a popular use case as such.

**Stack trace**
```
Error: Argument passed in must be a single String of 12 bytes or a string of 24 hex characters
    at new ObjectID (/app/node_modules/bson/lib/bson/objectid.js:59:11)
    at MongoPlatform.denormalizePrimaryKey (/app/node_modules/mikro-orm/dist/platforms/MongoPlatform.js:23:16)
    at EntityFactory.create (/app/node_modules/mikro-orm/dist/entity/EntityFactory.js:22:33)
    at EntityManager.create (/app/node_modules/mikro-orm/dist/EntityManager.js:227:40)
    at EntityRepository.create (/app/node_modules/mikro-orm/dist/entity/EntityRepository.js:75:24)
    at UserService.create (/app/dist/modules/user/user.service.js:27:42)
    at UserController.create (/app/dist/modules/user/user.controller.js:22:33)
    at /app/node_modules/@nestjs/core/router/router-execution-context.js:37:29
    at InterceptorsConsumer.intercept (/app/node_modules/@nestjs/core/interceptors/interceptors-consumer.js:10:20)
    at /app/node_modules/@nestjs/core/router/router-execution-context.js:45:60
    at /app/node_modules/@nestjs/core/router/router-proxy.js:8:23
    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
    at next (/app/node_modules/express/lib/router/route.js:137:13)
    at Route.dispatch (/app/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
    at /app/node_modules/express/lib/router/index.js:281:22
```

**Versions**

| Dependency | Version |
| - | - |
| node | 13.8.0 |
| typescript | 3.8.2 |
| mikro-orm | 3.2.1 |
| mongodb | 3.5.1 |

ayazhussein:
Closes #349

### DIFF
--- a/packages/core/src/entity/EntityFactory.ts
+++ b/packages/core/src/entity/EntityFactory.ts
@@ -1,5 +1,5 @@
 import { Utils } from '../utils';
-import { AnyEntity, EntityData, EntityMetadata, EntityName, Primary } from '../typings';
+import { AnyEntity, EntityData, EntityMetadata, EntityName, EntityProperty, Primary } from '../typings';
 import { UnitOfWork } from '../unit-of-work';
 import { ReferenceType } from './enums';
 import { EntityManager, wrap } from '..';
@@ -23,7 +23,7 @@ export class EntityFactory {
 
     entityName = Utils.className(entityName);
     const meta = this.metadata.get(entityName);
-    meta.primaryKeys.forEach(pk => this.denormalizePrimaryKey(data, pk));
+    meta.primaryKeys.forEach(pk => this.denormalizePrimaryKey(data, pk, meta.properties[pk]));
     const entity = this.createEntity(data, meta);
     const wrapped = wrap(entity, true);
 
@@ -103,12 +103,15 @@ export class EntityFactory {
   /**
    * denormalize PK to value required by driver (e.g. ObjectId)
    */
-  private denormalizePrimaryKey<T extends AnyEntity<T>>(data: EntityData<T>, primaryKey: string): void {
+  private denormalizePrimaryKey<T extends AnyEntity<T>>(data: EntityData<T>, primaryKey: string, metadata: EntityProperty ): void {
     const platform = this.driver.getPlatform();
     const pk = platform.getSerializedPrimaryKeyField(primaryKey);
 
     if (Utils.isDefined(data[pk], true) || Utils.isDefined(data[primaryKey], true)) {
-      const id = platform.denormalizePrimaryKey(data[pk] || data[primaryKey]);
+      let id = data[pk] || data[primaryKey];
+      if(metadata.type.toLowerCase() === 'objectid') {
+        id = platform.denormalizePrimaryKey(id);
+      }
       delete data[pk];
       data[primaryKey as keyof T] = id as Primary<T> & T[keyof T];
     }
--- a/packages/core/src/entity/EntityFactory.ts
+++ b/packages/core/src/entity/EntityFactory.ts
@@ -103,16 +103,17 @@ export class EntityFactory {
   /**
    * denormalize PK to value required by driver (e.g. ObjectId)
    */
-  private denormalizePrimaryKey<T extends AnyEntity<T>>(data: EntityData<T>, primaryKey: string, prop: EntityProperty<T> ): void {
+  private denormalizePrimaryKey<T extends AnyEntity<T>>(data: EntityData<T>, primaryKey: string, prop: EntityProperty<T>): void {
     const platform = this.driver.getPlatform();
     const pk = platform.getSerializedPrimaryKeyField(primaryKey);
 
     if (Utils.isDefined(data[pk], true) || Utils.isDefined(data[primaryKey], true)) {
       let id = data[pk] || data[primaryKey];
 
-      if(prop.type.toLowerCase() === 'objectid') {
+      if (prop.type.toLowerCase() === 'objectid') {
         id = platform.denormalizePrimaryKey(id);
       }
+
       delete data[pk];
       data[primaryKey as keyof T] = id as Primary<T> & T[keyof T];
     }
--- a/packages/core/src/entity/EntityFactory.ts
+++ b/packages/core/src/entity/EntityFactory.ts
@@ -103,13 +103,14 @@ export class EntityFactory {
   /**
    * denormalize PK to value required by driver (e.g. ObjectId)
    */
-  private denormalizePrimaryKey<T extends AnyEntity<T>>(data: EntityData<T>, primaryKey: string, metadata: EntityProperty ): void {
+  private denormalizePrimaryKey<T extends AnyEntity<T>>(data: EntityData<T>, primaryKey: string, prop: EntityProperty<T> ): void {
     const platform = this.driver.getPlatform();
     const pk = platform.getSerializedPrimaryKeyField(primaryKey);
 
     if (Utils.isDefined(data[pk], true) || Utils.isDefined(data[primaryKey], true)) {
       let id = data[pk] || data[primaryKey];
-      if(metadata.type.toLowerCase() === 'objectid') {
+
+      if(prop.type.toLowerCase() === 'objectid') {
         id = platform.denormalizePrimaryKey(id);
       }
       delete data[pk];
--- a/packages/mongodb/src/MongoDriver.ts
+++ b/packages/mongodb/src/MongoDriver.ts
@@ -261,9 +261,11 @@ export class MongoDriver extends DatabaseDriver<MongoConnection> {
 
   private buildFilterById<T extends AnyEntity<T>>(entityName: string, id: string): FilterQuery<T> {
     const meta = this.metadata.get(entityName);
+
     if (meta.properties[meta.primaryKeys[0]].type.toLowerCase() === 'objectid') {
       return { _id: new ObjectId(id) } as FilterQuery<T>;
     }
+
     return { _id: id } as FilterQuery<T>;
   }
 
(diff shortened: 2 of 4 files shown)


packages/mongodb/src/MongoPlatform.ts
```diff
@@ -19,7 +19,11 @@ export class MongoPlatform extends Platform {
   }
 
   denormalizePrimaryKey(data: number | string): IPrimaryKey {
-    return new ObjectId(data);
+    try {
```
B4nan, line 22: this is not really fixing anything, you are just adding a fallback :/ i am sure this won't be very performant.

also this will help with your use case where the string is not convertible to object id, but it wont help with those that are. so it does not fix the issue at all.

package.json
```diff
@@ -105,7 +105,7 @@
     "@types/node": "^13.11.0",
     "@types/pg": "^7.14.3",
     "@types/umzug": "^2.2.3",
-    "@types/uuid": "^7.0.2",
+    "@types/uuid": "7.0.3",
```
B4nan, line 108: drop all those changes, no need for them. same with `yarn.lock`.

ayazhussein:
What do you propose?

packages/mongodb/src/MongoPlatform.ts
ayazhussein: I was thinking that this would be flexible enough to handle cases where the _id could be any type if it's not an object Id. what do you propose?

B4nan:
Correct fix is to handle this based on the metadata. If the property type is not object id, we dont want to convert it. 

It will be much more complex change, I don't know what and where, if I knew, it would be already implemented... But changing the platform methods is wrong I think, we should instead make sure they are not called for properties with non object id type. Also we need tests for the actual issue, where the PK is string, but it is 24 char hex string that can be converted to object id without problem (that is what #349 is about). We also need tests with numeric PK (as numbers can be also converted to object id without problems).

B4nan:
I already did some changes in this manner here: https://github.com/mikro-orm/mikro-orm/commit/86cd0277d048021ee244c3cf3537afc187d69683

ayazhussein:
it doesn't cover all the cases, I did some changes, let me know if they are better

B4nan:
Good, much better! Will need more time to review it properly, but this looks like the correct way to me.

packages/core/src/entity/EntityFactory.ts
```diff
@@ -103,12 +103,15 @@ export class EntityFactory {
   /**
    * denormalize PK to value required by driver (e.g. ObjectId)
    */
-  private denormalizePrimaryKey<T extends AnyEntity<T>>(data: EntityData<T>, primaryKey: string): void {
+  private denormalizePrimaryKey<T extends AnyEntity<T>>(data: EntityData<T>, primaryKey: string, metadata: EntityProperty ): void {
```
B4nan, line 106:
```suggestion
  private denormalizePrimaryKey<T extends AnyEntity<T>>(data: EntityData<T>, primaryKey: string, prop: EntityProperty<T>): void {
```

please use `prop` as the variable name, as it holds metadata for single property (for metadata of single entity we use `meta`, and `metadata` is reserved for the instance of `MetadataStorage`, so all entities metadata).

packages/core/src/entity/EntityFactory.ts
```diff
@@ -103,12 +103,15 @@ export class EntityFactory {
...
     const platform = this.driver.getPlatform();
     const pk = platform.getSerializedPrimaryKeyField(primaryKey);
 
     if (Utils.isDefined(data[pk], true) || Utils.isDefined(data[primaryKey], true)) {
-      const id = platform.denormalizePrimaryKey(data[pk] || data[primaryKey]);
+      let id = data[pk] || data[primaryKey];
+      if(metadata.type.toLowerCase() === 'objectid') {
+        id = platform.denormalizePrimaryKey(id);
+      }
```
B4nan, line 114:
```suggestion
      let id = data[pk] || data[primaryKey];

      if (prop.type.toLowerCase() === 'objectid') {
        id = platform.denormalizePrimaryKey(id);
      }

```

packages/core/src/entity/EntityRepository.ts
```diff
@@ -26,32 +26,32 @@ export class EntityRepository<T extends AnyEntity<T>> {
   }
 
   async findOne(where: FilterQuery<T>, populate?: string[] | boolean, orderBy?: QueryOrderMap): Promise<T | null>;
-  async findOne(where: FilterQuery<T>, populate?: FindOneOptions, orderBy?: QueryOrderMap): Promise<T | null>;
-  async findOne(where: FilterQuery<T>, populate: string[] | boolean | FindOneOptions = [], orderBy?: QueryOrderMap): Promise<T | null> {
+  async findOne(where: FilterQuery<T>, populate?: FindOneOptions<T>, orderBy?: QueryOrderMap): Promise<T | null>;
```
B4nan, line 29: those changes should be already part of `dev` branch, so please rebase and they should disappear.

packages/mongodb/src/MongoDriver.ts
```diff
@@ -243,4 +259,12 @@ export class MongoDriver extends DatabaseDriver<MongoConnection> {
     return data;
   }
 
+  private buildFilterById<T extends AnyEntity<T>>(entityName: string, id: string): FilterQuery<T> {
+    const meta = this.metadata.get(entityName);
+    if (meta.properties[meta.primaryKeys[0]].type.toLowerCase() === 'objectid') {
+      return { _id: new ObjectId(id) } as FilterQuery<T>;
+    }
```
B4nan, line 266:
```suggestion

    if (meta.properties[meta.primaryKeys[0]].type.toLowerCase() === 'objectid') {
      return { _id: new ObjectId(id) } as FilterQuery<T>;
    }

```

packages/core/src/entity/EntityFactory.ts
```diff
@@ -103,12 +103,16 @@ export class EntityFactory {
   /**
    * denormalize PK to value required by driver (e.g. ObjectId)
    */
-  private denormalizePrimaryKey<T extends AnyEntity<T>>(data: EntityData<T>, primaryKey: string): void {
+  private denormalizePrimaryKey<T extends AnyEntity<T>>(data: EntityData<T>, primaryKey: string, prop: EntityProperty<T> ): void {
```
B4nan, line 106:
```suggestion
  private denormalizePrimaryKey<T extends AnyEntity<T>>(data: EntityData<T>, primaryKey: string, prop: EntityProperty<T>): void {
```

packages/core/src/entity/EntityFactory.ts
```diff
@@ -103,12 +103,16 @@ export class EntityFactory {
...
     const platform = this.driver.getPlatform();
     const pk = platform.getSerializedPrimaryKeyField(primaryKey);
 
     if (Utils.isDefined(data[pk], true) || Utils.isDefined(data[primaryKey], true)) {
-      const id = platform.denormalizePrimaryKey(data[pk] || data[primaryKey]);
+      let id = data[pk] || data[primaryKey];
+
+      if(prop.type.toLowerCase() === 'objectid') {
+        id = platform.denormalizePrimaryKey(id);
+      }
```
B4nan, line 115:
```suggestion
      if (prop.type.toLowerCase() === 'objectid') {
        id = platform.denormalizePrimaryKey(id);
      }

```

ayazhussein: maybe we could add this rule to [eslint](https://eslint.org/docs/rules/space-after-keywords)?

B4nan: yeah tbh i was expecting it to already catch spacing issues. if you know what rule is missing there, please add it (in separate PR) or ping me with the name (if there would be lot of changes required in the repo, i would rather handle that myself).

tests/issues/GH349.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  @Property()
+  name: string;
+
+  constructor(name: string) {
+    this.name = name;
+  }
+
+}
+
+describe('GH issue 349', () => {
```
B4nan, line 48: can you please add that extra entity with numeric PK so we can be sure it does work too? otherwise its ready to be merged

ayazhussein: done, once we merge I'll open the eslint rules pull request, I added to new rules:
1. [keyword-spacing](https://eslint.org/docs/rules/keyword-spacing)
2. [space-in-parens](https://eslint.org/docs/rules/space-in-parens)

B4nan:
Good, thanks again!

nickerlan:
@B4nan is it possible that this solution breaks em.refresh() behaviour? When i refresh entity with string id i get `input must be a 24 character hex string, 12 byte Uint8Array, or an integer` which I don't get when I refresh entity with ObjectId.

PR #574, opened 2020-05-14
fix(validation): throw when trying to call `em.flush()` from hooks

### LINKED ISSUE #493: Limitations of lifecycle hooks
**Describe the bug**
Limitations of lifecycle hooks are not documented.
I would like to be able to modify a related entity (`OneToOne`) in `@BeforeCreate` and `@BeforeDelete` hooks. This fails for various reasons:
* `@BeforeCreate` does not have EntityManager
* `@AfterCreate` calling `flush` fails with locking error, just `persist` is not flushed
* `@BeforeDelete` same `flush` and `persist` issue

**Stack trace**
```
(node:464192) UnhandledPromiseRejectionWarning: ValidationError: The optimistic lock on entity SocialActivity failed
    at Function.lockFailed (/home/dschoordsch/eventbadger/badger/node_modules/mikro-orm/dist/utils/ValidationError.js:80:16)
    at ChangeSetPersister.processOptimisticLock (/home/dschoordsch/eventbadger/badger/node_modules/mikro-orm/dist/unit-of-work/ChangeSetPersister.js:60:43)
    at ChangeSetPersister.persistEntity (/home/dschoordsch/eventbadger/badger/node_modules/mikro-orm/dist/unit-of-work/ChangeSetPersister.js:42:20)
    at process._tickCallback (internal/process/next_tick.js:178:7)
```

**To Reproduce**
modify a related entity in `@BeforeDelete` and `flush()`

**Expected behavior**
Ideally there would be a way to modify current entity in `@BeforeCreate` and `@BeforeDelete`, i.e. I could create a related entity and set the reference.
At least the limitations of the hooks should be documented, i.e. don't modify the changeset, don't flush.

**Versions**

| mikro-orm | 3.6.5 |

B4nan:
Hooks are executed inside the commit action of unit of work, after all change
sets are computed. This means that it is not possible to create new entities as
usual from inside the hook. Calling `em.flush()` from hooks will result in
validation error. Calling `em.persist()` can result in undefined behaviour like
locking errors.

> The **internal** instance of `EntityManager` accessible under `wrap(this).__em` is
> not meant for public usage.

Closes #493

### DIFF
--- a/docs/docs/lifecycle-hooks.md
+++ b/docs/docs/lifecycle-hooks.md
@@ -23,3 +23,14 @@ removing entity or entity reference, not when deleting records by query.
 the identity map.
 
 > `@OnInit` is not fired when you create the entity manually via its constructor (`new MyEntity()`)
+
+## Limitations of lifecycle hooks
+
+Hooks are executed inside the commit action of unit of work, after all change 
+sets are computed. This means that it is not possible to create new entities as
+usual from inside the hook. Calling `em.flush()` from hooks will result in 
+validation error. Calling `em.persist()` can result in undefined behaviour like
+locking errors. 
+
+> The **internal** instance of `EntityManager` accessible under `wrap(this).__em` is 
+> not meant for public usage. 
--- a/docs/versioned_docs/version-3.6/lifecycle-hooks.md
+++ b/docs/versioned_docs/version-3.6/lifecycle-hooks.md
@@ -23,3 +23,14 @@ removing entity or entity reference, not when deleting records by query.
 the identity map.
 
 > `@OnInit` is not fired when you create the entity manually via its constructor (`new MyEntity()`)
+
+## Limitations of lifecycle hooks
+
+Hooks are executed inside the commit action of unit of work, after all change 
+sets are computed. This means that it is not possible to create new entities as
+usual from inside the hook. Calling `em.flush()` from hooks will result in 
+validation error. Calling `em.persist()` can result in undefined behaviour like
+locking errors. 
+
+> The **internal** instance of `EntityManager` accessible under `wrap(this).__em` is 
+> not meant for public usage. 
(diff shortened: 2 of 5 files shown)


EvanPurkhiser:
I'm receiving this exception, _without_ using any of the lifecycle hooks.

https://github.com/mikro-orm/mikro-orm/blob/0bd93f83b106dcf5fc15d54968bfef77370a5d07/lib/unit-of-work/UnitOfWork.ts#L109-L111

I don't know much about the mikro-orm internals, but this looks like it's checking more than if you're trying to flush within a lifecycle.

In my case I'm flushing entities asynchronously (which is why I run into this)

https://github.com/EvanPurkhiser/prolink-connect/blob/0bfce80aad1cb0e4e333cf5073a60527e4cd1676/src/localdb/rekordbox.ts#L150-L173

Thought I would leave this comment here, but I'll likely work around it.

B4nan:
> In my case I'm flushing entities asynchronously (which is why I run into this)

You can't flush asynchronously, not with the current implementation. If you want to do that, you need to create separate contexts via `em.fork()` and flush those.

EvanPurkhiser:
Thanks for the quick response @B4nan.

That makes sense. Thanks.

B4nan:
Here is an example of this: https://github.com/mikro-orm/mikro-orm/blob/master/tests/EntityManager.mysql.test.ts#L1507

    const author = new Author2('name', 'email');
    await orm.em.persistAndFlush(author); // we need to flush here so the entity gets inside IM

    // fork EM without clearing the IM (once for each process), so author entity will be there
    await Promise.all([
      orm.em.fork(false).persistAndFlush(new Book2('b1', author)),
      orm.em.fork(false).persistAndFlush(new Book2('b2', author)),
      orm.em.fork(false).persistAndFlush(new Book2('b3', author)),
    ]);

If we did not flush the author outside like this, it would mess up the identity map. The `em.fork(false)` keeps current contents of the IM in the fork, so the author entity is known - otherwise it would produce insert query for it (if the IM would be empty).

PR #584, opened 2020-05-22
Bug: populating m:n relation where only one side hase composite PK causes Error: Expected 1 bindings, saw 2 

whocaresk:
when populating m:n relation in some cases was error, where wrong metadata used to check is there composite PK involved in query, which caused some cryptic errors


B4nan:
Closing this one as its just a repro, open new PR with the full code once you have some spare time :]

PR #586, opened 2020-05-24
chore: add prepublish hook

kibertoad:
refs #527


B4nan:
This needs to go to `dev` brach. Please fix the conflicts.

kibertoad:
@B4nan https://github.com/lerna/lerna/issues/1169
Looks like it needs to be inside each separate package. Should I move it there?

B4nan:
Not sure what is the best practice with lerna. I remember I read somewhere I still need to run `lerna build` manually before publishing, but it looks like `lerna publish` does traverse the packages in the correct order anyway, so having `prepublish` script inside the packages should work just fine and we could skip the manual `lerna build` call. We already check whether all packages are buildable, so it should be safe to do it like this when publishing.

kibertoad:
@B4nan Done.

B4nan:
Thanks!

PR #588, opened 2020-05-27
docs: add section for Relationship Loading Strategy

willsoto:
First pass. Open to feedback of course.

#440


docs/docs/loading-strategy.md
```diff
@@ -0,0 +1,68 @@
...
+  })
+  books!: Collection<Book>;
+}
+
+@Entity()
+export class Book {
+  @ManyToOne({
+    entity: 'Author',
+    cascade: [],
+  })
```
B4nan, line 30: `cascade: []` does disable all cascading, including merge and persist. that is usually not what anyone wants.

I would keep just empty decorator, as with rest of the docs. 

```suggestion
  @ManyToOne()
```

docs/docs/loading-strategy.md
```diff
@@ -0,0 +1,68 @@
...
+```typescript
+import { Entity, LoadStrategy, OneToMany, ManyToOne } from 'mikro-orm';
+
+@Entity()
+export class Author {
+  @OneToMany({
+    entity: () => Book,
+    mappedBy: 'author',
+  })
+  books!: Collection<Book>;
```
B4nan, line 22:
```suggestion
  @OneToMany(() => Book, b => b.author)
  books = new Collection<Book>(this);
```

docs/docs/loading-strategy.md
```diff
@@ -0,0 +1,68 @@
+---
+title: Relationship Loading Strategy
+---
+
```
B4nan, line 4: we should have a warning (formatted as a quote, so via `>`) that this is SQL only feature

docs/docs/loading-strategy.md
```diff
@@ -0,0 +1,68 @@
+---
+title: Relationship Loading Strategy
```
B4nan, line 2: use just `Loading Strategy` in the sidebar (keep this as a full title)

docs/docs/loading-strategy.md
```diff
@@ -0,0 +1,68 @@
...
+import { Entity, LoadStrategy, OneToMany } from 'mikro-orm';
+
+@Entity()
+export class Author {
+  @OneToMany({
+    entity: () => Book,
+    mappedBy: 'author',
+    strategy: LoadStrategy.JOINED,
+  })
+  books!: Collection<Book>;
```
B4nan, line 54:
```suggestion
  @OneToMany({
    entity: () => Book,
    mappedBy: b => b.author,
    strategy: LoadStrategy.JOINED,
  })
  books = new Collection<Book>(this);
```

docs/docs/loading-strategy.md
```diff
@@ -0,0 +1,68 @@
...
+The following will issue **one** SQL statement:
+
+```typescript
+const author = await orm.em.findOne(Author, 1, ['books']);
+```
+
+You can also specify the load strategy as needed. This will override whatever strategy is declared in the mapping:
+
+```typescript
+const author = await orm.em.findOne(Author, 1, [{ field: 'books', strategy: LoadStrategy.JOINED }]);
```
B4nan, line 67: we should also have an example how to do it with nested populate - that is the part I was adding myself recently

```typescript
// one level
const author = await orm.em.findOne(Author, 1, { populate: { books: LoadStrategy.JOINED } });

// two levels
const author = await orm.em.findOne(Author, 1, { populate: {
  books: [LoadStrategy.JOINED, { publisher: LoadStrategy.JOINED }]
} });
```

willsoto:
Feedback applied.

One thing I was thinking was whether or not we should have a section dedicated to describing all the various strategies. I know only 2 are supported now, but if there are plans to support more in the future it could be useful to have them all described up-front. I'm thinking something like [here](https://docs.sqlalchemy.org/en/13/orm/loading_relationships.html)

B4nan:
But that is the page you just added, isn't it? It's called loading strategy, we have 2 of those, and it describes both of them. Sure we could be more verbose if that is what you mean (feel free to do so :])

B4nan:
Maybe we should rename the page to loading strategies? WDYT?

willsoto:
> But that is the page you just added, isn't it? It's called loading strategy, we have 2 of those, and it describes both of them. Sure we could be more verbose if that is what you mean (feel free to do so :])

haha I suppose verbosity is one way to describe what I want. Yes, you are right that it is all there already but I am thinking more in the case that I already know about the feature, but I just need to look up the various supported strategies at a glance. Could be useful, but it was just a thought. I'll think on it more.

> Maybe we should rename the page to loading strategies? WDYT?

I don't have a strong preference either way. I titled it `Relationship Loading Strategies` because I think there are a whole lot of things one can "load" via an ORM, right? This document specifically refers to how _relationships_ get loaded. But that is certainly debatable :)

B4nan:
> I don't have a strong preference either way. I titled it Relationship Loading Strategies because I think there are a whole lot of things one can "load" via an ORM, right? This document specifically refers to how relationships get loaded. But that is certainly debatable :)

The name is fine, what I meant is to use `Loading Strategies` instead of the current `Loading Strategy`, so just singular -> plural :]

willsoto:
> > I don't have a strong preference either way. I titled it Relationship Loading Strategies because I think there are a whole lot of things one can "load" via an ORM, right? This document specifically refers to how relationships get loaded. But that is certainly debatable :)
> 
> The name is fine, what I meant is to use `Loading Strategies` instead of the current `Loading Strategy`, so just singular -> plural :]

Oh! Got it haha

Sure that is fine I think, I named it singular since most (all?) of the titles are singular (like `Naming Strategy`). But yeah I can make that change np

B4nan:
I don't think so :]

https://mikro-orm.io/docs/defining-entities
https://mikro-orm.io/docs/entity-references
https://mikro-orm.io/docs/entity-constructors
https://mikro-orm.io/docs/collections
https://mikro-orm.io/docs/transactions
https://mikro-orm.io/docs/composite-keys
https://mikro-orm.io/docs/decorators

... and more and more :]

willsoto:
Point taken. Will change.

B4nan:
Yes please, change it on all places, including the file name (and therefore url).

willsoto:
Should be good, I just did it via the github UI, I hope you plan on squash merging this too 😅

B4nan:
Sure, no worries, I am squash merging everything (I even accidentally squash merged the v3 PR so the history is totally fucked up with v3 changes being in single commit 🤣)

willsoto:
@B4nan I'm okay not adding that section for right now. You are free to merge this unless there are other changes you'd like made.

B4nan:
It's good to have it there, thanks! :]

PR #597, opened 2020-06-04
fix(cli): custom tsconfig gets wrong path

Atrox:
When using a custom tsconfig it fails on the ` require(tsConfigPath);` because the path is wrong:
`(node:34784) UnhandledPromiseRejectionWarning: Error: Cannot find module 'tsconfig.server.json'`

Before this fix only the default parameter had the correct path.


B4nan:
thanks!

PR #630, opened 2020-07-01
test: fix load embedded entities mysql test

quaternion:
Fix expected and received values in embedded entities mysql test.


B4nan:
Thanks!

PR #645, opened 2020-07-07
feat(core): refactor internal dependencies to support Yarn PnP

merceyz:
**What's the problem this PR addresses?**

MikroORM is currently trying to use dependencies without declaring them which doesn't work under strict dependency environments (Yarn PnP) and makes us rely on hoisting which is really fragile.

Example:
Running `yarn add @mikro-orm/sqlite knex@0.19.0` will produce a tree where `@mikro-orm/sqlite` won't get the same version of `knex` as `@mikro-orm/knex`
```
. -> knex@0.19.0
  -> @mikro-orm/sqlite
  -> @mikro-orm/knex -> knex@0.21.0
```

**How did you fix it?**

Add the undeclared dependencies and make sure the version of knex is required from `@mikro-orm/knex`


merceyz:
Now that the dependencies are declared correctly I can't get the build to pass, most likely because the dependencies are circular. Would love some input on this

B4nan:
I don't follow what your problem is, this definitely nothing mergable - the whole purpose of the `core` package is that it does not depend on anything, what you did is to include packages that will transitively include their dependencies. That totally beats the purpose of the split into multiple packages.

packages/mariadb/package.json
```diff
@@ -50,6 +50,8 @@
     "access": "public"
   },
   "dependencies": {
+    "@mikro-orm/core": "^4.0.0-alpha.4",
+    "@mikro-orm/knex": "^4.0.0-alpha.4",
```
B4nan, line 54: those are already transitive dependencies of `@mikro-orm/mysql-base`

packages/sqlite/package.json
```diff
@@ -49,6 +49,7 @@
   "dependencies": {
     "@mikro-orm/core": "^4.0.0-alpha.4",
     "@mikro-orm/knex": "^4.0.0-alpha.4",
+    "fs-extra": "^9.0.1",
```
B4nan, line 52: i can understand this one

packages/mariadb/package.json
arcanis: Transitive dependencies are not allowed. Consider the following case; which version of C is gonna be hoisted to the top in such a way that A can require it?

```
. -> A -> B -> C@1
       -> D -> C@2
```

B4nan:
> Running yarn add @mikro-orm/sqlite knex@0.19.0 will produce a tree where `@mikro-orm/sqlite` won't get the same version of knex as @mikro-orm/knex

knex should not be installed like this, it is already transitive dependency of `@mikro-orm/knex` (which is a dependency of `@mikro-orm/sqlite`). just like `sqlite3` is already a dependency of `@mikro-orm/sqlite` so you install only `@mikro-orm/sqlite`, not `knex`, not `sqlite3`.

packages/mariadb/package.json
B4nan: i dont follow, how are they not allowed? also i dont follow your example, can you use real package names there? 

i would say the dependent version will be always same, so not `@X` and `@Y`, but always `@X`, the versions will be aligned in each release

B4nan:
you should literally install only `@mikro-orm/sqlite`, at least that was the intension. it already includes `@mikro-orm/core` and `@mikro-orm/knex` as well as `sqlite3`.

@darkbasic thoughts?

packages/core/package.json
```diff
@@ -47,6 +47,9 @@
     "access": "public"
   },
   "dependencies": {
+    "@mikro-orm/entity-generator": "^4.0.0-alpha.4",
+    "@mikro-orm/knex": "^4.0.0-alpha.4",
+    "@mikro-orm/migrations": "^4.0.0-alpha.4",
```
B4nan, line 52: those could be optional peer dependencies, would that help with the probelm?

packages/core/package.json
```diff
@@ -56,6 +59,12 @@
     "fs-extra": "^9.0.1",
     "globby": "^11.0.1",
     "reflect-metadata": "^0.1.13",
+    "tsconfig-paths": "^3.9.0",
```
B4nan, line 62: this is actually again an optional peer dependency - one can use the ORM without `tsconfig-paths`

packages/mariadb/package.json
arcanis: > i dont follow, how are they not allowed?

They are undefined behaviors. Package managers guarantee that a package will be able to access dependencies that match what it explicitly requests. Nothing else. In the example I shared, there is two possible hoistings : either C@1 gets hoisted, or C@2. Which one it is is entirely unspecified, and may change from a package manager to another.

Note that it's not a matter of "but we're not doing this", or "this will never happen". It WILL happen. It already does. Ever wondered why you sometimes needed to trash your node_modules and run fresh installs? Or why a project may not work on npm/Yarn randomly? That's in large part because package boundaries aren't respected. 

Because it's honestly quite difficult to catch, Yarn now enforces boundaries. If your project accesses dependencies in a non-deterministic way, it'll get an exception in return.

@merceyz's diff fixes that by properly listing the dependencies in a way that guarantees that all package managers will always be able to install your project in a way that makes sense per your requirements.

packages/knex/src/index.ts
```diff
@@ -8,3 +8,7 @@ export * from './query';
 export * from './schema';
 export { SqlEntityManager as EntityManager } from './SqlEntityManager';
 export { SqlEntityRepository as EntityRepository } from './SqlEntityRepository';
+export * from './Utils';
+
+import * as Knex from 'knex';
+export { Knex };
```
B4nan, line 14: i am also fine with reexporting knex

packages/mariadb/package.json
B4nan: well, i do not think it does properly list dependencies. `core` package is not dependent of any other ORM package. there is not a single import like that. 

as mentioned in the root thread - most of the changes are actually optional peer dependencies if something (but the intention was different - you should not install `@mikro-orm/knex` or `knex` manually, they should be transitive dependencies of the single package you install - `@mikro-orm/sqlite`)

packages/knex/src/index.ts
darkbasic: This is something I thought about doing as well, because I needed a Knex handle while creating a custom type for postgis and I was forced to manually install knex to avoid extraneous imports.

darkbasic:
> > Running yarn add @mikro-orm/sqlite knex@0.19.0 will produce a tree where `@mikro-orm/sqlite` won't get the same version of knex as @mikro-orm/knex
> 
> knex should not be installed like this, it is already transitive dependency of `@mikro-orm/knex` (which is a dependency of `@mikro-orm/sqlite`). just like `sqlite3` is already a dependency of `@mikro-orm/sqlite` so you install only `@mikro-orm/sqlite`, not `knex`, not `sqlite3`.

I **had** to manually install knex because I needed to create a custom type for postgis and I needed a handle on `Knex`. I couldn't simply import it because it would be undefined behaviour (I have a lint rule which forbids extraneous imports). I think that re-exporting Knex could work for such use cases.

darkbasic:
> @darkbasic thoughts?

Unfortunately I have no experience with yarn pnp because it doesn't play well with tsc, but I agree that you should always declare the dependencies that you use. For example if in the mariadb package you need to access @mikro-orm/core you will have to list it, even if you depend on @mikro-orm/mysql-base which already depends on it: otherwise you will rely on an undefined behaviour because there could be multiple versions racing to be hoisted. Lock files and yarn resolutions help a lot to fix those issues, but they are annoying nonetheless. I personally use an eslint rule to forbid extranous imports, so everytime I try to import something which I didn't declare I direct dependency on I get an error.

B4nan:
As I said, I am fine with reexporting knex, as it makes sense to me. 

I also don't like the usage of `Knex` as a namespace (there is no real reason for that, all the usage is just about types, so nothing that would actually matter, right?).

The bottom line is, `core` package can't depend on anything. Knex is optional transitive dependency of `@mikro-orm/knex`, it should not be installed manually as it is not a peer dependency. 

> For example if in the mariadb package you need to access @mikro-orm/core you will have to list it, even if you depend on @mikro-orm/mysql-base which already depends on it: otherwise you will rely on an undefined behaviour because there could be multiple versions racing to be hoisted. Lock files and yarn resolutions help a lot to fix those issues, but they are annoying nonetheless. I personally use an eslint rule to forbid extranous imports, so every time I try to import something which I didn't declare a direct dependency on I get an error.

But that is the exact purpose of peer dependencies, if we want to ensure there is only one copy and not more. What you are proposing would make it actually worse for the regular users, or not? Every direct dependency would result in a new copy, even if the versions are matching, or not?

B4nan:
As I said, the idea was to import only the driver package, rest should be transitive dependencies, nothing users should install themselves in the root project. Maybe that was wrong (although I am not very convinced), and we should require also the `core` package, and set that as a peer dependency everywhere, that is something for discussion. 

But knex should not be installed like this, it was never meant to be peer dependency. 

Would be great to see a repository demonstrating the problem first, before firing a PR like this (that basically reverts all the outcomes of #475, as with those changes the `core` package is suddenly directly dependent on `knex`, `umzug`, `ts-morph` and `typescript`).

darkbasic:
> But that is the exact purpose of peer dependencies, if we want to ensure there is only one copy and not more. What you are proposing would make it actually worse for the regular users, or not? Every direct dependency would result in a new copy, even if the versions are matching, or not?

I think it's up to the package manager to hoist it or not, but I wouldn't rely on it. If a dep is meant to be a peer dep I agree that it would be better to omit it, but I would still declare it as a *dev* dependency to avoid undefined behaviour in the development environment (won't matter once packaged though). If it's a peer dep there is no chance for ub because the user will be forced to manually install it so there will be no need to hoist it because it already is. Anyway as I said I'm no pnp expert.

merceyz:
> the whole purpose of the core package is that it does not depend on anything

The thing is it does, it just doesn't declare it
https://github.com/mikro-orm/mikro-orm/blob/576e7fe798716521d1116dcad4fb669d939b883f/packages/core/src/MikroORM.ts#L110
https://github.com/mikro-orm/mikro-orm/blob/576e7fe798716521d1116dcad4fb669d939b883f/packages/core/src/MikroORM.ts#L119
https://github.com/mikro-orm/mikro-orm/blob/576e7fe798716521d1116dcad4fb669d939b883f/packages/core/src/MikroORM.ts#L128

> what you did is to include packages that will transitively include their dependencies

There is no guarantee that you'll get access to the transitive dependencies, it is undefined behaviour and just happens to work.

> knex should not be installed like this, it is already transitive dependency of @mikro-orm/knex (which is a dependency of @mikro-orm/sqlite). just like sqlite3 is already a dependency of @mikro-orm/sqlite so you install only @mikro-orm/sqlite, not knex, not sqlite3.

That is besides the point, I'll give a different example:
`yarn add company-acl @mikro-orm/sqlite`
```
. -> company-acl -> knex@0.19.0
  -> @mikro-orm/sqlite -> @mikro-orm/knex -> knex@0.21.0
```
Which version of knex `@mikro-orm/sqlite` gets access to here is undefined behaviour. This is fixed by re-exporting it from `@mikro-orm/knex` and using `createRequire` for the internal parts/patching

> because it doesn't play well with tsc,

@darkbasic If there are issues please report them back to us (yarn) and we'll take a look, it works perfectly fine for us https://github.com/yarnpkg/berry

packages/core/package.json
merceyz: As long as it's declared, yes, but if that happens everything that depends on core needs to declare it as well

B4nan:
During development, the `@mikro-orm/*` paths are translated to relative paths, we don't actually develop against those packages, so where is the undefined behaviour? we developed agains the local files in `packages` folder. 

Btw I guess the logic behind adding new `core` dependencies was those lines? 
https://github.com/mikro-orm/mikro-orm/blob/dev/packages/core/src/MikroORM.ts#L108

If one does not like this dynamic require, don't use the methods and go the other way (create schema generator directly and pass the `orm` instead of asking the `orm` to create it). 

By this logic, we would have to include **every** driver in the core package, as they are all required dynamically: 
https://github.com/mikro-orm/mikro-orm/blob/dev/packages/core/src/utils/Configuration.ts#L234

I hope you all understand that this is not an option. 

> If it's a peer dep there is no chance for ub because the user will be forced to manually install it so there will be no need to hoist it because it already is.

What I had in my mind is not to require user to install the peer deps like knex, they would still be transitive, but the child packages would state it is a peer dependency. So with the knex example, knex would be direct dependency of `@mikro-orm/knex` (as it already is), and a transitive dependency of all its dependent packages (like `@mikro-orm/mysql-base` or `@mikro-orm/mariadb`).

packages/core/package.json
arcanis: From what I understand, the ideal would be something like:

- `@mikro-orm/core` has an optional peer dependency on `@mikro-orm/sqlite`
- `@mikro-orm/sqlite` has a dependency on `@mikro-orm/knex` (already the case)
- Instead of accessing `@mikro-orm/knex`, the Core would instead access `@mikro-orm/sqlite/knex`, which would reexport

Do you think that would make sense, @B4nan ?

merceyz:
> If one does not like this dynamic require, don't use the methods and go the other way (create schema generator directly and pass the orm instead of asking the orm to create it).
By this logic, we would have to include every driver in the core package, as they are all required dynamically:

Didn't even spot that one, seems the solution to core is to require on behalf of the user using `createRequire`

packages/core/package.json
B4nan: i am fine with stating the driver packages as peer deps, i am not sure about the last bullet point

if [those shortcuts in MikroORM.ts](https://github.com/mikro-orm/mikro-orm/blob/dev/packages/core/src/MikroORM.ts#L108) are really a problem, we could get rid of them (i added them just to maintain BC, but as you can see from the implementation, its just about `orm.getSchemaGenerator()` vs `new SchemaGenerator(orm)`)

xeladejo:
I think `@mikro-orm/core` should be a peer dependency the all other packages, because all mikro-orm user need include it in his code to get `MikroORM` or `Entity`. For example `eslint-plugin-node`, with the recommended rules, show a error if the code has some import of a package not listed in `dependencies`.

B4nan:
>     . -> company-acl -> knex@0.19.0
>       -> @mikro-orm/sqlite -> @mikro-orm/knex -> knex@0.21.0
> 
> Which version of knex @mikro-orm/sqlite gets access to here is undefined behaviour. This is fixed by re-exporting it from @mikro-orm/knex and using createRequire for the internal parts/patching

I understand the point with sqlite package now, in general the sql driver packages do not depend on knex, they depend only on `@mikro-orm/knex`, but the problem here is the monkey patching of dialects mainly (as there we need to patch the right version, the one required by `@mikro-orm/knex`)... 

So maybe the right solution is to do the monkey patching in `@mikro-orm/knex` instead? Together with reexporting the `Knex` object from there, so we can use it for typings.

packages/core/package.json
merceyz: I assumed the intention was that the user didn't have to do `yarn add ts-node tsconfig-paths @mikro-orm/core` only `yarn add ts-node @mikro-orm/core` whenever this path is hit https://github.com/mikro-orm/mikro-orm/blob/030ab98bbee9e51e1ce0b73f640030c77932663f/packages/core/src/utils/ConfigurationLoader.ts#L60-L83

B4nan: user needs to install both of them, if they need them (if they want to use them). they are both optional peer deps. if one wants to execute TS code without compiling (like migrations), ts-node is needed, just like you need `tsconfig-paths` if you want to use the TS paths mapping in your entities. 

orm just tries to automatically register those so the user do not have to care about it.

merceyz:
> So maybe the right solution is to do the monkey patching in @mikro-orm/knex instead? Together with reexporting the Knex object from there, so we can use it for typings.

Either is fine, it was easier and cleaner to use the `requireModule` function I added to `@mikro-orm/knex` to require `knex` as `@mikro-orm/knex`. Initially used `createRequire` but it wasn't as clean.

packages/core/package.json
merceyz: Removed as dependencies
- the drivers now have a `getSchemaGenerator` function.
- `@mikro-orm/entity-generator` and `@mikro-orm/migrations` are required on behalf of the user, meaning the user is responsible for running `yarn add @mikro-orm/migrations @mikro-orm/entity-generator` for those two functions to work

packages/core/src/MikroORM.ts
```diff
@@ -106,17 +107,15 @@ export class MikroORM<D extends IDatabaseDriver = IDatabaseDriver> {
...
-    const SchemaGenerator = require('@mikro-orm/knex').SchemaGenerator;
-    return new SchemaGenerator(this.em);
+    return this.driver.getSchemaGenerator(this.em) as T;
   }
 
   /**
    * Gets the EntityGenerator.
    */
   getEntityGenerator<T extends IEntityGenerator = IEntityGenerator>(): T {
     // eslint-disable-next-line @typescript-eslint/no-var-requires
```
B4nan, line 117: we can get rid of the eslint ignore comment. we might want to replace it with instanbul ignore next (or is it easy to test both branches?)

what is the exact purpose of `createRequire || createRequireFromPath`? btw you are missing a space there, in all usages (`createRequire||` vs `createRequire ||`

in general i would like to see that as a helper in the `Utils` class (in `core` package) - is that possible or is it somehow bound the the context (the package where we import from `module`)?

packages/core/package.json
merceyz: Ah, alright. Fixed :+1:

packages/core/src/MikroORM.ts
```diff
@@ -106,17 +107,15 @@ export class MikroORM<D extends IDatabaseDriver = IDatabaseDriver> {
    * Gets the SchemaGenerator.
    */
   getSchemaGenerator<T extends ISchemaGenerator = ISchemaGenerator>(): T {
-    // eslint-disable-next-line @typescript-eslint/no-var-requires
-    const SchemaGenerator = require('@mikro-orm/knex').SchemaGenerator;
-    return new SchemaGenerator(this.em);
+    return this.driver.getSchemaGenerator(this.em) as T;
```
B4nan, line 110: this is actually something i was planning to do anyway, also for entity generator and migrator, to have better abstraction, as i wanted to implement also a mongo migrator and entity generator in near future

packages/core/src/drivers/IDatabaseDriver.ts
```diff
@@ -72,6 +72,7 @@ export interface IDatabaseDriver<C extends Connection = Connection> {
    */
   convertException(exception: Error): DriverException;
 
+  getSchemaGenerator(em: EntityManager): ISchemaGenerator;
```
B4nan, line 75:
```suggestion
  getSchemaGenerator(em: EntityManager): ISchemaGenerator;

```

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -436,6 +437,10 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
     await this.rethrow(qb.execute());
   }
 
+  getSchemaGenerator(em: SqlEntityManager): SchemaGenerator {
```
B4nan, line 440: platform layer would be probably better place to have this

packages/knex/src/Utils.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Require a module on behalf of @mikro-orm/knex
+ */
+export function requireModule(id: string) {
```
B4nan, line 4: please rename the file to `utils.ts`. the convention here is that files always export a class with the same name (`Utils.ts` exports `Utils` class), and this one just exports a function.

packages/mariadb/package.json
B4nan: i am still not convinced by this change. maybe i am wrong, but the way i understand how npm works is that this would result in multiple copies of the `@mikro-orm/knex` and `@mikro-orm/core` (as they would be dependency of the mariadb driver as well as the `mysql-base` package).

B4nan: namely `@mikro-orm/core` is used here _only_ to specify a TS type in one place, it is not used during runtime. and `@mikro-orm/knex` is there only for purpose of reexporting (so reexporting the copy that is made only for the mariadb package sounds not like what we want - we want to reexport the one from `@mikro-orm/mysql-base`).

merceyz: > but the way i understand how npm works is that this would result in multiple copies of the @mikro-orm/knex and @mikro-orm/core (as they would be dependency of the mariadb driver as well as the mysql-base package).

They'll most likely be hoisted to the top as one copy, they currently are which is why it happens works without declaring them.

> all sql driver packages do the reexport of @mikro-orm/knex, so this is not even consistent change, as you did it only for the mariadb one.

Because the other drivers already declares it or don't directly use it

B4nan: we could reexport the whole `mysql-base` instead of `knex` to get around this. and i though we already agreed on `@mikro-orm/core` to be peer dependency.

packages/core/src/MikroORM.ts
merceyz: > we might want to replace it with instanbul ignore next (or is it easy to test both branches?)
> what is the exact purpose of `createRequire || createRequireFromPath`?

`createRequire` was added in v12.2.0, `createRequireFromPath` was added in v10.12.0 and deprecated in v12.2.0 so it isn't possible to always hit both branches, i'll update the ignore comment.
https://nodejs.org/api/modules.html#modules_module_createrequire_filename
https://nodejs.org/api/modules.html#modules_module_createrequirefrompath_filename

> in general i would like to see that as a helper in the Utils class (in core package) - is that possible or is it somehow bound the the context (the package where we import from module)?

That's fully possible

packages/knex/src/Utils.ts
merceyz: Removed, used the helper I added in core instead

packages/core/src/MikroORM.ts
merceyz: Helper has been added to core

packages/core/src/MikroORM.ts
```diff
@@ -106,26 +106,22 @@ export class MikroORM<D extends IDatabaseDriver = IDatabaseDriver> {
...
   }
 
   /**
    * Gets the EntityGenerator.
    */
   getEntityGenerator<T extends IEntityGenerator = IEntityGenerator>(): T {
-    // eslint-disable-next-line @typescript-eslint/no-var-requires
-    const EntityGenerator = require('@mikro-orm/entity-generator').EntityGenerator;
+    const { EntityGenerator } = Utils.requireFrom('@mikro-orm/entity-generator', this.config.get('baseDir'));
```
B4nan, line 116: `baseDir` can be a subfolder of the actual project root (where `node_modules` folder is), is that fine with this approach? maybe it would be safe to add `rootDir` option that would default to `process.cwd()` for this purpose, which would be in most cases not touched by users

packages/core/src/utils/Utils.ts
```diff
@@ -624,4 +625,9 @@ export class Utils {
     return global[key];
   }
 
+  static requireFrom(id: string, from: string) {
```
B4nan, line 628: would be good to add a jsdoc with a brief summary of why this is needed

packages/core/src/MikroORM.ts
merceyz: > baseDir can be a subfolder of the actual project root (where node_modules folder is), is that fine with this approach?

Yes, that's perfectly fine. That's what it is in the project i'm testing this on as well.

packages/knex/src/AbstractSqlDriver.ts
merceyz: Moved :+1:

packages/mariadb/package.json
merceyz: > i though we already agreed on @mikro-orm/core to be peer dependency.

core is now a peer dependency

> we could reexport the whole mysql-base instead of knex to get around this

Not quite sure what you mean here, could you give an example?

B4nan: `@mikro-orm/knex` is currently used (in the mariadb package) only on one place - in the `src/index.ts`, where we reexport everything (so things from `@mikro-orm/knex` can be also imported from the `mariadb` package)

so if that is the only reason why we need it as a hard dependency, then we can reexport the whole `@mikro-orm/mysql-base`, which already includes `@mikro-orm/knex` (plus some MySQL driver classes, but i think its fine to export those)

so in the `src/index.ts` of mariadb we would do 

```diff
-export * from '@mikro-orm/knex';
+export * from '@mikro-orm/mysql-base';
```

which already is a hard dependency. this way we will be reexporting the same knex as the one that mysql-base is using, always, right?

merceyz: Updated to use `mysql-base` instead, it makes the patching a bit more verbose but I don't mind.

> this way we will be reexporting the same knex as the one that mysql-base is using, always, right?

Correct

packages/mariadb/src/MariaDbConnection.ts
```diff
@@ -1,17 +1,19 @@
...
-import { MySqlConnectionConfig } from 'knex';
-import { MySqlConnection } from '@mikro-orm/mysql-base';
+import { MySqlConnection, Knex } from '@mikro-orm/mysql-base';
+import { Utils } from '@mikro-orm/core';
 
-// @ts-ignore
-import Dialect from 'knex/lib/dialects/mysql/index.js';
+const Dialect = Utils.requireFrom(
+	'knex/lib/dialects/mysql/index.js',
+	require.resolve('@mikro-orm/knex', { paths: [require.resolve('@mikro-orm/mysql-base')] })
```
B4nan, line 7: while i was a long time tabs advocate, this project uses spaces for indentation so lets be consistent :] 

```suggestion
  'knex/lib/dialects/mysql/index.js',
  require.resolve('@mikro-orm/knex', { paths: [require.resolve('@mikro-orm/mysql-base')] })
```

merceyz:
> The code is LGTM, I feel like it now makes all sense to me, thanks for bearing with me

Perfect, no worries :)

packages/core/src/utils/Utils.ts
```diff
@@ -624,4 +625,14 @@ export class Utils {
...
   }
 
+  /**
+   * Require a module from a specific location
+   * @param id The module to require
+   * @param from Location to start the node resolution
+   */
+  static requireFrom(id: string, from: string) {
+    /* istanbul ignore next */
+    return (createRequire || createRequireFromPath)(Utils.absolutePath(from))(id);
```
merceyz, line 635: `Utils.absolutePath` doesn't produce a absolute path, which is what I expect it to. Is this a bug or do I missunderstand its use?

B4nan: Hmm not 100% sure now, maybe bad naming, I would say it should do it all the time, but it might be some quirk for testing.

What value is causing issues?

merceyz: Look at the failing tests, it's passing `../../../../../../../tests/tsconfig.json`

B4nan: thats probably caused by mocked `process.cwd()`. looking at the code, and it is just about prepending `baseDir` if the path is relative, so if the `baseDir` is relative, the result will be also relative.

B4nan:
Changed the title as I dont consider this as a fix (v4 is not yet released), hope it makes sense...

packages/core/src/utils/Utils.ts
merceyz: I'm not quite sure how to fix this, could you give me some pointers?

B4nan: either use `path.resolve` or adjust the method to use it (but as i said, it might be intentional, although it is not very probable and would be a quirk)

B4nan: looks like `path.resolve` also counts on `process.cwd` returning already an absolute path :D

merceyz: It does, seems it's changed to always be relative in here https://github.com/mikro-orm/mikro-orm/blob/767aa91c2fe761934c1675baf40b48fbc2992a7e/tests/cli/CLIHelper.test.ts

B4nan: instead of mocking the actual ts-node and tsconfig-paths `register` methods, i mocked the `Utils.requireFrom()` helper to get around that. cli helper tests should be all passing now.

B4nan: the commit is here: https://github.com/mikro-orm/mikro-orm/commit/4d412d7f1427015cf5ad6d432df264ccc6017fd4

B4nan: All tests are passing with that fix, we just need to cover few missing lines and branches, if you would struggle with that, let me know and I can finish that too. 

Other than that this is ready to be merged 👍

merceyz: You haven't pushed to my fork but to your own branch, could you fix that?

> if you would struggle with that, let me know and I can finish that too.

I'd figure it out, but you can have the pleasure of doing it if you want to :)

B4nan: I wont have much time for this today, also cant push to your branch now, already away from pc.

packages/cli/src/CLIHelper.ts
```diff
@@ -124,15 +124,12 @@ export class CLIHelper {
   }
 
   static async getModuleVersion(name: string): Promise<string> {
-    const path = process.cwd() + '/node_modules/' + name + '/package.json';
-
-    if (await pathExists(path)) {
-      // eslint-disable-next-line @typescript-eslint/no-var-requires
-      const pkg = require(path);
+    try {
+      const pkg = Utils.requireFrom(`${name}/package.json`, process.cwd());
```
B4nan, line 128: looks like this is not working

image

B4nan: apparently it does not work without specifying the path to `node_modules`, changing the line to this works:


```typescript
  static async getModuleVersion(name: string): Promise<string> {
    try {
      const pkg = Utils.requireFrom(`${name}/package.json`, process.cwd() + '/node_modules');
      return chalk.green(pkg.version);
    } catch {
      return chalk.red('not-found');
    }
  }
```

merceyz: Made a slight change to the helper, could you try again?

B4nan: yeah that helps, good. this is how i changed the test so it uses the real `cwd`:

```typescript
  test('getModuleVersion', async () => {
    (global as any).process.cwd = cwd;
    const pathExistsMock = jest.spyOn(require('fs-extra'), 'pathExists');
    pathExistsMock.mockResolvedValue(false);
    await expect(CLIHelper.getModuleVersion('pg')).resolves.not.toBe('not-found');
    await expect(CLIHelper.getModuleVersion('does-not-exist')).resolves.toBe('not-found');
    pathExistsMock.mockRestore();
    (global as any).process.cwd = () => '../../../../tests';
  });
```

then there is one more uncovered line, which is the default `Platform.getSchemaGenerator()` that throws, other than that this is ready to be merged

merceyz: Want me to update the test with that change?

> then there is one more uncovered line, which is the default Platform.getSchemaGenerator() that throws, other than that this is ready to be merged

Any specific place you'd like to have that test?

B4nan: yes please, update it. and the platform test should go to `SchemaGenerator.test.ts`, we already had similar one there (removed it as the place where we have thrown was removed too)

```typescript
  test('not supported [mongodb]', async () => {
    const mongoOrm = Object.create(MikroORM.prototype, { driver: new MongoDriver(new Configuration({} as any, false)) } as any);
    expect(() => mongoOrm.getSchemaGenerator()).toThrowError('Not supported by given driver');
  });
```

merceyz: Done :+1:

B4nan:
Thanks again!

merceyz:
No problem, thanks for this great ORM :+1: 
I've been using v3 in production at work for about three months now and it's been a pleasure to work with.

B4nan:
We could do `export { Knex, knex } from 'knex';` in v5, where knex is up to date. 

Btw I believe that WKT problem can be now solved without using knex raw, see https://mikro-orm.io/docs/custom-types/#advanced-example---pointtype-and-wkt

xeladejo:
Currently I am using custom types, but previously I was using this:
```typescript
import { Knex } from "@mikro-orm/knex";

const kx = Knex.default({ client: "mysql" });
```

PR #651, opened 2020-07-10
docs: improve primary key decorator section

0ttik:
Since mikro-orm has now composite PK support it's not true that every entity must have exactly one PK. 

Also I clarified a little bit about 'auto_increment' prop in MySQL (maybe all SQL drivers tho, I haven't checked that) since I assume it's not 100% clear, when it's set to 'auto_increment' and when it's not.

### DIFF
--- a/docs/docs/decorators.md
+++ b/docs/docs/decorators.md
@@ -70,7 +70,9 @@ registered = false;
 > `@PrimaryKey()` decorator extend the `@Property()` decorator, so you can use all 
 > its parameters.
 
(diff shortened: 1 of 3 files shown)


B4nan:
It works like that with all SQL drivers, also please update the next docs, as those changes are made only to the current version which will be overridden once v4 is merged.

B4nan:
Thanks!

0ttik:
Happy to leave my 1 penny here. Mikro-orm looks like a cool thing.

PR #656, opened 2020-07-14
refactor: move monkey patchable portions of Knex to Knex package

thekevinbrown:
Moves the portions of Knex that are being monkey patched from dynamic requires to hard requires so webpack can trace them. They're optional, but each driver insists that they're received.

### DIFF
--- a/packages/sqlite/src/SqliteConnection.ts
+++ b/packages/sqlite/src/SqliteConnection.ts
@@ -1,9 +1,7 @@
 import { ensureDir, readFile } from 'fs-extra';
 import { dirname } from 'path';
-import { AbstractSqlConnection, Knex } from '@mikro-orm/knex';
-import { Utils } from '@mikro-orm/core';
+import { AbstractSqlConnection, Knex, MonkeyPatchable } from '@mikro-orm/knex';
 
-const Dialect = Utils.requireFrom('knex/lib/dialects/sqlite3/index.js', require.resolve('@mikro-orm/knex'));
 
 export class SqliteConnection extends AbstractSqlConnection {
 
@@ -56,8 +54,9 @@ export class SqliteConnection extends AbstractSqlConnection {
    * monkey patch knex' sqlite Dialect so it returns inserted id when doing raw insert query
    */
   private getPatchedDialect() {
-    const processResponse = Dialect.prototype.processResponse;
-    Dialect.prototype.processResponse = (obj: any, runner: any) => {
+    const { Sqlite3Dialect } = MonkeyPatchable;
+    const processResponse = Sqlite3Dialect.prototype.processResponse;
+    Sqlite3Dialect.prototype.processResponse = (obj: any, runner: any) => {
       if (obj.method === 'raw' && obj.sql.trim().match('^insert into|update|delete')) {
         return obj.context;
       }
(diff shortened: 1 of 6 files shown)


B4nan:
@merceyz @arcanis will this work with yarn pnp?

merceyz:
Yes, it will work perfectly fine. I don't quite see the point of the try catch though as the files will exist, which also makes the null checks unnecessary.

B4nan:
Agreed, the whole knex will always be there, so it is safe to require all of it

B4nan:
should be even possible to use `import` instead of `require`, right?

thekevinbrown:
I wasn't sure if importing those knex files would also in turn hard require an optional dependency that users don't always have installed. For example if I'm using Postgres, and I require `'knex/lib/dialects/mysql'`, does that throw if I don't have the mysql package installed? If that's not the case then we don't need the try catches, I thought it might cause problems.

I'm wasn't sure about using `import` because it returns a promise, and we can't await before exporting, so we'd need to change the drivers to await the actual import. Can make that change if that's preferable.

B4nan:
> I wasn't sure if importing those knex files would also in turn hard require an optional dependency that users don't always have installed. For example if I'm using Postgres, and I require 'knex/lib/dialects/mysql', does that throw if I don't have the mysql package installed? If that's not the case then we don't need the try catches, I thought it might cause problems.

Looking at the source code, requiring just the dialect won't load the connector, so it should be safe. 

> I'm wasn't sure about using import because it returns a promise, and we can't await before exporting, so we'd need to change the drivers to await the actual import. Can make that change if that's preferable.

I mean imports like `import * as ... from '...'`

thekevinbrown:
Ah yup, totally can. I’ll make those changes tomorrow.

thekevinbrown:
@B4nan, all done, let me know if there's anything else on this one that needs doing.

packages/sqlite/src/SqliteConnection.ts
```diff
@@ -56,16 +54,21 @@ export class SqliteConnection extends AbstractSqlConnection {
    * monkey patch knex' sqlite Dialect so it returns inserted id when doing raw insert query
    */
   private getPatchedDialect() {
-    const processResponse = Dialect.prototype.processResponse;
-    Dialect.prototype.processResponse = (obj: any, runner: any) => {
+    const { Sqlite3Dialect } = MonkeyPatchable;
+    if (!Sqlite3Dialect) {
```
merceyz, line 58: These checks shouldn't be needed as the object is guaranteed to exist

thekevinbrown: Yup, my bad, I'll fix that.

thekevinbrown: Should be good to go now. Thanks!

thekevinbrown:
Actually, this isn't working for me: "Cannot destructure property 'PostgresDialectTableCompiler' of 'knex_1.MonkeyPatchable' as it is undefined."

Let me keep working and see what I can come up with.

B4nan:
Looking good, we just need to cover that one last branch in the `Utils.ts`, as apparently now we never call `Utils.requireFrom()` with a path to directory in the second param (I'd vote for adding simple unit test for that method into `Utils.test.ts` instead of refactoring the method to the minimum).

merceyz:
Just a note, when you force push it makes it really hard to see what changed so we need to go over all of the code again, the PR will be squash merged so you don't have to worry about keeping one commit

thekevinbrown:
How does that work with the angular commit messages?

B4nan:
There will be only one commit in the master (after merging), and that is what will be used for changelog.

B4nan:
Thanks again! And don't forget to let me know if we need to adjust the webpack docs (if there are changes needed in the config mainly).

PR #662, opened 2020-07-17
refactor(cli): warn users when cache is disabled

thekevinbrown:


B4nan:
This will require some changes in tests

PR #674, opened 2020-07-20
refactor(mysql-base): add missing optional peer dependency

merceyz:
**What's the problem this PR addresses?**

`@mikro-orm/mysql-base` is missing a optional peer dependency on `mysql2` which it inherits from `@mikro-orm/knex` (added in https://github.com/mikro-orm/mikro-orm/pull/673)

Continues https://github.com/mikro-orm/mikro-orm/pull/645 and https://github.com/mikro-orm/mikro-orm/pull/673

**How did you fix it?**

Add undeclared optional peer dependency `mysql2` to `

### DIFF
--- a/packages/mysql-base/package.json
+++ b/packages/mysql-base/package.json
@@ -54,14 +54,10 @@
   },
   "peerDependencies": {
     "@mikro-orm/core": "^4.0.0-alpha.6",
-    "mysql2": "^2.1.0",
-    "mariadb": "^2.4.0"
+    "mysql2": "^2.1.0"
   },
   "peerDependenciesMeta": {
     "mysql2": {
-      "optional": true
-    },
-    "mariadb": {
       "optional": true
     }
   }
--- a/packages/mysql-base/package.json
+++ b/packages/mysql-base/package.json
@@ -54,10 +54,14 @@
   },
   "peerDependencies": {
     "@mikro-orm/core": "^4.0.0-alpha.6",
-    "mysql2": "^2.1.0"
+    "mysql2": "^2.1.0",
+    "mariadb": "^2.4.0"
   },
   "peerDependenciesMeta": {
     "mysql2": {
+      "optional": true
+    },
+    "mariadb": {
       "optional": true
     }
   }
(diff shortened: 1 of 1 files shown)


merceyz:
With these changes (and https://github.com/mikro-orm/mikro-orm/pull/645) MikroORM works under Yarn PnP for both SQLite(https://github.com/mikro-orm/mikro-orm/pull/673) and MySQL

packages/mysql-base/package.json
```diff
@@ -53,6 +53,12 @@
     "@mikro-orm/core": "^4.0.0-alpha.6"
   },
   "peerDependencies": {
-    "@mikro-orm/core": "^4.0.0-alpha.6"
+    "@mikro-orm/core": "^4.0.0-alpha.6",
+    "mysql2": "^2.1.0"
```
B4nan, line 57: in this way, `mysql-base` have actually two optional peer deps - the second being `mariadb`

merceyz: I looked into it before sending the PR but couldn't find the require for it

B4nan: its required via this callback (by knex)

https://github.com/mikro-orm/mikro-orm/blob/dev/packages/mariadb/src/MariaDbConnection.ts#L20

merceyz: That require is in the context of that file, even though it's a callback. So if that's the only instance, there is no need for the peer dependency on mariadb

B4nan: should be the only place afaik

merceyz: Perfect, reverted

B4nan: so do i get it right that `require` in each file is basically defined as `const require = createRequire(__filename)`?

merceyz: You can look at it that way, each module(file) has its own require function that is specific to that module since node needs to know where to start searching for files, which is why relative imports work

The polyfill for `createRequire` looks like this:
```js
module.exports =
  Module.createRequire ||
  Module.createRequireFromPath ||
  function(filename) {
    const mod = new Module(filename, null)

    mod.filename = filename
    mod.paths = Module._nodeModulePaths(path.dirname(filename))
    mod._compile(`module.exports = require;`, filename)

    return mod.exports
  }
```

PR #679, opened 2020-07-22
refactor: test for Reference via __reference instead of instanceof #679

### LINKED ISSUE #676: Circular dependency in v4 causes problems with Webpack
**Describe the bug**

[Reference requires wrap](https://github.com/mikro-orm/mikro-orm/blob/dev/packages/core/src/entity/Reference.ts#L2), then [wrap immediately requires Reference](https://github.com/mikro-orm/mikro-orm/blob/dev/packages/core/src/entity/wrap.ts#L3).

This is a circular dependency, which is undefined behaviour.

This seems to work on vanilla Node, but in a Webpack context, Reference doesn't correctly get set to the same object everywhere, resulting in `TypeError: Cannot read property '__meta' of undefined`. This is because wrap doesn't see that the entity is an `instanceof Reference`, resulting in it falling through wrap without doing what's intended.

I added this to wrap:

```typescript
    console.log('entity: ', entity);
    console.log('entity instanceof Reference_1.Reference', entity instanceof Reference_1.Reference);
    console.trace();
    console.log('--------------------');
```

Then called `findAndCount` once. Both the successful output and the error output happen from the same call to `findAndCount`.

Successful output:
```
entity:  Reference { entity: Location { id: '4' } }
entity instanceof Reference_1.Reference true
Trace: 
    at Object.wrap (webpack:///[project dir]/node_modules/@mikro-orm/core/entity/wrap.js?:17:13)
    at Function.autoWireOneToOne (webpack:///[project dir]/node_modules/@mikro-orm/core/entity/EntityAssigner.js?:65:30)
    at ObjectHydrator.hydrateManyToOne (webpack:///[project dir]/node_modules/@mikro-orm/core/hydration/ObjectHydrator.js?:79:37)
    at ObjectHydrator.hydrateProperty (webpack:///[project dir]/node_modules/@mikro-orm/core/hydration/ObjectHydrator.js?:10:18)
    at ObjectHydrator.hydrate (webpack:///[project dir]/node_modules/@mikro-orm/core/hydration/Hydrator.js?:20:18)
    at EntityFactory.create ([project dir]/node_modules/@mikro-orm/core/entity/EntityFactory.js:27:27)
    at SqlEntityManager.merge ([project dir]/node_modules/@mikro-orm/core/EntityManager.js:258:80)
    at SqlEntityManager.find ([project dir]/node_modules/@mikro-orm/core/EntityManager.js:78:33)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Promise.all (index 0)
    at SqlEntityManager.findAndCount ([project dir]/node_modules/@mikro-orm/core/EntityManager.js:133:35)
(our code)
--------------------
```

Unsuccessful output:
```
entity:  Reference { entity: Organisation { id: '2' } }
entity instanceof Reference_1.Reference false
Trace: 
    at Object.wrap (webpack:///[project dir]/node_modules/@mikro-orm/core/entity/wrap.js?:17:13)
    at new Reference (webpack:///[project dir]/node_modules/@mikro-orm/core/entity/Reference.js?:9:32)
    at Function.create (webpack:///[project dir]/node_modules/@mikro-orm/core/entity/Reference.js?:29:16)
    at Function.wrapReference (webpack:///[project dir]/node_modules/@mikro-orm/core/utils/Utils.js?:409:39)
    at Shop.set [as organisation] (webpack:///[project dir]/node_modules/@mikro-orm/core/entity/EntityHelper.js?:99:56)
    at Function.assignReference ([project dir]/node_modules/@mikro-orm/core/entity/EntityAssigner.js:89:31)
    at [project dir]/node_modules/@mikro-orm/core/entity/EntityAssigner.js:33:39
    at Function.assign ([project dir]/node_modules/@mikro-orm/core/entity/EntityAssigner.js:20:27)
    at SqlEntityManager.merge ([project dir]/node_modules/@mikro-orm/core/EntityManager.js:261:33)
    at SqlEntityManager.find ([project dir]/node_modules/@mikro-orm/core/EntityManager.js:78:33)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Promise.all (index 0)
    at SqlEntityManager.findAndCount ([project dir]/node_modules/@mikro-orm/core/EntityManager.js:133:35)
(our code)
--------------------
```

**Stack trace**
```
TypeError: Cannot read property '__meta' of undefined
    at new Reference (webpack:///[project dir]/node_modules/@mikro-orm/core/entity/Reference.js?:10:17)
    at Function.create (webpack:///[project dir]/node_modules/@mikro-orm/core/entity/Reference.js?:29:16)
    at Function.wrapReference (webpack:///[project dir]/node_modules/@mikro-orm/core/utils/Utils.js?:409:39)
    at Shop.set [as organisation] (webpack:///[project dir]/node_modules/@mikro-orm/core/entity/EntityHelper.js?:99:56)
    at Function.assignReference ([project dir]/node_modules/@mikro-orm/core/entity/EntityAssigner.js:89:31)
    at [project dir]/node_modules/@mikro-orm/core/entity/EntityAssigner.js:33:39
    at Function.assign ([project dir]/node_modules/@mikro-orm/core/entity/EntityAssigner.js:20:27)
    at SqlEntityManager.merge ([project dir]/node_modules/@mikro-orm/core/EntityManager.js:261:33)
    at SqlEntityManager.find ([project dir]/node_modules/@mikro-orm/core/EntityManager.js:78:33)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Promise.all (index 0)
    at SqlEntityManager.findAndCount ([project dir]/node_modules/@mikro-orm/core/EntityManager.js:133:35)
(our code)
```

**To Reproduce**

I will provide a reproduction case, but I haven't had time to yet and wanted to get the issue some visibility so we can work towards eliminating circular dependencies in the project in general.

**Expected behavior**

Expected no circular dependency to exist as it's undefined behaviour.

**Versions**

| Dependency | Version |
| - | - |
| node | 12 |
| typescript | 3.9.5 |
| mikro-orm | 4.0.0-alpha.6 |
| your-driver | Postgresql |

thekevinbrown:
I think this completely fixes #676. Thanks for bearing with me.

### DIFF
--- a/packages/core/src/entity/Reference.ts
+++ b/packages/core/src/entity/Reference.ts
@@ -30,7 +30,7 @@ export class Reference<T> {
   }
 
   static create<T, PK extends keyof T>(entity: T | IdentifiedReference<T, PK>): IdentifiedReference<T, PK> {
-    if (entity instanceof Reference) {
+    if (this.isReference(entity)) {
       return entity;
     }
 
--- a/packages/core/src/entity/Reference.ts
+++ b/packages/core/src/entity/Reference.ts
@@ -30,8 +30,8 @@ export class Reference<T> {
   }
 
   static create<T, PK extends keyof T>(entity: T | IdentifiedReference<T, PK>): IdentifiedReference<T, PK> {
-    if (this.isReference(entity)) {
-      return entity;
+    if (Reference.isReference(entity)) {
+      return entity as IdentifiedReference<T, PK>;
     }
 
     return new Reference(entity) as IdentifiedReference<T, PK>;
--- a/packages/core/src/entity/Reference.ts
+++ b/packages/core/src/entity/Reference.ts
@@ -41,7 +41,7 @@ export class Reference<T> {
    * Checks whether the argument is instance or `Reference` wrapper.
    */
   static isReference<T extends AnyEntity<T>>(data: any): data is Reference<T> {
-    return data instanceof Reference;
+    return data && !!data.__reference;
   }
 
   /**
--- a/packages/core/src/unit-of-work/UnitOfWork.ts
+++ b/packages/core/src/unit-of-work/UnitOfWork.ts
@@ -474,7 +474,7 @@ export class UnitOfWork {
     const reference = entity[prop.name] as U;
 
     if (Reference.isReference(reference)) {
-      return reference.unwrap() as any as U;
+      return reference.unwrap() as unknown as U;
     }
 
     return reference;
--- a/packages/core/src/unit-of-work/UnitOfWork.ts
+++ b/packages/core/src/unit-of-work/UnitOfWork.ts
@@ -473,8 +473,8 @@ export class UnitOfWork {
   private unwrapReference<T extends AnyEntity<T>, U extends AnyEntity | Reference<T> | Collection<AnyEntity> | Primary<T> | (AnyEntity | Primary<T>)[]>(entity: T, prop: EntityProperty<T>): U {
     const reference = entity[prop.name] as U;
 
-    if (reference instanceof Reference) {
-      return reference.unwrap();
+    if (Reference.isReference(reference)) {
+      return reference.unwrap() as any as U;
     }
 
     return reference;
--- a/packages/core/src/unit-of-work/UnitOfWork.ts
+++ b/packages/core/src/unit-of-work/UnitOfWork.ts
@@ -474,7 +474,7 @@ export class UnitOfWork {
     const reference = entity[prop.name] as U;
 
     if (Reference.isReference(reference)) {
-      return reference.unwrap() as unknown as U;
+      return reference.unwrap() as U;
     }
 
     return reference;
--- a/packages/core/src/entity/EntityAssigner.ts
+++ b/packages/core/src/entity/EntityAssigner.ts
@@ -80,7 +80,7 @@ export class EntityAssigner {
     const prop2 = meta2.properties[prop.inversedBy || prop.mappedBy];
 
     if (prop2 && !entity[prop.name][prop2.name]) {
-      if (entity[prop.name] instanceof Reference) {
+      if (Reference.isReference(entity[prop.name])) {
         entity[prop.name].unwrap()[prop2.name] = Reference.wrapReference(entity, prop2);
       } else {
         entity[prop.name][prop2.name] = Reference.wrapReference(entity, prop2);
--- a/packages/core/src/entity/EntityLoader.ts
+++ b/packages/core/src/entity/EntityLoader.ts
@@ -194,7 +194,7 @@ export class EntityLoader {
     for (const entity of entities) {
       if (Utils.isEntity(entity[populate.field])) {
         children.push(entity[populate.field]);
-      } else if (entity[populate.field] instanceof Reference) {
+      } else if (Reference.isReference(entity[populate.field])) {
         children.push(entity[populate.field].unwrap());
       } else if (entity[populate.field] as unknown instanceof Collection) {
         children.push(...entity[populate.field].getItems());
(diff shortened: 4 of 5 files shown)


packages/core/src/entity/Reference.ts
```diff
@@ -41,7 +41,7 @@ export class Reference<T> {
    * Checks whether the argument is instance or `Reference` wrapper.
    */
   static isReference<T extends AnyEntity<T>>(data: any): data is Reference<T> {
-    return data instanceof Reference;
+    return data && !!data.__reference;
```
B4nan, line 44: weird that this is actually what solves it :]

packages/core/src/unit-of-work/UnitOfWork.ts
```diff
@@ -473,8 +473,8 @@ export class UnitOfWork {
   private unwrapReference<T extends AnyEntity<T>, U extends AnyEntity | Reference<T> | Collection<AnyEntity> | Primary<T> | (AnyEntity | Primary<T>)[]>(entity: T, prop: EntityProperty<T>): U {
     const reference = entity[prop.name] as U;
 
-    if (reference instanceof Reference) {
-      return reference.unwrap();
+    if (Reference.isReference(reference)) {
+      return reference.unwrap() as any as U;
```
B4nan, line 477: same here, as we are still referencing the `Reference` class

```suggestion
      return reference.unwrap() as unknown as U;
```

packages/core/src/entity/Reference.ts
thekevinbrown: Yeah I think we get two instances of the Reference class at runtime, but that's ok if we test against that __reference property

packages/core/src/unit-of-work/UnitOfWork.ts
thekevinbrown: That seems to break the TypeScript build
```
src/entity/Reference.ts(34,7): error TS2322: Type '(T & Reference<AnyEntity<unknown, never>>) | ({ [K in PK]: T[K]; } & Reference<T> & Reference<AnyEntity<unknown, never>>)' is not assignable to type 'IdentifiedReference<T, PK>'.
  Type 'T & Reference<AnyEntity<unknown, never>>' is not assignable to type 'IdentifiedReference<T, PK>'.
    Type 'T & Reference<AnyEntity<unknown, never>>' is not assignable to type 'Reference<T>'.
      Types of property '__helper' are incompatible.
        Type 'IWrappedEntityInternal<AnyEntity<unknown, never>, unique symbol> | undefined' is not assignable to type 'IWrappedEntityInternal<T, keyof T> | undefined'.
          Type 'IWrappedEntityInternal<AnyEntity<unknown, never>, unique symbol>' is not assignable to type 'IWrappedEntityInternal<T, keyof T>'.
            The types of '__meta.properties' are incompatible between these types.
              Type '{}' is not assignable to type '{ [K in keyof T & string]: EntityProperty<T>; }'.
```

B4nan: uh what? casting as `unknown` breaks the build? the error is from different line than this

B4nan: its weird that we now get this error, but it should not be connected to this change, its about `Reference.create()` method

i guess we need to do this to fix it:

```typescript
if (entity instanceof Reference) {
  return entity as IdentifiedReference<T, PK>; // or maybe even `as unknown` first
}
```

thekevinbrown: Yup, that fixes it, thanks!

thekevinbrown:
I'm not sure what's up with coveralls. :-(

B4nan:
Nothing new, its going up and down for last week or so, and in general it's causing vast majority of CI fails... I guess I should find a replacement, or maybe ditch it as we also report coverage to codeclimate (but that does not fail when we go down from 100%...).

PR #680, opened 2020-07-23
enhancement(core): Improve query performance when logging is disabled

thekevinbrown:
Logger doesn't actually log anything by default, and highlight is enabled by default, so this combination resulted in colorizing logs that are then not actually logged. In order to keep the default as highlight: true but perform better in the default case,  we now bail out of logQuery before doing the highlighting if the logger won't log anything anyway.

### DIFF
--- a/packages/core/src/connections/Connection.ts
+++ b/packages/core/src/connections/Connection.ts
@@ -84,6 +84,13 @@ export abstract class Connection {
   }
 
   protected logQuery(query: string, took?: number, language?: string): void {
+    const logger = this.config.getLogger();
+
+    // We only actually log something when debugMode is enabled. If it's not enabled,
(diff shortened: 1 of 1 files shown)


packages/core/src/connections/Connection.ts
```diff
@@ -84,6 +84,13 @@ export abstract class Connection {
   }
 
   protected logQuery(query: string, took?: number, language?: string): void {
+    const logger = this.config.getLogger();
+
+    // We only actually log something when debugMode is enabled. If it's not enabled,
+    // we can jump out here as a performance optimization and save ourselves some cycles
+    // preparing a message that won't get used.
+    if (!logger.debugMode) return;
```
B4nan, line 92: mmm how come this is not checked via eslint :]

```suggestion
    if (!logger.debugMode) {
      return;
    }
```

PR #682, opened 2020-07-24
refactor: switch driver requires so webpack can rewrite them

thekevinbrown:


packages/core/src/utils/Configuration.ts
```diff
@@ -69,11 +69,11 @@ export class Configuration<D extends IDatabaseDriver = IDatabaseDriver> {
   };
 
   static readonly PLATFORMS = {
-    mongo: ['MongoDriver', '@mikro-orm/mongodb'],
-    mysql: ['MySqlDriver', '@mikro-orm/mysql'],
-    mariadb: ['MariaDbDriver', '@mikro-orm/mariadb'],
-    postgresql: ['PostgreSqlDriver', '@mikro-orm/postgresql'],
-    sqlite: ['SqliteDriver', '@mikro-orm/sqlite'],
+    mongo: { className: 'MongoDriver', module: () => require('@mikro-orm/mongodb') },
```
B4nan, line 72: won't this break yarn pnp? cc @merceyz

its a peer dependency, so we need to use `Utils.requireFrom()`, right?

merceyz: It will break under Yarn PnP and be subject to hoisting under the node resolution unless declared as a peer dependency.

> its a peer dependency, so we need to use Utils.requireFrom(), right?

Correct, either that or declare it as a peer dependency

thekevinbrown: They look like optional peer dependencies to me.

merceyz: Indeed, they can be declared as such or use `requireFrom`, both work

B4nan: They are indeed optional peer dependencies, so let's declare them. I guess we need to do it for both `@mikro-orm/core` and `mikro-orm` packages, right?

merceyz: Yes, though I haven't looked that much into the `mikro-orm` package.

Might want to replace these as well while you're at it
https://github.com/mikro-orm/mikro-orm/blob/071edc72d173575db586b9854f8d5bba0ab68d85/packages/core/src/MikroORM.ts#L116
https://github.com/mikro-orm/mikro-orm/blob/071edc72d173575db586b9854f8d5bba0ab68d85/packages/core/src/MikroORM.ts#L124

B4nan: Yop, also had those on my mind.

PR #687, opened 2020-07-28
refactor(dev-experience): turn on curly eslint rule to match preferred project styling

thekevinbrown:
Turn on the curly eslint rule so that the linter matches the project's preferred style.


B4nan:
It is already there, line 137 :]

PR #708, opened 2020-08-05
Explain usage

oliversalzburg:
Felt to me like half the story is missing :D


B4nan:
Well, this happen automatically, if the context is registered, EM (and therefore repositories too) will prefer this context. One should generally not obtain the EM from `RequestContext` helper (or better say should not need).

oliversalzburg:
Alright, even better. That aspect was unclear to me though, getting started. Maybe I skipped over another relevant part of the documentation. Thanks for the help :)

PR #721, opened 2020-08-09
docs: update cli output and describe importing sql dump files

Langstra:
Cli output now reflects the current state again.
Also before it was unclear how to import multiline dump files.

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -329,3 +329,16 @@ MikroORM.init({
 ```
 
 Read more about this in [Metadata Cache](metadata-cache.md) section.
+
+## Importing database dump files (MySQL and Postgress)
+
+Using the `mikro-orm database:import db-file.sql` you can import a database dump file. This can be useful when kickstarting an application or could be used in tests to reset the database. Database dumps often have queries spread over multiple lines and therefore you need the following configuration.
+
+```typescript
+MikroORM.init({
+  ...
+  multipleStatements: true,
+  ...
+});
+```
+ > This should be disabled in production environments for added security.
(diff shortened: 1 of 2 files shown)


docs/docs/configuration.md
```diff
@@ -329,3 +329,13 @@ MikroORM.init({
...
+
+## Importing database dump files (MySQL and Postgress)
+Using the `mikro-orm database:import db-file.sql` you can import a database dump file. This can be useful when kickstarting an application or could be used in tests to reset the database. Database dumps often have queries spread over multiple lines and therefore you need the following configuration.
+```
+export interface ConnectionOptions {
+  ...
+  multipleStatements: true;
+  ...
+}
+```
```
B4nan, line 341: add few blank lines and line breaks

```suggestion
## Importing database dump files (MySQL and Postgress)

Using the `mikro-orm database:import db-file.sql` you can import a database dump file. 
This can be useful when kickstarting an application or could be used in tests to reset 
the database. Database dumps often have queries spread over multiple lines and therefore 
you need the following configuration.

```typescript
MikroORM.init({
  ...
  multipleStatements: true,
  ...
});
```

> This should be disabled in production environments for added security. 
```

File: docs/docs/installation.md [hunk 2]
```diff
@@ -247,4 +260,21 @@ To verify your setup, you can use `mikro-orm debug` command.
...
+
+```typescript
+const app = express();
+
+app.use((req, res, next) => {
+  RequestContext.create(orm.em, next);
+});
+```
+
+More info about `RequestContext` is described [here](identity-map.md#request-context).
```
B4nan, line 278: this section should not be here (i moved it above the CLI section), probably bad rebase?
```suggestion
```

Langstra: Yes, you are correct, bad rebase. Guess I missed it, my bad.

B4nan:
Thanks!

Issue #711, opened 2020-08-05
Audit logging


Langstra:
When using mikro-orm with nestjs (in my case) I want to use the EventSubscriber to act on Insert, Update and Delete events. However, I cannot create new AuditLog entities inside the EventSubscriber. I cannot inject my AuditLogService into the Subscriber

I'd like to be able to inject a service into the EventSubscriber which can then be used to insert new AuditLog entities into the db. 

I used typeorm before, which did support something similar. With Mikro-Orm I have tried to use the @Injectable decorator and added the subscriber to the provider array of nestjs App Module.

B4nan:
It should work fine, all you need to do is to register the subscriber in the `EventManager` when the subscriber gets instanced. Something like this could work:

```typescript
@Injectable()
export class MySubscriber {

  constructor(em: EntityManager) {
    em.getEventManager().registerSubscriber(this);
  }

}
```

But this won't allow you to flush new entities from inside the event handlers (not in the same UoW, you can fire native queries for sure). For that you need to use flush events:

https://github.com/mikro-orm/mikro-orm/blob/dev/docs/docs/lifecycle-hooks.md#flush-events

Langstra:
Does this mean that you can inject your own service into a subscriber along with the EntityManager and the subscribe the thing as an EventSubscriber? And what if in that case the service would try to flush a new (audit log) entity?

Langstra:
I created a repository displaying the problem I am having. If you have time would you mind looking at it? https://github.com/RiskChallenger/mikro-orm-issue (Second example in the repo)

B4nan:
As I already said, you can't use lifecycle events for this, you need to use flush events. Lifecycle events are fired when the transaction is already running, calling flush inside them is forbidden. What you can do is to create a new fork (via `em.fork()`) and use that - which will create separate transaction. Or you can use `em.nativeInsert()` (which is equivalent to using a QB). 

Btw all event handlers will get EM instance in the `EventArgs`. You should use that one, as there can be multiple instances of EM (e.g. when doing nested transactions), and the one in `EventArgs` is always the one that triggered the event.

Will take a look at the repo, I believe it should be working, but I haven't tested this myself yet, so there might be some problem with this lazy registration (maybe you are registering it to the request scoped EM, but you need to register it to the global one).

Langstra:
[...]

PR #726, opened 2020-08-09
fix: use lazy registered EventSubscribers

Langstra:
Before for every new context an EM was created with a new EventManager.
This EventManager would only use EventSubscribers that were annotated or explicitly defined in the
config.
Now the EventManager from the global EM is reused in every context specific EM.
Thereby lazy registeren EventSubscribers are also listening to events.

I did take a slight different approach then was suggested here https://github.com/mikro-orm/mikro-orm/issues/711#issuecomment-671087399. This I did to preserve the readonly property and keep the code cleaner (in my perspective). Hope you agree and otherwise we'll make changes.


Langstra:
Thanks for accepting. Any idea when this will be released, either as rc.1 or final (though probably too soon for that yet).

B4nan:
Later this week. For now as a workaround you could do this:

```typescript
@Injectable()
export class MySubscriber {

  constructor(em: EntityManager) {
    em.getEventManager().registerSubscriber(this);
    em.config.get('subscribers').push(this); // here
  }

}
```

PR #728, opened 2020-08-10
docs: correct postgresql package name

tamatomas:
The postgres package is @mikro-orm/postgresql:  https://www.npmjs.com/package/@mikro-orm/postgresql


B4nan:
thanks

PR #741, opened 2020-08-13
feat(migrations): allow specifying list of migrations

thomaschaaf:


packages/core/src/utils/Configuration.ts
```diff
@@ -256,6 +256,11 @@ export interface ConnectionOptions {
   pool?: PoolConfig;
 }
 
+export type MigrationObject = {
```
B4nan, line 259: move this to `typings.ts`

packages/migrations/src/Migrator.ts
```diff
@@ -18,14 +18,27 @@ export class Migrator {
...
+      path: Utils.absolutePath(this.options.path!, this.config.get('baseDir')),
+      pattern: this.options.pattern,
+      customResolver: (file) => this.resolve(file),
+    };
+
+    if (this.options.migrationsList?.length) {
+      migrations = migrationsList(
+        this.options.migrationsList.map((migration: MigrationObject) =>
+          this.initialize(
+            migration.class as unknown as Constructor<Migration>,
```
B4nan, line 31: why don't you type it like `Constructor<Migration>` directly in the `MigrationObject`?

packages/migrations/src/Migrator.ts
```diff
@@ -67,9 +80,14 @@ export class Migrator {
     // eslint-disable-next-line @typescript-eslint/no-var-requires
     const migration = require(file);
     const MigrationClass = Object.values(migration)[0] as Constructor<Migration>;
+    return this.initialize(MigrationClass);
```
B4nan, line 83: add blank line before return statement

packages/core/src/typings.ts
```diff
@@ -276,3 +276,8 @@ export type New<T, P = string[]> = Loaded<T, P>;
 export interface Highlighter {
   highlight(text: string): string;
 }
+
+export interface Migration {
```
B4nan, line 280: move this above, where the `IMigrator` is defined, so its on the same place

docs/docs/migrations.md
```diff
@@ -105,6 +105,51 @@ Then run this script via `ts-node` (or compile it to plain JS and use `node`):
...
+## Importing migrations statically
+
+```typescript
+import Migration20191019195930 from "../migrations/Migration20191019195930.ts";
+
+await MikroORM.init({
+  migrations: {
+    migrationsList: [
+      {
+        name: "Migration20191019195930.ts",
```
B4nan, line 117: use single quotes (multiple places here)

docs/docs/migrations.md
```diff
@@ -105,6 +105,51 @@ Then run this script via `ts-node` (or compile it to plain JS and use `node`):
 $ ts-node migrate
 ```
 
+## Importing migrations statically
+
```
B4nan, line 109: there should be at least one sentence here, e.g.  "this is how you do it with webpack, where this and that is not possible, so we...)

docs/docs/migrations.md
```diff
@@ -105,6 +105,51 @@ Then run this script via `ts-node` (or compile it to plain JS and use `node`):
...
+```typescript
+import Migration20191019195930 from "../migrations/Migration20191019195930.ts";
```
B4nan, line 111: migrations are using regular exports, not default exports, so this is probably invalid statement, right?

packages/migrations/src/Migrator.ts
```diff
@@ -1,5 +1,5 @@
-import umzug, { Umzug } from 'umzug';
-import { Utils, Constructor } from '@mikro-orm/core';
+import umzug, { Umzug, migrationsList } from 'umzug';
```
B4nan, line 1: this fails in the CI as well as on my end, probably bad/outdated typings? the `migrationsList` is in the sources, but not in the d.ts

maybe we could access it from the `umzug` default export, or we would need to augment the typings (via declaration merging)

packages/migrations/src/Migrator.ts
thomaschaaf: Migration in Core is not identical to the Migration from this package. Thus it clashes.

packages/migrations/src/Migrator.ts
thomaschaaf: I tried everything I could but ultimately this seems to be the only way I am able to fix this until the typings are merged.

packages/migrations/src/Migrator.ts
B4nan: but it should be the same thing, right? so instead of type casting we should fix that interface in core i guess?

packages/migrations/src/Migrator.ts
B4nan: maybe we could use your fork as a dependency? 

or as i mentioned before, we could use declaration merging

thomaschaaf: I do not know how to do this. Would you mind showing how to do this?

B4nan: https://www.typescriptlang.org/docs/handbook/declaration-merging.html
https://stackoverflow.com/questions/37377731/extend-express-request-object-using-typescript

(the use case is pretty much the same as adding custom props to expess.req object)

packages/migrations/src/Migrator.ts
thomaschaaf: I do not think this is possible. I tried changing the interface to:

```
export interface Migration {
  private readonly queries: string[];

  protected readonly driver: AbstractSqlDriver;
  protected readonly config: Configuration;

  up(): Promise<void>;
  down(): Promise<void>;

  isTransactional(): boolean;

  addSql(sql: string): void;

  reset(): void;

  getQueries(): string[];
}
``` in core but am getting weird errors like: ``` Property 'queries' is private in type 'Migration' but not in type 'Migration'.``` Maybe you are able to show me how to do this properly.

Comment (B4nan) [reply packages/migrations/src/Migrator.ts:32]: right, because `Migration` is actual class, not just an interface, and it has private `queries` array. 

then i'd say we should not have that `queries` array in the interface at all, as it is private

Comment (B4nan) [reply packages/migrations/src/Migrator.ts:32]: let's forget about this for now, its a not a big deal, it was just looking weird. i can play with that myself once its merged

**Comment by B4nan** [comment]
Ok, tests are green, let's merge this, will take a look at the typings later

**Comment by thomaschaaf** [comment]
I fixed it here: https://github.com/mikro-orm/mikro-orm/pull/744/commits/c380d99dd424935852c8883cda8db87ed59284c7

**Comment by B4nan** [comment]
Hmm damn umzug and their mediocre typings... 

```
node_modules/@mikro-orm/migrations/Migrator.d.ts:4:16 - error TS2665: Invalid module name in augmentation. Module 'umzug' resolves to an untyped module at '/usr/local/var/www/vision.gl/vision-api/node_modules/umzug/lib/index.js', which cannot be augmented.

4 declare module 'umzug' {
                 ~~~~~~~


Found 1 error.
```

What about that PR you made against definitely typed, is that merged already?

merceyz:
If you move that declare to it's own file it should work

PR #742, opened 2020-08-13
chore(deps): Upgrade umzug to v3

thomaschaaf:


packages/migrations/package.json
```diff
@@ -50,10 +50,12 @@
     "@mikro-orm/knex": "^4.0.0-rc.1",
     "fs-extra": "^9.0.1",
     "ts-morph": "^7.1.2",
-    "umzug": "^2.3.0"
+    "umzug": "^3.0.0-beta.5"
   },
   "devDependencies": {
-    "@mikro-orm/core": "^4.0.0-rc.1"
+    "@mikro-orm/core": "^4.0.0-rc.1",
+    "@types/validator": "^13.1.0",
```
thomaschaaf, line 57: else tests failed 🤷

packages/migrations/package.json
```diff
@@ -50,10 +50,12 @@
...
+    "sequelize": "^6.3.4"
```
thomaschaaf, line 58: else tests failed 🤷

B4nan: oh jeez, and i though they will make it right when migrating to TS 🤷. nope, i dont want sequalize as dependency, not even dev one. 

what were the failures? this sounds like something to report to their project

will try the upgrade myself too to see it in action, if its just about those tiny changes

thomaschaaf:
@B4nan Okay I am sorry I will stop with the upgrade then. There are some unit tests failing and I'd have to do some deeper digging.

B4nan:
no need to be sorry, its not even your project :] i will take a look, maybe it could work just by having `@types/sequalize`? I see them included too when looking at current node_modules.

B4nan:
What were the issues without having `sequalize` and `@types/validator` included? I did the changes as you did, without including them, and 2 tests are failing, but nothing about those dependencies, the error i see is:

```
TypeError: The "path" argument must be of type string. Received an instance of Array
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received an instance of Array
    at validateString (internal/validators.js:120:11)
    at Object.basename (path.js:1156:5)
    at Object.nameFormatter (/usr/local/var/www/b4nan/mikro-orm/node_modules/umzug/src/migration.ts:21:44)
    at Migration.testFileName (/usr/local/var/www/b4nan/mikro-orm/node_modules/umzug/src/migration.ts:75:55)
    at /usr/local/var/www/b4nan/mikro-orm/node_modules/umzug/src/umzug.ts:352:40
    at Array.find (<anonymous>)
    at Umzug._findMigration (/usr/local/var/www/b4nan/mikro-orm/node_modules/umzug/src/umzug.ts:352:28)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

And that is caused by this line in the tests:

```
await migrator.up({ migrations: [migration.fileName] });
```

Which was working fine with umzug 2.

thomaschaaf:
I tried only importing the typing but that doesn't work either because it requires sequelize ^5 for the storage adapter.

merceyz:
They have two undeclared dependencies right here https://github.com/sequelize/umzug/blob/493e4e8d803b327655ec4b0322bf47cd248b747a/src/storage/sequelize.ts#L2-L3 granted they're only used for types so runtime should work

B4nan:
Let's close this now, as #741 is working without the need to ugprade, so better to wait for stable release (that will hopefully have that hidden dependency fixed).

Issue #792, opened 2020-08-30
Given a single filter on an $or operation the resulted sql will be wrong.


charmpitz:
**Describe the bug**
I'm dynamically creating some filters and it seems that some queries are wrong when the `$or` operator contains an array with a single element like in the scenario below.

**Stack trace**
Given the following arguments:
```json
{
    "$and": [
        { "user": { "id": { "$eq": 1 } } },
        { "$or": [{ "$and": [{ "id": { "$lt": 6 } }] }] }
    ]
}
{ "limit": 3, "orderBy": { "id": "desc" } }
```
We receive the following SQL:
```sql
select "e0".* from "schema"."article" as "e0" 
    left join "schema"."user" as "e1" 
        on "e0"."user_id" = "e1"."id" 
where 
    "e1"."id" = 1 or 
    ("e0"."id" < 6) 
order by 
    "e0"."id" desc 
limit 3
```

**To Reproduce**
1. Instantiate a query builder.
2. Use the input from above.

**Expected behavior**
Even though the query is over-complicated (due to being dynamically generated) it should be correctly parsed and transformed into the following SQL instead (using an `and` instead of an `or`):
```sql
select "e0".* from "schema"."article" as "e0" 
    left join "schema"."user" as "e1" 
        on "e0"."user_id" = "e1"."id" 
where 
    "e1"."id" = 1 and 
    ("e0"."id" < 6) 
order by 
    "e0"."id" desc 
limit 3
```

**Additional context**
For example, in a more complex scenario where the `$or` has an array with more than 1 element everything seems to be ok:
Given:
```json
{
    "$and": [
        { "user": { "id": { "$eq": 1 } } },
        {
            "$or": [
                { "$and": [{ "id": { "$lt": 6 } }] },
                {
                    "$and": [
                        { "id": { "$eq": 6 } },
                        { "createdAt": { "$gt": "2020-08-26T20:01:48.863Z" } }
                    ]
                }
            ]
        }
    ]
}
{ "limit": 3, "orderBy": { "id": "desc" } }
```
It is correctly transformed into:
```sql
select "e0".* from "schema"."article" as "e0" 
    left join "schema"."user" as "e1" 
        on "e0"."user_id" = "e1"."id" 
where 
    "e1"."id" = 1 and 
    (
        ("e0"."id" < 6) or 
        (
            "e0"."id" = 6 and 
            "e0"."created_at" > '2020-08-26T20:01:48.863Z'
        )
    ) 
[...]

PR #794, opened 2020-08-30
fix(query-builder): make sure `$or` and `$and` combined works correctly

charmpitz:
Fix for https://github.com/mikro-orm/mikro-orm/issues/792

### DIFF
--- a/tests/QueryBuilder.test.ts
+++ b/tests/QueryBuilder.test.ts
@@ -1240,6 +1240,30 @@ describe('QueryBuilder', () => {
     const qb2 = orm.em.createQueryBuilder(Author2, 'a');
     qb2.select('*').where({ $or: items.map(i => ({ $and: [i] })) });
     expect(qb2.getQuery()).toEqual('select `a`.* from `author2` as `a` where ((`a`.`name` = ? and `a`.`email` = ?) or (`a`.`name` = ? and `a`.`email` = ?))');
+
+    const qb3 = orm.em.createQueryBuilder(Author2, 'a');
+    qb3.select('*').where({
+      $and: [
+        { $or: [{ name: 'value1' }] },
+        { $or: [{ name: 'value3' }] },
+      ],
+    });
+    expect(qb3.getQuery()).toEqual(
+      'select `a`.* from `author2` as `a` ' +
+      'where (`a`.`name` = ?) and (`a`.`name` = ?)');
+
+
+    const qb4 = orm.em.createQueryBuilder(Author2, 'a');
+    qb4.select('*').where({
+      $and: [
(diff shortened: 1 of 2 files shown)


B4nan:
Thanks!

tests/QueryBuilder.test.ts
```diff
@@ -1240,6 +1240,30 @@ describe('QueryBuilder', () => {
...
+    });
+    expect(qb3.getQuery()).toEqual(
+      'select `a`.* from `author2` as `a` ' +
+      'where (`a`.`name` = ?) and (`a`.`name` = ?)');
+
+
+    const qb4 = orm.em.createQueryBuilder(Author2, 'a');
+    qb4.select('*').where({
+      $and: [
+        { $or: [{ name: 'value1', email: 'value2' }] },
```
B4nan, line 1259: uh wait a minute, this looks wrong. you have single item in the `$or` array, which is object. the objet properties should be `$and`ed, not `$or`ed.

this should produce this query:

```sql
select `a`.* from `author2` as `a` 
where (`a`.`name` = ? and `a`.`email` = ?) and (`a`.`name` = ? and `a`.`email` = ?)
```

to build the query you are asserting, you would have to have multiple items in the `$or` array:

```typescript
    const qb4 = orm.em.createQueryBuilder(Author2, 'a');
    qb4.select('*').where({
      $and: [
        { $or: [{ name: 'value1' }, { email: 'value2' }] }, // now we have two objects in the $or array
        { $or: [{ name: 'value3' }, { email: 'value4' }] }, // now we have two objects in the $or array
      ],
    });
```

charmpitz: Indeed, you're right, I'll fix it in a new PR and incorporate a few more tests to make sure it works accordingly. Let me know if you have more insights :v:

charmpitz: I had hoped this would be easier to fix but unfortunately it took me all day to make it work properly. I've added more test case scenarios and I've found a lot more corner cases. Before moving further there are still a few quirks we should discuss:
1. There are extra lots of extra parenthesis and frankly I don't think we can get rid of them. They seem to be the main reason why the queries weren't working properly. Because of this, sadly, there are lots of tests that need to be fixed as well..
2. I'm not sure how `method` argument should be treated

I've added a PR (https://github.com/mikro-orm/mikro-orm/pull/798), let me know what you think. It would probably be best to play around with it too.

B4nan: Thanks, will take a look at what you did in more detail and see if I can finalize that. Btw its better to move the conversation to the new PR.

charmpitz: You're welcome, let me know if I can be of any help.

PR #798, opened 2020-09-01
fix(query-builder): added more tests for 'or' and 'and' combined

charmpitz:
NOT READY TO BE MERGED.

### DIFF
--- a/packages/knex/src/query/QueryBuilderHelper.ts
+++ b/packages/knex/src/query/QueryBuilderHelper.ts
@@ -219,26 +219,47 @@ export class QueryBuilderHelper {
     return `%${value}%`;
   }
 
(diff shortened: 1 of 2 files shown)


B4nan:
So I managed to get all tests passing, there are some extra parens but only in those complex and/or combinations (I had to adjust only single query test in the end).

B4nan:
Btw you had one thing wrong in the tests - `$not` operator can't be used with arrays, if you want `not (... and ...)` then you need to use `$not: { $and: [...] }`, or just `$not: { ... }` as object keys will be `$and`ed too. Following is equal:

```ts
qb.select('*').where({
  $or: [
    { $not: { name: 'value3', email: 'value4' } },
    { $not: { $and: [{ name: 'value3', email: 'value4' }] } },
  ],
});
```

PR #818, opened 2020-09-06
feat(migrations): add support for initial migrations

### LINKED ISSUE #772: Feat: Add features like seeding and fake migrations in the cli
The first case will be to insert values into the database with some useful data after a successful migration.
For reference: https://sequelize.org/master/manual/migrations.html#creating-the-first-seed 

Fake migrations will be useful when a database has been created and altered manually and later you want to manage all things with migrations.
Reference: https://docs.djangoproject.com/en/3.1/topics/migrations/#initial-migrations

B4nan:
If you want to start using migrations, and you already have the schema generated,
you can do so by creating so called initial migration:

> Initial migration can be created only if there are no migrations previously
> generated or executed.

```sh
npx mikro-orm migration:create --initial
```

This will create the initial migration, containing the schema dump from
`schema:create` command. The migration will be automatically marked as executed.

Closes #772

### DIFF
--- a/docs/docs/migrations.md
+++ b/docs/docs/migrations.md
@@ -29,6 +29,26 @@ per migration basis by implementing the `isTransactional(): boolean` method.
 
 `Configuration` object and driver instance are available in the `Migration` class context.
 
+You can execute queries in the migration via `Migration.execute()` method, which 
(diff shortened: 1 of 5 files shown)


rubiin:
cool, this was what i needed

PR #825, opened 2020-09-07
feat(oracle): adds initial support to the oracle database (WIP)

guilhermedalleprane:
This PR brings some initial work for compatibility with oracle databases. It's still a WIP and it is not yet stable. Still lacks automated testing and review of some features.


B4nan:
What's the status? I still don't see any test cases in the PR.

guicovred:
Hi, Martin!
I had to solve some bugs on the side of the knex that was affecting the oracle driver and after that I went on vacation from work and decided to take some time off for myself. 
I will keep working on it soon!

guicovred:
I'm back!

guicovred:
I thought it would be easier to bring support to oracle, but oracle driver (and the oracle knex dialect) has several details incompatible with the current implementation, especially when it comes to SchemaGenerator.

An example is the execution of raw sql, where the oracle driver does not accept the termination with semicolons, except in statements ending with `END;`

Another example is that to load a .sql file, I had to change [here](https://github.com/mikro-orm/mikro-orm/blob/39818caf5a6f7cac46b77d79f8fa6710ef9806a1/packages/knex/src/AbstractSqlConnection.ts#L69) to run it line by line.

I think it's necessary to take a look at all the necessary changes to the existing code for everything to work properly (not just for all tests to pass, but to ensure if there is no loss of performance or unexpected effects). I confess that I am little afraid to change the existing framework code for this.

For now, as this is taking much longer than I imagined, I am using other alternatives to meet my initial need (which was relatively small compared to the amount of work to be done here), but it will be a pleasure to continue helping if more people want to help implement this in the future, otherwise it is not extremely priority for me too.

B4nan:
> An example is the execution of raw sql, where the oracle driver does not accept the termination with semicolons, except in statements ending with END;

Do you have some links about that? As from what I searched I cannot find a place that would say so. It sounds very weird, hard to believe it :]

guicovred:
I explained it very badly, sorry.

> statements ending with a END;

 I mean [ Oracle PL/SQL blocks](https://docs.oracle.com/cd/A97630_01/appdev.920/a96624/a_samps.htm).

Semicolons are for separating multiple statements, except when the semicolon is part of the statement as is the case with Oracle PL/SQL blocks. In this case, the semicolon at the end, is part of a single statement.

Knex internally uses the [execute()](https://oracle.github.io/node-oracledb/doc/api.html#execute) method from oracle-nodedb, that does not allow the execution of multiple statements, and consequently, I believe, statements ending with semicolons.

This also causes SchemaGenerator.createSchema() to not work, since it first calls getCreateSchemaSQL(), and then executes the returned multiline sql). SchemaGenerator.dump() would also add an extra semicolon at the end [here](https://github.com/mikro-orm/mikro-orm/blob/a7da03d2a2a937a1a2642cb34f7583a333fd50da/packages/knex/src/schema/SchemaGenerator.ts#L603). (double semicolons in case of PL/SQL blocks, which syntactically already have a semicolon at the end). 

Ok, now why are PL/SQL blocks important? Because Knex [wrap some instructions with it](https://github.com/knex/knex/blob/master/lib/dialects/oracle/utils.js#L25), to catch exceptions (like in [dropTableIfExists](https://github.com/knex/knex/blob/69167b6fa9c9fe54b81dd6cb0ea49e4e1481c012/lib/dialects/oracle/schema/compiler.js#L72
)). Knex had a similar problem and I solved it [here](https://github.com/knex/knex/issues/4045).

> As from what I searched I cannot find a place that would say so.

If you search, you will not find examples also where knex is using semicolons in single statements too. Except when the driver itself supports, like as is the case with mysql, when the parameter [multipleStatements](https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/utils/Configuration.ts#L288) is added. But this is not something officially supported by knex, is merely supported by each driver, as stated [here](https://github.com/knex/knex/issues/257). 

So if knex does not officially support it, is not a guarantee of operation in all cases and it's not so hard to believe that it doesn't work in oracle. I haven't tested if removing multipleStatements, it would work with single statements ending with semicolons for MySQL, but I would guess it wouldn't.

hinogi:
@guilhermedalleprane maybe with oracledb 6 it might be easier because you don't necessarily need the fat client

B4nan:
Closing in favor of #6704. I'd say that it already covers most of what this PR was doing (and many other things). 

I must say that after spending a few weeks working on this, I finally understand all the weird things (semicolons, multi inserts, plsql blocks, ...) we were discussing back in 2020 🙃 This is really a weird one.

Anyway, thank you for the initial work on this, it did help.

hinogi:
If possible, packages like sql-formatter or sql-highligher also struggle with weirdness  of procedural sql if your work can help with that maybe there is a way those packages can benefit from that. just an idea

B4nan:
Our highlighter works fine as far as I can tell, no plans to fix other formatting libraries since I got my own (which is tiny, unlike all the other options, seeing `sql-formatter`, it has install size of 3mb, while `@mikro-orm/sql-highlighter` is just 60kb).

hinogi:
I was speaking of https://github.com/scriptcoded/sql-highlight and https://github.com/sql-formatter-org/sql-formatter of course, I haven't tried the mikro-rom highglighter. I am wondering, is there some treesitter supported option out there for syntax highlighting sql?

guidalleprane:
> we were discussing back in 2020

Wow, hard to believe it's been that long...
Awesome work on this, @B4nan! 🎉

Issue #858, opened 2020-09-22
findAndCount returning NaN for count


followben:
**Describe the bug**
When using findAndCount on Postgres with a filterQuery that includes inverse relations, count can incorrectly return NaN

**To Reproduce**
1. check out https://github.com/followben/mikro-orm/tree/gh858
2. run `yarn test GH858`

**Expected behavior**
Count returns a number greater than or equal to zero (or findAndCount() throws an error).

**Versions**

| Dependency | Version |
| - | - |
| node | 12.x |
| typescript | 3.9.7 |
| mikro-orm | 4.0.5 |
| pg | 8.3.0 |

B4nan:
Add a reproduction

followben:
Hah, you're too quick... I was upping one with the GH issue number attached.

B4nan:
Good, ty :] 

Anyway, I never merge those "repro PRs", the idea was that this "single file repro" should be part of the issue - I will be the one adding it to the codebase together with a fix.

followben:
Why is that @B4nan? I mean, why aren't test cases worthy of merging?

B4nan:
Why would I merge something that will cause build failures?

followben:
Because someone went to the trouble of creating it as per the contribution guidelines. You can push to the PR to fix, correct?

PR #859, opened 2020-09-22
test(core): verify findAndCount for inverse relations

followben:
Demonstrates #858.

### DIFF
--- a/tests/issues/GH858.test.ts
+++ b/tests/issues/GH858.test.ts
@@ -1,0 +1,97 @@
+import { BaseEntity, Collection, Entity, IdentifiedReference, MikroORM, OneToOne, ManyToOne, OneToMany, PrimaryKey, Property } from '@mikro-orm/core';
+import { PostgreSqlDriver } from '@mikro-orm/postgresql';
+import { v4 } from 'uuid';
+
+@Entity()
+class Transaction extends BaseEntity<Transaction, 'id'> {
+
+  @PrimaryKey({ type: 'uuid' })
+  id = v4();
+
+  @ManyToOne('Account')
+  debitAccount!: IdentifiedReference<Account>;
+
+  @ManyToOne('Account')
+  creditAccount!: IdentifiedReference<Account>;
+
+}
+
+@Entity()
+class Voucher extends BaseEntity<Voucher, 'id'> {
+
+  @PrimaryKey({ type: 'uuid' })
+  id = v4();
+
+  @OneToOne('Account')
+  account!: IdentifiedReference<Account>;
+
+}
+
+@Entity()
+class Account extends BaseEntity<Account, 'id'>  {
+
+  @PrimaryKey({ type: 'uuid' })
+  id = v4();
+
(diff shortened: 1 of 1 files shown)


followben:
Ive been looking into root cause with a view to adding a fix.

So qb in `AbstractSqlDriver.count()` doesn't actually generate a count statement all. This is the statement it produces:

```sql
select distinct
    "e0"."id",
    "e2"."id" as "voucher_id",
    "e4"."id" as "voucher_id"
from "transaction" as "e0"
left join "account" as "e1" on "e0"."credit_account_id" = "e1"."id"
left join "voucher" as "e2" on "e1"."id" = "e2"."account_id"
left join "account" as "e3" on "e0"."debit_account_id" = "e3"."id"
left join "voucher" as "e4" on "e3"."id" = "e4"."account_id"
where ("e2"."id" = $1 or "e4"."id" = $2)
```

That looks correct in terms of joins and the where, but because or aliases both have the same name, the value returned in that column (`"voucher_id"` in this case) can intermittently be null depending on the run.

Given the statement is identical to that generated by the qb in `AbstractSqlDriver.find()`, I wonder if this isn't a wider problem?

B4nan:
That sounds very improbable, `driver.count()` is literally creating QB in count mode. I guess you are just observing it wrongly, those queries are fired in parallel...

Better to debug just the `em.count()` part, as `findAndCount()` is as simple as running `find()` and `count()`:

image

followben:
Yup, I'm calling `qb.getQuery()` on the qbs in `AbstractSqlDriver.count()` and `AbstractSqlDriver.find()` (specifically just prior to the execution of [L171](https://github.com/mikro-orm/mikro-orm/blob/9799e70bd7235695f4f1e55b25fe61bbc158eb38/packages/knex/src/AbstractSqlDriver.ts#L171) and [L60](https://github.com/mikro-orm/mikro-orm/blob/9799e70bd7235695f4f1e55b25fe61bbc158eb38/packages/knex/src/AbstractSqlDriver.ts#L60) respectively).

They're almost exactly the same except the latter selects all cols from the primary entity, and while the former just selects the id.

B4nan:
Hmm if so, then the issue must be with resetting the qb type to `select` probably in the `CriteriaNode` where autojoining happens. 

My guess is that `qb.addSelect()` is called on the count qb instance, which should be ignored if the qb is in select mode (I'd say it's ok to keep the `addSelect` call, but inside that method we should check if the type is `select`, if not, ignore).

B4nan:
It's indeed caused by the `qb.addSelect`, fixed that already via 9b8056c7.

The other thing with dual `voucher_id` alias should not really matter, in fact this bit should not be part of the query at all. Will deal with that too, it looks confusing, but the query should select only props from the root entity - the transaction - and that entity do not have a `voucher` property (this clause is used for selecting 1:1 owner when querying from the inverse side, but that happens one level deeper).

B4nan:
The latter is now fixed via 86c3032, the auto-select is now happening only for root entity.

Added the tests only at the query builder level to make them fast as possible (as well as truly minimal).

Thanks for the repro, helped a lot!

PR #892, opened 2020-10-01
fix(core): define the primary key of the entity regardless of the hook in which it was defined

justkey007:
When registering a new entity, if the primary key is defined in the constructor or in the @BeforeCreate hook of the entity class, it returns an undefined value as an identifier to its related entities unless the primary key has been initialized through a function on the onCreate option of the PrimaryKey decorator.

```typescript
const publisher = new Publisher('Bastoz', PublisherType.GLOBAL);
const book = new Book('Le manguier', '1');
publisher.books.add(book);
```
Each entity inherits from the same base class which defines the primary key during the @BeforeCreate hook.

Here is the (incorrect) result obtained in debug mode:
```sql
INSERT INTO `publisher` (`created_at`, `id`, `name`, `owner_id`, `type`, `updated_at`)
VALUES ('2020-09-30 19:11:47.789', '5896064505599379', 'Bastoz', '1', 'global', NULL);


INSERT INTO `book` (`author_id`, `created_at`, `id`, `owner_id`, `publisher_id`, `title`, `updated_at`)
VALUES ('1', '2020-09-30 19:11:47.820', '8304010352419691', '1', NULL, 'Le manguier', NULL);

```

Normally expected result:
```sql
INSERT INTO `publisher` (`created_at`, `id`, `name`, `owner_id`, `type`, `updated_at`)
VALUES ('2020-09-30 19:11:47.789', '5896064505599379', 'Bastoz', '1', 'global', NULL);


INSERT INTO `book` (`author_id`, `created_at`, `id`, `owner_id`, `publisher_id`, `title`, `updated_at`)
VALUES ('1', '2020-09-30 19:11:47.820', '8304010352419691', '1', '5896064505599379', 'Le manguier', NULL);


```
After analyzing the source code I found the value of the identifier was not set if it was not set by the onCreate function of the Primarykey decorator option. Everything has been done better so as not to break the existing tests. Thank you!

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -49,8 +49,7 @@
   "dependencies": {
     "ansi-colors": "^4.1.1",
     "clone": "^2.1.2",
-    "escaya": "^0.0.51",
-    "fast-deep-equal": "^3.1.3",
+    "escaya": "^0.0.55",
     "fs-extra": "^9.0.1",
     "globby": "^11.0.1",
     "reflect-metadata": "^0.1.13",
--- a/packages/core/src/drivers/DatabaseDriver.ts
+++ b/packages/core/src/drivers/DatabaseDriver.ts
@@ -33,6 +33,10 @@ export abstract class DatabaseDriver<C extends Connection> implements IDatabaseD
 
   abstract async nativeUpdate<T extends AnyEntity<T>>(entityName: string, where: FilterQuery<T>, data: EntityData<T>, ctx?: Transaction): Promise<QueryResult>;
 
+  async nativeUpdateMany<T extends AnyEntity<T>>(entityName: string, where: FilterQuery<T>[], data: EntityData<T>[], ctx?: Transaction): Promise<QueryResult> {
+    throw new Error(`Batch updates are not supported by ${this.constructor.name} driver`);
+  }
+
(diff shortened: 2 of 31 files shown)


B4nan:
Add some tests please, the 100% (branch) coverage is mandatory, especially for changes in UoW/persistence internals. Maybe the place where you handled this is not the best one, I'd like to first see the (failing) test on my end to see what is actually happening. 

Not sure why have you skipped the `mapPrimaryKey` call altogether, the method handles custom types, and I would say it is needed. You might not break the existing tests, but only few entities in the tests are using custom types for PKs and none of them are assigning the PK value thru a hook... 

In general I would expect this to be handled somewhere else - where the hooks are processed (`UoW.runHooks()` method).

justkey007:
> Not sure why have you skipped the `mapPrimaryKey` call altogether, the method handles custom types, and I would say it is needed. You might not break the existing tests, but only few entities in the tests are using custom types for PKs and none of them are assigning the PK value thru a hook...
> 
I tried it before but the tests are broken and I got around it

justkey007:
> Add some tests please, the 100% (branch) coverage is mandatory, especially for changes in UoW/persistence internals. Maybe the place where you handled this is not the best one, I'd like to first see the (failing) test on my end to see what is actually happening.

I just created issue #893 with a related test that will allow you to better see what I'm talking about. Thank you

B4nan:
Thanks for pointing this out. I picked a bit different way to fix this - by setting the identifier value based on the diff right after we execute the hook.

PR #917, opened 2020-10-13
docs: add a note about circular dependencies in custom repositories

roderik:


B4nan:
thanks! can you add that to v4.1 too? its in the `versioned_docs` subfolder

PR #925, opened 2020-10-15
fix(schema): improve column type equality check

francisco-sanchez-molina:
This PR changes hasSameType function in order of add support for type as:

```
numeric(18,4)
```

so that they do not constantly change during migration


B4nan:
Thanks!

PR #929, opened 2020-10-19
feat(seeder): add seeder package

### LINKED ISSUE #251: Create seeds as in Eloquent (Laravel)
https://laravel.com/docs/6.x/seeding

Langstra:
Currently this is still a WIP, but I am opening the PR already to gather some feedback. Nothing from this code has been tested yet and I am not even sure it will transpile or pass eslint and/or prettier.

todo:
- [x] Add documentation for the other CLI commands
- [x] Add configurable seeder class for `--seed` option for migration and schema fresh commands
- [x] Add configurable default seed class for fresh commands
- [x] Add documentation on the config option for the path for seed files
- [x] Refactor seed commands to `seed:action`
- [x] Add `seed:create` command for creating a seeder class

Closes #251

### DIFF
--- a/docs/docs/seeding.md
+++ b/docs/docs/seeding.md
@@ -1,0 +1,194 @@
+---
+title: Seeding
+---
+
+When initializing your application or testing it can be exhausting to create sample data for your database. The solution is to use seeding. Create factories for your entities and use them in the seed script or combine multiple seed scripts.
+
+## Seeders
+A seeder class contains one method `run`. This method is called when you use the command `npx mikro-orm seeder:run`. In the `run` method you define how and what data you want to insert into the database. You can create entities using the [EntityManager](http://mikro-orm.io/docs/entity-manager) or you can use [Factories](#using-entity-factories).
+
+You can create your own seeder classes using the following CLI command:
+```shell
+npx mikro-orm seeder:create DatabaseSeeder
+```
+This creates a new seeder class. By default it will be generated in the `./database/seeder/` directory. You can configure the directory in the config with the key `seeder.path`.
+
+As an example we will look at a very basic seeder.
+```typescript
+// ./database/seeder/database.seeder.ts
+
+import { EntityManager } from '@mikro-orm/core';
+import { Seeder } from '@mikro-orm/seeder';
+import { Author } from './author'
+
+export class DatabaseSeeder extends Seeder {
+
+  async run(em: EntityManager): Promise<void> {
+    const author = em.create(Author, {
+      name: 'John Snow',
+      email: 'snow@wall.st'
+    });
+  }
+}
+```
+> Running a seeder from the command line or programmatically will automatically call `flush` and `clear` after the `run` method has completed.
+
+### Using entity factories
+Instead of specifying all the attributes for every entity, you can also use [entity factories](#entity-factories). These can be used to generate large amounts of database records. Please read the [documentation on how to define factories](#entity-factories) to learn how to define your factories. 
+
+As an example we will generate 10 authors.
+```typescript
+import { EntityManager } from '@mikro-orm/core';
+import { Seeder } from '@mikro-orm/seeder';
+import { AuthorFactory } from '../factories/author.factory'
+
+export class DatabaseSeeder extends Seeder {
+
+  async run(em: EntityManager): Promise<void> {
+    new AuthorFactory(em).count(10).make()
+  }
+}
+```
+
+### Calling additional seeders
+Inside the `run` method you can specify other seeder classes. You can use the `call` method to breakup the database seeder into multiple files to prevent a seeder file from becoming too large. The `call` method accepts an array of seeder classes.
+```typescript
+import { EntityManager } from '@mikro-orm/core';
+import { Seeder } from '@mikro-orm/seeder';
+import { AuthorSeeder, BookSeeder } from '../seeders'
+
+export class DatabaseSeeder extends Seeder {
+
+  run(em: EntityManager): Promise<void> {
+    return this.call(em, [
+      AuthorSeeder,
+      BookSeeder
+    ]);
+  }
+}
+```
+
+## Entity factories
+When testing you may insert entities in the database before starting a test. Instead of specifying every attribute of every entity by hand, you could also use a `Factory` to define a set of default attributes for an entity using entity factories.
+
+Lets have a look at an example factory for an [Author entity](http://mikro-orm.io/docs/defining-entities).
+```typescript
+import { Factory } from '@mikro-orm/seeder';
+import Faker from 'faker';
+import { Author } from './entities/author.entity';
+
+export class AuthorFactory extends Factory<Author> {
+  model = Author;
+
+  definition(faker: typeof Faker): Partial<Author> {
+    return {
+      name: faker.person.findName(),
+      email: faker.internet.email(),
+      age: faker.random.number(18, 99)
+    };
+  }
+}
+``` 
+Basically you extend the base `Factory` class, define a `model` property and a `definition` method. The `model` defines for which entity the factory generates entity instances. The `definition` method returns the default set of attribute values that should be applied when creating an entity using the factory.
+
+Via the faker property, factories have access to the [Faker library](https://github.com/marak/Faker.js/), which allows you to conveniently generate various kinds of random data for testing.
+
+### Creating entities using factories
+Once you defined your factories you can use them to generate entities. Simply import the factory, instantiate it and call the `makeOne` method.
+```typescript
+const author: Author = new AuthorFactory(orm.em).makeOne();
+```
+
+#### Generate multiple entities
+Generate multiple entities by calling the `make` method. The parameter of the `make` method is the number of entities you generate.
+```typescript
+// Generate 5 authors
+const authors: Author[] = new AuthorFactory(orm.em).make(5);
+``` 
+
+#### Overriding attributes
+If you would like to override some of the default values of your factories, you may pass an object to the make method. Only the specified attributes will be replaced while the rest of the attributes remain set to their default values as specified by the factory.
+```typescript
+const author: Author = new AuthorFactory(orm.em).make({
+    name: 'John Snow'
+});
+``` 
+
+### Persisting entities
+The `create` method instantiates entities and persists them to the database using the `persistAndFlush` method of the EntityManager.
+```typescript
+// Make and persist a single author
+const author: Author = await new AuthorFactory(orm.em).createOne();
+
+// Make and persist 5 authors
+const authors: Author[] = await new AuthorFactory(orm.em).create(5);
+```
+You can override the default values of your factories by passing an object to the `create` method.
+```typescript
+// Make and persist a single author
+const author: Author = await new AuthorFactory(orm.em).createOne({
+    name: 'John Snow'
+});
+
+// Make and persist a 5 authors
+const authors: Author[] = await new AuthorFactory(orm.em).create(5, {
+    name: 'John Snow'
+});
+```
+### Factory relationships
+It is nice to create large quantities of data for one entity, but most of the time we want to create data for multiple entities and also have relations between these. For this we can use the `each` method which can be chained on a factory. The `each` method can be called with a function that transforms output entity from the factory before returning it. Lets look at some examples for the different relations.
+
+#### ManyToOne and OneToOne relations
+```typescript
+const books: Book[] = new BookFactory(orm.em).each(book => {
+    book.author = new AuthorFactory(orm.em).makeOne();
+}).make(5);
+``` 
+
+#### OneToMany and ManyToMany
+```typescript
+const books: Book[] = new BookFactory(orm.em).each(book => {
+    book.owners.set(new OwnerFactory(orm.em).make(5));
+}).make(5);
+``` 
+
+## Use with CLI
+You may execute the `seeder:run` MikroORM CLI command to seed your database. By default, the `seeder:run` command runs the DatabaseSeeder class, which may in turn invoke other seed classes. However, you may use the `--class` option to specify a specific seeder class to run individually:
+```shell script
+npx mikro-orm seeder:run
+
+npx mikro-orm seeder:run --class=BookSeeder
+```
+
+You may also seed your database using the [`migrate:fresh`](migrations.md#using-via-cli) or [`schema:fresh`](schema-generator.md) command in combination with the `--seed` option, which will drop all tables and re-run all of your migrations or generate the database based on the current entities. This command is useful for completely re-building your database:
+```shell script
+npx mikro-orm migration:fresh --seed
+
+npx mikro-orm schema:fresh --seed
+```
+
+
+## Use in tests
+Now we know how to create seeders and factories, but how can we effectively use them in tests. We will show an example how it can be used.
+
+```typescript
+beforeAll(async () => {
+    // Get seeder from MikroORM
+    const seeder = orm.getSeeder();
+    
+    // Refresh the database to start clean
+    await seeder.refreshDatabase();
+    
+    // Seed using a seeder defined by you
+    await seeder.seed(DatabaseSeeder);
+})
+
+test(() => {
+    // Do tests
+});
+
+afterAll(async () => {
+    // Close connection
+    await orm.close();
+})
+```
--- a/packages/seeder/package.json
+++ b/packages/seeder/package.json
@@ -1,0 +1,60 @@
+{
+  "name": "@mikro-orm/seeder",
+  "version": "4.4.4",
+  "description": "Seeder package for MikroORM.",
+  "main": "dist/index.js",
+  "typings": "dist/index.d.ts",
+  "repository": {
+    "type": "git",
+    "url": "git+ssh://git@github.com/mikro-orm/mikro-orm.git"
+  },
+  "keywords": [
+    "orm",
+    "mongo",
+    "mongodb",
+    "mysql",
+    "mariadb",
+    "postgresql",
+    "sqlite",
+    "sqlite3",
+    "ts",
+    "typescript",
+    "js",
+    "javascript",
+    "entity",
+    "ddd",
+    "mikro-orm",
+    "unit-of-work",
+    "data-mapper",
+    "identity-map",
+    "seeder"
+  ],
+  "author": "Wybren Kortstra",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/mikro-orm/mikro-orm/issues"
+  },
+  "homepage": "https://mikro-orm.io",
+  "engines": {
+    "node": ">= 10.13.0"
+  },
+  "scripts": {
+    "build": "yarn clean && yarn compile && yarn copy",
+    "clean": "rimraf ./dist",
+    "compile": "tsc -p tsconfig.build.json",
+    "copy": "ts-node -T ../../scripts/copy.ts"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "faker": "5.4.0"
+  },
+  "peerDependencies": {
+    "@mikro-orm/core": "^4.4.4"
+  },
+  "devDependencies": {
+    "@mikro-orm/core": "^4.4.4",
+    "@types/faker": "^5.1.6"
+  }
+}
--- a/packages/seeder/src/factory.ts
+++ b/packages/seeder/src/factory.ts
@@ -1,0 +1,68 @@
+import Faker from 'faker';
+import { EntityManager } from '@mikro-orm/core';
+
+export abstract class Factory<C> {
+
+  abstract readonly model: { new(): C };
+  private eachFunction?: (entity: C) => void;
+
+  constructor(private readonly em: EntityManager) {
+  }
+
+  protected abstract definition(faker: typeof Faker): Partial<C>;
+
+  /**
+   * Make a single entity
+   * @param overrideParameters Object specifying what default attributes of the entity factory should be overridden
+   */
+  makeOne(overrideParameters?: Partial<C>): C {
+    const entity = this.em.create(this.model, Object.assign({}, this.definition(Faker), overrideParameters));
+    if (this.eachFunction) {
+      this.eachFunction(entity);
+    }
+    return entity;
+  }
+
+  /**
+   * Make multiple entities
+   * @param amount Number of entities that should be generated
+   * @param overrideParameters Object specifying what default attributes of the entity factory should be overridden
+   */
+  make(amount: number, overrideParameters?: Partial<C>): C[] {
+    return [...Array(amount)].map(() => {
+      return this.makeOne(overrideParameters);
+    });
+  }
+
+  /**
+   * Create (and persist) a single entity
+   * @param overrideParameters Object specifying what default attributes of the entity factory should be overridden
+   */
+  async createOne(overrideParameters?: Partial<C>): Promise<C> {
+    const entity = this.makeOne(overrideParameters);
+    await this.em.persistAndFlush(entity);
+    return entity;
+  }
+
+  /**
+   * Create (and persist) multiple entities
+   * @param amount Number of entities that should be generated
+   * @param overrideParameters Object specifying what default attributes of the entity factory should be overridden
+   */
+  async create(amount: number, overrideParameters?: Partial<C>): Promise<C[]> {
+    const entities = this.make(amount, overrideParameters);
+    await this.em.persistAndFlush(entities);
+    return entities;
+  }
+
+  /**
+   * Set a function that is applied to each entity before it is returned
+   * In case of `createOne` or `create` it is applied before the entity is persisted
+   * @param eachFunction The function that is applied on every entity
+   */
+  each(eachFunction: (entity: C) => void) {
+    this.eachFunction = eachFunction;
+    return this;
+  }
+
+}
--- a/packages/seeder/src/seeder.ts
+++ b/packages/seeder/src/seeder.ts
@@ -1,0 +1,16 @@
+import { EntityManager } from '@mikro-orm/core';
+
+export abstract class Seeder {
+
+  abstract run(em: EntityManager): Promise<void>;
+
+  protected call(em: EntityManager, seeders: { new(): Seeder }[]): Promise<void> {
+    return new Promise((resolve, reject) => {
+      Promise.all(seeders.map(s => {
+        return (new s()).run(em.fork());
+      })).then(() => resolve()).catch(reject);
+    });
+  }
+
+}
+
--- a/packages/seeder/tsconfig.json
+++ b/packages/seeder/tsconfig.json
@@ -1,0 +1,4 @@
+{
+  "extends": "../../tsconfig.json",
+  "include": ["src/**/*"]
+}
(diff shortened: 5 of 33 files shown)


package.json
```diff
@@ -45,7 +46,7 @@
     "build": "lerna run build",
     "release": "yarn build && lerna publish from-package --contents dist",
     "release:next": "yarn build && lerna publish from-package --dist-tag next --contents dist",
-    "test": "jest --runInBand",
+    "test": "jest Logger --runInBand",
```
B4nan, line 49: this should not be here

packages/seeder/package.json
```diff
@@ -0,0 +1,61 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@mikro-orm/core": "^4.0.7",
+    "faker": "^4.1.0",
+    "yargs": "^15.4.1"
+  },
+  "peerDependencies": {
+    "typescript-mix": "^3.1.3"
```
B4nan, line 56: looks like its not used anywhere, is it needed?

packages/seeder/src/factory.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  constructor(private orm: MikroORM) {
+  }
+
+  public abstract definition(): Record<string, unknown>;
+
+  public make(overrideParameters: Record<string, unknown>): C | C[] {
+    const objects = [...Array(this.amount)].map(() => {
+      const o = new this.model();
+      Object.assign(o, Object.assign({}, this.definition(), overrideParameters));
```
B4nan, line 17: using Object.assign is usually wrong, if `o` is an entity,  you should use `em.assign` that will handle collections and references

packages/seeder/src/factory.ts
```diff
@@ -0,0 +1,39 @@
...
+      return o;
+    });
+    return this.amount === 1 ? objects[0] : objects;
+  }
+
+  public async create(overrideParameters: Record<string, unknown>): Promise<C | C[]> {
+    const objects = [...Array(this.amount)].map(() => {
+      const o = new this.model();
+      wrap(o).assign(Object.assign({}, this.definition(), overrideParameters),
```
B4nan, line 26: use `em.assign()` instead of `wrap().assign()` (the same but more readable)

packages/seeder/src/factory.ts
```diff
@@ -0,0 +1,39 @@
...
+    const objects = [...Array(this.amount)].map(() => {
+      const o = new this.model();
+      Object.assign(o, Object.assign({}, this.definition(), overrideParameters));
+      return o;
+    });
+    return this.amount === 1 ? objects[0] : objects;
+  }
+
+  public async create(overrideParameters: Record<string, unknown>): Promise<C | C[]> {
+    const objects = [...Array(this.amount)].map(() => {
```
B4nan, line 24: using `Array` without `new` looks quite weird to me :]

packages/seeder/src/has-factory.decorator.ts
```diff
@@ -0,0 +1,28 @@
+import { Factory } from './factory';
+import { importFiles, loadFiles } from './utils/file.util';
+
+export class HasFactory {
+
+  public static async factory<C>(): Promise<Factory<C>> {
+    if (!(global as any).imported) {
```
B4nan, line 7: why is `global` needed here? why cant we use static variables?

if it really needs to be stored on `global`, use `Utils.getGlobalStorage()` at least, but it looks wrong in general to need global state here. and we should have some object instance there, instead of writing to the global object directly. 

but without tests or documentation i am not sure how things are supposed to be used, so that is definitely needed for better review

packages/seeder/src/refresh-database.decorator.ts
```diff
@@ -0,0 +1,16 @@
+import { MikroORM } from '@mikro-orm/core';
+import { Seeder } from './seeder';
+
+export const refreshDatabase = async (orm: MikroORM): Promise<void> => {
```
B4nan, line 4: those do not really look like decorators :]

packages/seeder/src/refresh-database.decorator.ts
```diff
@@ -0,0 +1,16 @@
...
+  const generator = orm.getSchemaGenerator();
+  await generator.dropSchema();
+  orm.config.getLogger().log('info', 'Dropped schema');
```
B4nan, line 7: afaik schema generator already logs success messages so this looks redundant

packages/seeder/src/refresh-database.decorator.ts
```diff
@@ -0,0 +1,16 @@
...
+  await generator.createSchema();
+  orm.config.getLogger().log('info', 'Recreated schema');
+};
+
+export const seed = async (orm: MikroORM, seeder: Seeder): Promise<void> => {
+  (global as any).orm = orm;
```
B4nan, line 13: this sounds wrong, as i mentioned above, i'd prefer not to have global state

packages/seeder/src/seeder.ts
```diff
@@ -0,0 +1,12 @@
+import { MikroORM } from '@mikro-orm/core';
+
+export abstract class Seeder {
+
+  public abstract run(): Promise<void>;
+
+  protected call(orm: MikroORM, seeders: Seeder[]): Promise<void[]> {
+    return Promise.all(seeders.map(s => s.run()));
```
B4nan, line 8: i would say this should be executed serially (in general that is how i would expect things to work as a user). if we wanted to execute it in parallel, each "thread" would need its own context (fork) probably.

packages/seeder/src/utils/file.util.ts
```diff
@@ -0,0 +1,12 @@
+import * as glob from 'glob';
+import * as path from 'path';
+
+export const importFiles = async (filePaths: string[]): Promise<any> => {
+  return await Promise.all(filePaths.map(filePath => import(filePath)));
+};
+
+export const loadFiles = (filePattern: string[]): string[] => {
+  return filePattern
+    .map(pattern => glob.sync(path.resolve(process.cwd(), pattern)))
```
B4nan, line 10: let's not use sync methods

B4nan (changes requested):
Nice, thanks! left some comments here, but in general without tests or docs its hard to suggest what to change how, so please add some tests or at least adjust the description with some example usage. I don't really like the global storage, I'd expect this to work similarly to migrations or entity generator - you as a user need to initialize the ORM and get the seed manager from that - that is where the context should be, no need to have anything globally.

packages/seeder/src/factory.ts
AsyncLegs: or maybe Array.from

B4nan:
Do you plan to finalize this?

packages/seeder/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "build": "yarn clean && yarn compile && yarn copy",
+    "clean": "rimraf ./dist",
+    "compile": "tsc -p tsconfig.build.json",
+    "copy": "ts-node -T ../../scripts/copy.ts"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@mikro-orm/core": "^4.0.7",
```
merceyz, line 51: Should be a peer dependency and for lerna to work it needs to be a devDependency as well

packages/seeder/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "faker": "^4.1.0",
+    "yargs": "^15.4.1"
```
merceyz, line 53: Seems to be unused

packages/seeder/src/utils/file.util.ts
```diff
@@ -0,0 +1,12 @@
+import * as glob from 'glob';
```
merceyz, line 1: This is an undeclared dependency, but since the core uses `globby` should probably use that here as well

merceyz (changes requested):
Have some comments on dependencies, staying out of the implementation part

B4nan:
Do you plan to continue working on this? I could pick it up myself, but I really need some docs (for the public api) and tests (or example usage) for what is done already.

rubiin:
i think it would be better https://github.com/CyriacBr/mikro-resources/tree/develop/packages/fixtures as its documeneted

B4nan:
@CyriacBr could you present a bit what's there and how it fits the #251? 

I don't really want to add more decorators and drive it that way, the idea was to have something similar to laravel. Maybe your `FixturesFactory` is exactly that?

(sorry for dumb questions, never used any kind of seeder myself, same with laravel, so this issue is rather abstract to me)

CyriacBr:
> @CyriacBr could you present a bit what's there and how it fits the #251?
> 
> I don't really want to add more decorators and drive it that way, the idea was to have something similar to laravel. Maybe your `FixturesFactory` is exactly that?
> 
> (sorry for dumb questions, never used any kind of seeder myself, same with laravel, so this issue is rather abstract to me)

Hello. I haven't used Laravel myself so I don't know how its Seeder behaves, but basically my package allows you to generate random entities on the fly using `FixturesFactory`. Thanks to MikroORM metadatas, decorators are only there for customization. (e.g if you do not want random values). I opted for decorators because it fits well within Typescript and MikroORM entities and everything is centralized.

It was written more for testing but can be used as a CLI by creating a `.ts` file and use the API there.
```ts
const orm = await MikroORM.init(ormConfig);
const factory = new FixturesFactory(ormInstance);
// Generate a fixture
const author = await factory.make(Author).oneAndPersist();
// Generate multiple fixtures
const authors = await factory.make(Author).manyAndPersist(10);
```

But note that I haven't used MikroORM in a while so I do not know if the package is up to date with the latest release.

Langstra:
Yeah I am sorry for not working on this anymore.work took up a lot of time lately. I'll try and make some documentation and code for this in the coming weekends. I'll keep you posted.

rubiin:
> Yeah I am sorry for not working on this anymore.work took up a lot of time lately. I'll try and make some documentation and code for this in the coming weekends. I'll keep you posted.
A documentation upto now would be great so that me or anyone else can pickup things where you left off

Langstra:
I am still working on it and it is far from done, but I pushed my progress so you get a bit more feeling for and know I am working on it ;)

Langstra:
I think right now would be a good moment for some feedback. In my opinion it is mostly done.

packages/seeder/src/factory.ts
Langstra: https://tc39.es/ecma262/#sec-array-constructor
>also creates and initializes a new Array object when called as a function rather than as a constructor. Thus the function call Array(…) is equivalent to the object creation expression new Array(…) with the same arguments.

B4nan:
Thanks a lot, gave it a quick scan and it looks good! Will review the code in detail soon, and ideally we can get this out in v4.5

docs/docs/seeding.md
```diff
@@ -0,0 +1,188 @@
...
+const book = await (new BookFactory(orm.em)).map(async (book) => {
+    book.owners = new Collection(book, await (new OwnerFactory(orm.em)).make() as Owner[]);
+    return book;
+}) as Book;
+``` 
+
+## Use with CLI
+You may execute the `db:seed` Mikro-ORM CLI command to seed your database. By default, the db:seed command runs the DatabaseSeeder class, which may in turn invoke other seed classes. However, you may use the --class option to specify a specific seeder class to run individually:
+```shell script
+mikro-orm database:seed
```
B4nan, line 151: i'd remove the `database:` prefix, also in the previous text you call it `db:` instead

also add `npx` in front of all calls to be consistent

docs/docs/seeding.md
```diff
@@ -0,0 +1,188 @@
...
+    await seeder.seed(DatabaseSeeder);
+})
+
+test(() => {
+    // Do tests
+});
+
+afterAll(() => {
+    // Close connection
+    seeder.closeConnection();
```
B4nan, line 186: should be awaited

packages/seeder/src/mikro-orm-seeder.ts
```diff
@@ -0,0 +1,50 @@
+import { Configuration, MikroORM, Options } from '@mikro-orm/core';
+import { Seeder } from './seeder';
+
+export class MikroOrmSeeder {
```
B4nan, line 4: `SeedManager` or something down that path sounds better, right? i'd rather not use `MikroORM` in class names

packages/seeder/src/seeder.ts
```diff
@@ -0,0 +1,16 @@
+import { EntityManager } from '@mikro-orm/core';
+
+export abstract class Seeder {
+
+  public abstract run(em: EntityManager): Promise<void>;
```
B4nan, line 5: remove `public` modifier (will need to review why this is not caught by eslint)

packages/seeder/tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "../../tsconfig.json",
+  "include": ["src/**/*"]
+}
```
B4nan, line 4: not sure what the `run-seeders` file is, probably should not be there?

(cant comment on it as it has no content)

tests/cli/CLIHelper.test.ts
```diff
@@ -26,14 +26,17 @@ describe('CLIHelper', () => {
       'cache:generate',
       'generate-entities',
       'database:import',
+      'database:seed',
```
B4nan, line 29: to add a bit more reasoning why i dont like the prefix - the command will work with entities, while `import` command (the only one with that prefix) works with raw SQL dumps only.

tests/cli/DatabaseSeedCommand.test.ts
```diff
@@ -0,0 +1,58 @@
...
+import { DatabaseSeedCommand } from '../../packages/cli/src/commands/DatabaseSeedCommand';
+import { initORMSqlite } from '../bootstrap';
+
+describe('DatabaseSeedCommand', () => {
+
+  let seeder: MikroOrmSeeder;
+
+  beforeAll(async () => {
+    const getORMMock = jest.spyOn(MikroOrmSeeder, 'init');
+    seeder = await MikroOrmSeeder.init(await initORMSqlite());
```
B4nan, line 30: i'd go the same way as with schema generator and migrator - having a getter on the `MikroORM` instance, so we would do `orm.getSeeder()` or `orm.getSeedManager()` or something like that.

packages/seeder/src/mikro-orm-seeder.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  orm!: MikroORM;
+
+  static async init(config: Configuration | Options | MikroORM | undefined): Promise<MikroOrmSeeder> {
+    const seeder = new MikroOrmSeeder();
+    seeder.orm = config instanceof MikroORM ? config : await MikroORM.init(config);
+    return seeder;
```
B4nan, line 11: do we really need to be that flexible here? i'd rather always require the orm intance, as as noted elsewhere, we should have a getter on the orm object

also that way there is no need to have this method async

packages/seeder/src/mikro-orm-seeder.ts
```diff
@@ -0,0 +1,50 @@
...
+    for (const seederClass of seederClasses) {
+      const seeder = new seederClass();
+      await seeder.run(this.orm.em);
+    }
+  }
+
+  async seedString(...seederClasses: string[]): Promise<void> {
+    for (const seederClass of seederClasses) {
+      await import(`${this.orm.config.get('seeder')}/${this.getFileName(seederClass)}`);
+      await this.seed(eval(seederClass));
```
B4nan, line 30: ouch, `eval`? why dont you just require the class (and as said, export the seeder from it). i would even drop the dynamic import, especially if it is causing headaches when mocking

just like with entities during folder based discovery

packages/seeder/src/mikro-orm-seeder.ts
```diff
@@ -0,0 +1,50 @@
...
+  static async init(config: Configuration | Options | MikroORM | undefined): Promise<MikroOrmSeeder> {
+    const seeder = new MikroOrmSeeder();
+    seeder.orm = config instanceof MikroORM ? config : await MikroORM.init(config);
+    return seeder;
+  }
+
+  async refreshDatabase(): Promise<void> {
+    const generator = this.orm.getSchemaGenerator();
+    await generator.dropSchema();
+    await generator.createSchema();
```
B4nan, line 17: do we need the create call?

B4nan (changes requested):
looking good, but i'd like to change the api to be sync as much as possible (while allowing to have async callbacks probably) to truly leverage the UoW

also the `eval` should go, i dont understand why it should be needed - seeders should be exported and we should require them as we do with entities or the orm config. dont worry about the eslint rule (add ignore comment for it) - this is the right place to do it.

B4nan:
anyway thanks a lot for this, especially for the docs. if you would struggle with the requests, testing or with time in general, this is already in a shape where i will be happy to take over myself. if so, i'd suggest to merge this to `dev` branch instead of master, so i can work with it directly.

totomakers:
Sorry if im wrong.
Seeder would be very nice \o/
I have made a proposal for Factory API here: https://github.com/totomakers/mikro-orm/tree/seeder-propose
Quite bit simpler with less type casting.

Langstra:
Thanks for the comments, I'll work on it and let you know when I have finished ;)

docs/docs/seeding.md
```diff
@@ -0,0 +1,188 @@
...
+Basically you extend the base `Factory` class, define a `model` property and a `definition` method. The `model` defines for which entity the factory generates entity instances. The `definition` method returns the default set of attribute values that should be applied when creating an entity using the factory.
+
+Via the faker property, factories have access to the [Faker library](https://www.npmjs.com/package/faker), which allows you to conveniently generate various kinds of random data for testing.
+
+### Creating entities using factories
+Once you defined your factories you can use them to generate entities. Simply import the factory, instantiate it and call the `make` or `create` method.
+```typescript
+const author = await (new AuthorFactory(orm.em)).make() as Author;
+``` 
+The `make` method returns the type `Author | Author[]`, so be sure to cast it to the correct type.
```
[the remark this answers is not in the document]
Langstra: My suggestion would be this
```ts
public async make(overrideParameters?: Partial<C>, count?: number): Promise<C[]>
  public async make(overrideParameters?: Partial<C>, count?: 1): Promise<C>
  public async make(overrideParameters?: Partial<C>, count?: 1 | number) {
```
Based on the count parameter you know the return type. This seems better than having multiple methods.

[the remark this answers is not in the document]
Langstra: Oh well, in my opinion an even better way is to change the class generic signature after calling the `count()` method.
```ts
public make(this: Factory<C>, overrideParameters?: Partial<C>): C
public make(this: Factory<C[]>, overrideParameters?: Partial<C>): C[]
public make(overrideParameters?: Partial<C>) {

public count(this: Factory<C>, amount: number): Factory<C[]> {
```
Before calling `count()`,  `make()` will return a non-array entity e.g. `Project`. After calling `count()`, `make()` will return an array of entities e.g. `Project[]`

docs/docs/seeding.md
```diff
@@ -0,0 +1,188 @@
...
+    return book;
+}) as Book;
+``` 
+
+#### OneToMany and ManyToMany
+```typescript
+const book = await (new BookFactory(orm.em)).map(async (book) => {
+    book.owners = new Collection(book, await (new OwnerFactory(orm.em)).make() as Owner[]);
+    return book;
+}) as Book;
```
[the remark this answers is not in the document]
Langstra: Agreed on this, definitely cleans up the code and makes it more readable. Will change.

packages/seeder/tsconfig.json
Langstra: No idea, something must have gone wrong here ;)

packages/seeder/src/mikro-orm-seeder.ts
Langstra: Shouldn't we need to recreate the schema after dropping it? Otherwise we cannot seed it again, right? The idea of the `refreshDatabase` method is that you get a clean/fresh database without data, but with the schema.

docs/docs/seeding.md
```diff
@@ -0,0 +1,188 @@
...
+
+class DatabaseSeeder extends Seeder {
+
+  async run(em: EntityManager): Promise<void> {
+    const author = em.create(Author, {
+      name: 'John Snow',
+      email: 'snow@wall.st'
+    });
+    await em.persistAndFlush(author);
+    em.clear();
```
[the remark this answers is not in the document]
Langstra: I think it is up to the user if they want to `flush` and `clear`. Most of the time after using a seeder you'd like to clear the em to prevent having a large memory foot print. If you seed a lot of entities and after that start running some tests you'd be using a lot of memory where it is probably not necessary.

Yes should be exported.

Langstra:
> also the `eval` should go, i dont understand why it should be needed - seeders should be exported and we should require them as we do with entities or the orm config. dont worry about the eslint rule (add ignore comment for it) - this is the right place to do it.

The cli has a command to seed the database and you can the seederclass as argument.
`npx mikro-orm seed --class=ProjectsSeeder`
In this case ProjectsSeeder is a string. This class must then be imported and instantiated. Agree that eval() is evil(), but I could not find a suitable solution.

tests/cli/CLIHelper.test.ts
Langstra: Personally I think the prefix is correct. The entities define the schema of the database. The instances of entities define the data which is saved in the database. This kind of package would also be called a database seeder. Also when looking at other frameworks this kind of functionality is called database seeding. So to keep it straight forward en recognizable I think this is the right command. However I'd like to hear your opinion and suggestion for the command you'd like to see.

packages/seeder/src/mikro-orm-seeder.ts
Langstra: What about `DatabaseSeeder`?

tests/seeder/factory.test.ts
```diff
@@ -0,0 +1,118 @@
+import { Collection, MikroORM } from '@mikro-orm/core';
+import { Factory } from '@mikro-orm/seeder';
+import { SqliteDriver } from '@mikro-orm/sqlite';
+import * as Faker from 'faker';
```
B4nan, line 4: we have `@types/faker` installed, is this really needed (using `* as Faker`)? and the `typeof Faker`?

B4nan (changes requested):

docs/docs/seeding.md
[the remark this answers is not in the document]
Langstra: It does decrease the complexity of the abstract factory class implementation, but I think it also increases the ease with which it can be implemented by a developer wanting to create a factory. Personally I do not like the `makeOne` and `makeMany` methods, if really necessary I'd go with `make` and `makeMany` or `makeOne` and `make`, but I'd prefer only to have the `make`.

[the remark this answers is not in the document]
B4nan: lets agree to disagree. this has already bitten me more than once, i would go for `make` and `makeOne`, just like we have `em.find` and `em.findOne`.

docs/docs/seeding.md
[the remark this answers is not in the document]
Langstra: Been giving this some thought, but I am not sure how we could `flush` and `clear` on the level above. This is something a user should do when implementing the seeder, right? The factory does this automatically when called in a seeder (at least if they call `create` and not `make`).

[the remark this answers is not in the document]
B4nan: Nope, this is something to be called in the seeder manager (or however you call it), where the `run` method is called - that is the one level above I am referring to. As mentioned, this should happen only in the top level seeder, so not in those invoked via `call` method.

docs/docs/seeding.md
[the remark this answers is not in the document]
Langstra: Aight, guess I'll have to settle ;)

tests/cli/CLIHelper.test.ts
Langstra: I think we could have some great discussions :sunglasses:

tests/seeder/factory.test.ts
Langstra: Yes I do believe so, there is no exported member `Faker`

B4nan: this project uses `esModuleInterop` so it should be always possible to do just `import Faker from 'faker';`

B4nan:
So what is it? draft? wip? or ready for review? :] I see you cracked the coverage 👍 

One thing I can say right ahead - I don't like the `seedString` method, why should we have something like that? Also from the implementation you require default export, not cool as well. Looks like its not documented (so probably internal), and I kinda understand why its there (but the name is pretty bad 😉), will give it a bit more thinking...

Another quick note, it looks like you have the `DatabaseSeeder` name for the default seeder hardcoded in some places (`handleFreshCommand`, `handleSchemaCommand`), this should be probably configurable.

Langstra:
Yeah, yesterday I thought I finished it, but then this morning I was like "oh shit this and this needs to be done".
So yeah, things I still need to work on:
- [ ] Add documentation for the other CLI commands
- [ ] Add configurable seeder class for `--seed` option for migration and schema fresh commands
- [ ] Add configurable default seed class for fresh commands
- [ ] Add documentation on the config option for the path for seed files
- [ ] Refactor seed commands to `seed:action`
- [ ] Add `seed:create` command for creating a seeder class

We need the `seedString` method for commands like `npx database:seed --class=BookSeeder`, because in the command `BookSeeder` is a string. So we load the seeder class file using `require` and then execute the seeder.

RDeluxe:
Any chance you took a look at @totomakers proposal ?

Langstra:
> Any chance you took a look at @totomakers proposal ?

Yeah, I used some ideas from it, but I do not like loading all SeederClasses and running them I'd like to give the user the option to run whatever seeder he/she wants.

B4nan:
Yeah we want this "main seeder" that orchestrates child ones, having an array of classes is not that powerful - we need to control the flow.

totomakers:
@Langstra i make some improvement in my company project. I can't share the whole project but i can let you take a peek.

Langstra:
> One thing I can say right ahead - I don't like the `seedString` method, why should we have something like that? Also from the implementation you require default export, not cool as well. Looks like its not documented (so probably internal), and I kinda understand why its there (but the name is pretty bad ), will give it a bit more thinking...

Any suggestions for this? I believe this would be the last part of the feature to fix. Otherwise I am quite confident that the feature is ready.

B4nan:
Don't worry too much about that, I am fine having it there, we should just pick a better name. 

Will try to look into this in more detail during weekend.

B4nan:
Time has come, please rebase this and let's merge it, master is now officially 5.x. I might tweak some things/naming directly there.

Langstra:
> Time has come, please rebase this and let's merge it, master is now officially 5.x. I might tweak some things/naming directly there.

Will work on it asap, won't have time for it this weekend, but I am aiming to fix it this week.

B4nan:
What about this? The rebase should be rather easy, right? :] 

Or did I miss it and we have new conflicts? If so, sorry about that, I guess (force) pushing won't trigger any notifications...

Langstra:
Yeah I already had, but now I have rebased it again. Should be fine and pass the pipeline again.

rubiin:
cool waiting for new version soon to use this

edorgeville:
Great job @Langstra !

Issue #907, opened 2020-10-07
the migration folder should be created if not found


rubiin:
If we set a migration folder to a non existent folder `for example: path: ./migrations` , 
we get something like this. It would be rather convienient to create a folder if not exists. Typeorm does that
`
[OperationalError: ENOENT: no such file or directory, scandir '/home/devina/nestjs-starter/migrations'] {
  cause: [Error: ENOENT: no such file or directory, scandir '/home/devina/nestjs-starter/migrations'] {
    errno: -2,
    code: 'ENOENT',
    syscall: 'scandir',
    path: '/home/devina/nestjs-starter/migrations'
  },
`

rubiin:
Another, thing i noticed was when path set to ./migrations , why is the folder being addressed to the root of the project. Shouldn't it should be on to the same directory as mikrorom config

Langstra:
> Another, thing i noticed was when path set to ./migrations , why is the folder being addressed to the root of the project. Shouldn't it should be on to the same directory as mikrorom config

I'd rather have the root be the actual root of the project. That way I can better predict where my folders are. Also I think this is logical when comparing it to other frameworks and libraries. For instance Nestjs, the cli generates all its stuff from the root of the project.

rubiin:
> > Another, thing i noticed was when path set to ./migrations , why is the folder being addressed to the root of the project. Shouldn't it should be on to the same directory as mikrorom config
> 
> I'd rather have the root be the actual root of the project. That way I can better predict where my folders are. Also I think this is logical when comparing it to other frameworks and libraries. For instance Nestjs, the cli generates all its stuff from the root of the project.

Also I see the temp folder with json of entities being created. Shouldnt this folder be automatically deleted after the migration is ran

B4nan:
> Also I see the temp folder with json of entities being created. Shouldnt this folder be automatically deleted after the migration is ran

Nope, its a cache folder, why would we wipe it? :] It has nothing to do with migrations, and it will be reused every time you initialize the ORM. You can disable caching if you want (it is "required" only for ts-morph provider, in v4 other providers have cache disabled by default, 
 and the default is reflect-metedata).

https://mikro-orm.io/docs/metadata-cache

rubiin:
> > Also I see the temp folder with json of entities being created. Shouldnt this folder be automatically deleted after the migration is ran
> 
[...]

PR #941, opened 2020-10-21
fix: only create migrations folder if migrationsList is not used

thomaschaaf:
https://github.com/mikro-orm/mikro-orm/issues/907#issuecomment-713532649

### DIFF
--- a/packages/migrations/src/Migrator.ts
+++ b/packages/migrations/src/Migrator.ts
@@ -38,14 +38,6 @@ export class Migrator {
     });
   }
 
-  async ensurePrerequisites() {
-    if (this.options.migrationsList?.length) {
-      await ensureDir(Utils.normalizePath(this.options.path!));
-    }
-
-    await this.storage.ensureTable();
-  }
-
   async createMigration(path?: string, blank = false, initial = false): Promise<MigrationResult> {
     await this.ensurePrerequisites();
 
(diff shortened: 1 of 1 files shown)


packages/migrations/src/Migrator.ts
```diff
@@ -38,8 +38,16 @@ export class Migrator {
     });
   }
 
+  async ensurePrerequisites() {
```
B4nan, line 41: let's make this private (and move it to the end of the file)

also the optional chaining (on following line) might need `/* istanbul ignore next */` as it usually reports missed branch even if both set/unset paths are taken (due to how it is transpiled)

B4nan:
Looking good, lets wait for the CI.

packages/migrations/src/Migrator.ts
```diff
@@ -39,7 +39,7 @@ export class Migrator {
   }
 
   async createMigration(path?: string, blank = false, initial = false): Promise<MigrationResult> {
-    await ensureDir(Utils.normalizePath(this.options.path!));
+    await this.ensurePrerequisites();
```
thomaschaaf, line 42: is it okay that this calls ```this.storage.ensureTable```?

thomaschaaf:
@B4nan I don't think the failing code climate has anything to do with my changes.

B4nan:
Yeah that's fine, some configuration mismatch. Tests are passing and full coverage is remained, that is what is important.

PR #946, opened 2020-10-22
docs: update deployment webpack config due to v4 release

alexstefanski:
- devDependencies defaults to empty object - @mikro-orm/core doesn't have devDependecies in its package.json so webpack throws error
- check is optional package is not like @mikro-orm/... - like all database adapters dependencies - if so than check extended name  (@mikro-orm/mssql) in optional dependencies


B4nan:
Thanks! Can you also update the next version of docs? It's in `docs/docs`.

PR #952, opened 2020-10-22
fix(core): ensure global filters are enabled by default

Langstra:
Add documentation for all parameters and describe when global filters
are default enabled or not

### DIFF
--- a/docs/docs/filters.md
+++ b/docs/docs/filters.md
@@ -33,6 +33,13 @@ const books2 = await orm.em.find(Book, {}, {
 });
 ```
 
+## Properties of filter
+
+There are three parameters you can use:
+- `name` - can be used to enable a filter on the query can also used to pass a parameter
+- `cond` - is the condition that should be added to the query when the filter is enabled. This can be a callback, even async
+- `default` - indicates if the filter is enabled by default on the query
+
 ## Parameters
 
 You can define the `cond` dynamically as a callback. This callback can be also 
--- a/packages/core/src/utils/Configuration.ts
+++ b/packages/core/src/utils/Configuration.ts
@@ -232,6 +232,10 @@ export class Configuration<D extends IDatabaseDriver = IDatabaseDriver> {
       this.options.charset = this.platform.getDefaultCharset();
     }
 
+    Object.keys(this.options.filters).forEach(key => {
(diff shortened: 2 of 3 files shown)


Langstra:
Probably more of a bug that the Filter is not enabled by default when added through the configuration. Will try to fix that and then update this PR.

docs/docs/filters.md
```diff
@@ -35,6 +35,11 @@ const books2 = await orm.em.find(Book, {}, {
 
 ## Parameters
 
+There are three parameters you can use:
```
B4nan, line 38: For me this is actually adding confusion. The paragraph is about filter parameters, and you start the sentence describing filter properties as if they are the parameters the paragraph is about. But that is not true, it's about the user params, that is the `args` in the `cond` callback. 

If this should be described somewhere, its the section above.

also the description is missing some important information

- `name` is used also for passing parameters
- `cond` can be callback, possibly async
- the default desc is weird (the word order), maybe just drop the ` to the query` at the end

and use `-` instead of `*` to be consistent

packages/core/src/utils/Configuration.ts
```diff
@@ -232,6 +232,10 @@ export class Configuration<D extends IDatabaseDriver = IDatabaseDriver> {
       this.options.charset = this.platform.getDefaultCharset();
     }
 
+    Object.keys(this.options.filters).forEach(key => {
+      this.options.filters[key].default = this.options.filters[key].default ?? true;
```
B4nan, line 236: this branch is not tested

B4nan (changes requested):

docs/docs/filters.md
Langstra: Ah alright, in that case I must have misunderstood the thing. Maybe we can rename the title `# Parameters` to `# Parameters of cond`, making it even more clear that we are talking about those?

B4nan: idk, to me it is not that confusing, `Parameters of cond` sounds weird. maybe better name would be `Parametric filters`

Langstra: Guess it is fine like it is then. I'll resolve all comments.

B4nan:
Looking good, thanks!

Issue #889, opened 2020-09-29
SQLITE: wrong number of arguments to function count()


njeirath:
**Describe the bug**
I had submitted #887 yesterday and ran into another issue with the same setup.  I'm not sure if they're related and if this is a sqlite only issue like the other one.  

For reference, the entity I'm using:

```
export class FriendLink {
    requestor: User
    requestee: User
    created: Date

    [PrimaryKeyType]: [string, string]   //User is using a string as it's ID

    constructor(requestor: User, requestee: User) {
        this.requestor = requestor
        this.requestee = requestee
        this.created = new Date()
    }
}

export const schema = new EntitySchema<FriendLink>({
    // Some other stuff here
    properties: {
        requestor: {
            reference: 'm:1',
            entity: 'User',
            nullable: false,
            primary: true,
            joinColumn: 'requestor_user_id',
        },
        requestee: {
            reference: 'm:1',
            entity: 'User',
            nullable: false,
            primary: true,
            joinColumn: 'requestee_user_id',
        },
        created: {type: 'datetime', nullable: false},
    },
})

export const entity = FriendLink
```

And when I try to use this entity to perform a count operation as in the following code:

```
const totalCount = await em.count(FriendLink, {requestee: user})
```

I get an error like:

**Stack trace**
```
DriverException: select count(distinct `e0`.`requestor_user_id`, `e0`.`requestee_user_id`) as `count` from `friend_link` as `e0` where `e0`.`requestee_user_id` = 'user123' - SQLITE_ERROR: wrong number of arguments to function count()
```

**Expected behavior**
I'm assuming, but haven't validated, that the issue is that the distinct needs parenthesis like:
```
select count(distinct (`e0`.`requestor_user_id`, `e0`.`requestee_user_id`)) as `count ...`
```

Alternatively, since both those columns make up a composite key I'd expect them to already be unique so rather than using distinct it could likely just be `select count(*) as count ...`

I was able to work around this for the time being by executing my own SQL query to get what I needed.

**Versions**

| Dependency | Version |
| - | - |
| node | v14.5.0 |
| typescript | 3.8.3 |
| mikro-orm | 4.0.7 |
| your-driver | sqlite |

B4nan:
mmmm would be better to address this in knex, apparently they do not support distinct count with composite keys in sqlite...

https://github.com/knex/knex/pull/2449#issuecomment-450590918

reported, and PR'ed almost 2 years ago

[...]

PR #955, opened 2020-10-23
feat(count): initial implementation of loadCount

AzariasB:
Instead of populating a collection to get the number of items, fetches the count directly from the database. (`collection.loadCount()`)
An optional parameter `refresh` (false by default) can be passed to the function to force a database query instead of using the cache.

I still need to update the documentation.

Any feedback would be greatly appreciated :)


B4nan:
please remove all code style changes.

AzariasB:
Ow, didn't see all the changes made by the vscode plugin :|, will fix that

packages/core/src/entity/Collection.ts
```diff
@@ -47,6 +47,23 @@ export class Collection<T, O = unknown> extends ArrayCollection<T, O> {
     return super.getItems();
   }
 
+  /**
+   * Counts the number of entries in the database
```
B4nan, line 51: this sounds a bit misleading, it's count of items of this collection, loaded from database. 

```suggestion
   * Gets the count of collection items from database instead of counting loaded items.
   * The value is cached, use `refresh = true` to force reload it.
```

packages/core/src/entity/ArrayCollection.ts
```diff
@@ -80,10 +88,14 @@ export class ArrayCollection<T, O> {
    */
   hydrate(items: T[]): void {
     this.items.clear();
+    if (Utils.isDefined(this._count)) {
+      this._count = 0;
+    }
```
B4nan, line 93: i would prefer raw `typeof` instead of the `Utils` helper (in this particular file), as it could result in circular requires easily.

but here the condition does not make sense anyway, you would be re-adding to the cached value every time its re-hydrated

```suggestion
    this._count = 0;
```

packages/core/src/entity/ArrayCollection.ts
```diff
@@ -2,6 +2,7 @@ import { AnyEntity, Dictionary, EntityProperty, IPrimaryKey, Primary } from '../
 import { Reference } from './Reference';
 import { wrap } from './wrap';
 import { ReferenceType } from '../enums';
+import { Utils } from '../utils';
```
B4nan, line 5: i would prefer raw typeof instead of the Utils helper (in this particular file), as it could result in circular requires easily.

```suggestion
```

tests/issues/GH949.test.ts
```diff
@@ -0,0 +1,88 @@
...
+class B {
+
+  @PrimaryKey()
+  id!: number;
+
+  @ManyToOne(() => A)
+  a!: A;
+
+}
+describe('GH issue 949', () => {
```
B4nan, line 26: i would prefer to test this in existing test cases, there are many entities with collections. this issues folder is meant for bugs that are hard to reproduce with what we already have, but in general the more files we have here, the slower the tests will run.

use author.books for 1:m, book.tags for m:n owner and tag.books for inverse, i'd put it in one of the sqlite EM tests. 

also use real description, instead of just pointing to issue. again, this is mainly for bugs, where its hard to describe it.

tests/issues/GH949.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  test(`GH issue 949`, async () => {
+    let aEntity = new A();
+    aEntity.bItems.add(new B());
+    // Entity not managed yet
+    expect(() => aEntity.bItems.loadCount()).rejects.toThrow(ValidationError);
+
+    await orm.em.persistAndFlush(aEntity);
+
+    if (!aEntity) { return; }
```
B4nan, line 50: uh why? you just defined it, this will be always true

tests/issues/GH949.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    // Force refresh
+    expect(await aEntity.bItems.loadCount(true)).toEqual(1);
+    // Testing array collection implemntation
+    await orm.em.flush();
+    orm.em.clear();
+
+
+    // Updates when removing an item
+    aEntity = (await orm.em.findOne(A, aEntity.id))!;
```
B4nan, line 68: instead of the parens and `!`, we can use `findOneOrFail`

tests/issues/GH949.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    expect(await aEntity.bItems.loadCount()).toEqual(3);
```
B4nan, line 69: in general the async expectations should be written this way:

```suggestion
    await expect(aEntity.bItems.loadCount()).resolves.toEqual(3);
```

tests/issues/GH949.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    // Adding new items
+    const laterRemoved = new B();
+    aEntity.bItems.add(laterRemoved, new B());
+    const threeItms = await aEntity.bItems.loadCount();
+    expect(threeItms).toEqual(3);
+
+    // Force refresh
+    expect(await aEntity.bItems.loadCount(true)).toEqual(1);
+    // Testing array collection implemntation
```
B4nan, line 62: typo

B4nan (changes requested):
left some comments, but in general its looking good, ty!

AzariasB:
I'm having an issue that has been fixed for #956, should I merge mikro-orm master's branch into my own ?

B4nan:
Yes, you need to rebase and resolve conflicts. Or close it and create new PR if you struggle with that, and just pick the changes you made.

packages/core/src/entity/Collection.ts
```diff
@@ -47,6 +47,24 @@ export class Collection<T, O = unknown> extends ArrayCollection<T, O> {
...
+   */
+  async loadCount(refresh = false): Promise<number> {
+    const em = this.owner.__helper!.__em;
+
+    if (!em) {
+      throw ValidationError.entityNotManaged(this.owner);
+    }
+
+    if (refresh || !Utils.isDefined(this._count)) {
+      this._count = await em.count(this.property.type,{ [this.property.mappedBy || this.property.inversedBy]: this.owner.__helper!.getPrimaryKey() });
```
AzariasB, line 62: So, `createCondition`  was incorrect for n:m relations as it just returned a condition dictionary with empty ids: `{id : {$in : [] } }`
I'm not sure what I did is the correct way of creating the condition though ...

B4nan: `createCondition` works fine, the collection items need to be set upfront to use it. in other words, it is not the method you should be using for this use case.

take a look at `loadFromPivotTable`, that is where this happens (but it loads the items - their PKs - so its again not what you should use, but a method for inspiration).

AzariasB: Are there existing entities with composite primary keys and many-to-many relationships so that I can test this case ?

B4nan: https://github.com/mikro-orm/mikro-orm/blob/master/tests/composite-keys.sqlite.test.ts
https://github.com/mikro-orm/mikro-orm/blob/master/tests/composite-keys.mysql.test.ts

AzariasB:
I'm having an issue when testing the loadCount with many-to-many relationShips :
With composite keys, the count queries [looks like](https://github.com/mikro-orm/mikro-orm/pull/955/checks?check_run_id=1302792939#step:12:139) : ``select count(distinct `e0`.`first_name`, `e0`.`last_name`) as count ...`` which is not a valid SQL query (too many arguments for count). I'm not sure how to deal with this ?

B4nan:
That's a bug in knex, sqlite dialect does not support distinct count with multiple fields. I'd say ignore that for now, maybe comment that test out and point to this issue as this is blocked by knex:

https://github.com/mikro-orm/mikro-orm/issues/889

In the long run I am planning to remove the dependency on knex and build the queries myself to have absolute control, there are more and more issue popping up that are hard or impossible to work around...

B4nan:
There are few minor things I'd do differently, but let's not clutter this up :] 

Last thing worth testing is how this works with mongo, as the m:n collections are implemented differently (no pivot tables). 

Also there is still one brach not tested in `Collection.ts:270`.

AzariasB:
Yes, I still need to add more tests (mysql composite keys)
For mongo, is there an existing test file  for composite keys ?

B4nan:
There are no composite keys in mongo, don't worry about that.

AzariasB:
I'm having an issue with the count method of the entitymanager for mongod-db :
[The tests fails when using count with a filter](https://github.com/mikro-orm/mikro-orm/pull/955/checks?check_run_id=1308372632#step:12:145). Am I using the filters incorrectly ?

B4nan:
There are no joins in mongo, it is not possible to query anything from the inverse side, so yes, this is correct behaviour in mongo unfortunately. 

Also M:N collections are implemented as arrays on the owning entity, so the actual count on the entity will be always the correct one. You should make it conditional in a similar way to the `Collection.init()` (based on reference type and `Platform.usesPivotTable` method).

AzariasB:
All right, I think everything's good now :)

packages/core/src/entity/Collection.ts
```diff
@@ -47,6 +47,28 @@ export class Collection<T, O = unknown> extends ArrayCollection<T, O> {
...
+
+    if (!em) {
+      throw ValidationError.entityNotManaged(this.owner);
+    }
+
+    if (refresh || !Utils.isDefined(this._count)) {
+      if (em.getDriver().getPlatform().usesPivotTable()) {
+        this._count = await em.count(this.property.type, this.createLoadCountCondition({}));
+      } else {
+        this._count = this.length;
```
B4nan, line 65: this will not work for 1:m relations in mongo (that should be handled just like in SQL)

B4nan:
Ok, let's merge it, thanks!

AzariasB:
Perfect, thanks !
It was a bit toughter than I expected
Thank a lot for the help throughout this pull request :), and thank you very much for the work you put into mikro-orm !

PR #959, opened 2020-10-25
feat(core): maintain transaction context automatically

B4nan:
Domain API is now used behind the scenes to automatically bare the tx context. This means
we can now use any EM, repository or a service (from DI), regardless of what EM instance
they were created from.

```ts
// this repo is based on `orm.em`, but thanks `TransactionContext` helper,
// it will use the right EM behind the scenes when used inside `em.transactional()`
const repo = orm.em.getRepository(Author4);

await orm.em.transactional(async () => {
  const a = await repo.findOneOrFail(god1);
  a.name = 'abc';
});
```

### DIFF
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -1,6 +1,6 @@
 import { inspect } from 'util';
 
-import { Configuration, QueryHelper, Utils } from './utils';
+import { Configuration, QueryHelper, TransactionContext, Utils } from './utils';
 import { AssignOptions, EntityAssigner, EntityFactory, EntityLoader, EntityRepository, EntityValidator, IdentifiedReference, Reference } from './entity';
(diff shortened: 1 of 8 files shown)


merceyz:
This will override the current `domain.active` meaning any custom things in there will be lost when using `transactional`, could it instead attach itself to the current active domain or copy the data from the active domain?

```ts
import domain from 'domain';

(async () => {
	const userDomain = domain.create();
	userDomain.__requestId = 1;

	userDomain.run(async () => {
		console.log(domain.active.__requestId);
		// 1

		const transactionalDomain = domain.create();
		transactionalDomain.__transactional = 2;
		transactionalDomain.run(async () => {
			console.log(domain.active.__requestId, domain.active.__transactional);
			// undefined, 2
		});
	});
})();

```

B4nan:
Sure, good idea

PR #998, opened 2020-10-27
fix(validation): don't validate inherited STI props

### LINKED ISSUE #997: ValidationError: Child2.rel is not initialized, define it as 'rel = new Collection<Child1Specific>(this);' when using STI and 'problematic' relation is defined on other child
**Describe the bug**
When querying table, that uses STI, with `QueryBuilder` things broke up and orm starts complaining about uninitialized relation, although it is initialized and defined, but on other children 

**Stack trace**
```
ValidationError: Child2.rel is not initialized, define it as 'rel = new Collection<Child1Specific>(this);'

    at Function.fromCollectionNotInitialized (/home/k/WebstormProjects/mikro-orm/packages/core/src/errors.ts:32:12)
    at EntityValidator.validateCollection (/home/k/WebstormProjects/mikro-orm/packages/core/src/entity/EntityValidator.ts:106:29)
    at /home/k/WebstormProjects/mikro-orm/packages/core/src/entity/EntityValidator.ts:13:14
    at EntityValidator.validate (/home/k/WebstormProjects/mikro-orm/packages/core/src/entity/EntityValidator.ts:11:16)
    at SqlEntityManager.merge (/home/k/WebstormProjects/mikro-orm/packages/core/src/EntityManager.ts:479:20)
    at SqlEntityManager.map (/home/k/WebstormProjects/mikro-orm/packages/core/src/EntityManager.ts:446:17)
    at /home/k/WebstormProjects/mikro-orm/packages/knex/src/query/QueryBuilder.ts:389:34
    at Array.map (<anonymous>)
    at QueryBuilder.getResultList (/home/k/WebstormProjects/mikro-orm/packages/knex/src/query/QueryBuilder.ts:389:16)
    at processTicksAndRejections (internal/process/task_queues.js:94:5)
    at Object.<anonymous> (/home/k/WebstormProjects/mikro-orm/tests/issues/GH8455.test.ts:93:21)
```

**To Reproduce**
1. Run test case GH997.test.ts

**Expected behavior**
Should work like it works without `QueryBuilder`

**Versions**

| Dependency | Version |
| - | - |
| node | v12.14.1 |
| typescript | 4.0.5 |
| mikro-orm | 4.2 |
| mysql | dunno |

whocaresk:
Inherited props of STI entities shouldn't be validated because they don't exist in real entity definition.

Closes #997

### DIFF
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -475,8 +475,15 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
       return entity;
     }
 
+    let meta = this.metadata.find(entityName);
+
+    if (meta!.discriminatorColumn) {
+      const childMeta = this.metadata.getByDiscriminatorColumn(meta!, data as EntityData<T>);
+      meta = childMeta ?? meta;
+    }
+
     entity = Utils.isEntity<T>(data) ? data : this.getEntityFactory().create<T>(entityName, data as EntityData<T>, { merge: true, refresh, convertCustomTypes });
-    this.validator.validate(entity, data, this.metadata.find(entityName)!);
+    this.validator.validate(entity, data, meta!);
     this.getUnitOfWork().merge(entity);
 
     return entity;
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -475,15 +475,11 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
       return entity;
     }
 
-    let meta = this.metadata.find(entityName);
-
-    if (meta!.discriminatorColumn) {
-      const childMeta = this.metadata.getByDiscriminatorColumn(meta!, data as EntityData<T>);
-      meta = childMeta ?? meta;
-    }
+    const meta = this.metadata.get(entityName);
+    const childMeta = this.metadata.getByDiscriminatorColumn(meta!, data as EntityData<T>);
 
     entity = Utils.isEntity<T>(data) ? data : this.getEntityFactory().create<T>(entityName, data as EntityData<T>, { merge: true, refresh, convertCustomTypes });
-    this.validator.validate(entity, data, meta!);
+    this.validator.validate(entity, data, childMeta !== undefined ? childMeta : meta);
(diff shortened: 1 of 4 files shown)


packages/core/src/metadata/MetadataStorage.ts
```diff
@@ -60,6 +60,24 @@ export class MetadataStorage {
...
   }
 
+  getByDiscriminatorColumn<T>(meta: EntityMetadata<T>, data: EntityData<T>): EntityMetadata<T> | undefined {
+    const prop = meta.properties[meta.root.discriminatorColumn!];
+
+    if (!prop) {
+      return undefined;
+    }
+
+    const value = data[prop.name];
```
B4nan, line 70:
```suggestion
    const value = data[meta.root.discriminatorColumn!];
```

packages/core/src/EntityManager.ts
```diff
@@ -475,8 +475,11 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
       return entity;
     }
 
+    const meta = this.metadata.get(entityName);
```
B4nan, line 478: it is ok to use `find()!` here (faster than `get()`), as the validation check is already done few lines above (on 471)

```suggestion
    const meta = this.metadata.find(entityName)!;
```

packages/core/src/EntityManager.ts
```diff
@@ -475,8 +475,11 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
...
+    const childMeta = this.metadata.getByDiscriminatorColumn(meta, data as EntityData<T>);
+
     entity = Utils.isEntity<T>(data) ? data : this.getEntityFactory().create<T>(entityName, data as EntityData<T>, { merge: true, refresh, convertCustomTypes });
-    this.validator.validate(entity, data, this.metadata.find(entityName)!);
+    this.validator.validate(entity, data, childMeta !== undefined ? childMeta : meta);
```
B4nan, line 482: the original approach with `??` was better :] 

```suggestion
    this.validator.validate(entity, data, childMeta ?? meta);
```

B4nan (changes requested):

PR #1000, opened 2020-10-27
fix(core): do not override child class properties

patricklx:
when the base class has properties that are also defined in the child class with different options (e.g. nullable or not nullable). the base class property would just override the child properties.
with this change the properties will be filtered out by name and allow the child class to keep its own property definitions while inheriting others from the parent


B4nan:
The change itself looks fine to me, but the empty PR description don't :] So please provide some motivation for this change.

B4nan:
Thanks!

PR #1001, opened 2020-10-27
feat(sql): add support for schemas (multi tenancy)

patricklx:
this is to support schemas in e.g . postgres/mysql (multi tenancy)
This should also work for creating migrations.
Each ORM instance can have another schema assigned to work with multiple schemas. (schema passed to MikroORM.init options)

the initial migration files will also support dynamic schema by passing in the schema variable.
e.g.
```
async up(): Promise<void> {
    const schema = this.config.get('schema' as any);

    this.addSql(`create index "${schema}"."BOOKS_KEY_index" on "${schema}"."BOOKS" ("KEY");`);
  }
  ```

### DIFF
--- a/packages/knex/src/AbstractSqlConnection.ts
+++ b/packages/knex/src/AbstractSqlConnection.ts
@@ -16,7 +16,6 @@ export abstract class AbstractSqlConnection extends Connection {
 
   protected platform!: AbstractSqlPlatform;
   protected client!: Knex;
-  knexSchemaGetter: (() => any) | undefined;
 
   constructor(config: Configuration, options?: ConnectionOptions, type?: 'read' | 'write') {
     super(config, options, type);
--- a/packages/knex/src/AbstractSqlConnection.ts
+++ b/packages/knex/src/AbstractSqlConnection.ts
@@ -16,14 +16,22 @@ export abstract class AbstractSqlConnection extends Connection {
 
   protected platform!: AbstractSqlPlatform;
   protected client!: Knex;
+  knexSchemaGetter: (() => any) | undefined;
 
   constructor(config: Configuration, options?: ConnectionOptions, type?: 'read' | 'write') {
     super(config, options, type);
     this.patchKnexClient();
   }
 
+  /**
+  * this is required to support schemas for migrations
+  */
   getKnex(): Knex {
     return this.client;
+  }
+
+  getSchema(): string {
+    return this.config.get('schema');
   }
 
   async close(force?: boolean): Promise<void> {
@@ -101,6 +109,7 @@ export abstract class AbstractSqlConnection extends Connection {
   async execute<T extends QueryResult | EntityData<AnyEntity> | EntityData<AnyEntity>[] = EntityData<AnyEntity>[]>(queryOrKnex: string | QueryBuilder | Raw, params: any[] = [], method: 'all' | 'get' | 'run' = 'all', ctx?: Transaction): Promise<T> {
     if (Utils.isObject<QueryBuilder | Raw>(queryOrKnex)) {
       ctx = ctx ?? ((queryOrKnex as any).client.transacting ? queryOrKnex : null);
+      (queryOrKnex as QueryBuilder).withSchema?.(this.config.get('schema'));
       const q = queryOrKnex.toSQL();
       queryOrKnex = q.sql;
       params = q.bindings as any[];
--- a/packages/knex/src/AbstractSqlConnection.ts
+++ b/packages/knex/src/AbstractSqlConnection.ts
@@ -22,9 +22,6 @@ export abstract class AbstractSqlConnection extends Connection {
     this.patchKnexClient();
   }
 
-  /**
-  * this is required to support schemas for migrations
-  */
   getKnex(): Knex {
     return this.client;
   }
--- a/packages/knex/src/schema/SchemaGenerator.ts
+++ b/packages/knex/src/schema/SchemaGenerator.ts
@@ -48,10 +48,6 @@ export class SchemaGenerator {
     const createdColumns: string[] = [];
     let ret = '';
 
-    if (this.connection.getSchema()) {
-      ret += `create schema "${this.connection.getSchema()}"`;
-    }
-
     for (const meta of metadata) {
       ret += this.dump(this.createTable(meta, createdColumns));
     }
@@ -156,7 +152,8 @@ export class SchemaGenerator {
   }
 
   private getUpdateTableSQL(meta: EntityMetadata, schema: DatabaseSchema, safe: boolean, createdColumns: string[]): string {
-    const table = schema.getTable(meta.collection);
+    const tblName = this.connection.getSchema() ? this.connection.getSchema() + '.' + meta.collection : meta.collection;
+    const table = schema.getTable(tblName);
 
     if (!table) {
       return this.dump(this.createTable(meta, createdColumns));
@@ -166,7 +163,8 @@ export class SchemaGenerator {
   }
 
   private getUpdateTableFKsSQL(meta: EntityMetadata, schema: DatabaseSchema, createdColumns: string[]): string {
-    const table = schema.getTable(meta.collection);
+    const tblName = this.connection.getSchema() ? this.connection.getSchema() + '.' + meta.collection : meta.collection;
+    const table = schema.getTable(tblName);
 
     if (!table) {
       return this.dump(this.knex.schema.withSchema(this.connection.getSchema()).alterTable(meta.collection, table => this.createForeignKeys(table, meta)));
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

File: packages/knex/src/AbstractSqlConnection.ts [hunk 1]
```diff
@@ -7,14 +7,26 @@ export abstract class AbstractSqlConnection extends Connection {
 
   protected platform!: AbstractSqlPlatform;
   protected client!: Knex;
+  schema: string;
 
   constructor(config: Configuration, options?: ConnectionOptions, type?: 'read' | 'write') {
     super(config, options, type);
     this.patchKnexClient();
+    this.schema = this.config.get('driverOptions').schema;
```
B4nan, line 15: just one note, will ignore the rest until you provide some description and tests so i can understand it better...

`driverOptions` is wrong place for this, if we handle something in the ORM, the configuration should be above (probably in `ConnectionOptions` interface). `driverOptions` is for users to be able to override things that the ORM does, that is the only purpose.

tests/bootstrap.ts
```diff
@@ -91,10 +91,11 @@ export async function initORMMySql<D extends MySqlDriver | MariaDbDriver = MySql
   return orm as MikroORM<D>;
 }
 
-export async function initORMPostgreSql(loadStrategy = LoadStrategy.SELECT_IN) {
+export async function initORMPostgreSql(loadStrategy = LoadStrategy.SELECT_IN, schema?: string) {
   const orm = await MikroORM.init<PostgreSqlDriver>({
     entities: [Author2, Address2, Book2, BookTag2, Publisher2, Test2, FooBar2, FooBaz2, FooParam2, Label2, Configuration2, BaseEntity2, BaseEntity22],
     dbName: `mikro_orm_test`,
+    schema: schema,
```
B4nan, line 98:
```suggestion
    schema,
```

tests/bootstrap.ts
```diff
@@ -108,6 +109,9 @@ export async function initORMPostgreSql(loadStrategy = LoadStrategy.SELECT_IN) {
   const schemaGenerator = new SchemaGenerator(orm.em);
   await schemaGenerator.ensureDatabase();
   const connection = orm.em.getConnection();
+  if (connection.getSchema()) {
+    await connection.execute('SET search_path TO ??', [connection.getSchema()]);
```
B4nan, line 113: could this happen automatically?

B4nan (changes requested):
thanks! added few suggestions to the code. here are some general remarks:

- the tests are failing with `no schema has been selected to create`
- we dont need all the pg tests twice, so please prune it to minimum that is actually needed to test the feature
- don't forget to add docs :] maybe new page in recipes section about multi tenancy, that would be great. doesn't have to be very verbose, just few notes about the setup with few code snippets

packages/knex/src/AbstractSqlConnection.ts
```diff
@@ -7,14 +7,31 @@ export abstract class AbstractSqlConnection extends Connection {
...
   protected client!: Knex;
+  knexSchemaGetter: (() => any) | undefined;
 
   constructor(config: Configuration, options?: ConnectionOptions, type?: 'read' | 'write') {
     super(config, options, type);
     this.patchKnexClient();
   }
 
+  /**
+  * this is required to support schemas for migrations
```
B4nan, line 18: I'd like to hear more about this, why is it required and how exactly it works? Is it only for migrations? If so, maybe we could do it in the `MigrationRunner` class instead? 

Also that comment belongs to the implementation, its not about the method itself.

patricklx: basically this is required so that when generating the migration files it also prefixes the table names with the schema name. this is done by calling `withSchema`. But yes, this is an ugly hack. I would need to replace all `knex.schema.createTable/alterTable` and so on with `knex.schema.withSchema().createTable` in schema generator i think

B4nan: sounds like better solution, there are only few of those and all of them in the same file...

packages/knex/src/schema/SchemaGenerator.ts
```diff
@@ -48,6 +48,10 @@ export class SchemaGenerator {
     const createdColumns: string[] = [];
     let ret = '';
 
+    if (this.connection.getSchema()) {
+      ret += `create schema "${this.connection.getSchema()}"`;
```
B4nan, line 52: why is the semicolon missing? also i believe i already said you should use `platform.quoteIdentified()` here

packages/postgresql/src/PostgreSqlConnection.ts
```diff
@@ -7,6 +7,9 @@ export class PostgreSqlConnection extends AbstractSqlConnection {
   async connect(): Promise<void> {
     this.patchKnex();
     this.client = this.createKnexClient('pg');
+    if (this.config.get('schema')) {
+      await this.execute('set search_path to ??', [this.config.get('schema')]);
```
B4nan, line 11: is this really enough to run on connection instance? the ORM connection class is singleton, i'd say we need to run this on each knex connection (e.g. for each one in the pool)

B4nan (changes requested):
> according to postgresql.org/docs/9.3/ddl-schemas.html#DDL-SCHEMAS-PATH
it is best to specify the search path instead of prefixing all identifiers

But then we are talking about postgres only support, but this should work just fine in mysql too (with hard coded schema names), right?

I see you removed the second postgres test file and just used schema in the existing one. i am not sure if that is the way to go, sounds like you are now testing only one edge case - usage with single (non standard) schema name. tests for implicit (public) schema are now gone this way, and you did not even test usage with multiple schemas (which this PR is about, right?)

My point was that you should test the feature you are adding, in the way it is supposed to be used. I'd say we should have a dedicated test file for multi schema support, just like we have for embeddables or composite keys. I'd rather not touch existing tests.

Also please rebase.

What are your plans with this? Want to ship v4.3 later this week finally, so asking if I should wait for this to be ready? But don't want to postpone for longer than few days, there are some bugfixes that need to go out :]

packages/postgresql/src/PostgreSqlConnection.ts
patricklx: but each orm will receive another connection instance or not?
so, if MikoORM.init() is called and then another time call MikoORM.init() with different schema name, the connection instance for those will be different?

patricklx:
> > according to postgresql.org/docs/9.3/ddl-schemas.html#DDL-SCHEMAS-PATH
> > it is best to specify the search path instead of prefixing all identifiers
> 
> But then we are talking about postgres only support, but this should work just fine in mysql too (with hard coded schema names), right?
> 
> I see you removed the second postgres test file and just used schema in the existing one. i am not sure if that is the way to go, sounds like you are now testing only one edge case - usage with single (non standard) schema name. tests for implicit (public) schema are now gone this way, and you did not even test usage with multiple schemas (which this PR is about, right?)
> 
> My point was that you should test the feature you are adding, in the way it is supposed to be used. I'd say we should have a dedicated test file for multi schema support, just like we have for embeddables or composite keys. I'd rather not touch existing tests.
> 
> Also please rebase.
> 
> What are your plans with this? Want to ship v4.3 later this week finally, so asking if I should wait for this to be ready? But don't want to postpone for longer than few days, there are some bugfixes that need to go out :]

i'm busy with other things right now... :[
it will not work for mysql since they do not have a similar thing as schemas which allows specific user permissions and so on

B4nan:
> i'm busy with other things right now... :[

Sure, no probs, just wanted to know if I should wait a bit more. 

> it will not work for mysql since they do not have a similar thing as schemas which allows specific user permissions and so on

In mysql, database is the same as schema (it is actually an alias). Users can have access to multiple databases. I don't see a reason why this could not work with mysql (obviously it won't work via `set search_path`, but it could with prefixed table names). Actually there are tests in mysql that are using multiple schemas already.

packages/postgresql/src/PostgreSqlConnection.ts
B4nan: yes, but is that enough? there will be many actual connections, and you are running it only on a single one. 

this looks like correct way to do it:

> Note: Knex's PostgreSQL client allows you to set the initial search path for each connection automatically using an additional option "searchPath" as shown below.

```
var pg = require('knex')({
  client: 'pg',
  connection: process.env.PG_CONNECTION_STRING,
  searchPath: ['knex', 'public'],
});
```

http://knexjs.org/#Installation-client

demsey2:
hey @patricklx are you planning to continue with that PR? I need that change too but for MySQL

B4nan:
There is another PR improving the multi schema support - #1424 - so would be good to collaborate.

B4nan:
Closing as the PR became stale and most of the changes are implemented in #2296.

PR #1030, opened 2020-11-04
fix(schema): fix diffing FKs in MySQL 8

jsprw:
It is exactly the same issue as #398 with exactly the same solution. This caused https://github.com/mikro-orm/mikro-orm/blob/master/packages/knex/src/schema/SchemaHelper.ts#L112 to return

```
{
  "undefined": {
    "columnName": undefined,
    "constraintName": undefined,
    "referencedTableName": undefined,
    "referencedColumnName":  undefined,
    "updateRule": undefined,
    "deleteRule": undefined
  }
}
``` 

for all found foreign keys causing all foreign keys to be dropped and re-added.

### DIFF
--- a/packages/mysql-base/src/MySqlSchemaHelper.ts
+++ b/packages/mysql-base/src/MySqlSchemaHelper.ts
@@ -62,7 +62,7 @@ export class MySqlSchemaHelper extends SchemaHelper {
   }
 
   getForeignKeysSQL(tableName: string, schemaName?: string): string {
-    return `select distinct k.constraint_name, k.column_name, k.referenced_table_name, k.referenced_column_name, c.update_rule, c.delete_rule `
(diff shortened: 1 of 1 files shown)


jsprw:
How to retry checks? It failed on preparing a docker.. I have also checked the tests and there doesn't seem to be any tests for this code to fix.

B4nan:
Not sure if you can do it, it might be restricted to the repo maintainers only. Tried to rerun the CI, it failed the same way, weird. Even more weird that it is passing in node 10/12/15, but fails in 14. Anyway, the change is clear to me, will merge it, let's see if the CI will pass on the master.

B4nan:
You can test the dev release via `4.2.4-dev.92` (for mysql package only, other packages have different versions, so better to install latest `next` version).

PR #1035, opened 2020-11-04
docs: add example of nullable string enum definition

### LINKED ISSUE #1034: Should add an explicit example of Enum defined in another file, with nullable: true
I had trouble generating a migration with the correct SQL for a `nullable: true` string enum.

I suggest adding this 4th example in User class
https://mikro-orm.io/docs/defining-entities/#enums :

```ts
@Enum({ items: () => OutsideNullableEnum, nullable: true })
outsideNullable?: OutsideNullableEnum; // string enum defined outside of this file, may be null
```

pacome35220:
close #1034

### DIFF
--- a/docs/docs/defining-entities.md
+++ b/docs/docs/defining-entities.md
@@ -233,6 +233,8 @@ export class User {
   @Enum(() => OutsideEnum)
   outside!: OutsideEnum; // string enum defined outside of this file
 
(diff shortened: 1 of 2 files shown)


B4nan:
Please add it also to the next docs too (`./docs/docs`)

B4nan:
Thanks!

PR #1051, opened 2020-11-07
feat(core): allow extending embeddables

### LINKED ISSUE #1049: Embeddable class extends do not inherit properties
When using embeddable and extending it to another, fields are not created in database as it cant find the properties. 

```
@Embeddable()
export class ResolvedExpression extends Expression {

/// No properties get added to database.

}

@Embeddable()
export class Expression implements ExpressionInterface {
  
  @Property()
  value?: string;

}
```

dvlalex:
This fixes the inheritance issue on embeddable types.

Closes #1049


packages/core/src/metadata/MetadataDiscovery.ts
```diff
@@ -39,11 +39,10 @@ export class MetadataDiscovery {
 
     // ignore base entities (not annotated with @Entity)
     const filtered = this.discovered.filter(meta => meta.name);
-    filtered.forEach(meta => Object.values(meta.properties).forEach(prop => this.initEmbeddables(meta, prop)));
     filtered.forEach(meta => this.initSingleTableInheritance(meta, filtered));
     filtered.forEach(meta => this.defineBaseEntityProperties(meta));
     filtered.forEach(meta => this.metadata.set(meta.className, EntitySchema.fromMetadata(meta).init().meta));
-    filtered.forEach(meta => this.defineBaseEntityProperties(meta));
```
B4nan, line 46: there was a reason to do this twice, but as the tests are passing, let's try to remove it now

tests/issues/GH1049.test.ts
```diff
@@ -0,0 +1,118 @@
...
+  @Property()
+  legs!: number;
+
+  protected constructor(legs?: number) {
+    if (legs) { this.legs = legs; }
+  }
+
+}
+
+abstract class Pet extends Animal {
```
B4nan, line 15: the issue was about extending embeddables, here you are testing extending a class to build an embeddable.

add `@Embeddable()` decorators and make those parent classes concrete implementations to truly test what the issue was about (i guess it's ok to keep `Animal` as is, and change only `Pet`).

also this should go to the `embedded-entities.mongo.test.ts` file (for performance reasons, each added test file makes everything a bit slower)

B4nan (changes requested):
nice, thanks! only one small suggestion to truly test what the issue was about

packages/core/src/metadata/MetadataDiscovery.ts
dvlalex: I thought so, but I went through it twice and it seemed as yielding the same output, at least for my input. I'll go through it again once i move the test.

B4nan: yeah i was checking the commit that added the line, and it was composite keys support. from that time, i improved lots of internals in the discovery process, so it is probably not needed anymore.

PR #1052, opened 2020-11-07
chore: deeply nested assign tests

ballwood:
Added a few deeply nested tests around assign. Some are failing (removal and update) and I'm not sure about 'attaches to entity currently in map when using em'. Feel free to take over the PR and make any changes you want to it

### DIFF
--- a/tests/EntityAssigner.mysql.test.ts
+++ b/tests/EntityAssigner.mysql.test.ts
@@ -218,7 +218,7 @@ describe('EntityAssignerMySql', () => {
       ],
     };
 
-    wrap(toSave).assign(inputUpdated);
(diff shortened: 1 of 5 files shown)


B4nan:
I would rather not add any SQL entities that are used in the rest of the tests. It means you would have to adjust all the schema and entity generator snapshots. In general this sounds like something that should not require additional entities. 

But I guess I can take over and cherry pick what I want from this myself :] Thanks, appreciated!

ballwood:
@B4nan no problem - apologies r.e. the new entities. I didn't really know what to modify so I just set up the most minimal example I knew that caused issues to try and trigger some errors

B4nan:
No probs, as I said, this is enough, I will pick the actual tests and use existing entities when working on it.

PR #1087, opened 2020-11-17
fix(core): assign embedded properties from class objects

### LINKED ISSUE #1083: Wrap Assign does not work if object has prototype.
**Describe the bug**
Wrap Assign does not work if object has a prototype.
Object Assign works perfectly, i think that this is a bug and it should perform the same as Object Assign.

**To Reproduce**
```
@Embeddable()
export class PersonName {
  @Property()
  first: string;

  @Property({ nullable: true })
  last: string;
}

@Entity()
export class Person {
  @PrimaryKey()
  id: number;

  @Embedded({ prefix: "name_" })
  name: PersonName;
}
```
```
const person = await ctx.em.findOne(Person, { id: 1 });

const data: any = {};
data.name = new PersonName();
data.name.first = "Test";
ctx.em.assign(person, data);
await ctx.em.persistAndFlush(person);
```
It performs:
`update "person" set "name_first" = NULL where "id" = 1`

**Versions**

| Dependency | Version |
| - | - |
| node | v12.16.3 |
| typescript | v4.0.5 |
| mikro-orm | v4.3.0 |

dvlalex:
This fixes entity assign not merging embedded properties when the source has a prototype.

Closes #1083

### DIFF
--- a/packages/core/src/entity/EntityAssigner.ts
+++ b/packages/core/src/entity/EntityAssigner.ts
@@ -49,7 +49,7 @@ export class EntityAssigner {
       if (props[prop]?.reference === ReferenceType.EMBEDDED) {
         const Embeddable = props[prop].embeddable;
         entity[props[prop].name] = Object.create(Embeddable.prototype);
-        Utils.merge(entity[prop as keyof T], value);
+        Utils.merge(entity[prop as keyof T], Utils.isPlainObject(value) ? value : { ...value });
         return;
       }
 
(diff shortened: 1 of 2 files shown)


dvlalex:
The issue here is that Utils.merge only accepts POJO sources, and since #1087 only affects embeddables, I just destructured it when otherwise.

I noticed the util is used in multiple places, so I didn't want to change it and add some weird behavior that might not be accounted for.

B4nan:
Thanks!

PR #1135, opened 2020-11-27
Count work when no result is found

ramiel:
This should fix when grouping and no result is found

### DIFF
--- a/packages/knex/src/AbstractSqlDriver.ts
+++ b/packages/knex/src/AbstractSqlDriver.ts
@@ -174,7 +174,7 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
       .where(where);
     const res = await this.rethrow(qb.execute('get', false));
 
-    return +res.count;
+    return +res?.count || 0;
   }
 
   async nativeInsert<T extends AnyEntity<T>>(entityName: string, data: EntityData<T>, ctx?: Transaction<KnexTransaction>, convertCustomTypes = true): Promise<QueryResult> {


B4nan:
To maintain the full branch coverage, this will either need a test case (that would otherwise fail without that change) or at least ignore hint for istanbul (as using optional chaining usually reports missed branches due to how it is transpiled).

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -174,7 +174,7 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
       .where(where);
     const res = await this.rethrow(qb.execute('get', false));
 
-    return +res.count;
+    return +res?.count || 0;
```
B4nan, line 177: Is it even possible that `res` will not be defined? I don't think so...

ramiel: this is what happened to me. Sorry I should have filed a bug actually

ramiel: I'm running this

```
await this.myRepository.count(
        { externalId: externalId },
        { groupBy: 'user' },
      );
```

this fails if there's no entry in myRepository that satisfies the condition. The error says `cannot read count of undefined`

B4nan: ok then add the test case so we have that covered, to one of the `EntityManager` tests, `em.count` should be covered in all of them, so just extend the test that is already there

B4nan: we might still need that istanbul ignore hint, as covering lines with optional chaining is buggy

B4nan: do you plan to finish this or should I take over?

ramiel: Oh sorry, I had no time to look at this. If you can take over would be great

B4nan: Cant replicate this. More importantly, using `count` with `groupBy` will pretty much never give you what you want. Grouping will result in multiple rows, but `em.count()` will give you only the first one. It already uses `count distinct` query, what do you need the group by for?

PR #1136, opened 2020-11-27
docs: extra flags on Terser/Webpack configuration

hilkeheremans:
I've found that Terser, in some cases, also changes function and class names during its compression phase, even though mangling is disabled. This can cause trouble in some relations and table naming.

This is apparently also documented over at https://terser.org/docs/api-reference.html#mangle-options.

This PR is a minuscule addition to the docs that provides the proper options to Terser to avoid any issues occurring in production.

### DIFF
--- a/docs/docs/deployment.md
+++ b/docs/docs/deployment.md
@@ -197,6 +197,14 @@ module.exports = {
           // https://terser.org/docs/api-reference.html#mangle-options) but the safest default
           // config is that we simply disable mangling altogether but allow minification to proceed.
           mangle: false,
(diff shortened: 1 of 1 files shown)


B4nan:
Can you distribute that change to other v4 docs versions? (at least the current stable one, which is 4.3)

hilkeheremans:
Gladly, though I'm not sure I follow. What do you want me to do?

Sorry, feeling dumb for a second 🤷‍♂️

B4nan:
You changed only the next docs, here are the stable docs: 

https://github.com/mikro-orm/mikro-orm/tree/master/docs/versioned_docs

PR #1155, opened 2020-12-03
feat(config): add support for Typescript config files when running via ts-jest

### LINKED ISSUE #1151: Support running tests via ts-jest
Currently, MikroORM only supports Typescript config files if the process is being run through `ts-node`

Typescript config files should be able to be used when running tests via `ts-jest`

Our current alternative is to transpile to project to Javascript, then copy across our config file (which is now Javascript) to the source directory

**Additional context**
I believe it should be possible to determine if ts-jest is being used by (a) checking if the current process is being run by Jest (`process.argv[1].includes('jest-worker')`), and then checking if "ts-jest" is present as a transformer in any Jest config (withing `package.json` or within `jest.config.json`). Just like with `ts-node`, there should also be an option in the MikroORM config within `package.json` which is required to be enabled by users before this check happens, to avoid messing up any existing configuration that users have

I'm happy to work on a PR, but let me know if anyone knows of a better way of achieving this functionality

Ezard:
Closes #1151

### DIFF
--- a/packages/core/src/utils/Utils.ts
+++ b/packages/core/src/utils/Utils.ts
@@ -1,3 +1,4 @@
+import { ConfigurationLoader } from '@mikro-orm/core';
 import { createRequire, createRequireFromPath } from 'module';
 import clone from 'clone';
 import globby, { GlobbyOptions } from 'globby';
@@ -512,6 +513,35 @@ export class Utils {
       || process.argv.slice(1).some(arg => arg.includes('ts-node')) // registering ts-node runner
       || (require.extensions && !!require.extensions['.ts']) // check if the extension is registered
       || !!new Error().stack!.split('\n').find(line => line.match(/\w\.ts:\d/)); // as a last resort, try to find a TS file in the stack trace
+  }
+
+  private static async getJestConfig(): Promise<Dictionary> {
+    if (await pathExists(process.cwd() + '/jest.config.json')) {
+      return require(process.cwd() + '/jest.config.json');
+    }
+    return {};
+  }
+
+  private static jestConfigUsesTsJest(jestConfig: Dictionary): boolean {
+    return jestConfig.preset === 'ts-jest'
+      || (Object.values(jestConfig.transform ?? {}).some(transformer => transformer === 'ts-jest'));
+  }
+
+  /**
+   * Tries to detect `ts-jest` runtime.
+   */
+  static async detectTsJest(): Promise<boolean> {
+    if (process.argv.length > 1 && process.argv[1].includes('jest-worker')) {
+      const packageConfig = await ConfigurationLoader.getPackageConfig();
+      const jestConfig = await Utils.getJestConfig();
+      if (
+        (packageConfig.jest && Utils.jestConfigUsesTsJest(packageConfig.jest))
+        || Utils.jestConfigUsesTsJest(jestConfig)
+      ) {
+        return true;
+      }
+    }
+    return false;
   }
 
   /**
--- a/packages/core/src/utils/Utils.ts
+++ b/packages/core/src/utils/Utils.ts
@@ -1,4 +1,4 @@
-import { ConfigurationLoader } from '@mikro-orm/core';
+import { ConfigurationLoader } from './ConfigurationLoader';
 import { createRequire, createRequireFromPath } from 'module';
 import clone from 'clone';
 import globby, { GlobbyOptions } from 'globby';
--- a/packages/core/src/utils/Utils.ts
+++ b/packages/core/src/utils/Utils.ts
@@ -522,26 +522,21 @@ export class Utils {
     return {};
   }
 
-  private static jestConfigUsesTsJest(jestConfig: Dictionary): boolean {
-    return jestConfig.preset === 'ts-jest'
-      || (Object.values(jestConfig.transform ?? {}).some(transformer => transformer === 'ts-jest'));
+  private static jestConfigUsesTsJest(jestConfig?: Dictionary): boolean {
+    const usesJestTransformer = Object.values(jestConfig?.transform ?? {}).some(transformer => transformer === 'ts-jest');
+    return jestConfig?.preset === 'ts-jest' || usesJestTransformer;
   }
 
   /**
    * Tries to detect `ts-jest` runtime.
    */
   static async detectTsJest(): Promise<boolean> {
-    if (process.argv.length > 1 && process.argv[1].includes('jest-worker')) {
-      const packageConfig = await ConfigurationLoader.getPackageConfig();
-      const jestConfig = await Utils.getJestConfig();
-      if (
-        (packageConfig.jest && Utils.jestConfigUsesTsJest(packageConfig.jest))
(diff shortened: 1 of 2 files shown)


Ezard:
Just need to figure out how to test this properly

packages/core/src/utils/Utils.ts
```diff
@@ -514,6 +515,35 @@ export class Utils {
...
+      const packageConfig = await ConfigurationLoader.getPackageConfig();
+      const jestConfig = await Utils.getJestConfig();
+      if (
+        (packageConfig.jest && Utils.jestConfigUsesTsJest(packageConfig.jest))
+        || Utils.jestConfigUsesTsJest(jestConfig)
+      ) {
+        return true;
+      }
+    }
+    return false;
```
B4nan, line 544: this could be re-written to reduce complexity:

```suggestion
    if (process.argv.length <= 1 || !process.argv[1].includes('jest-worker')) {
        return false;
    }

    const packageConfig = await ConfigurationLoader.getPackageConfig();
    const jestConfig = await Utils.getJestConfig();

    return (packageConfig.jest && Utils.jestConfigUsesTsJest(packageConfig.jest)) || Utils.jestConfigUsesTsJest(jestConfig);
```

even better `(packageConfig.jest && Utils.jestConfigUsesTsJest(packageConfig.jest))` - the first condition should be handled in `jestConfigUsesTsJest` method

packages/core/src/utils/Utils.ts
```diff
@@ -514,6 +515,35 @@ export class Utils {
...
+  private static async getJestConfig(): Promise<Dictionary> {
+    if (await pathExists(process.cwd() + '/jest.config.json')) {
+      return require(process.cwd() + '/jest.config.json');
+    }
+    return {};
+  }
+
+  private static jestConfigUsesTsJest(jestConfig: Dictionary): boolean {
+    return jestConfig.preset === 'ts-jest'
+      || (Object.values(jestConfig.transform ?? {}).some(transformer => transformer === 'ts-jest'));
```
B4nan, line 527: not a big fan of splitting stuff to multiple lines randomly (prettier style, meh). much better to create new local variable (human style, lol):

```suggestion
    const usesJestTransformer = Object.values(jestConfig.transform ?? {}).some(transformer => transformer === 'ts-jest');
    return jestConfig.preset === 'ts-jest' || usesJestTransformer;
```

B4nan (changes requested):
One missing thing is the jest config inside `package.json` (the way used here in the repository too).

packages/core/src/utils/Utils.ts
```diff
@@ -1,3 +1,4 @@
+import { ConfigurationLoader } from '@mikro-orm/core';
```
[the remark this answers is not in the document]
Ezard: @B4nan ah thanks - didn't notice the IDE had done that

Ezard:
Thanks for the fast responses - I'll update the PR this weekend

Ezard:
> One missing thing is the jest config inside `package.json` (the way used here in the repository too).

This was already handled :)

I've implemented the rest of your suggestions though, so I just need to sort out some tests now

B4nan:
Do you plan to finish this? Should be enough to cover this via unit tests.

Ezard:
@B4nan sorry for dragging this out - I do intend to finish it, I've just been super busy
I'm rotating back onto a MikroORM-based project within the next week, so I'll finish it then

Ezard:
@B4nan I'm attempting to sort out tests; however, the fact that the code is written in Typescript is making it kinda hard 🤦‍♂️ 

For example, I wrote a test to ensure that `ConfigurationLoader.getConfigPaths()` returns the correct values when running via `ts-jest`, but since the tests are running via `ts-node`, that gets picked up instead and makes my test pointless

I traced how you're detecting the presence of `ts-node`, and for me at least it's picked up by `Utils.detectTsNode()` here:
```
  static detectTsNode(): boolean {
    return process.argv[0].endsWith('ts-node')
      || !!process[Symbol.for('ts-node.register.instance')]
      || process.argv.slice(1).some(arg => arg.includes('ts-node'))
      || (require.extensions && !!require.extensions['.ts'])
      || !!new Error().stack!.split('\n').find(line => line.match(/\w\.ts:\d/));  <------
  }
```

Should I just mock the `Error` object so I can manipulate its `stack` field? It seems a bit of a thing to do, so if you have any better suggestions then I'm all ears :)

packages/core/src/utils/Utils.ts
```diff
@@ -514,6 +515,30 @@ export class Utils {
       || !!new Error().stack!.split('\n').find(line => line.match(/\w\.ts:\d/)); // as a last resort, try to find a TS file in the stack trace
   }
 
+  private static async getJestConfig(): Promise<Dictionary> {
```
B4nan, line 518: it feels more and more weird to even have a method that refers to `jest` (and now we even have 3 of them)... 

this all seems overcomplicated and the bit with jest config even seems wrong to me - why would existance of a config file (or anything inside it) mean we are running via ts-jest? 

there should be just one synchoronous method if something, and it feels like we should just improve the `detectTsNode` method instead of adding new ones - we don't really need to know if the user is running via ts-node or ts-jest, we care about the TS support in general.

B4nan (changes requested):

packages/core/src/utils/Utils.ts
Ezard: @B4nan the reason that the config needs to be read is because, without reading them, we're only able to determine if the user is running their code via Jest, and we're not able to tell if that code is JS or TS

Therefore, once we've determined that the user is running their code through Jest, we check the Jest config to see if that config supports TS files

If there's something that I've missed and there's a better way of detecting it, then I'd be more than happy to re-implement it

B4nan: My point is that this solution is fragile and makes the ORM dependent on jest somehow (which is indeed wrong). Furthermore you are dealing with some ways jest can be configured, but not all of them (e.g. it can be basically any JSON file as long as you point to it when running jest).

TBH I am yet to understand why do you need this change. I am jest user myself, still idk :] Could you create some repository that demonstrates the need?

Again the whole point of this is to decide whether the app is running with TS support or not. It is not about ts-node, babel, ts-jest, whatever - we don't care how it is done, we care about being able to require TS file in current context. 

Also the `useTsJest` flag in package.json does not make sense to me - those flags from package.json are only for CLI, ts-jest has nothing to do with CLI.

Ezard: @B4nan it was to allow for MikroORM config files to be written in Typescript

Like you said though, people have been doing just fine without support for this, so maybe I'm just an edge-case haha :)

I'll close the PR

B4nan: I'd still like too see a repro for this, I am not against dealing with that, but it should be probably an extension to `detectTsNode` method (another heuristic), we might also rename the method to something like `detectTypeScriptSupport` so its more descriptive

PR #1177, opened 2020-12-07
fix(knex): reject in `commit()` method if commit statement fails

### LINKED ISSUE #1176: em.commit doesn't throw when a sql transaction commit fails
Extracting this small issue out of #1160 and providing fix via PR.

Basically, when using explicit transactions and explicit `begin`, `commit`, and `rollback` methods, if the `commit` statement fails on the database side, this is not currently being caught by mikroorm and `em.commit` resolves incorrectly. It should rethrow the error given by the knex transaction context.

rhyek:
Closes #1176


B4nan:
Thanks, interesting that this is intended behaviour in knex, lol...

B4nan:
Available in `4.3.4-dev.8` for testing

PR #1181, opened 2020-12-08
feat[core]: partially update embedded object properties

dvlalex:
I've added in some logic for when an embedded property is set as { object: true }.

Currently an old query that was:
``[query] db.getCollection('user').updateMany({ _id: ObjectId('5fcf95d54f6a9653930f3c8b') }, { '$set': { address1_city: 'new city', address4: { street: null, postalCode: null, city: 'new city 4', country: null } } }, { session: undefined }); [took 2 ms]``

now, looks like:
``[query] db.getCollection('user').updateMany({ _id: ObjectId('5fcf96cd2c05055470c9edff') }, { '$set': { address1_city: 'new city', address4: { city: 'new city 4' } } }, { session: undefined }); [took 2 ms]``

So the changes are computed properly for object types, but I'm not sure where to go from now. I'd have to tap into each specific driver and make sure that for embedded objects on update it doesnt replace the value.

Any pointers / ideas?

### DIFF
--- a/packages/core/src/utils/EntityComparator.ts
+++ b/packages/core/src/utils/EntityComparator.ts
@@ -309,6 +309,7 @@ export class EntityComparator {
 
   private getPropertyComparator<T>(prop: EntityProperty<T>): string {
     let type = prop.type.toLowerCase();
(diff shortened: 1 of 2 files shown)


B4nan:
Closing as stale

PR #1213, opened 2020-12-15
feat(core): implement transaction lifecycle hooks

### LINKED ISSUE #1175: [Feature]: Implement afterTransactionCompletion, afterCommitCreate, et al similar to Hibernate counterparts
Our project is making use of explicit transactions with explicit use of `begin`, `commit`, and `rollback` methods. We need to register hook handlers for `afterFlush`, `afterCreate`, etc, to run some business logic that will notify users and/or integrated systems of such events.
Our problem as of now is that these are run after UoW commit phase and not transaction commit phase so in the case where a `commit` fails on the database side (deferred constraint checks, concurrency problems, etc), it is important to have a way from those hook handlers to be aware of this and therefore not execute their code *OR* have separate hooks that are guaranteed to only run on successful transactions. Currently those hooks are executed regardless and this presents a serious problem for our use case.

Implement hooks similar to [afterTransactionCompleted](https://docs.jboss.org/hibernate/orm/5.2/javadocs/org/hibernate/EmptyInterceptor.html#afterTransactionCompletion-org.hibernate.Transaction-) and events such as [POST_COMMIT_INSERT](https://docs.jboss.org/hibernate/orm/5.2/javadocs/org/hibernate/event/spi/EventType.html#POST_COMMIT_INSERT), et al, which are part of the Hibernate ORM (I understand this project is inspired in part by it) so that we can make use of those hooks instead of the current ones. The behavior of the current ones would not be altered.

Our team has concluded that the only way for us to guarantee our business logic is run on successful transactions only is to make use of the CDC (Change Data Capture) functionalities of our preferred database in conjunction with perhaps something like Kafka Connect by way of a Debezium Kafka Source Connector and then additionally a Kafka Consumer that would execute our aforementioned business logic. This seems overkill to us.

**Additional context**
I had previously opened an issue regarding our findings: #1160. I am willing to write a PR implementing this feature, but would first like to hear from @B4nan if that is something that would be taken into consideration or not.

rhyek:
Current lifecycle hooks cover entity and unit of work events, but
transaction-related ones are missing. Add beforeTransactionStart,
afterTransactionStart, beforeTransactionCommit, afterTransactionCommit,
beforeTransactionRollback, and afterTransactionRollback

Closes #1175

---
Although #1175 mentions implementing "after" entity hooks post commit, they are not included in this PR as they are not needed. Refer to the `UserSubscriber` example class in test file `GH1175.test.ts` to see how to implement an ad-hoc `afterCommitCreate` hook using a combination of `afterTransactionStart`, `afterCreate`, and `afterTransactionCommit`.

Currently, all this PR is missing right now is documentation. Please let me know if you need me to do that as well, but I was thinking maybe @B4nan could lend a hand with that as he might have a better idea regarding form, style, and content.

### DIFF
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -8,7 +8,7 @@ import { AnyEntity, Dictionary, EntityData, EntityMetadata, EntityName, FilterDe
 import { LoadStrategy, LockMode, QueryOrderMap, ReferenceType, SCALAR_TYPES } from './enums';
 import { MetadataStorage } from './metadata';
 import { Transaction } from './connections';
-import { EventManager } from './events';
+import { EventManager, TransactionEventBroadcaster } from './events';
 import { EntityComparator } from './utils/EntityComparator';
 import { OptimisticLockError, ValidationError } from './errors';
 
@@ -335,7 +335,7 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
         await em.flush();
 
         return ret;
-      }, ctx);
+      }, ctx, new TransactionEventBroadcaster(em));
     });
   }
 
@@ -343,7 +343,7 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
    * Starts new transaction bound to this EntityManager. Use `ctx` parameter to provide the parent when nesting transactions.
    */
   async begin(ctx?: Transaction): Promise<void> {
-    this.transactionContext = await this.getConnection('write').begin(ctx);
+    this.transactionContext = await this.getConnection('write').begin(ctx, new TransactionEventBroadcaster(this));
   }
 
   /**
@@ -351,7 +351,7 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
    */
   async commit(): Promise<void> {
     await this.flush();
-    await this.getConnection('write').commit(this.transactionContext);
+    await this.getConnection('write').commit(this.transactionContext, new TransactionEventBroadcaster(this));
     delete this.transactionContext;
   }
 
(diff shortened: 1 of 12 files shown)


rhyek:
Btw, these hooks work correctly with nested transactions. Testing is very exhaustive and covers many cases including those.

B4nan:
Looking good, thanks! Will have few code style hints, might change it myself as I want to fiddle around with this locally anyway to understand it better.

Also few lines in `MongoConnection` are not covered (probably just branches caused by the optional chaining - if so, let's just force ignore those via istanbul ignore comments).

That `TransactionEventBroadcaster` thing looks a bit unneeded, but will need to dig deeper to better understand the code, might be missig something now.

One thing that looks stinky is that you are using `isRootTransaction` only in the SQL driver, not in mongo one.

rhyek:
@B4nan glad to help.

> Also few lines in MongoConnection are not covered (probably just branches caused by the optional chaining - if so, let's just force ignore those via istanbul ignore comments).

I will check this today.

> That TransactionEventBroadcaster thing looks a bit unneeded, but will need to dig deeper to better understand the code, might be missig something now.

The idea here is that what the broadcaster is is agnostic to the connection and we could provide any extra context we want from the call site in the future. For example, we could choose to provide the unit of work when doing implicit transactions. I actually had planned to define the EventBroadcaster interface in the Connection package. I will do that today.

> One thing that looks stinky is that you are using isRootTransaction only in the SQL driver, not in mongo one.

Currently mongo doesn't allow for child transactions (save points), but they might in the future and in that case checking to see if it is a root transaction or not sounds like a generic concern. Let me know your thoughts. I can easily move that to AbstractSqlConnection.

rhyek:
Pushed some of the changes mentioned, added tests for mongo, and fixed coverage errors.

rhyek:
Thank you, @B4nan.

mcasarrubios:
Hi @rhyek, I was struggling with afterTransactionCommit hook because I didn't understand why `uow` param was undefined. 
Watching the `GH1175.test.ts` file I realize that it only is defined in implicit transactions. 

Is there a reason why `uow`is undefined in explicit transactions? I would love that it was defined as in afterFlush event. The thing is that I don't know if it is a constraint of mikro-orm or if it could be done.

I've created the question in SO too: https://stackoverflow.com/questions/76851480/in-mikro-orm-is-it-possible-to-track-unit-of-work-changes-in-the-aftertransacti

Thanks in advance.

PR #1225, opened 2020-12-21
fix(core): support FK as PK in `Collection.getIdentifiers()`

### LINKED ISSUE #1224: PK as FK with References & getIdentifiers
**Describe the bug**
The collection method "getIdentifiers" returns an array of references instead of the expecting ids:
~~~ts
      [ Reference { entity: Ref<Node> { id: 1 } } ]
~~~
This derivates from the type declaration.

**To Reproduce**
#1224

**Expected behavior**
~~~ts
      [ 1 ]
~~~
 instead of
~~~ts
      [ Reference { entity: Ref<Node> { id: 1 } } ]
~~~

**Versions**
| Dependency | Version |
| node | 15.4.0 |
| typescript | 4.1.2 |
| mikro-orm | 4.3.4  |
| postgresql |  8.5.1 |

ml1nk:
Closes #1224

### DIFF
--- a/packages/core/src/entity/ArrayCollection.ts
+++ b/packages/core/src/entity/ArrayCollection.ts
@@ -2,6 +2,7 @@ import { AnyEntity, Dictionary, EntityProperty, IPrimaryKey, Primary } from '../
 import { Reference } from './Reference';
 import { wrap } from './wrap';
 import { ReferenceType } from '../enums';
+import { Utils } from '../utils/Utils';
 
 export class ArrayCollection<T, O> {
 
@@ -60,8 +61,8 @@ export class ArrayCollection<T, O> {
 
     return this.getItems().map(i => {
       let e = i[field as keyof T] as unknown as U;
-      while (e instanceof Reference) {
-        e = e.unwrap()[this.property.targetMeta!.properties[field as string].targetMeta!.serializedPrimaryKey];
+      while (Utils.isEntity(e, true)) {
+        e = Reference.unwrapReference(e as AnyEntity)[this.property.targetMeta!.properties[field as string].targetMeta!.serializedPrimaryKey];
       }
       return e;
     });
--- a/packages/core/src/entity/ArrayCollection.ts
+++ b/packages/core/src/entity/ArrayCollection.ts
@@ -58,7 +58,13 @@ export class ArrayCollection<T, O> {
 
     field = field ?? this.property.targetMeta!.serializedPrimaryKey;
 
-    return this.getItems().map(i => i[field as keyof T]) as unknown as U[];
+    return items.map(i => {
+      let e = i[field as keyof T] as unknown as U;
+      while (e instanceof Reference) {
+        e = e.unwrap()[this.property.targetMeta!.properties[field as string].targetMeta!.serializedPrimaryKey];
+      }
+      return e;
+    });
   }
 
   add(...items: (T | Reference<T>)[]): void {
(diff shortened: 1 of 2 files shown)


packages/core/src/entity/ArrayCollection.ts
```diff
@@ -58,7 +58,13 @@ export class ArrayCollection<T, O> {
 
     field = field ?? this.property.targetMeta!.serializedPrimaryKey;
 
-    return this.getItems().map(i => i[field as keyof T]) as unknown as U[];
```
ml1nk, line 61: Replaced this.getItems() with items as it seems to be faster.

packages/core/src/entity/ArrayCollection.ts
```diff
@@ -58,7 +58,13 @@ export class ArrayCollection<T, O> {
...
+    return items.map(i => {
+      let e = i[field as keyof T] as unknown as U;
+      while (e instanceof Reference) {
```
ml1nk, line 63: I have never tested nested PK as FK with References, but think it is possible.

packages/core/src/entity/ArrayCollection.ts
B4nan: please revert that, first of all its unrelated to this issue, second this would disable check in `Collection.getItems()`.

ml1nk: It's reverted. I understand the first point, but not the second. What's the difference of calling getItems() once (few lines above this one) or twice in the same method/tick? The check in Collection is called anyways.

B4nan (changes requested):
what is wrong with using `wrap(e).getPrimaryKey()`? it should handle the recursion automatically, as well as the reference wrapper

packages/core/src/entity/ArrayCollection.ts
ml1nk: I make a separate refactor PR for this, thanks.

packages/core/src/entity/ArrayCollection.ts
ml1nk: Refactored

packages/core/src/entity/ArrayCollection.ts
```diff
@@ -58,7 +58,13 @@ export class ArrayCollection<T, O> {
 
     field = field ?? this.property.targetMeta!.serializedPrimaryKey;
 
-    return this.getItems().map(i => i[field as keyof T]) as unknown as U[];
+    return this.getItems().map(i => {
+      let e = i[field as keyof T] as unknown as U;
+      while (e instanceof Reference) {
+        e = e.unwrap()[this.property.targetMeta!.properties[field as string].targetMeta!.serializedPrimaryKey];
```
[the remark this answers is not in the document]
ml1nk: > wrap(e).getPrimaryKey()

I couldn't access getPrimaryKey() like this and i assume that, like in my first try, it returns the raw object even if it should be serialized (like a test expects).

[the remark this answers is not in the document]
B4nan: This way it works fine for me (and your test case is passing):

```ts
  getIdentifiers<U extends IPrimaryKey = Primary<T> & IPrimaryKey>(field?: string): U[] {
    const items = this.getItems();

    if (items.length === 0) {
      return [];
    }

    field = field ?? this.property.targetMeta!.serializedPrimaryKey;

    return items.map(i => {
      if (Utils.isEntity(i[field as keyof T], true)) {
        return wrap(i[field as keyof T], true).getPrimaryKey();
      }

      return i[field as keyof T];
    }) as unknown as U[];
  }
```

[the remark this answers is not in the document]
ml1nk: I was missing wrap(...,true). Thanks for your help, i updated my pr appropriately.

B4nan:
Thanks!

PR #1257, opened 2021-01-03
feat(sql): modify connection options to support 'dateStrings' parameter

demsey2:


B4nan:
what is the motivation? you should prefer custom types to have control over how dates are serialized

demsey2:
so I work on a legacy application where each tenant has their own schema and all dates are stored according to their timezone since mikro-orm doesn't support multi connections I want to fetch all dates as strings and then manage dates with timezones manually 

`driverOptions` doesn't support `dateStrings` option, `dateStrings` is hardcoded to `['Date']` but according to mysql2 or knex driver we should be able to set that as a boolean

B4nan:
> driverOptions doesn't support dateStrings option, dateStrings is hardcoded to ['Date'] but according to mysql2 or knex driver we should be able to set that as a boolean

It does support everything - it is applied as the very last thing, overriding everything that the ORM does internally. 

https://github.com/mikro-orm/mikro-orm/blob/master/packages/knex/src/AbstractSqlConnection.ts#L150

B4nan:
I will close this, I really don't see a need for new top level ORM configuration, when:
- it is possible to do it other way, via `{ driverOptions: { connection: { dateStrings: true } } }`
- when it is not really a global toggle, in fact it affects only two drivers

If the `driverOptions` approach don't work for you, open issue with reproduction, but it should. The idea behind it is to be able to have the very last word on the form and shape of the knex/mongo config.

It is important to understad what `driverOptions` are, because if you were trying `{ driverOptions: { dateStrings: true } }`, that is indeed not going to work. And not because something is hardcoded, but because you are not overriding the value on the right place. It needs to be `{ driverOptions: { connection: { dateStrings: true } } }`, so under `connection` object. `driverOptions` is for overriding the whole knex configuration object, and `dateStrings` are not living on the top level.

demsey2:
ok that is fine, thanks for your help

Issue #1294, opened 2021-01-11
Parts of data model passed to `.create` calls is destroyed


oliversalzburg:
**Describe the bug**
When calling `em.create( dataModel )`, some or all elements will be missing from `dataModel` after the call.

**To Reproduce**
I'm sending a PR with a failing test in a moment.

**Expected behavior**
The passed data model should be treated as read-only. Having it manipulated by this call is unexpected and can cause a myriad of issues down the line.

**Versions**

| Dependency | Version |
| - | - |
| node | 14 |
| typescript | 4.1.3 |
| mikro-orm | HEAD |
| your-driver | postgresql |

PR #1295, opened 2021-01-11
test: input data model must remain untouched

oliversalzburg:
The root cause seems to be https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/entity/EntityFactory.ts#L91

Keys, that are also parameters of the constructor, will be deleted from the input object.

Relates to #1294


oliversalzburg:
Awesome. Thanks for the fix 😀

PR #1300, opened 2021-01-13
test: @Entity() should provide BaseEntity implementation

oliversalzburg:
It is my understanding that using `@Entity()` should be equivalent to extending `BaseEntity` (https://mikro-orm.io/docs/defining-entities).

This would imply to me, that my entities decorated with `@Entity()` also have methods like `.assign()`, but they do not. They _do_ have a `.toJSON` added, but not _all_ methods I would expect.

I believe a second stage of tests/assertions is required to see if these methods, once they exist, actually work as expected.


oliversalzburg:
This is to be expected. The decoration and extending `BaseEntity` is not identical in behavior.

PR #1306, opened 2021-01-15
feat(core): add support for nested partial loading

### LINKED ISSUE #221: The fields property doesn't work with nested fields
**Describe the bug**
For example wanting to retrieve the field `ent1.prop1` with the FindOne option creates a wrong query (it has support for nested fields as it does not turn ent1.prop1 into e0.ent1.prop1 which happens when you dont specify anything before the dot). 

It will create this query which does not work. 
```select `e0`.`id`, `ent1`.`prop1` from `ent0` as `e0` where `e0`.`id` = 9 and `e0`.`prop2` = 73 limit 1```

Is it possible to have that option work the same way as the FilterQuery does?

### PR DESCRIPTION
**Pull request by B4nan**
Adds support for nested partial loading:

```ts
const author = await orm.em.findOne(Author, '...', { fields: ['name', 'books.title', 'books.author', 'books.price'] });
```

Or with alternative object syntax:

```ts
const author = await orm.em.findOne(Author, '...', { fields: ['name', { books: ['title', 'author', 'price'] }] });
```

It is also possible to use multiple levels:

```ts
const author = await orm.em.findOne(Author, '...', { fields: ['name', { books: ['title', 'price', 'author', { author: ['email'] }] }] });
```

Primary keys are always selected even if you omit them. On the other hand, you are responsible 
for selecting the FKs - if you omit such property, the relation might not be loaded properly.
In the following example the books would not be linked the author, because we did not specify 
the `books.author` field to be loaded.

```ts
// this will load both author and book entities, but they won't be connected due to the missing FK in select
const author = await orm.em.findOne(Author, '...', { fields: ['name', { books: ['title', 'price'] });
```

> Same problem can occur in mongo with M:N collections - those are stored as array property 
> on the owning entity, so you need to make sure to mark such properties too.

Closes #221

### DIFF
--- a/docs/docs/entity-manager.md
+++ b/docs/docs/entity-manager.md
@@ -186,14 +186,47 @@ console.log(author.books[0].tags[0].isInitialized()); // true, because it was po
 
 ### Fetching Partial Entities
 
+> This feature is supported only for `SELECT_IN` loading strategy.
+
 When fetching single entity, you can choose to select only parts of an entity via `options.fields`:
 
-```typescript
+```ts
 const author = await orm.em.findOne(Author, '...', { fields: ['name', 'born'] });
 console.log(author.id); // PK is always selected
 console.log(author.name); // Jon Snow
 console.log(author.email); // undefined
 ```
+
+From v4.4 it is also possible to specify fields for nested relations:
+
+```ts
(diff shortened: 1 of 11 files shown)


astahmer:
That's a great addition ! Thank you for this.

PR #1307, opened 2021-01-15
feat(core): allow disabling identity map and change set tracking

### LINKED ISSUE #1267: add feature for disable tracking on query result entities for better performance similar to .AsNoTracking in entity framework
hi I think this is a good feature for improve performance specially for heavy queries
add feature for disabling tracking on query result entities for better performance like .AsNoTracking in entity framework

B4nan:
Sometimes we might want to disable identity map and change set tracking for some query.
This is possible via `disableIdentityMap` option. Behind the scenes, it will create new
context, load the entities inside that, and clear it afterwards, so the main identity map
will stay clean.

```ts
const users = await orm.em.find(User, { email: 'foo@bar.baz' }, {
  disableIdentityMap: true,
  populate: { cars: { brand: true } },
});
users[0].name = 'changed';
await orm.em.flush(); // calling flush have no effect, as the entity is not managed
```

> Keep in mind that this can also have
> [negative effect on the performance](https://stackoverflow.com/questions/9259480/entity-framework-mergeoption-notracking-bad-performance).

Closes #1267

### DIFF
--- a/docs/docs/entity-manager.md
+++ b/docs/docs/entity-manager.md
@@ -291,6 +291,28 @@ where lower(email) = 'foo@bar.baz'
 order by (point(loc_latitude, loclongitude) <@> point(0, 0)) asc
 ```
 
+## Disabling identity map and change set tracking
+
+Sometimes we might want to disable identity map and change set tracking for some query.
+This is possible via `disableIdentityMap` option. Behind the scenes, it will create new 
+context, load the entities inside that, and clear it afterwards, so the main identity map
+will stay clean.
+
+> As opposed to _managed_ entities, such entities are called _detached_. 
+> To be able to work with them, you first need to merge them via `em.registerManaged()`. 
+
+```ts
+const users = await orm.em.find(User, { email: 'foo@bar.baz' }, {
+  disableIdentityMap: true,
+  populate: { cars: { brand: true } },
+});
+users[0].name = 'changed';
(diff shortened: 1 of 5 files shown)


dbettini:
@B4nan hello, I just found this issue while searching for a way to populate only specific relations on a model, without the possibility of loading more data as a consequence of using identity map.
What is the reason for using `this.fork(false)` instead of `this.fork(true)`? Looking at the [docs](https://mikro-orm.io/docs/entity-manager-api/#forkclear-boolean-usecontext-boolean-dtypeof-entitymanagertype), it seems like we don't get a clear identity map this way, but clone the existing one. In other words, the current implementation uses the existing identity map, it just doesn't update it, which will result in sometimes getting more data than required. I.e.:
```
const one = repo.findOne(id, ['firstRelation'])
const two = repo.findOne(id, { populate: ['secondRelation], disableIdentityMap: true });
```
In the above example, `two` will also have `firstRelation` loaded, which would not be expected by the naming of `disableIdentityMap`

B4nan:
I though it was required for the population to work, but looks like the tests are passing without clearning too. I am fine with the change in v5, for v4 it seems semi breaking.

dbettini:
Cool, I'll stick to manual forking until v5 then. Thanks for the quick reply!

mittche:
why is not global.
can't use with create this.userRepository.create(createUserDto, { disableIdentityMap: true })

PR #1320, opened 2021-01-19
Added explicit types in class based entity schema

ReactReduxRouterDOM:
Reference https://stackoverflow.com/a/65787951/9795654.
This question was asked by me, where the author also described that types should be added explicitly unless we use ts-morph, which is not added in the overview section, and beginners may not know about this limitation until they visit: https://mikro-orm.io/docs/metadata-providers/#limitations-and-requirements

### DIFF
--- a/docs/versioned_docs/version-4.3/defining-entities.md
+++ b/docs/versioned_docs/version-4.3/defining-entities.md
@@ -58,10 +58,10 @@ export class Book {
   id!: number;
 
   @Property()
-  createdAt = new Date();
+  createdAt: Date = new Date();
 
   @Property({ onUpdate: () => new Date() })
-  updatedAt = new Date();
+  updatedAt: Date = new Date();
 
   @Property()
   title!: string;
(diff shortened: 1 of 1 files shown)


B4nan:
That person was me, lol :] Updating one place won't help much, especially when you change only one historic version of the docs. The entity definitions in the docs are still valid but they respect the ts-morph provider. 

I would much rather add a link to the limitations section than adjust everything. The end goal would be to ideally use code tabs in all the places, and show how the valid definition looks when using one of the possible ways to define entity (reflect-metadata, ts-morph, EntitySchema). But that feels like a loooot of work :]

ReactReduxRouterDOM:
I didn't recognize you are the maintainer of this project :D. This seems a lot of work though. I might be able to add code tabs and create a PR. That would take time, but I am motivated to contribute to this project.

B4nan:
Closing as the section is now rewritten with code tabs.

Issue #1334, opened 2021-01-23
Unnecessary  update on entity when using LoadStrategy.JOINED


Langstra:
**Describe the bug**
Example:
Project has Collection of Radio

We load a radio (call it radio1) from the database with together with a project (call it project1). Then we create a new project or do something else and we persistAndFlush it. Then it also finds a diff on radio1 that the project_id has changed to project1.id. This is weird, since it has not changed.

This triggers an additional and unnecessary query to update radio1.

**To Reproduce**
1. Checkout PR 1333
2. yarn test GH1334
3. ...

**Expected behavior**
Should not trigger an update query on radio1

**Additional context**
When setting the LoadStrategy to SELECT_IN it does not trigger the additional update query

B4nan:
The issue here is with `originalEntity` data being populated with the fetch joined result:

```ts
      ChangeSet {
        entity: Radio {
          id: 1,
          order: 0,
          question: 'ufzmqkk',
          project: [Reference],
          options: [Collection]
        },
        type: 'update',
        payload: { project: 1 },
        name: 'Radio',
        rootName: 'Radio',
        collection: 'radio',
        originalEntity: {
          id: 1,
          order: 0,
          question: 'ufzmqkk',
          project: [Object],
          o2__id: null,
          o2__value: null,
          o2__order: null,
          o2__radio_id: null
        }
      }
```

We should have something like this instead:

```ts
        originalEntity: {
          id: 1,
          order: 0,
          question: 'ufzmqkk',
          project: 1
        }
```

PR #1335, opened 2021-01-23
test(loadstrategy): added failing test for wrong diff

Langstra:
Failing test case for #1334


B4nan:
Closing as fixed in master

PR #1336, opened 2021-01-23
fix(sql): sort fetch-joined properties on their orderBy

### LINKED ISSUE #1331: OrderBy does not work for nested collections with LoadStrategy.JOINED
**Describe the bug**
When loading nested collections using the LoadStrategy.JOINED it only orders by the first collection. For instance:
Project has Collection<Radio> 
Radio has Collection<Option>

both radio and option have a property order. When using LoadStrategy.JOINED it creates one query to load all entities and it only orders by the order of Radio. When using LoadStrategy.SELECT_IN it creates multiple queries and it orders by the order of Radio and the order of Option.

**To Reproduce**
1. Checkout PR xxx
2. yarn test GHxxxx
3. 

**Expected behavior**
Should sort on both orders.

**Versions**

| Dependency | Version |
| - | - |
| node | ? |
| typescript | ? |
| mikro-orm | ? |
| your-driver | ? |

TiesB:
Merges a joined relation's orderBy with the parent's, keeping tableAliases in mind. Fixes #1331

### DIFF
--- a/packages/knex/src/AbstractSqlDriver.ts
+++ b/packages/knex/src/AbstractSqlDriver.ts
@@ -40,6 +40,7 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
     const joinedProps = this.joinedProps(meta, populate);
     const qb = this.createQueryBuilder<T>(entityName, ctx, !!ctx, false);
     const fields = this.buildFields(meta, populate, joinedProps, qb, options.fields as Field<T>[]);
+    const joinedPropsOrderBy = this.buildJoinedPropsOrderBy(entityName, qb, meta, joinedProps);
 
     if (Utils.isPrimaryKey(where, meta.compositePK)) {
       where = { [Utils.getPrimaryKeyHash(meta.primaryKeys)]: where } as FilterQuery<T>;
@@ -48,7 +49,7 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
     qb.select(fields)
       .populate(populate)
       .where(where)
-      .orderBy(options.orderBy!)
+      .orderBy({ ...options.orderBy!, ...joinedPropsOrderBy })
       .groupBy(options.groupBy!)
       .having(options.having!)
       .withSchema(options.schema);
@@ -581,6 +582,30 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
     const cond = Utils.getPrimaryKeyCond(entity, meta.primaryKeys);
     qb.select('1').where(cond!).setLockMode(mode);
     await this.rethrow(qb.execute());
+  }
+
+  protected buildJoinedPropsOrderBy<T extends AnyEntity<T>>(entityName: string, qb: QueryBuilder<T>, meta: EntityMetadata<T>, populate: PopulateOptions<T>[], parentPath?: string): QueryOrderMap {
+    const orderBy: QueryOrderMap = {};
+    const joinedProps = this.joinedProps(meta, populate);
+
+    joinedProps.forEach(relation => {
+      const prop = meta.properties[relation.field];
+      const propOrderBy = prop.orderBy;
+
+      const path = `${parentPath ? parentPath : entityName}.${relation.field}`;
+      const propAlias = qb.getAliasForJoinPath(path);
+      if (propOrderBy) {
+        Object.keys(propOrderBy).forEach(field => {
+          Object.assign(orderBy, { [`${propAlias}.${field}`]: propOrderBy[field] });
+        });
+      }
+
+      if (relation.children) {
+        const meta2 = this.metadata.find<T>(prop.type)!;
+        Object.assign(orderBy, { ...this.buildJoinedPropsOrderBy(prop.name, qb, meta2, relation.children, path) });
+      }
+    });
+    return orderBy;
   }
 
   protected buildFields<T extends AnyEntity<T>>(meta: EntityMetadata<T>, populate: PopulateOptions<T>[], joinedProps: PopulateOptions<T>[], qb: QueryBuilder<T>, fields?: Field<T>[]): Field<T>[] {
--- a/tests/issues/GH1331.test.ts
+++ b/tests/issues/GH1331.test.ts
@@ -1,0 +1,133 @@
+import {
+  Collection,
+  Entity,
+  IdentifiedReference,
+  LoadStrategy,
+  ManyToOne,
+  MikroORM,
+  OneToMany,
+  PrimaryKey,
+  Property,
+  QueryOrder,
+} from '@mikro-orm/core';
+import { SqliteDriver } from '@mikro-orm/sqlite';
+
+@Entity()
+export class RadioOption {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property()
+  order!: number;
+
+  @ManyToOne({
+    // eslint-disable-next-line @typescript-eslint/no-use-before-define
+    entity: () => Radio,
+    wrappedReference: true,
+  })
+  radio!: IdentifiedReference<Radio>;
+
+}
+
+@Entity()
+export class Radio {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property()
+  order!: number;
+
+  @ManyToOne({
+    // eslint-disable-next-line @typescript-eslint/no-use-before-define
+    entity: () => Project,
+    wrappedReference: true,
+  })
+  project!: IdentifiedReference<Project>;
+
+  @OneToMany(
+    () => RadioOption,
+    option => option.radio,
+    {
+      eager: true,
+      orderBy: { order: QueryOrder.ASC, id: QueryOrder.ASC },
+    },
+  )
+  options = new Collection<RadioOption>(this);
+
+}
+
+@Entity()
+export class Project {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property()
+  name!: string;
+
+  @OneToMany(
+    () => Radio,
+    radio => radio.project,
+    {
+      eager: true,
+      orderBy: { order: QueryOrder.ASC, id: QueryOrder.ASC },
+    },
+  )
+  radios = new Collection<Radio>(this);
+
+}
+
+describe('GH issue 1331', () => {
+
+  let orm: MikroORM<SqliteDriver>;
+
+  beforeAll(async () => {
+    orm = await MikroORM.init({
+      type: 'sqlite',
+      dbName: ':memory:',
+      entities: [Project, Radio, RadioOption],
+      loadStrategy: LoadStrategy.JOINED,
+    });
+    await orm.getSchemaGenerator().createSchema();
+  });
+
+  afterAll(async () => {
+    await orm.close(true);
+  });
+
+  test(`relations' orderBy should be respectend when using LoadStrategy.JOINED`, async () => {
+    const project = orm.em.create(Project, { name: 'project name' });
+    const radio1 = orm.em.create(Radio, { order: 0 });
+    const radio2 = orm.em.create(Radio, { order: 2 });
+    const radio3 = orm.em.create(Radio, { order: 1 });
+
+    radio1.options.add(orm.em.create(RadioOption, { order: 3 }));
+    radio1.options.add(orm.em.create(RadioOption, { order: 2 }));
+    radio1.options.add(orm.em.create(RadioOption, { order: 4 }));
+    radio1.options.add(orm.em.create(RadioOption, { order: 1 }));
+
+    radio2.options.add(orm.em.create(RadioOption, { order: 5 }));
+    radio2.options.add(orm.em.create(RadioOption, { order: 2 }));
+    radio2.options.add(orm.em.create(RadioOption, { order: 11 }));
+    radio2.options.add(orm.em.create(RadioOption, { order: 12 }));
+
+    radio3.options.add(orm.em.create(RadioOption, { order: 0 }));
+    radio3.options.add(orm.em.create(RadioOption, { order: 2 }));
+    radio3.options.add(orm.em.create(RadioOption, { order: 4 }));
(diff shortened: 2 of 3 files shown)


packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -39,7 +39,8 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
     const populate = this.autoJoinOneToOneOwner(meta, options.populate as PopulateOptions<T>[], options.fields);
     const joinedProps = this.joinedProps(meta, populate);
     const qb = this.createQueryBuilder<T>(entityName, ctx, !!ctx, false);
-    const fields = this.buildFields(meta, populate, joinedProps, qb, options.fields as Field<T>[]);
+    const fieldsAndJoinedPropsOrderBy = this.buildFields(meta, populate, joinedProps, qb, options.fields as Field<T>[]);
```
B4nan, line 42: I don't really like huge property names, especially if they contain conjunctions. Why dont we destruct instead?

```suggestion
    const { fields, orderBy } = this.buildFields(meta, populate, joinedProps, qb, options.fields as Field<T>[]);
```

But I would much rather see a dedicated method - this one is called `buildFields`, it should not have anything to do with ordering.

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -393,8 +394,8 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
     orderBy = this.getPivotOrderBy(prop, orderBy);
     const qb = this.createQueryBuilder<T>(prop.type, ctx, !!ctx).unsetFlag(QueryFlag.CONVERT_CUSTOM_TYPES);
     const populate = this.autoJoinOneToOneOwner(targetMeta, [{ field: prop.pivotTable }]);
-    const fields = this.buildFields(targetMeta, (options?.populate ?? []) as PopulateOptions<T>[], [], qb, options?.fields as Field<T>[]);
-    qb.select(fields).populate(populate).where(where).orderBy(orderBy!);
+    const fieldsAndJoinedPropsOrderBy = this.buildFields(targetMeta, (options?.populate ?? []) as PopulateOptions<T>[], [], qb, options?.fields as Field<T>[]);
```
B4nan, line 397: same here

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -457,9 +458,10 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
     return res;
   }
 
-  protected getFieldsForJoinedLoad<T extends AnyEntity<T>>(qb: QueryBuilder<T>, meta: EntityMetadata<T>, populate: PopulateOptions<T>[] = [], parentTableAlias?: string, parentJoinPath?: string): Field<T>[] {
+  protected getFieldsAndOrderByForJoinedLoad<T extends AnyEntity<T>>(qb: QueryBuilder<T>, meta: EntityMetadata<T>, populate: PopulateOptions<T>[] = [], parentTableAlias?: string, parentJoinPath?: string): { fields: Field<T>[]; orderBy: QueryOrderMap } {
```
B4nan, line 461: same here, having conjunctions inside method name feels even worse that in a property name...

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -473,10 +475,18 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
...
-      fields.push(...this.getFieldsForJoinedLoad(qb, meta2, relation.children, tableAlias, path));
+      const propFieldsAndOrderBy = this.getFieldsAndOrderByForJoinedLoad(qb, meta2, relation.children, tableAlias, path);
+      const propOrderBy = {};
+      if (prop.orderBy) {
+        for (const childField in prop.orderBy) {
+          propOrderBy[`${tableAlias}.${childField}`] = prop.orderBy[childField];
+        }
+      }
+      fields.push(...propFieldsAndOrderBy.fields);
+      orderBy = { ...orderBy, ...propOrderBy, ...propFieldsAndOrderBy.orderBy };
```
B4nan, line 486: you should not need to reassign the variable here. this is equal to the following, right?

```suggestion
      Object.assign(orderBy, { ...propOrderBy, ...propFieldsAndOrderBy.orderBy };
```

tests/issues/GH1331.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  @Property()
+  value!: string;
+
+  @Property()
+  order!: number;
+
+  @ManyToOne({
+    // eslint-disable-next-line @typescript-eslint/no-use-before-define
+    entity: () => Radio,
```
B4nan, line 30: this is invalid reference definition, you are missing `wrappedReference: true` here.

tests/issues/GH1331.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  @Property()
+  order!: number;
+
+  @Property()
+  question!: string;
+
+  @ManyToOne({
+    // eslint-disable-next-line @typescript-eslint/no-use-before-define
+    entity: () => Project,
```
B4nan, line 50: same here, `wrappedReference: true` missing

tests/issues/GH1331.test.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+function createOption(order: number): RadioOption {
+  const radioOption = new RadioOption();
+  radioOption.order = order;
+  radioOption.value = Math.random().toString(36).replace(/[^a-z]+/g, '').substr(0, 10);
+  return radioOption;
+}
+
+describe('GH issue xxxx', () => {
```
B4nan, line 107: `xxxx`

tests/issues/GH1331.test.ts
```diff
@@ -0,0 +1,162 @@
...
+      loadStrategy: LoadStrategy.JOINED,
+    });
+    await orm.getSchemaGenerator().createSchema();
+  });
+
+  afterAll(async () => {
+    await orm.close(true);
+  });
+
+  test(`GH issue 1278`, async () => {
```
B4nan, line 125: wrong issue number

tests/issues/GH1331.test.ts
```diff
@@ -0,0 +1,162 @@
...
+  radio.question = Math.random().toString(36).replace(/[^a-z]+/g, '').substr(0, 10);
+  return radio;
+}
+
+function createOption(order: number): RadioOption {
+  const radioOption = new RadioOption();
+  radioOption.order = order;
+  radioOption.value = Math.random().toString(36).replace(/[^a-z]+/g, '').substr(0, 10);
+  return radioOption;
+}
```
B4nan, line 105: not that it matters, but this is exactly what should be implemented in the entity ctors - that is what MikroORM unlocks, having methods like this seems a bit cumbersome :]

tests/issues/GH1331.test.ts
```diff
@@ -0,0 +1,162 @@
...
+  });
+
+  afterAll(async () => {
+    await orm.close(true);
+  });
+
+  test(`GH issue 1278`, async () => {
+    const mock = jest.fn();
+    const logger = new Logger(mock, ['query']);
+    Object.assign(orm.config, { logger });
```
B4nan, line 128: this code is not used anyhow. i am curious, as large portion of repros contain it - where/why did you copy it from?
```suggestion
```

tests/issues/GH1331.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    radio3.options.add(createOption(2));
+    radio3.options.add(createOption(4));
+    radio3.options.add(createOption(1));
+
+    project.radios.add(radio1, radio2, radio3);
+
+    await orm.em.persistAndFlush(project);
+    orm.em.clear();
+
+    const loadedProject = await orm.em.findOne(Project, { id: 1 });
```
B4nan, line 155: use `findOneOrFail` so the `!` is not needed
```suggestion
    const loadedProject = await orm.em.findOneOrFail(Project, project.id);
```

tests/issues/GH1331.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    expect(loadedProject?.radios.getItems().map(r => r.order)).toStrictEqual([0, 1, 2]);
+    expect(loadedProject?.radios.getItems()[0].options.getItems().map(r => r.order).toString()).toBe([1, 2, 3, 4].toString());
```
B4nan, line 157: this looks also very hard to read...

```suggestion
    expect(loadedProject.radios[0].options.getIdentifiers('order').toEqual([1, 2, 3, 4]);
```

B4nan (changes requested):
thanks! the fix itself looks fine, but i don't really like the way it is coded inside existing methods with different purposes. should be refactored. based on my comments, but if you would struggle with that, I will be happy to merge as is and refactor it myself

TiesB:
Thanks a lot for the feedback! I refactored it to one single function and applied your other suggestions (I hope correctly ;-) ). I was not sure about how to deal with the orderBy in the loadFromPivotTable method; it would be great if you could look at that.

tests/issues/GH1331.test.ts
Langstra: Oh yeah I copied GH1278.test.ts and the code was in there. I have no idea who originally wrote it.

I probably forgot to remove it myself. Thanks @tiesb for fixing.

B4nan: I am quite sure it was me :] I am using this exact snippet on many places in tests, not just in repros (oh yeah, I should have a helper method long time ago, lol). But as long as there are no asserts on the mocked logger, there is no need for mocking it.

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -40,6 +40,7 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
     const joinedProps = this.joinedProps(meta, populate);
     const qb = this.createQueryBuilder<T>(entityName, ctx, !!ctx, false);
     const fields = this.buildFields(meta, populate, joinedProps, qb, options.fields as Field<T>[]);
+    const orderBy = this.fillInOrderBy(entityName, qb, meta, options.orderBy!, joinedProps);
```
B4nan, line 43: I was expecting something like `buildOrderBy`, and I'd rather not mutate the exiting `options.orderBy` there.

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -583,6 +584,19 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
     await this.rethrow(qb.execute());
   }
 
+  protected fillInOrderBy<T extends AnyEntity<T>>(entityName: string, qb: QueryBuilder<T>, meta: EntityMetadata<T>, orderBy: QueryOrderMap, joinedProps: PopulateOptions<T>[]): QueryOrderMap {
+    joinedProps.forEach(prop => {
+      const propAlias = qb.getAliasForJoinPath(`${entityName}.${prop.field}`);
```
B4nan, line 589: this probably won't work with more than one level. the join path always starts with root entity name, but for if you would load author, its books and their tags, it would be something like `Author.books.tags`

would be good to add some tests for such use case

B4nan (changes requested):
looking much cleaner now!

B4nan:
Regarding the `loadFromPivotTable()` method, it should be used only for select-in strategy (or when lazy loading m:n collections via `init()` method).

packages/knex/src/AbstractSqlDriver.ts
TiesB: Very good point! I changed it such that it can handle nested relations (I looked at `getFieldsForJoinedLoad()` for reference) and extended the test with a level deeper relation

TiesB: What is intriguing me is the amount of queries executed; a query is added for retrieving the 'ds', even though they are also retrieved with the 2nd massive query. Do you know why this is and whether it's needed?

```
select `e0`.* from `a` as `e0` where `e0`.`id` = ? limit ?

select `e0`.`id`, `e0`.`order`, `e0`.`a_id`, `c1`.`id` as `c1__id`, `c1`.`order` as `c1__order`, `c1`.`b_id` as `c1__b_id`, `d2`.`id` as `d2__id`, `d2`.`order` as `d2__order`, `d2`.`c_id` as `d2__c_id` from `b` as `e0` left join `c` as `c1` on `e0`.`id` = `c1`.`b_id` left join `d` as `d2` on `c1`.`id` = `d2`.`c_id` where `e0`.`a_id` in (?) order by `e0`.`order` asc, `e0`.`id` asc, `c1`.`order` asc, `c1`.`id` asc, `d2`.`order` asc, `d2`.`id` asc

select `e0`.* from `d` as `e0` where `e0`.`c_id` in (?, ?, ?, ?, ?, ?, ?, ?) order by `e0`.`order` asc, `e0`.`id` asc
```

B4nan: hard to say, might be some other issue, either with reference wrapper, or `eager: true` combined with joined strategy

packages/knex/src/AbstractSqlDriver.ts
TiesB: Renamed it to `buildJoinedPropsOrderBy` now, since I feel `buildOrderBy` is not right because it doesn't touch the `options.orderBy` anymore.

TiesB:
I am running into the 500loc limit in codeclimate. I'd like to leave the refactoring to get the loc within limits to you, since you know the code waaayy better than I do :-)


> Regarding the `loadFromPivotTable()` method, it should be used only for select-in strategy (or when lazy loading m:n collections via `init()` method).

Cool. Then I'll leave that untouched.

B4nan:
Sure, don't worry too much about codeclimate.

B4nan (approved):

PR #1353, opened 2021-01-26
test(joined-strategy): added failing test case

Langstra:
Added a test case to show that the JOINED strategy gives different results than the SELECT_IN strategy. This is the case when entity A has a ManyToOne relation with entity B. This entity B has OneToMany relations that are eager loaded with C and D. 

When loading this entity A and then populating relation B using the SELECT_IN strategy it does eager load C and D. When using the JOINED strategy the relations C and D are not eager loaded.

### DIFF
--- a/tests/issues/GH1352.test.ts
+++ b/tests/issues/GH1352.test.ts
@@ -1,0 +1,151 @@
+import {
+  Collection,
+  Entity,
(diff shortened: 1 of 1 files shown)


B4nan:
Closing as fixed in master

PR #1365, opened 2021-01-29
docs: add additional sub-query example

ggriffin:
I found this example in tests particularly helpful.


B4nan:
Please add it to the next version too (so in `docs/docs` folder), this change would be forgotten when releasing v4.5.

ggriffin:
Here's a more thoughtful update. Feedback welcome!

PR #1375, opened 2021-02-01
feat(mssql): add MS SQL Server driver

### LINKED ISSUE #771: Mssql support
I think the orm should include mssql as a supported database type by default instead of doing a custom database setup

B4nan:
Adds a new `@mikro-orm/mssql` package.

Closes #771


UTGuy:
@B4nan Would love to see more work done on this

B4nan:
> @B4nan Would love to see more work done on this

Well, then do the work on this :] For me this have very low priority. I did gave it a try and rebased to latest v5 but it did not really help, there are more blockers than just the schema generator, and for me there are too much unknowns as I never used mssql.

UTGuy:
You asked... I've delivered.
https://github.com/mikro-orm/mikro-orm/pull/3023

B4nan:
The `clearDatabase` calls are taking a lot of time (>500ms). Its pretty much all from the `truncate.sql` script. Will try the other suggestions from https://stackoverflow.com/questions/253849/cannot-truncate-table-because-it-is-being-referenced-by-a-foreign-key-constraint

B4nan:
All the mapped types you added seems to be not needed, I guess that was for some initial implementation, right? Will remove them, all the tests are passing just fine without them. If they are needed, we need a test case that shows why we need them.

Will try to get the schema generator working, for now with disabled FKs, given we can't even define all of them in the `mssql-schema.sql` file.

UTGuy:
@B4nan 

> All the mapped types you added seems to be not needed, I guess that was for some initial implementation, right?
They are absolutely needed!

MSSql uses N-types (Unicode types) which is different from regular strings (Non-unicode types). If you have no ability to apply Unicode types then languages like Chinese will fail. This is incredibly important.

B4nan:
> They are absolutely needed!

I guess you are missing my point. Those types should be applied to all strings/dates, right? We dont need them if things work out of box the same way for all strings/dates. 

If you think they are still needed (with my latest changes), please provide test case that proves it. I already removed them, everything is passing just fine. I dont want to bend indernals via custom types - that is an extension for users, internally we can find more performant ways to do things.

UTGuy:
Yes, I will add some tests that will cause the default assumption of Unicode to fail.
You either need to have a Unicode type or a NonUnicode type as MSSQL commonly uses both.

B4nan:
@UTGuy I guess I got confused by what the unicode strings mean. Initially I thought the other "regular" strings will support only ASCII, but thats probably not true? I mean you can put things like the czech characters there or not? e.g. `ěščřžýáíéůú`? Those (and similar chars used e.g. in Slovakia, France, etc) are in my eyes something to support out of box with the default `string` mapping. If its just about chinese and special chars or even emoji, we don't need that and having the opt in type implementation sounds good to me.

RobinKemna:
@B4nan is this still work in progress and or other plans to support mssql in the near future?

Issue #446, opened 2020-04-02
Column types of chained OneToOne relations with primary keys are not handled properly


Haringat:
**Describe the bug**
When you chain OneToOne relations with primary keys the column types are not set properly for all entity tables and are actualy invalid. In the example below the class EntityC will get column type "varchar(255)" instead of "UUID".

I have created a repl.it for this: https://repl.it/repls/UprightStupidBuckets

**Stack trace**
this is a misbehaviour that does not throw a bug. Thus, no stack trace can be provided.

**To Reproduce**
1. open the repl.it link
2. wait for execution
3. see the wrong sql

**Expected behavior**
Type of entity_c.id should be "UUID".

**Additional context**

**Versions**

| Dependency | Version |
| - | - |
| node | 10.19.0 |
| typescript | 3.8.3 |
| mikro-orm | 3.6.0 |
| your-driver | pg (8.0.0) |

PR #1416, opened 2021-02-11
Add custom type mapping for find method

jbmikk:
I'm not sure if this is right or not, but my custom types are failing to load when using `findOne` or `find` methods.

Debugging I realized the `find` method in `AbstractSqlDriver` creates a `QueryBuilder` with the `convertCustomTypes` flag set to `false`. 

I changed it to `true` and now I get the results as expected.

Am I missing something? 

Shouldn't the `find` methods return the mapped custom types as I defined them in the entities?

### DIFF
--- a/packages/knex/src/AbstractSqlDriver.ts
+++ b/packages/knex/src/AbstractSqlDriver.ts
@@ -38,7 +38,7 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
     const meta = this.metadata.find<T>(entityName)!;
     const populate = this.autoJoinOneToOneOwner(meta, options.populate as PopulateOptions<T>[], options.fields);
     const joinedProps = this.joinedProps(meta, populate);
-    const qb = this.createQueryBuilder<T>(entityName, ctx, !!ctx, false);
+    const qb = this.createQueryBuilder<T>(entityName, ctx, !!ctx, true);
     const fields = this.buildFields(meta, populate, joinedProps, qb, options.fields as Field<T>[]);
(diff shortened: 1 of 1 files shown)


B4nan:
As you can see in the tests, your change would break something (namely #446).

Custom types should be already converted at the driver level. 

Please provide a repro for your problem first.

jbmikk:
Sorry, I tried to run the tests but I could not get them to work.
Anyway, I figured something like this would happen.

What I don't understand is, aren't `findOne` and `find` supposed to convert values also? How are they going to convert values if the driver method they use is hardcoded to `false`?

Shouldn't the driver method at least have a parameter just like `nativeUpdate` indicating whether we want conversion or not?

B4nan:
As I already said, custom types are converted at the EM level, what gets to driver should be already converted. Most probably you have a different issue - that is why I am asking for a reproduction. 

One way or the other, I won't merge anything that does not have a test case to prove the change is needed. You should always first build the failing test before you start changing the code.

jbmikk:
Sure, I'm not asking you to re-open this nor merge something without tests, I was just wondering.

Anyway, if you are curious I found the problem, it was the other thing I fixed in #1415, it looks like because of that the EM was not doing its job properly, but with that fix in everything works and I don't need this flag anymore.

B4nan:
Makes sense, if the property was retrived from db with wrong casing, the ORM will see it as undefined, hence ignore it.

Btw you can already try the dev version with your fix: `4.4.3-dev.19`

jbmikk:
I noticed the next package on npm, thanks.
I'm already using it.

PR #1424, opened 2021-02-12
feat: add explicit (multi) schema support

Aeolun:
Since I need this feature for our own usage, and I've found MikroORM to be a serious improvement over TypeORM. I'll build it regardless, but I hope it would be a good thing to have as part of the core, so I figured it'd be better to coordinate before it's fully done so I can adjust course to match conventions/wishes.

I realize someone else had started work on something like this before, but I couldn't really figure out what was going on in that PR, so I started from scratch.

Please let me know what you think.


packages/core/src/decorators/Entity.ts
```diff
@@ -4,10 +4,10 @@ import { Utils } from '../utils';
 import { Constructor, Dictionary } from '../typings';
 
 export function Entity(options: EntityOptions<any> = {}) {
-  return function <T>(target: T & Dictionary) {
+  return function<T>(target: T & Dictionary) {
```
B4nan, line 7: remove unrelated code style changes

packages/core/src/decorators/Entity.ts
```diff
@@ -4,10 +4,10 @@ import { Utils } from '../utils';
...
     const meta = MetadataStorage.getMetadataFromDecorator(target);
     Utils.merge(meta, options);
-    meta.class = target as unknown as Constructor<T>;
+    meta.class = (target as unknown) as Constructor<T>;
```
B4nan, line 10: remove unrelated code style changes

packages/core/src/decorators/Entity.ts
```diff
@@ -20,6 +20,8 @@ export function Entity(options: EntityOptions<any> = {}) {
 export type EntityOptions<T> = {
   tableName?: string;
   collection?: string;
+  schema?: string;
+  database?: string;
```
B4nan, line 24: this looks like its not used anywhere

packages/core/src/metadata/EntitySchema.ts
```diff
@@ -22,7 +22,7 @@ type Property<T, O> =
 type PropertyKey<T, U> = NonFunctionPropertyNames<Omit<T, keyof U>>;
 type Metadata<T, U> =
   & Omit<Partial<EntityMetadata<T>>, 'name' | 'properties'>
-  & ({ name: string } | { class: Constructor<T>; name?: string })
+  & ({ name: string } | { class: Constructor<T>; name?: string } | { name: string; schema: string })
```
B4nan, line 25: this looks wrong, I'd say it should be this instead (otherwise you'd not allow using `class` with `schema`)
```suggestion
  & ({ name: string; schema?: string } | { class: Constructor<T>; name?: string; schema?: string })
```

packages/core/src/platforms/Platform.ts
```diff
@@ -19,6 +19,10 @@ export abstract class Platform {
     return true;
   }
 
+  supportsMultipleSchemas(): boolean {
+    return true;
```
B4nan, line 23: sounds like a bad default value to me, I'd rather override where it is actually possible than making all drivers enabled by default.

packages/core/src/typings.ts
```diff
@@ -124,6 +124,7 @@ export type GetRepository<T extends AnyEntity<T>, U> = T[typeof EntityRepository
 
 export interface EntityProperty<T extends AnyEntity<T> = any> {
   name: string & keyof T;
+  schema: string;
```
B4nan, line 127: should be optional, or not?

tests/__snapshots__/SchemaGenerator.test.ts.snap
```diff
@@ -716,7 +766,7 @@ drop table if exists \`book_tag3\`;
 drop table if exists \`publisher3_tests\`;
 drop table if exists \`publisher3\`;
 drop table if exists \`test3\`;
-drop table if exists \`mikro_orm_migrations\`;
+drop table if exists \`:memory:\`.\`mikro_orm_migrations\`;
```
B4nan, line 769: this is obvious wrong :] i guess sqlite will need special treatment for the in memory db

tests/__snapshots__/SchemaGenerator.test.ts.snap
```diff
@@ -847,7 +897,7 @@ drop table if exists \`publisher4\`;
 drop table if exists \`test4\`;
 drop table if exists \`foo_bar4\`;
 drop table if exists \`foo_baz4\`;
-drop table if exists \`mikro_orm_migrations\`;
+drop table if exists \`:memory:\`.\`mikro_orm_migrations\`;
```
B4nan, line 900: same here

B4nan (changes requested):
Interesting, thanks a lot, gave it a quick look and if you can finalize it, will be happy to merge it. Few general hints:

- we need some documentation, right now this is just code and tests (but as you said, its a WIP, we can deal with that at the end - but a progress status in the PR description would be great to see what you plan and what is done already)
- do you plan to support postgres as well? not sure if things are possible with sqlite (i'd say nope?)
- what about schema diffing?
- the tests look a bit suspicious, e.g. entity generator snapshots are missing postgres entities?
- remove unrelated code style changes (e.g. the very first thing on the diff, I am not interested in wrapping each ternary expression in parens...)

packages/core/src/decorators/Entity.ts
Aeolun: I think my autoformatter is a bit forward in this regard. I turned it off at some point. I'll look through the changelist to remove the extraneous changes. 

Do you you use a standard that I can set it to so it won't happen in the future?

packages/core/src/decorators/Entity.ts
Aeolun: `schema` is currently used, `database` isn't. I was planning to do a similar tableName/collection thing where you can use either, but I didn't know enough about the code yet to be sure I'd add it in the correct place.

packages/core/src/metadata/EntitySchema.ts
Aeolun: Eh? I think I got some TS error somewhere in this regards. Maybe I added it here before I added it to `EntityMetadata<>`.

packages/core/src/platforms/Platform.ts
Aeolun: I thought about the existing drivers and how many of them support multiple schemas (most RDBMS do), then settled on this. No issues flipping it around though.

tests/__snapshots__/SchemaGenerator.test.ts.snap
Aeolun: Haha, yeah, I noticed this when I submitted it. Will be fixed in the next push.

packages/core/src/typings.ts
Aeolun: Not certain of the top of my head. I'll check.

Aeolun:
> Interesting, thanks a lot, gave it a quick look and if you can finalize it, will be happy to merge it. Few general hints:
> 
>     * we need some documentation, right now this is just code and tests (but as you said, its a WIP, we can deal with that at the end - but a progress status in the PR description would be great to see what you plan and what is done already)
> 
>     * do you plan to support postgres as well? not sure if things are possible with sqlite (i'd say nope?)
> 
>     * what about schema diffing?
> 
>     * the tests look a bit suspicious, e.g. entity generator snapshots are missing postgres entities?
> 
>     * remove unrelated code style changes (e.g. the very first thing on the diff, I am not interested in wrapping each ternary expression in parens...)

Indeed, it's not nearly done. Things that I can immediately think of that still need to be done:

- Support all currently supported databases (where possible)/features, otherwise gracefully fail. I started with MySQL because that's my use case and easiest to test.
- Write documentation. Better to do after the structure is finalized though.
- Tests for all the databases/scenarios. Might need you help pointing me towards things that need to be tested, as I'm not yet familiar with all possible permutations.

> * the tests look a bit suspicious, e.g. entity generator snapshots are missing postgres entities?

I'll have a look at this. Doesn't look like something I expected to happen.

Aeolun:
Just FYI, I may not respond to this every day, I get pulled in a bunch of different directions every week.

packages/migrations/src/Migrator.ts
```diff
@@ -98,14 +99,18 @@ export class Migrator {
...
 
     schema.getTables().forEach(table => {
-      /* istanbul ignore next */
-      const tableName = table.schema ? `${table.schema}.${table.name}` : table.name;
-
+      // Tables will always have a schema when the DB supports it, since they come directly from the getListTablesSQL
+      // but this isn't guaranteed for the metadata, which might have an empty schema and just use the default from
+      // the connection.
+      const tableName = table.name;
+      const tableNameWithSchema = table.schema ? `${table.schema}.${table.name}` : '';
```
B4nan, line 109: you call this `tableNameWithSchema` but when the schema is not set, it is empty string? :]

packages/mysql-base/src/MySqlSchemaHelper.ts
```diff
@@ -32,11 +32,19 @@ export class MySqlSchemaHelper extends SchemaHelper {
   };
 
   getSchemaBeginning(charset: string): string {
-    return `set names ${charset};\nset foreign_key_checks = 0;\n\n`;
+    return `set names ${charset};\n${this.disableForeignKeysSQL()}\n\n`;
```
B4nan, line 35: i kinda like this change, but it seems unrelated - i'd prefer to have a separate PR that would introduce this method - as long as the tests are passing, we can merge that right ahead

B4nan (changes requested):

Aeolun:
Not dead yet. Just trying to find time to work on it again :)

packages/core/src/typings.ts
Aeolun: Yeah, point. I misunderstood what EntityProperty was.

B4nan:
Just a small heads up - if you want to see this being merged at some point, get rid of the unrelated changes, remove the NPM lock files (this project uses yarn, and I don't see why you should do any changes to its dependencies anyway), and rebase with master (as there is already some basic support for schemas in schema generator). I'd suggest to do it rather sooner than later, especially the rebase (you should rebase it regularly, otherwise it will get harder and harder).

Aeolun:
@B4nan Yeah, I merged with master again already.  A lot of the changes were implemented in a more or less compatible way, so it was much less work than I expected.

I’ll clean up the whole thing afterwards (and pull out the unrelated changes you mentioned before), it’s still strictly WIP. I just didn’t want the changes to be only on my machine so I pushed it up. 

On the plus side, it seems like everything related to entity generation works fine for postgres, mariadb and sqlite, now just need to check mongo.

Aeolun:
It's alive! Need to add a little bit more testing, and update the docs (mainly around entitygenerator), but it seems it's now doing what I expect in postgres, mysql and sqlite (where it's expected it does nothing).

B4nan (commented):
do you still plan to finalize this? would be great to have it in v5

added few issues i spotted when peeking at the code that would be better resoved before actual review. also rebase is needed.

Aeolun:
@B4nan Yeah, it’s still on my todo list. Can hopefully finish all the outstanding issues the next time I work on it.

B4nan:
Not sure if this PR is still relevant, given the changes from #2296

Aeolun:
@B4nan Probably not relevant any more. It looks like that does what I was trying to do with this one. Nice way of doing both multi-schema and multi-tenancy in one go.

If I find that #2296 doesn't do what I need, I'll submit a different PR based on that, but for now let's close this.

PR #1428, opened 2021-02-12
Fix documentation for orderBy

jbmikk:
I found it was documented as `sort:` but the only way I managed to compile it is with `orderBy:`.
I'm not sure if it ever was `sort:` in the past so older versions of the docs may be wrong as well.


B4nan:
Good catch, I think it was always `orderBy`.

PR #1430, opened 2021-02-13
fix: Avoid mutation of Property options object

nspaeth:
I have a number of properties that I pass the same options to so I made a single object that I import and pass to fields to reduce some boilerplate, but this caused bugs.

The previous implementation mutated the passed in object so I've updated it to avoid that.

### DIFF
--- a/packages/core/src/decorators/Property.ts
+++ b/packages/core/src/decorators/Property.ts
@@ -11,14 +11,17 @@ export function Property<T>(options: PropertyOptions<T> = {}) {
     MetadataValidator.validateSingleDecorator(meta, propertyName, ReferenceType.SCALAR);
     const name = options.name || propertyName;
 
-    if (propertyName !== name && !(desc.value instanceof Function)) {
-      Utils.renameKey(options, 'name', 'fieldName');
-    }
+    const prop = {
+      reference: ReferenceType.SCALAR,
+      ...options,
+      name: propertyName,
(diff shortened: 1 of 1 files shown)


nspaeth:
`fieldName` is not a property of `EntityProperty`, but I don't think that was type-checked correctly with `renameKey`. In any case, I don't see anywhere that it is used, so maybe that could be removed altogether?

B4nan:
> fieldName is not a property of EntityProperty, but I don't think that was type-checked correctly with renameKey. In any case, I don't see anywhere that it is used, so maybe that could be removed altogether?

Nope, this works as expected, the property is called `fieldNames`, it is an array and the singular forms are also supported - the conversion then happens elsewhere (in the `EntitySchema` where all the normalization of properties happen). Removing it is a breaking change, not a fix. `name`, `fieldName` and `fieldNames` are all aliases from the decorator options point of view, internally we store and use only the plural form. None of that is used to set the property name - you cannot set that from decorator options, it is always implicit. `prop.fieldName` = what is in the database, `prop.name` = what is in the entity prototype.

Also reusing decorator input sounds like a bad idea to me. You should have a base entity instead. Many other decorators are mutating the options object too (although this is the only one that actually rewrites existing properties - still it is bad idea to me).

If you care about the mutation, correct fix is to copy the input object, not to restructure the code (so just prepend `option = { ...options }` and remove the rest of your changes).

PR #1450, opened 2021-02-16
perf(core): improve processing of 1:m relations

yousufhasan:
I was having a performance issue while saving a large list of entities having oneToMany Collection. on Debugging it turned out that processToMany method inside ChangeSetComputer.ts is taking a lot of time due to unnecessary looping through filter even if there are not items in removeStack. This quick fix will prevent running it unnecessarily. I have noticed a major performance improvement after introducing this check.

### DIFF
--- a/packages/core/src/unit-of-work/ChangeSetComputer.ts
+++ b/packages/core/src/unit-of-work/ChangeSetComputer.ts
@@ -84,7 +84,7 @@ export class ChangeSetComputer {
     const target = changeSet.entity[prop.name] as unknown as Collection<any>;
 
     // remove items from collection based on removeStack
-    if (target.isInitialized()) {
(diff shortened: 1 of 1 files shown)


packages/core/src/unit-of-work/ChangeSetComputer.ts
```diff
@@ -84,7 +84,7 @@ export class ChangeSetComputer {
     const target = changeSet.entity[prop.name] as unknown as Collection<any>;
 
     // remove items from collection based on removeStack
-    if (target.isInitialized()) {
+    if (target.isInitialized() && this.removeStack.size > 0) {
       target.getItems()
```
B4nan, line 88: we can improve this bit too - as we already know the collection is initialized, we can skip the init check inside `getItems()`

```suggestion
      target.getItems(false)
```

PR #1476, opened 2021-02-22
MariaDB: wrong Date value during select

pepakriz:
Hi,

I'm facing a strange bug with MariaDB driver. When I'm trying to load `date` property, the value is not shifted. It's still in UTC. Saving works correctly. The mysql driver contains a test for this scenario, but the MariaDB driver not. So I copy&pasted them and It's really failing just for MariaDB.

So here is PR with a failing code. Any idea, how to fix it?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -67,6 +67,8 @@ jobs:
         yarn run-rs-in-background
         yarn coverage > COVERAGE_RESULT
         echo "$(cat COVERAGE_RESULT)"
+      env:
+        TZ: Europe/Prague # use different timezone than UTC, because we need to validate that time zone shifting works
 
     - name: Release dev version for testing
       if: github.ref == 'refs/heads/master' && matrix.node-version == '15.x' && steps.changed_packages.outputs.changed_packages != '0'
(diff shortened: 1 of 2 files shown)


pepakriz:
Heh, Github pipelines are probably running in UTC, so everything seems OK. But try to run this branch locally and set `process.env.TZ = 'Europe/Prague'`.

pepakriz:
The difference between mariadb and mysql driver is that MySQL converts datetime type to native `Date` type. But MariaDB uses string representation. It is because the `dateStrings` option as a `string[]` is not supported by mariadb connector.

B4nan:
So better to raise the issue there - or a PR right ahead. I had to implement this myself too in the mysql2 package. 

https://github.com/sidorares/node-mysql2/pull/1200

pepakriz:
@B4nan nice. Ok, I'll try to prepare an equivalent solution for mariadb.

btw why you prefer this solution over [custom types](https://mikro-orm.io/docs/custom-types)? I am just curious about that. I feel more confident with encoding/decoding on my own instead of random magic converts provided by DB drivers. But I'm not so familiar with internal things, so not sure about all consequences

B4nan:
This took way longer than it should have, but v6 should be finally good after #4362 (although the result mapped would deserve a bigger rewrite, I can see at least one more edge case that won't work properly). FYI doing this only via mapped types won't work, as we need to have normalized db value to compare, and using `Date` objects for that is actually pretty handy.

PR #1495, opened 2021-02-26
feat(cli): allow mikro-orm config to return Promise

nikita-yaroshevich:
The use case is - when I need to get my connection information from a remote system (Secret Store, Microservice etc.)

### DIFF
--- a/packages/core/src/utils/ConfigurationLoader.ts
+++ b/packages/core/src/utils/ConfigurationLoader.ts
@@ -18,7 +18,7 @@ export class ConfigurationLoader {
       if (await pathExists(path)) {
         // eslint-disable-next-line @typescript-eslint/no-var-requires
         const config = require(path);
-        return new Configuration({ ...(config.default || config), ...options }, validate);
+        return new Configuration({ ...(await (config.default || config)), ...options }, validate);
       }
     }
 
--- a/docs/docs/installation.md
+++ b/docs/docs/installation.md
@@ -188,7 +188,12 @@ $ yarn mikro-orm
 ```
 
 For CLI to be able to access your database, you will need to create `mikro-orm.config.js` file that 
-exports your ORM configuration. TypeScript is also supported, just enable `useTsNode` flag in your
+exports your ORM configuration. 
+
+> Your ORM configuration file can export the Promise, like:
+> `export default Promise.resolve({...});`.
+
+TypeScript is also supported, just enable `useTsNode` flag in your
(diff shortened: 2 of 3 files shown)


packages/core/src/utils/ConfigurationLoader.ts
```diff
@@ -17,7 +17,11 @@ export class ConfigurationLoader {
 
       if (await pathExists(path)) {
         // eslint-disable-next-line @typescript-eslint/no-var-requires
-        const config = require(path);
+        const configImport = require(path);
+        let config = configImport.default || configImport;
+        if (config.then && config.catch) {
+          config = await config;
+        }
         return new Configuration({ ...(config.default || config), ...options }, validate);
```
B4nan, line 25: let's do it this way:

```suggestion
        const configImport = require(path);
        let config = configImport.default || configImport;

        if (config instanceof Promise) {
          config = await config;
        }

        return new Configuration({ ...config, ...options }, validate);
```

B4nan (changes requested):
Why do you call this performance improvement? :] In my eyes this is actually the opposite, it will make the config loading tiny bit slower (ofc the difference is probably not even measurable for a one time operation like this).

Anyway, I am fine with this. Please add a note to the docs too (https://mikro-orm.io/docs/installation#setting-up-the-commandline-tool).

packages/core/src/utils/ConfigurationLoader.ts
```diff
@@ -18,7 +18,7 @@ export class ConfigurationLoader {
       if (await pathExists(path)) {
         // eslint-disable-next-line @typescript-eslint/no-var-requires
         const config = require(path);
-        return new Configuration({ ...(config.default || config), ...options }, validate);
+        return new Configuration({ ...(await (config.default || config)), ...options }, validate);
```
B4nan, line 21: right, it can be actually as simple as that 👍

B4nan (approved):
Thanks, LGTM! lets wait for the green light from CI

packages/core/src/utils/ConfigurationLoader.ts
nikita-yaroshevich: actually, it can be even simpler :-), like:

`const config = require(path);
return new Configuration({ ...(await (config.default || config)), ...options }, validate);`

I've updated the docs

nikita-yaroshevich:
Waw! You respond faster than I wrote the comment :-)
Btw, thanks a lot for the project. I've just started to use it but looks really great.

PR #1496, opened 2021-02-26
feat(core): support embeddable arrays

### LINKED ISSUE #1369: Array of embeddables
I'm not sure wether it's possible to have an embeddable as array, e.g.

```
  @Embedded(() => [Item], { object: true })
  items: Array<Item> = [];
```
where Item is an embeddable (like the above or using Collection). On DB, items would be an array of Item objects. At the moment this seems to be not possible but maybe I'm doing something wrong

The same described above or with a Collection

B4nan:
```ts
  @Embedded(() => Address, { array: true })
  addresses: Address[] = [];
```

Closes #1369

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -3,13 +3,13 @@ version: "2"
 checks:
   file-lines:
     config:
-      threshold: 500
+      threshold: 1000
   method-lines:
     config:
-      threshold: 50
+      threshold: 200
   method-count:
     config:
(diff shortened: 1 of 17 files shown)


AaronPerezPerez:
Hi @B4nan! Are there any estimated date for this to be released? We currently need this in production. Thanks for the hard and amazing work! 😄

B4nan:
Use the dev version if you need it now.

PR #1525, opened 2021-03-04
[WIP] feat(core): poc concurrency check impl

kbkk:
Todo:
- [ ] tests
- [ ] should not issue an UPDATE if concurrency check field hasn't been changed
- [ ] version logic is similar, perhaps version can be a flavor of concurrency check
- [ ] batch support

edit: Whoops, just realized I opened a PR against the real thing and not my fork. Embarassing! :(

### DIFF
--- a/tests/entities-sql/ConcurrencyCheckUser.ts
+++ b/tests/entities-sql/ConcurrencyCheckUser.ts
@@ -1,0 +1,21 @@
+import { Collection, Entity, ManyToMany, OneToOne, PrimaryKey, Property } from '@mikro-orm/core';
+
+@Entity()
+export class ConcurrencyCheckUser {
+
+  @PrimaryKey({ length: 100 })
+  firstName: string;
+
+  @PrimaryKey({ length: 100, concurrencyCheck: true })
+  lastName: string;
+
+  @Property({ concurrencyCheck: true })
+  age: number;
+
+  constructor(firstName: string, lastName: string, age: number) {
+    this.firstName = firstName;
+    this.lastName = lastName;
+    this.age = age;
+  }
+
+}
--- a/packages/core/src/decorators/Property.ts
+++ b/packages/core/src/decorators/Property.ts
@@ -60,6 +60,7 @@ export type PropertyOptions<T> = {
   serializer?: (value: any) => any;
   serializedName?: string;
   comment?: string;
+  concurrencyCheck?: boolean;
 };
 
 export interface ReferenceOptions<T, O> extends PropertyOptions<O> {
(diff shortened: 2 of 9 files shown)


B4nan:
Hey there, do you plan to finish this? Will be happy to take over if not, I'd like to have this in v5.

kbkk:
@B4nan  I asked you some questions on slack back in the day, not sure how to proceed: https://mikroorm.slack.com/archives/CFSAF8DEX/p1615563861110300?thread_ts=1614698470.136900&cid=CFSAF8DEX

B4nan:
Oh sorry about that.

> What does quoteVersionValue do? Do I need any special escaping for my Concurrency Checks? I would assume the values are escaped at knex layer

I don't think so, you will be using regular entity properties, so things should be handled just fine. `quoteVersionValue` is there to allow fixing something with timezones in sqlite when date properties are used. If you git blame the method, you should find more information about that, but I guess nothing you should worry about.

> I get duplicate values in concurrencyCheckKeys, both in real app and the tests. This doesn't hurt the logic but probably slightly hurts performance. Should I use a Set to store the keys? Is the metadata expected to be loaded multiple times?

Yes, some parts of the discovery are executed twice to support random order of entities (and maybe more). Using `Set` sounds good to me.

> Version logic seems very similar to Concurrency Check. I think all persisting code would be same across them, with the exception a Version is increased automatically and CC by the developer. Do you think it would make sense to internally represent Version as Concurrency Check?

Version fields have special treatment to allow atomic updates, so that will need to stay as is probably, but the conditioning logic should be reused ideally. But I am fine merging it as a separate thing too and we can iterate on that. 

> One more thing in addition to the above, trying to understand the whole thing:
> What's the purpose of that last select query? It tries to update the version in memory with db state, right? I think the calculation here is simple enough not to make that query. Am I missing something? 

Mmmmm I believe I based the implemention on doctrine and it was done this way there. But you are probably right that it is not needed for updates, as we know the version value was correct, otherwise the update would fail (and we would throw anyway).

tests/entities-sql/ConcurrencyCheckUser.ts
```diff
@@ -0,0 +1,21 @@
+import { Collection, Entity, ManyToMany, OneToOne, PrimaryKey, Property } from '@mikro-orm/core';
+
+@Entity()
+export class ConcurrencyCheckUser {
```
B4nan, line 4: adding entities to this folder will require updating all the snapshots for schema/entity generators - i'd rather create new test in `tests/features` for this that will use its own entities. especially because i am finishing #1641 which will rewrite most of the schema generator tests, so this would be a hard rebasing for me.

B4nan:
Regular reminder - do you plan to finish this? Will be happy to take over if not, I'd like to have this in v5.

B4nan:
>  should not issue an UPDATE if concurrency check field hasn't been changed

We should throw `OptimisticLockError.lockFailed()` in such case, right?

B4nan:
Closing in favour of #2437

PR #1535, opened 2021-03-08
feat(core): allow updating nested 1:1 and m:1 references with EntityAssigner

jsprw:
Currently, when you do the following:
```
const authorEntity = new Author('Jon2 Snow', 'snow3@wall.st');
book.author = authorEntity;

wrap(book).assign({ author: { name: 'Jon Snow2' } }, { mergeObjects: true });
console.log(book.author); // { ... name: 'Jon Snow2' ... }
console.log(book.author === authorEntity) // false
```
... it will try to create a new author entity with the name Jon Snow2 altough it can assign these values to the existing m:1 reference. Was doubting to fix the functionality for the mergeObjects flag, however that may be a breaking change.

Right now, it will only update loaded references; should I also implement initializing the enitity and then asssign the new values?

### DIFF
--- a/packages/core/src/entity/EntityAssigner.ts
+++ b/packages/core/src/entity/EntityAssigner.ts
@@ -43,7 +43,7 @@ export class EntityAssigner {
       }
 
       if ([ReferenceType.MANY_TO_ONE, ReferenceType.ONE_TO_ONE].includes(props[prop]?.reference) && Utils.isDefined(value, true) && EntityAssigner.validateEM(em)) {
-          return EntityAssigner.assignReference<T>(entity, value, props[prop], em!, options);
+        return EntityAssigner.assignReference<T>(entity, value, props[prop], em!, options);
       }
 
       if (props[prop]?.reference === ReferenceType.SCALAR && SCALAR_TYPES.includes(props[prop].type) && (props[prop].setter || !props[prop].getter)) {
--- a/packages/core/src/entity/EntityAssigner.ts
+++ b/packages/core/src/entity/EntityAssigner.ts
@@ -47,7 +47,7 @@ export class EntityAssigner {
 
         // eslint-disable-next-line no-prototype-builtins
         if (options.updateNestedEntities && entity.hasOwnProperty(prop) && (Utils.isEntity(entity[prop]) || Reference.isReference(entity[prop])) && Utils.isPlainObject(value)) {
-          const unwrappedEntity = Reference.isReference(entity[prop]) ? entity[prop].unwrap() : entity[prop];
+          const unwrappedEntity = Reference.unwrapReference(entity[prop]);
 
           if (wrap(unwrappedEntity).isInitialized()) {
             return EntityAssigner.assign(unwrappedEntity, value, options);
--- a/packages/core/src/entity/EntityAssigner.ts
+++ b/packages/core/src/entity/EntityAssigner.ts
@@ -6,6 +6,7 @@ import { Utils } from '../utils/Utils';
 import { Reference } from './Reference';
 import { ReferenceType, SCALAR_TYPES } from '../enums';
 import { EntityValidator } from './EntityValidator';
+import { wrap } from './wrap';
 
 const validator = new EntityValidator(false);
 
@@ -43,7 +44,8 @@ export class EntityAssigner {
       }
 
       if ([ReferenceType.MANY_TO_ONE, ReferenceType.ONE_TO_ONE].includes(props[prop]?.reference) && Utils.isDefined(value, true) && EntityAssigner.validateEM(em)) {
-        if (options.updateNestedEntities && Utils.isEntity(entity[prop], false) && Utils.isPlainObject(value)) {
+        // eslint-disable-next-line no-prototype-builtins
+        if (options.updateNestedEntities && entity.hasOwnProperty(prop) && Utils.isEntity(entity[prop]) && wrap(entity[prop]).isInitialized() && Utils.isPlainObject(value)) {
           return EntityAssigner.assign(entity[prop], value, options);
         }
 
@@ -60,8 +62,7 @@ export class EntityAssigner {
 
       if (options.mergeObjects && Utils.isPlainObject(value)) {
         Utils.merge(entity[prop as keyof T], value);
-      // eslint-disable-next-line no-prototype-builtins
-      } else if (entity.hasOwnProperty(prop) && (!props[prop] || props[prop].setter || !props[prop].getter)) {
+      } else if (!props[prop] || props[prop].setter || !props[prop].getter) {
         entity[prop as keyof T] = value;
       }
     });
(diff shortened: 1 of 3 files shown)


packages/core/src/entity/EntityAssigner.ts
```diff
@@ -43,7 +43,11 @@ export class EntityAssigner {
       }
 
       if ([ReferenceType.MANY_TO_ONE, ReferenceType.ONE_TO_ONE].includes(props[prop]?.reference) && Utils.isDefined(value, true) && EntityAssigner.validateEM(em)) {
-        return EntityAssigner.assignReference<T>(entity, value, props[prop], em!, options);
+        if (options.updateNestedEntities && Utils.isEntity(entity[prop], false) && Utils.isPlainObject(value)) {
+          return EntityAssigner.assign(entity[prop], value, options);
+        }
+          return EntityAssigner.assignReference<T>(entity, value, props[prop], em!, options);
+
```
B4nan, line 50: bad ident, how come this was not cought by eslint 🤫 
```suggestion

        return EntityAssigner.assignReference<T>(entity, value, props[prop], em!, options);
```

packages/core/src/entity/EntityAssigner.ts
```diff
@@ -43,7 +43,11 @@ export class EntityAssigner {
       }
 
       if ([ReferenceType.MANY_TO_ONE, ReferenceType.ONE_TO_ONE].includes(props[prop]?.reference) && Utils.isDefined(value, true) && EntityAssigner.validateEM(em)) {
-        return EntityAssigner.assignReference<T>(entity, value, props[prop], em!, options);
+        if (options.updateNestedEntities && Utils.isEntity(entity[prop], false) && Utils.isPlainObject(value)) {
```
B4nan, line 46: we should handle identified references here too, also we first need to check if the entity is initialized

B4nan, line 46: also please try to get around that prototype pollution, it is probably not valid here, but let's play it safe

should be just about adding a check for `hasOwnProperty` at the `assign` method

tests/EntityAssigner.mysql.test.ts
```diff
@@ -76,6 +76,24 @@ describe('EntityAssignerMySql', () => {
     expect(book2.tags.getIdentifiers()).toMatchObject([tag2.id]);
   });
 
+  test('assign() should update m:1 or 1:1 nested entities [mysql]', async () => {
```
B4nan, line 79: we need to test what happens if the relation is not there, if it is not populated, if it is a reference wrapper, ...

B4nan (changes requested):

B4nan:
> should I also implement initializing the enitity and then asssign the new values?

The method is synchronous, you can't lazy load from there. 

But as I mentioned somewhere, I have a WIP that allows updating references (not loaded entities), see https://github.com/mikro-orm/mikro-orm/commit/78e5ef2efc2d38e179a379f890b03bd325245c99

I might postpone that change for v5, as it might be somehow breaking.

packages/core/src/entity/EntityAssigner.ts
jsprw: Doesn't it already do that?

options.updateNestedEntities: if the newly introduced flag is set
Utils.isEntity(entity[prop], false): whether the prop to be updated is an entity and not a reference
Utils.isPlainObject(value): whether the value to be set is an object and not a reference (number)

As it is a to 1 enitity, it is either initialized (object form) or else a reference right?

else: 

do the original behavior of the assign method.

jsprw: I don't know why that error comes up now. As I did not change any of the property assignments. I just added an if-statement to check whether the to-be updated prop is an entity or not and if so, it just calls the same function with the entity and data.

Ill add a check but I thought that was not within the scope of this MR.

B4nan: > Utils.isEntity(entity[prop], false): whether the prop to be updated is an entity and not a reference

exactly, it does not allow references with the second `false` param, and that is what i am missing here (it will be more than just switching the flag, you will need to call `unwrap` on the wrapper first - it needs to be separate branch)

anyway, instead of guessing, add tests so you clearly see what it does and what not :] but i can assure you, this will ignore identified references.

B4nan: i think your confusion comes from the "reference" and "reference wrapper". a reference is an entity instance that is not initialized. still it is an entity instance.

on the other hand, reference wrapper = identified reference

B4nan: you actually did change the code paths, GH does a good job showing this. just click on `Show paths` to see yourself

image

jsprw: Ah, I see, but I don't understand why it has never shown that error before, see https://github.com/mikro-orm/mikro-orm/blob/master/tests/EntityAssigner.mysql.test.ts#L22. This test will fail now with that hasOwnProperty check.

B4nan, line 46: the `hasOwnProperty` check could be here, that should not cause any test failures. the problem is due to this newly introduced recursive assign in the end.

jsprw: Yeah, understood. But doesn't that test demonstrate prototype pollution?

B4nan: nope, don't confuse prototype and instance properties

```ts
class A {}
const a = new A();
a.prop = 1; // instance
A.prototype.prop = 2; // prototype
```

jsprw: Still confused how this in a second call may pollute the prototype. Anyways, moved the check.

jsprw: would wrap(entity[prop]).isInitialized() cover the references? in my test cases, i saw it correctly being redirected to the assign reference method.

B4nan: yes it would, but the rest won't work, you can't call `assign(Reference, data)`. that is why i said you should have another branch that will handle reference wrapper and call `unwrap` on it first

something like this 

```ts
    if (Reference.isReference(value)) {
      value = value.unwrap();
    }

```

jsprw: Ah, excuse me, got it.

packages/core/src/entity/EntityAssigner.ts
```diff
@@ -43,6 +44,16 @@ export class EntityAssigner {
       }
 
       if ([ReferenceType.MANY_TO_ONE, ReferenceType.ONE_TO_ONE].includes(props[prop]?.reference) && Utils.isDefined(value, true) && EntityAssigner.validateEM(em)) {
+
+        // eslint-disable-next-line no-prototype-builtins
+        if (options.updateNestedEntities && entity.hasOwnProperty(prop) && (Utils.isEntity(entity[prop]) || Reference.isReference(entity[prop])) && Utils.isPlainObject(value)) {
+          const unwrappedEntity = Reference.isReference(entity[prop]) ? entity[prop].unwrap() : entity[prop];
```
B4nan, line 50: there is a helper for this already
```suggestion
          const unwrappedEntity = Reference.unwrapReference(entity[prop]);
```

B4nan:
Thanks!

PR #1546, opened 2021-03-10
docs: update cascading example

huynhninhbinh:
Remove a duplicated example (the last one is exactly the same).


B4nan:
oops, wrong button :] thanks, yeah this comes from history as there was one more cascading type - merge, but that is not always enabled as people did not really understand how that works and it just added a lot of confusion.

huynhninhbinh:
Thanks for your amazing work, just glad that I could help! :)

PR #1547, opened 2021-03-10
Use mysql2 instead of mysql for MonkeyPatcher

thomaschaaf:
This might be incorrect but why is the monkeypatcher using the `mysql` instead of the `mysql2` reference?


B4nan:
The patching is for [mariadb driver](https://github.com/mikro-orm/mikro-orm/blob/master/packages/mariadb/src/MariaDbConnection.ts#L18), that will in fact use neither mysql nor mysql2, but mariadb driver. If you check the knex dialects, mysql2 dialect in an extension of mysql dialect, so that is why the tests are passing probably. 

I'd say this does not really matter, but I'd prefer not to change this, just for the sake of added safety.

PR #1551, opened 2021-03-12
chore: upgrade knex to 0.95.2

thomaschaaf:


B4nan:
This is a BC, as knex now requires TS 4.1 and a TS bump for us is breaking (new features that are not backcompatible, like template literal types, or the ts-morph support).

Also please stick to the code style and rules in the repo - I don't want namespaced imports (`Knex.Transaction` etc), and the version should stay pinned as it was (`0.95`, not `^0.95`). 

Build is failing, so would be nice if you first get it passing locally (as each push sends an email to me). I would actually expect more changes to be needed...

packages/knex/src/AbstractSqlConnection.ts
```diff
@@ -170,10 +170,10 @@ export abstract class AbstractSqlConnection extends Connection {
    * support edge cases like `\\?` strings (as `positionBindings` was removing the `\\`)
    */
   private patchKnexClient(): void {
-    const query = Client.prototype.query;
+    const query = Knex.Client.prototype.query;
```
thomaschaaf, line 173: can you check that this is correct?

B4nan: as I said, i am not interested in using namespaced imports. keep it as it was.

thomaschaaf:
> This is a BC, as knex now requires TS 4.1 and a TS bump for us is breaking (new features that are not backcompatible, like template literal types, or the ts-morph support).
> 
> Also please stick to the code style and rules in the repo - I don't want namespaced imports (`Knex.Transaction` etc), and the version should stay pinned as it was (`0.95`, not `^0.95`).

I pinned the version now. But I don't think named imports work anymore: https://github.com/knex/knex/blob/master/UPGRADING.md#upgrading-to-version-0950

> 
> Build is failing, so would be nice if you first get it passing locally (as each push sends an email to me). I would actually expect more changes to be needed...

Sorry I didn't mean to bug you with emails..


I'm actually not sure why the tests are currently failing. So I probably should close the pr :(

B4nan:
Lol I see, what a weird decesion with the namespaces, I guess we will need to live with that. For me this is a bad move, I feel like in TS world namespaces are quite discouraged. Would be great to understand the reason for it...

Let's keep the update for later, I am planning to start with v5 quite soon (let's say in few weeks).

PR #1555, opened 2021-03-12
Update entity-generator.md

yakryder:
Update the instructions for running entity generator without local installation

### DIFF
--- a/docs/docs/entity-generator.md
+++ b/docs/docs/entity-generator.md
@@ -7,8 +7,8 @@ To generate entities from existing database schema, you can use `EntityGenerator
 You can use it via CLI: 
 
 ```sh
-npx mikro-orm generate-entities --dump  # Dumps all generated entities
-npx mikro-orm generate-entities --save --path=./my-entities  # Saves entities into given directory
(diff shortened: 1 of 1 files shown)


B4nan:
Nope, [the script is called `mikro-orm`](https://github.com/mikro-orm/mikro-orm/blob/master/packages/cli/package.json#L38). And you still need to install things locally, otherwise the global installation would not see locally defined packages like the driver.

yakryder:
@B4nan Indeed, that's what I ended up having to do. The local install is clearly indicated in the FAQ if you know to look there, but I did burn a little time here (sticking point was ultimately the driver issue).

I will close this, do you feel positive/negative/neutral about a link from this section of the docs to the relevant FAQ blurb? 

Like 

> You can use it via CLI, which [must be installed locally](https://github.com/mikro-orm/mikro-orm/blob/5151b216f8c4400678de6a77ff5f6d8267edca63/docs/docs/faq.md#i-cannot-run-the-cli):

B4nan:
Sure, lets add a note there

PR #1557, opened 2021-03-13
Added custom type reproduction case.

thekevinbrown:
test(custom types): added custom type reproduction case

When a custom type uses object values, we used to incorrectly assume
these were nested where parameters when they could actually be an
instance of a custom type. Also added support for the $sameAs operator.

### DIFF
--- a/docs/docs/query-conditions.md
+++ b/docs/docs/query-conditions.md
@@ -82,22 +82,23 @@ const res = await orm.em.find(Author, [1, 2, 7]);
 
 ### Comparison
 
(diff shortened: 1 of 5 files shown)


thekevinbrown:
Never mind, I fixed it.

If someone else finds this in the future, you can't use plain old Javascript objects and expect Mikro to convert them to your custom type for you. Make a class, then ensure your filter passes instances of that class, e.g.

```typescript
const [result] = await orm.em.find(A, { prop: { $sameAs: new Point(5, 9) } });
```

PR #1574, opened 2021-03-17
docs: improve nestjs integration docs

Langstra:
After the docs are improved I'll upload the same to docs.nestjs.com repo so MikroORM is represented there as well. Hopefully more people will then make the right call.

### DIFF
--- a/docs/docs/usage-with-nestjs.md
+++ b/docs/docs/usage-with-nestjs.md
@@ -44,22 +44,42 @@ Once the installation process is completed, we can import the `MikroOrmModule` i
 export class AppModule {}
 ```
 
-The `forRoot()` method accepts the same configuration object as `init()` from the MikroORM package. 
-You can also omit the parameter to use the CLI config.
+The `forRoot()` method accepts the same configuration object as `init()` from the MikroORM package. Check [this page](/configuration) for the complete configuration documentation.
+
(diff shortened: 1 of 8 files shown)


docs/docs/usage-with-nestjs.md
```diff
@@ -44,22 +44,42 @@ Once the installation process is completed, we can import the `MikroOrmModule` i
 export class AppModule {}
 ```
 
-The `forRoot()` method accepts the same configuration object as `init()` from the MikroORM package. 
-You can also omit the parameter to use the CLI config.
+The `forRoot()` method accepts the same configuration object as `init()` from the MikroORM package. Check [this page](/configuration) for the complete configuration documentation.
+
+Alternatively we can [configure the CLI](/installation#setting-up-the-commandline-tool) by creating a configuration file `mikro-orm.config.ts` and then call the `forRoot()` without any arguments.
```
B4nan, line 49: should be noted this won't work with build tools that do tree shaking (there is a note already on https://mikro-orm.io/docs/installation/#setting-up-the-commandline-tool), so just a one line about this could be enough

B4nan (changes requested):
Looking good, we should also propagate the changes at least to the 4.4 stable docs, maybe even for all the 4.x versions.

B4nan (approved):

B4nan:
There is some wrong link so the build fails, can you fix that pls? 

https://github.com/mikro-orm/mikro-orm/runs/2141633805?check_suite_focus=true

Langstra:
Yes, I'll fix it and create a new PR.

PR #1591, opened 2021-03-23
bug: add reproduction of nested em create with TsMorphMetadataProvider

Langstra:


tests/issues/GH1592.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property()
+  question: string = Math.random().toString(36).replace(/[^a-z]+/g, '').substr(0, 10);
+
+  @OneToOne(
+    {
+      type: RadioOption,
```
B4nan, line 40: why are you using ts-morph and providing so many details anyway? but still partial? :]

i think that will be the issue, as if the `type` is set, ts-morph will skip the property. but you are missing the `wrappedReference` flag here which would be otherwise added.

so if you would remove the `type` here, it would probably work fine

Langstra: Removing the `type` attribute here does not fix the problem. Even though I am unsure what it does, I tried adding the `wrappedReference` flag and set it to `true`, but that did not help either.

I updated the reproduction.

PR #1630, opened 2021-03-31
fix(knex): find entity by advanced custom types

Vakrim:
Fixes SQL queries using custom types as find conditions.

Field `mapper` was adding unnecessary aliases in where clause creating invalid SQL syntax:
```
... where T_AsText(`e0`.`point`) as point = 'point(1 1)' 
```
and now it's
```
... where T_AsText(`e0`.`point`) = 'point(1 1)' 
```

### DIFF
--- a/packages/knex/src/query/QueryBuilderHelper.ts
+++ b/packages/knex/src/query/QueryBuilderHelper.ts
@@ -54,6 +54,7 @@ export class QueryBuilderHelper {
         return this.knex.raw(valueSQL);
       }
 
+      /* istanbul ignore next */
       return this.knex.raw(valueSQL + ' as ' + this.platform.quoteIdentifier(alias ?? prop.fieldNames[0]));
     }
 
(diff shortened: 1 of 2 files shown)


packages/knex/src/query/QueryBuilderHelper.ts
```diff
@@ -48,8 +48,13 @@ export class QueryBuilderHelper {
...
       const prefixed = this.prefix(field, true, true);
-      /* istanbul ignore next */
-      return this.knex.raw(prop.customType.convertToJSValueSQL!(prefixed, this.platform) + ' as ' + this.platform.quoteIdentifier(alias ?? prop.fieldNames[0]));
+      const valueSQL = prop.customType.convertToJSValueSQL!(prefixed, this.platform);
+
+      if (alias === null) {
+        return this.knex.raw(valueSQL);
+      }
+
+      return this.knex.raw(valueSQL + ' as ' + this.platform.quoteIdentifier(alias ?? prop.fieldNames[0]));
```
B4nan, line 57: let's keep the coverage ignore hint as it was to maintain branch coverage
```suggestion
      /* istanbul ignore next */
      return this.knex.raw(valueSQL + ' as ' + this.platform.quoteIdentifier(alias ?? prop.fieldNames[0]));
```

PR #1639, opened 2021-04-02
fix(core): fix querying embeddables over cast fields

francisco-sanchez-molina:
When within an embedded entity a numeric query is performed on a numeric field, the query is malformed:

`SyntaxErrorException: select "e0".* from "user" as "e0" where "e0".("address4"->>'number')::float8 > 2 limit 1 - syntax error at or near "("`

in this change the query builder prefix function detects if the field name has parenthesis, to add the alias inside

### DIFF
--- a/packages/knex/src/query/QueryBuilderHelper.ts
+++ b/packages/knex/src/query/QueryBuilderHelper.ts
@@ -20,12 +20,12 @@ import { JoinOptions } from '../typings';
 export class QueryBuilderHelper {
 
   constructor(private readonly entityName: string,
-              private readonly alias: string,
-              private readonly aliasMap: Dictionary<string>,
-              private readonly subQueries: Dictionary<string>,
-              private readonly metadata: MetadataStorage,
-              private readonly knex: Knex,
-              private readonly platform: Platform) { }
+    private readonly alias: string,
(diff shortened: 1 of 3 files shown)


packages/knex/src/query/QueryBuilderHelper.ts
```diff
@@ -20,12 +20,12 @@ import { JoinOptions } from '../typings';
...
-              private readonly subQueries: Dictionary<string>,
-              private readonly metadata: MetadataStorage,
-              private readonly knex: Knex,
-              private readonly platform: Platform) { }
+    private readonly alias: string,
+    private readonly aliasMap: Dictionary<string>,
+    private readonly subQueries: Dictionary<string>,
+    private readonly metadata: MetadataStorage,
+    private readonly knex: Knex,
+    private readonly platform: Platform) { }
```
B4nan, line 28: lets remove unrelated code style changes

```suggestion
              private readonly alias: string,
              private readonly aliasMap: Dictionary<string>,
              private readonly subQueries: Dictionary<string>,
              private readonly metadata: MetadataStorage,
              private readonly knex: Knex,
              private readonly platform: Platform) { }
```

B4nan (approved):

PR #1650, opened 2021-04-06
fix(core): fix aliasing of embeddables in update query

francisco-sanchez-molina:
…query over embedded field

When I perform a nativeUpdate as
```typescript
await orm.em.nativeUpdate(User, {
      address4: {
        number: {
          $gt: 2,
        },
      },
    }, {
      after: 2,
    });
```
the result is:
```
TableNotFoundException: update "user" set "after" = 2 where ("e0"."address4"->>'number')::float8 > 2 - missing FROM-clause entry for table "e0"
```

this pr refactors some checks on whether it is necessary to include a prefix based on the type of operation and it puts the decision in a single method.

Thank you for your excellent work!


B4nan:
Thanks!

PR #1656, opened 2021-04-06
fix(knex): find by custom types with object subconditions

jbmikk:
When using queries with conditions in custom types an alias gets appended in the `where` expression for the type.

```
    const foundLocation = await orm.em.findOne(Location, {
      point: new Point(1, 1),
      extendedPoint: { $ne: null },
    });
```


B4nan:
Thanks!

PR #1710, opened 2021-04-20
doc: fix typos in NestJS docs

hkdobrev:
- [x] Change "automatatic" to "automatic".
- [x] Change "thru" to "through".


B4nan:
Thanks! The changes are made in the next docs only (`docs/docs`), which is fine, but if you'd like to see them in the current (or older) versions, you'd also have to fix it in the `docs/versioned_docs` folder.

Will leave it up to you :]

PR #1728, opened 2021-04-26
docs: fix typo

PlatinBae:


B4nan:
Thanks! Can you please remove the permission changes (`100755 → 100644`)?

PlatinBae:
Should I change the permission from 100644 to 100755 again?

B4nan:
Yes, the diff should be clean, containing only the files you actually changed. You could also soft reset the commit, add only the files you changed, and force push the result. Changing the permissions back explicitly will probably do the trick too.

B4nan:
Perfect, thanks!

B4nan (approved):

PR #1729, opened 2021-04-26
added missing cli command on migrations

rubiin:


B4nan:
I don't understand why this should go to migrations docs. Installation of CLI is described in the installation section, and it also does mention the CLI needs to be installed locally. We also recently added the installation part to the code block to make it more explicit/visible.

https://mikro-orm.io/docs/installation/#setting-up-the-commandline-tool

If you want, we can have a quote there, but single line is enough, migrations docs should be about using migrations, not about installing CLI. 

```
> To work with the CLI, first install @mikro-orm/cli package locally. The version needs to be aligned with the @mikro-orm/core package.
```

rubiin:
yeah we can add quote . This will be a heads up for user so they would install cli

PR #1741, opened 2021-04-28
feat(core): fork event manager when forking entity managers

jbmikk:
This forks the event manager when forking entity managers.

Global subscribers are available in the forked entity manager, but subscribers added to the fork itself are only available in the fork.

```
const fork = orm.em.fork();
fork.getEventManager().registerSubscriber(subscriber)
```

This is useful in scoped DI containers for entity managers to add isolated subscribers for each container.

The only doubt I have is I get an `onInit` event after forking and I'm not sure if that should be happening because it wasn't happening in the original test I copied this one from.


B4nan:
This is in conflict with the commit that introduced the shared instance, originally it was like that: https://github.com/mikro-orm/mikro-orm/pull/726

B4nan:
Please fix the linting errors. Also I still don't really understand the motivation.

> This is useful in scoped DI containers for entity managers to add isolated subscribers for each container.

This is not really explaining the _why_... In general there should be one global EM and one (or more) forks for each request, why would you want to isolate subscribers across forks? They are stateless by definition, maybe you are just trying to pass some state into where it shouldn't be? 

cc @Langstra as the original change was made by you

jbmikk:
@B4nan Consider this work in progress, I have not even tested the last changes yet, I just pushed them so we can discuss about it and see what everyone thinks. And we defintely need more tests. The fact that my original PR was breaking desired behavior without any failing tests certainly indicates we need them.

The *why* is I'm working on a kind of change-set logger to keep a history of changes to my tables. This change set logger allows undoing old history by reverting changes, but I need to treat different the undo operations in the history log. I can't just log those changes as new changes. And not only that, I also need to add the user ID to those change sets so I need request information.

So yes, my subscribers are stateful. But more important than that is that they are **scoped**, they belong to the request (and its forked EM). I don't see why a scoped subscriber should listen to changes in another forked request. I want each subscriber to only listen to changes in that scoped (forked) EM transaction. So what I'm doing is adding a new subscriber instance to each fork.

If you want to listen to changes globally you can still do it, you can always subscribe to the global EM.
Now, if @Langstra  needs to subscribe globally from the child EM, then maybe we could have a parameter in the `registerSubscriber` indicating whether you want to subscribe to this EM instance, or to the parent (global) EM.

Does any of this make sense or is this too weird?

How would you solve the scoping problem?

My current hack is that I have a global subscriber with an internal transaction map. So each time a request arrives I can add a bunch of information to the transaction map and then when the subscribers listens to an event it looks up information in the map related to that specific transaction. It's kind of ugly, and prone to memory leaks, but it's the best I can do for now and it works.

StrikeForceZero:
I'm in the same boat as @jbmikk needing to have context of what user performed a change for logging change sets.

I was about to experiment having an extended method `registerSubscriber(context)` - `context` being the current request id but I might explore this PR

StrikeForceZero:
Here is a workaround snippet that should help with the "why"

```ts
export class FooEventSubscriber implements EventSubscriber<AnyEntity> {
    constructor(
        private readonly currentRequest: Request,
        private readonly entityManager: EntityManager,
    ) {
        this.currentRequest.once('close', () => this.unregister());
    }
    // would be nice if EventManager had this helper method
    private unregister() {
        const { listeners, entities } = (this.entityManager.getEventManager() as unknown as ExposedEventManager)
        Object.keys(EventType)
            .filter(event => event in this)
            .forEach(event => {
                listeners[event] = listeners[event].filter(listener => listener !== this)
            });
        entities.delete(this);
    }
    async onFlush(args: FlushEventArgs): Promise<void> {
        const { id: userId } = (this.currentRequest.user ?? {});
        const requestId = this.currentRequest['id'];
        // entityManager.EventManager is global so to cheat, we make sure the entity manager is the same before proceeding
        if (this.entityManager !== args.em) {
            // wrong entityManager so ignore
            return;
        }
        // now that we know we have an event that triggered from the forked entity manager, continue logic here
    }
}    
```
```ts
@Injectable()
export class AuditEntityMiddleware {
    constructor(
        // required for @UseRequestContext()
        private readonly orm: MikroORM,
    ) {}

    @UseRequestContext()
    use(req: Request, res: Response, next: NextFunction) {
        const em = RequestContext.getEntityManager();
        // useless check to make sure no one tries cleaning up "unused" objects
        if (!this.orm || !em) {
            throw new Error('missing entity manager - was @UseRequestContext removed from this scope?')
        }
        const subscriber = new FooEventSubscriber(req, em);
        em.getEventManager().registerSubscriber(subscriber)
        next();
    }
}
```

jbmikk:
@StrikeForceZero I didn't have time to fix this, but I will when I can.
I don't know if it's going to be merged or not, but regardless of that I think it is worth exploring.

B4nan:
I was hoping to get some feedback from @Langstra as the original change was made by him.

B4nan:
I think this could be solved in much easier way, we could add an option to `em.fork()` to force new instance of `EventManager`. 

```ts
const fork = orm.em.fork({ freshEventManager: true });
fork.getEventManager().registerSubscriber(subscriber)
```

wdyt?

jbmikk:
@B4nan sounds good

PR #1814, opened 2021-05-15
refactor: extract methods for disabling FKs

Aeolun:
Another small change from multi-schema support PR.

### DIFF
--- a/packages/postgresql/src/PostgreSqlSchemaHelper.ts
+++ b/packages/postgresql/src/PostgreSqlSchemaHelper.ts
@@ -15,11 +15,11 @@ export class PostgreSqlSchemaHelper extends SchemaHelper {
   };
 
   getSchemaBeginning(charset: string): string {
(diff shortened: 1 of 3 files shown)


packages/postgresql/src/PostgreSqlSchemaHelper.ts
```diff
@@ -211,6 +211,14 @@ export class PostgreSqlSchemaHelper extends SchemaHelper {
     return 'postgres';
   }
 
+  disableForeignKeysSQL(): string {
+    return `SET session_replication_role = 'replica';`;
+  }
+
+  enableForeignKeysSQL(): string {
+    return `SET session_replication_role = 'origin';`;
```
B4nan, line 219:
```suggestion
    return `set session_replication_role = 'replica';`;
  }

  enableForeignKeysSQL(): string {
    return `set session_replication_role = 'origin';`;
```

Aeolun:
Drat, guess those functions aren't covered by any test individually yet.

PR #1837, opened 2021-05-19
feat(core): add PlainObject class that DTO's can extend to treat class as POJO

Langstra:


docs/docs/entity-manager.md
```diff
@@ -137,7 +137,12 @@ const user1 = await orm.em.findOne(User, 1);
 
 As you can see in the fifth example, one can also use operators like `$and`, `$or`, `$gte`, 
 `$gt`, `$lte`, `$lt`, `$in`, `$nin`, `$eq`, `$ne`, `$like`, `$re`. More about that can be found in 
-[Query Conditions](query-conditions.md) section. 
+[Query Conditions](query-conditions.md) section.
+
+#### Using custom FilterQuery DTO
+If you decide to abstract the filter options in your own object then you might run into the problem that the find option does not return the results you'd expect. This is due to the fact that the FilterQuery should be provided as a POJO (Plain Old Json Object `{}`).
```
B4nan, line 143: POJO comes from java and means plain old java object, here i'd call it plain old js object, but calling it json object seems weird :]

docs/docs/entity-manager.md
```diff
@@ -137,7 +137,12 @@ const user1 = await orm.em.findOne(User, 1);
...
+
+If you want to provide your own FilterQuery DTO, then your DTO class should extend the `PlainJson` class. This way MikroORM knows it should be treated as such.
```
B4nan, line 145: the class is called `PlainObject` but you are referring to it in docs as `PlainJson`. I like `PlainObject` better, so lets just fix the docs.

B4nan (changes requested):
looking good, just two small docs issues

B4nan (approved):

Langstra:
Can we merge this?

PR #1853, opened 2021-05-24
docs: fix references loading with key name

jbmikk:
The docs are wrong.


B4nan:
Well, yes, the docs were dated, but your change is not really fixing them (read the whole sentence now, `load` is calling `load`?)

B4nan:
Perfect, thanks!

PR #1854, opened 2021-05-24
fix(core): use tsconfig-paths loadConfig function

### LINKED ISSUE #1849: 4.5.5 doesn't support  "extends": "@tsconfig/node12/tsconfig.json",
**Describe the bug**
I updated from 4.5.4 to 4.5.5 and migrations stopped working

**Stack trace**
```
> node node_modules/.bin/mikro-orm migration:up

(node:12927) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open '/Users/aklamer/Documents/GitHub/integrationhub/@tsconfig/node12/tsconfig.json'
(node:12927) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:12927) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
tsconfig:
```
{
  "extends": "@tsconfig/node12/tsconfig.json",
  "compilerOptions": {
    "experimentalDecorators": true,
    "emitDecoratorMetadata": true,
    "types": ["mocha"],
    "sourceMap": true
  },
  "exclude": ["node_modules"]
}

```
**To Reproduce**

1. npm install @tsconfig/node12 --save-dev
2. use tsconfig from example above with extends
3. use mikro-orm 4.5.5 
4. try to apply migrations

**Expected behavior**
Successfully migrated up to the latest version

**Additional context**

**Versions**
```
 "@mikro-orm/core": "4.5.5",
    "@mikro-orm/migrations": "4.5.5",
    "@mikro-orm/postgresql": "4.5.5",
   "@mikro-orm/cli": "4.5.5",
    "@tsconfig/node12": "^1.0.7",
 "typescript": "^4.2.4"
node -v: v12.16.3```

### PR DESCRIPTION
**Pull request by Langstra**
This way the extended property of the tsconfig file can be read
property.

Fixes #1849

### DIFF
--- a/packages/core/src/utils/ConfigurationLoader.ts
+++ b/packages/core/src/utils/ConfigurationLoader.ts
@@ -1,7 +1,6 @@
 import dotenv from 'dotenv';
-import { pathExists, readFile } from 'fs-extra';
-import { dirname, join, isAbsolute } from 'path';
-import stripJsonComments from 'strip-json-comments';
+import { pathExists } from 'fs-extra';
+import { join, isAbsolute } from 'path';
 import { IDatabaseDriver } from '../drivers';
 import { Configuration, Options } from './Configuration';
 import { Utils } from './Utils';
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/core/src/utils/ConfigurationLoader.ts [hunk 1]
```diff
@@ -77,31 +76,22 @@ export class ConfigurationLoader {
...
-    const tsConfigExtendsPath = isAbsolute(tsConfig.extends) ? tsConfig.extends : join(tsConfigFolderPath, tsConfig.extends);
-    const baseTsConfig = await ConfigurationLoader.getTsConfig(tsConfigExtendsPath);
-
-    return Utils.merge(baseTsConfig, tsConfig);
+  static async getTsConfig(tsConfigPath: string): Promise<{
+    resultType: 'success';
+    absoluteBaseUrl: string;
+    paths: { [key: string]: string[] };
+  }> {
+    return Utils.requireFrom('tsconfig-paths', tsConfigPath).loadConfig(tsConfigPath);
```
B4nan, line 94: the thing is, `tsconfig-paths` might not be there, its optional dependency needed only if you are using paths mapping - but we need to load the tsconfig to find out whether there is a paths mapping. 

in other words, this code can fail as the dependency won't be there unless explicitly installed by the user. and I'd rather not add a 3rd party depedency for such thing. 

on the other hand, `tsconfig-paths` is using JSON5 behind the scenes, which i was also considering as it would solve the problems with trailing commas in tsconfig files.

packages/core/src/utils/ConfigurationLoader.ts
```diff
@@ -77,31 +76,22 @@ export class ConfigurationLoader {
...
       const tsConfig = await this.getTsConfig(tsConfigPath);
-      /* istanbul ignore next */
-      const paths = tsConfig.compilerOptions?.paths;
 
-      if (paths) {
+      if (Object.entries(tsConfig.paths).length > 0) {
         Utils.requireFrom('tsconfig-paths', tsConfigPath).register({
-          baseUrl: tsConfig.compilerOptions.baseUrl,
-          paths: tsConfig.compilerOptions.paths,
+          baseUrl: tsConfig.absoluteBaseUrl,
```
B4nan, line 82: really `absoluteBaseUrl`? it is not documented in TS docs, is it something paths specific?

B4nan (changes requested):

Langstra:
How about it this way? The `register` method returns the `tsconfig` which can be used to check for paths. Then the paths can be registered if they are there. This is even a lot easier than the first implementation where the tsconfig was loaded manually.

I'll have to fix the tests later, but wanted to leave this for feedback first.

B4nan:
Interesting! As long as it works :]

Langstra:
It does work, but I do not know how to mock the tsconfig-file. `jest.mock()` only mocks a module, but it does not load it is a module. I'd like some feedback on that or a suggestion on how to fix it.

B4nan:
No worries, I can look into that later myself, don't have much to advice without playing with it myself first :]

PR #1866, opened 2021-05-26
fix(schema): fix diffing mysql enum columns

dmtar:
In case of multiple schemas, by default the enum column definition was taken from the first schema in alphabetical order where the enum column may be in completely different state and causing non needed alters.

### DIFF
--- a/packages/mysql-base/src/MySqlSchemaHelper.ts
+++ b/packages/mysql-base/src/MySqlSchemaHelper.ts
@@ -91,7 +91,7 @@ export class MySqlSchemaHelper extends SchemaHelper {
 
   async getEnumDefinitions(connection: AbstractSqlConnection, tableName: string, schemaName?: string): Promise<Dictionary> {
     const sql =  `select column_name as column_name, column_type as column_type from information_schema.columns
-      where data_type = 'enum' and table_name = '${tableName}'`;
+      where table_schema = database() and data_type = 'enum' and table_name = '${tableName}'`;
     const enums = await connection.execute<any[]>(sql);
 
     return enums.reduce((o, item) => {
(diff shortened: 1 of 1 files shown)


packages/mysql-base/src/MySqlSchemaHelper.ts
```diff
@@ -91,7 +91,7 @@ export class MySqlSchemaHelper extends SchemaHelper {
 
   async getEnumDefinitions(connection: AbstractSqlConnection, tableName: string, schemaName?: string): Promise<Dictionary> {
     const sql =  `select column_name as column_name, column_type as column_type from information_schema.columns
-      where data_type = 'enum' and table_name = '${tableName}'`;
+      where table_schema = database() and data_type = 'enum' and table_name = '${tableName}'`;
```
B4nan, line 94: shouldn't we use the `schemaName` from parameter if its available (and fallback to `database()`)?

B4nan: agreed

B4nan (changes requested):
looks like we need more tests to get the coverage back to 100%

B4nan:
do you plan to finish this?

dmtar:
> do you plan to finish this?
Yes, was busy with lots of other stuff. I will try to do it this week

B4nan:
Closing this as I did the change in https://github.com/mikro-orm/mikro-orm/commit/ecf79adc9e3fc1b3361649e88c5cdac80adb1626 and the multi schema will be handled in #1424, there is much more to change to have it working with multiple schemas and that PR will cover it.

Anyway, thanks for pointing this out!

PR #2043, opened 2021-07-19
docs: fix typo

janacm:


B4nan:
please fix this also in next version (`./docs/docs`), this change would be forgotten when next version will be released

PR #2045, opened 2021-07-19
Typo

janacm:


B4nan:
same as with the other PR, your change would affect only the 4.5 docs and would be forgotten when new docs are released. please fix it in the `docs/docs` too

PR #1964, opened 2021-06-25
Fix: Read replicas connection settings replaced by master settings

rbrishabh:
fixes #1963 

Hey @B4nan! How does this look?
let me know if this needs any changes!
Thanks

Issue #1963, opened 2021-06-25
Read replicas connection settings replaced by "master" settings


sbesedkov:
**Describe the bug**
Read replicas connection settings replaced by "master" settings, when connection settings specified as object.

**To Reproduce**
You may use example provided in the manual: https://mikro-orm.io/docs/read-connections
In this case replicas host and user parameter values will be replaced by top-level (master) configuration parameter values.

This issue occurs because `Connection` class instance `options` property value is replaced by the corresponding value from top-level config in `getConnectionOptions()`.

It works as expected when replica and master connection settings are specified in the `clientUrl` parameter.

```
  getConnectionOptions(): ConnectionConfig {
    const ret: ConnectionConfig = {};
    const url = new URL(this.options!.clientUrl || this.config.getClientUrl());
    this.options!.host = ret.host = this.config.get('host', url.hostname);
    this.options!.port = ret.port = this.config.get('port', +url.port);
    this.options!.user = ret.user = this.config.get('user', url.username);
    this.options!.password = ret.password = this.config.get('password', url.password);
    this.options!.dbName = ret.database = this.config.get('dbName', url.pathname.replace(/^\//, ''));

    return ret;
  }
```
See:
https://github.com/mikro-orm/mikro-orm/blob/091e60d5636959477176d851ef112ab53f1c6609/packages/core/src/connections/Connection.ts#L64

**Versions**

| Dependency | Version |
| - | - |
| node | 14.17.0 |
| typescript | 4.1.2 |
| mikro-orm | 4.5.5 |
| your-driver | postgresql |

B4nan:
So it should be enough to prefer existing values in `this.options`, right? 

Feel free to send a PR if that helps.

rbrishabh:
I can help with this!

erik-induro:
@B4nan would I be able to get a dev release for this? I'm dead in the water on performance testing since I can't scale horizontally.

PR #2081, opened 2021-07-31
Fix: Read replicas connection settings replaced by master settings (continuation)

ShinteiMai:
Based on the [previous pull request](https://github.com/mikro-orm/mikro-orm/pull/1964) for fixing #1963

This PR adds the unit test to check if MikroORM is initialized with replicas, when getting the connection with the params of `read`, it should return the configuration with the replicas db configuration.

### DIFF
--- a/packages/core/src/connections/Connection.ts
+++ b/packages/core/src/connections/Connection.ts
@@ -64,11 +64,11 @@ export abstract class Connection {
   getConnectionOptions(): ConnectionConfig {
     const ret: ConnectionConfig = {};
     const url = new URL(this.options!.clientUrl || this.config.getClientUrl());
-    this.options!.host = ret.host = (this.options!.host || this.config.get('host', url.hostname));
-    this.options!.port = ret.port = (this.options!.port || this.config.get('port', +url.port));
-    this.options!.user = ret.user = (this.options!.user || this.config.get('user', url.username));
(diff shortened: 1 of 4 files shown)


B4nan:
Are the tests passing for you locally? Because I see you did the very same change as in the old PR but many tests were failing.

ShinteiMai:
@B4nan Sometimes i do get an error like `database_name of x is not found` when running the changes. So, I thought that those tests were failing because of my testing setup or something else. I also tried testing only on the unit tests that I made, and they worked perfectly fine.

Do you have any hints regarding why those errors are showing? because I also tried running the tests in the master branch, and all the tests passed.

B4nan:
Well, CI tests are not passing, and they were not passing with the previous PR too, so this will be most probably something wrong with the change you made in code. I don't have any hints, the tests should create databases automatically.

Or it might as well be some other part of code that just relied on the old (and buggy) behaviour.

ShinteiMai:
> Well, CI tests are not passing, and they were not passing with the previous PR too, so this will be most probably something wrong with the change you made in code. I don't have any hints, the tests should create databases automatically.
> 
> Or it might as well be some other part of code that just relied on the old (and buggy) behaviour.

I will investigate further on what's causing this and find a fix

B4nan:
Let me know if you struggle with that, I can take a look locally too. I'd expect it will be something in the `DatabaseDriver.createReplicas` method.

Issue #2074, opened 2021-07-28
PostgreSQL schemas for multitenancy


Uminily:
**Is your feature request related to a problem? Please describe.**
Hi, first of all, thank you very much for your huge work on this ORM. I am using it today in my libraries and I am very happy with it. However, I have a small problem with multi-tenancy via postgresql schemas. No problem to have something dynamic with FindOptions (on find methods), it works incredibly well, but when inserting/updating/deleting, it is not possible to specify a schema.

**Describe the solution you'd like**
I checked a little bit the code of your library, it would have been great if on `*flush()` methods could take options, notably `{schema: 'mytenant'}`, and `nativeInsert() / nativeUpdate() / nativeDelete()` on `AbstractSqlDriver`, test before `isPlatformSchemaAuthorized()` or something like that.

Maybe applying `withSchema(options.schema)`on this [line](https://github.com/mikro-orm/mikro-orm/blob/b9d181061b8b015f91960db4bf65e8a671caad76/packages/knex/src/AbstractSqlDriver.ts#L195).

I've already tried with a monkey patch (on my current projet node_modules folder... wooohooo) to specify a schema on queryBuilder at this [line](https://github.com/mikro-orm/mikro-orm/blob/b9d181061b8b015f91960db4bf65e8a671caad76/packages/knex/src/query/QueryBuilder.ts#L651), and update/insert/delete with schema work like a charm.

I wish I could implement this with a PR to help but I'm too afraid to break whole parts of the library. If you did it on the Findable part but not on the insert etc., there must be a reason. If you give me a little help, I'll take a look. By the way, I will sponsor you in any case for your excellent work. This lib is much better than TypeORM and I'm already promoting it to my friends.

**Describe alternatives you've considered**
I've looked into creating a MikroORM instance (with the `next` version) that allows you to take a schema in the initial configuration, then destroy the instance at the end of the request.  I have the feeling that this is very heavy and I'm afraid that the number of connections to the database will explode and that it will slow down my application.

Currently I'm thinking of adding isolation fields per client on each entity and using `@Filter()` and compound indexes but it makes my upstream validations more complex, plus I'm not safe from missing something in my hooks (etc.) that may impact the data of another client.

**Additional context**
Use on Postgresql and platforms that can support database schema, maybe dropping schemas on mongo ?

PR #2093, opened 2021-08-05
feat: support multitenancy insert/update/delete on native & from uow with flush options

Uminily:
Hello,

as discussed here #2074 I propose you a very basic PR for the moment (I'm really struggling with the tests on migrations in particular). I also know that other people have worked on PRs or that there are topics on multitenancy (I tried to read the whole thing before deciding to write an issue and a PR).

The idea behind what I did is to be able to specify at the time of a `nativeInsert/nativeUpdate` or via the EM on a `flush`, the schema on which the modifications will apply. This avoids having several open connections or several instances of mikro-orm. This allows, in the context of a multi-tenancy API, to be able to specify upstream or at the last moment on which schema the modifications will be applied. I tried to respect the logic of `FindOptions`.

With this PR, a single Mikro-ORM instance, we would be able to search in several schemas. I saw that anothers PRs #1001 #1424 were proposed, especially on the migration part which I didn't deal with at all, should I integrate it here?

I think there are some shortcomings in what I did, because I don't master Mikro-ORM at all, and I could see the huge work you did on this lib :clap:

Don't validate this PR, but could you help me and guide me on what I could do to reach a level where multitenancy would be a feature of Mikro-ORM? Sorry for the very raw nature of my questions and my way of doing things... but I need to move forward on these topics on my side. In any case, thank you very much for your time and energy on reading this PR and for your work on this lib.

Xavier (Uminily).

### DIFF
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -3,7 +3,7 @@ import { inspect } from 'util';
 import { Configuration, QueryHelper, TransactionContext, Utils } from './utils';
 import { AssignOptions, EntityAssigner, EntityFactory, EntityLoader, EntityLoaderOptions, EntityRepository, EntityValidator, IdentifiedReference, Reference } from './entity';
 import { UnitOfWork } from './unit-of-work';
-import { CountOptions, DeleteOptions, EntityManagerType, FindOneOptions, FindOneOrFailOptions, FindOptions, IDatabaseDriver, UpdateOptions } from './drivers';
+import { CountOptions, DeleteOptions, EntityManagerType, FindOneOptions, FindOneOrFailOptions, FindOptions, IDatabaseDriver, InsertOptions, UpdateOptions } from './drivers';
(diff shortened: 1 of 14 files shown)


B4nan:
I have no time to review this right now, but as said several times already on various places, I don't think `flush` should have any parameters, it just feels wrong. The entity should bare the schema name which was used to load it instead. It should be most probably also used in the identity map key, so it is possible to have multiple instances of the same entity (but in different schemas).

B4nan:
As said above, I'll be happy to merge the part with `schema` param in `native*` methods, could you open new PR with that, or remove the flush options from this one? We can then discuss how to deal with UoW separately - I believe the idea I noted above (with storing the schema in entity instance, as well as using it in the identity map key, is the way to go - I can elaborate on that if you'd be interested in implementing that).

dzcpy:
Another thing to consider:
How to deal with relations? Is it possible to specify the schema of the relation in entity metadata? If not specified, it will use the schema from the query.

Issue #1843, opened 2021-05-20
Issues with seeder: TypeError: seederClass is not a constructor


unematiii:
**Describe the bug**

I was trying out the seeder functionality, I and I couldn't get it working. Issues I encountered:

a) Seeder is unable to instantiate seeder class:

```
npx mikro-orm seeder:create DatabaseSeeder
npx mikro-orm seeder:run [--class=DatabaseSeeder]
```

I am unable to run the seed, regardless whether the class name is provider or not. I tried exporting class (including default export), seeder path is correct and resolved by `require`. Also, seed created with CLI always uses `DatabaseSeeder` as a class name (ignoring specified name) and exports nothing.

Stack trace:

```
TypeError: seederClass is not a constructor
```

b) Creating database schema also triggers seeder

Running ```mikro-orm schema:create --run``` also triggers seeder. 

**To Reproduce**
1. Create and run seeder using CLI
2. Create DB schema using ``mikro-orm schema:create --run```

**Expected behavior**
a) I should be able to run seeders using CLI. Class name specified on seeder creation should be respected.
b) Seeds should not be ran on schema creation unless explicitly specified.

**Additional context**
I was able to get it working with [following changes](https://github.com/unematiii/mikro-orm/pull/1/files).

**Versions**

| Dependency | Version |
| - | - |
| node | v14.16.0 |
| typescript | 4.2.4 |
| mikro-orm | 5.0.0-dev.135 |
| your-driver | postgresql |

B4nan:
cc @Langstra

Langstra:
Thanks for trying it out and debugging it. Fixed the bug as you suggested and improved the tests.

unematiii:
@Langstra Thank you!

monarchwadia:
Hi! Just ran across this issue in `5.0.0-dev.292`. Here's my package.json:

```
{
  "name": "my-app",
  "version": "0.1.0",
  "private": true,
  "scripts": {
    "dev": "next dev",
    "build": "next build",
    "start": "next start",
    "lint": "next lint",
    "mikro": "MIKRO_ORM_CLI=./lib/db/mikro-orm.cli.config.ts mikro-orm"
  },
  "dependencies": {
    "@mikro-orm/cli": "5.0.0-dev.292",
    "@mikro-orm/core": "5.0.0-dev.292",
    "@mikro-orm/postgresql": "5.0.0-dev.292",
[...]

PR #2140, opened 2021-08-23
refactor: make seed option optional

Langstra:
Seed option when executing migration:fresh defaulted to empty string
and so it always tried to seed after migrating.


Langstra:
Should fix this https://github.com/mikro-orm/mikro-orm/issues/1843#issuecomment-903175177

PR #2181, opened 2021-09-07
feat: throw error in ChangeSetComputer.computeChangeSet when no meta …

hcharley:
…is found

### DIFF
--- a/packages/core/src/unit-of-work/ChangeSetComputer.ts
+++ b/packages/core/src/unit-of-work/ChangeSetComputer.ts
@@ -19,8 +19,12 @@ export class ChangeSetComputer {
               private readonly platform: Platform,
               private readonly config: Configuration) { }
 
-  computeChangeSet<T extends AnyEntity<T>>(entity: T): ChangeSet<T> | null {
-    const meta = this.metadata.find(entity.constructor.name)!;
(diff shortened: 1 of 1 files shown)


B4nan:
What is the motivation for this? The validation is done on higher level already, so this would make sense only if used directly (and the API is rather internal).

hcharley:
@B4nan I'm not sure how I slipped past the initial validation, but perhaps I hit a bug when this line was executed:

```ts
if (meta.readonly) {
```

Meta was undefined, and so this PR would have helped me diagnose the problem and find the right model.

It might be some problem with how I wrote my code (maybe circular references, or the wrong `this`, or something), so perhaps this error wouldn't be useful to others.

B4nan:
Well, would be good if you first share the stack trace so we can see how the code path was triggered. 

There are two methods to obtain the metadata, `find()` that does not validate the result (and is fast), and `get()` that will throw if it does not exist (and is therefore slower). I am using `find()` where I can to make things a tiny bit faster, but maybe it's not really relevant, will try to measure it here. 

But to me it feels a bit wrong to have it here (this place is triggered for every single managed entity instance, while we handle entities by groups). Do you at least remember what you were doing, was it a regular `flush` call that triggered this, or do you use flush events? 

Either way, if we are merge this, it should be just a one liner changing `find()` to `get()`. Also note that this project is not using `public` modifier` on purpose.

PR #2282, opened 2021-10-17
docs: fix logo size

Chantelsky:
Just a suggestion, but removing the svg links would make the navbar align better. I've just removed them in the browser to show an example.

image


B4nan:
Thanks! Yeah they shouldn't be there, they appeared with some docusaurus update.

Chantelsky:
@B4nan If I have time after work, i'm happy to fix that up for you :)

PR #2296, opened 2021-10-24
feat(core): add support for multiple schemas (including UoW)

### LINKED ISSUE #2074: PostgreSQL schemas for multitenancy

B4nan:
Entity instances now hold schema name (as part of `WrappedEntity`). Managed entities will have the schema from `FindOptions` or metadata. Methods that create new entity instances like `em.create()` or `em.getReference()` now have an options parameter to allow setting the schema. We can also use `wrap(entity).setSchema()`.

Entities can now specify `@Entity({ schema: '*' })`, that way they will be ignored in `SchemaGenerator` unless `schema` option is specified.
    
- if we have schema specified on entity level, it only exists in that schema
- if we have * schema on entity, it can exist in any schema, always controlled by the parameter
- no schema on entity - default schema or from global orm config

Closes #2074

BREAKING CHANGE:
`em.getReference()` now has options parameter.

### DIFF
--- a/docs/docs/entity-manager-api.md
+++ b/docs/docs/entity-manager-api.md
@@ -203,9 +203,9 @@ Shortcut for `wrap(entity).assign(data, { em })`
 
 ----
 
-#### `getReference(entityName: EntityName<T>, id: Primary<T>, wrapped?: boolean, convertCustomTypes?: boolean): T | Reference<T>`
+#### `getReference(entityName: EntityName<T>, id: Primary<T>, options?: GetReferenceOptions): T | Reference<T>`
 
-Gets a reference to the entity identified by the given type and identifier without actually loading it, if the entity is not yet loaded
+Gets a reference to the entity identified by the given type and identifier without actually loading it, if the entity is not yet loaded.
 
 ----
 
--- a/docs/docs/entity-references.md
+++ b/docs/docs/entity-references.md
@@ -170,13 +170,13 @@ author!: IdentifiedReference<Author>;
 
 When you define the property as `Reference` wrapper, you will need to assign the `Reference`
 to it instead of the entity. You can create it via `Reference.create()` factory, or use `wrapped`
-parameter of `em.getReference()`:
+option of `em.getReference()`:
 
 ```typescript
 const book = await orm.em.findOne(Book, 1);
 const repo = orm.em.getRepository(Author);
 
-book.author = repo.getReference(2, true);
+book.author = repo.getReference(2, { wrapped: true });
 
 // same as:
 book.author = Reference.create(repo.getReference(2));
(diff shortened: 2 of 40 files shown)


Uminily:
Thanks a lot for this feature!

Langstra:
> BREAKING CHANGE:
> `em.getReference()` now has options parameter.

Should this breaking change also be described in the `Upgrading from v4 to v5` docs?

B4nan:
It is there https://github.com/mikro-orm/mikro-orm/pull/2296/files#diff-828c0a2eecb6c19bee076c97a68e6251de3f61a41a5f14b042101eae7947c777R29

Langstra:
Oh yes, I must have missed that, sorry. Good job!

PR #2328, opened 2021-11-01
refactor: fix type of `Collection.init`

evantrimboli:
BREAKING CHANGE:
`Collection.init()` now has only options parameter

### DIFF
--- a/packages/core/src/entity/Collection.ts
+++ b/packages/core/src/entity/Collection.ts
@@ -206,11 +206,11 @@ export class Collection<T, O = unknown> extends ArrayCollection<T, O> {
     this.dirty = dirty;
   }
 
-  async init<P extends string = never>(options?: InitOptions<T, P>): Promise<this> {
+  async init<P extends string = never>(options: InitOptions<T, P> = {}): Promise<this> {
     const em = this.getEntityManager();
 
     if (!this.initialized && this.property.reference === ReferenceType.MANY_TO_MANY && em.getPlatform().usesPivotTable()) {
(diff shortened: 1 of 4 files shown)


packages/core/src/entity/Collection.ts
```diff
@@ -206,14 +206,11 @@ export class Collection<T, O = unknown> extends ArrayCollection<T, O> {
     this.dirty = dirty;
   }
 
-  async init(options?: InitOptions<T>): Promise<this>;
-  async init(populate?: string[], where?: FilterQuery<T>, orderBy?: QueryOrderMap<T>): Promise<this>;
-  async init(populate: string[] | InitOptions<T> = [], where?: FilterQuery<T>, orderBy?: QueryOrderMap<T>): Promise<this> {
-    const options = Utils.isObject<InitOptions<T>>(populate) ? populate : { populate, where, orderBy };
+  async init<P extends string = never>(options?: InitOptions<T, P>): Promise<this> {
```
B4nan, line 209: let's add a default value here so we don't need to use optional chaining everywhere (unfortunately it breaks coverage)
```suggestion
  async init<P extends string = never>(options: InitOptions<T, P> = {}): Promise<this> {
```

(and please remove the optional chaining usage underneath)

B4nan (changes requested):
looking good, just one nitpick

also please add this method to the list of breaking changes (there is a list of method signature changes, so just add it there)

https://github.com/mikro-orm/mikro-orm/blob/master/docs/docs/upgrading-v4-to-v5.md#options-parameters

B4nan (approved):

Issue #2343, opened 2021-11-04
Enum foreign key generates invalid check constraint


evantrimboli:
**Describe the bug**
When referring to a foreign key that is an enum, the generated SQL refers to the wrong column. 

Happy to look into this if I can be pointed in the right direction.

**Stack trace**
N/a

**To Reproduce**
Using the following entity definitions:

```
export enum BrandType { Google = 'google', Microsoft = 'microsoft', Mozilla = 'mozilla' }

@Entity()
export class Brand {
  @Enum({ primary: true })
  id!: BrandType;

  @Property()
  label!: string;
}

@Entity()
export class Product {
  @PrimaryKey()
  id!: number;

  @Property()
  label!: string;

  @ManyToOne({ onDelete: 'cascade' })
  brand!: Brand;
}
```

The generated SQL it produces is:
```
this.addSql('create table "brand" ("id" text check ("id" in (\'google\', \'microsoft\', \'mozilla\')) not null, "label" varchar(255) not null);');
this.addSql('alter table "brand" add constraint "brand_pkey" primary key ("id");');

this.addSql('create table "product" ("id" serial primary key, "label" varchar(255) not null, "brand_id" text check ("id" in (\'google\', \'microsoft\', \'mozilla\')) not null);');

this.addSql('alter table "product" add constraint "product_brand_id_foreign" foreign key ("brand_id") references "brand" ("id") on update cascade on delete cascade;');
```

**Expected behavior**
In the product table, the check constraint should refer to `brand_id`, not `id`.

**Additional context**
In 4.x, the constraint would not be placed on the foreign key field.

During metadata discovery, it seems like the `columnTypes` are being generated (copied perhaps?) incorrectly:

```
{
  name: 'brand',
  reference: 'm:1',
  cascade: [ 'persist' ],
  onDelete: 'cascade',
  type: 'Brand',
  nullable: false,
  fieldNames: [ 'brand_id' ],
  unsigned: false,
  length: undefined,
  precision: undefined,
  scale: undefined,
  columnTypes: [ `text check ("id" in ('google', 'microsoft', 'mozilla'))` ],
  customType: undefined,
  referencedTableName: 'brand',
  joinColumns: [ 'brand_id' ],
  referencedColumnNames: [ 'id' ]
}
```

**Versions**

| Dependency | Version |
| - | - |
| node | v14.15.4 |
| typescript | 4.4.0 |
| mikro-orm | 5.latest |
| your-driver | pg |

PR #2347, opened 2021-11-04
refactor: don't include enum check constraint in type

evantrimboli:
Resolves: #2343

I'm not super confident about this particular fix, however it does seem to revert the original behaviour back to how it worked in 4.x.

### DIFF
--- a/packages/knex/src/schema/SchemaComparator.ts
+++ b/packages/knex/src/schema/SchemaComparator.ts
@@ -5,8 +5,6 @@ import type { DatabaseSchema } from './DatabaseSchema';
 import type { DatabaseTable } from './DatabaseTable';
 import type { AbstractSqlPlatform } from '../AbstractSqlPlatform';
 
-const empty: readonly string[] = [];
-
 /**
  * Compares two Schemas and return an instance of SchemaDifference.
  */
@@ -375,7 +373,7 @@ export class SchemaComparator {
     return changedProperties;
   }
 
-  diffEnumItems(items1: readonly string[] = empty, items2: readonly string[] = empty): boolean {
+  diffEnumItems(items1: string[] = [], items2: string[] = []): boolean {
     return items1.length !== items2.length || items1.some((v, i) => v !== items2[i]);
   }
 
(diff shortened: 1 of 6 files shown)


packages/knex/src/schema/SchemaComparator.ts
```diff
@@ -366,9 +368,17 @@ export class SchemaComparator {
...
     }
 
+    if (this.diffEnumItems(column1.enumItems, column2.enumItems)) {
+      changedProperties.add('enumItems');
+    }
+
     return changedProperties;
   }
 
+  diffEnumItems(items1: readonly string[] = empty, items2: readonly string[] = empty): boolean {
```
B4nan, line 378: why do you bother with the `readonly` here? i kinda don't like that `empty` default value here, and the method is kinda internal, so we dont really need to care about this i think

B4nan (commented):
interesting, i dont really remember the particular reason for how it is done in v5, i guess it was to let the diffing work out of box instead of having dedicated `diffEnumItems` method. 

will need to spend some time testing this locally, but as the tests are passing, i guess i am fine with the change itself.

packages/knex/src/schema/SchemaComparator.ts
evantrimboli: - The `readonly` is there because we don't intend to modify it.
- The reason for using the default value is that makes the logic simpler, without a default it would need to be something like:

```
if (!items1 && !items2) {
  return false;
}

if (items1 && items2) {
  return items1.length !== items2.length .... etc
}
return true;

B4nan: > The readonly is there because we don't intend to modify it.

Yes, but as I just said, this is basically internal API, do don't need to care.

> The reason for using the default value is that makes the logic simpler, without a default it would need to be something like:

I have no issues with default value, what I don't like is that you need a separate variable to hold it.

evantrimboli: Fair enough. As a convention I've started using readonly in all cases to communicate intent. I'll make the relevant changes.

PR #2351, opened 2021-11-05
fix(query-builder): fix nested ordered pagination

### LINKED ISSUE #2095: OrderBy doesn't work properly using populate, paginate flag, filters and semi-complex hierarchies
**Describe the bug**

Using Postgresql:

Attempting to fetch an offset list of entities using `QueryFlag.PAGINATE`, `populate`, entity-level filters ordered by a field will return incorrectly-ordered results - e.g. results sorted by date will return the first 20 results ordered correctly, but the second 20 results will include results that should've appeared in the first set (in terms of order).

This is basically to be used by server-side limit/offset pagination into a data grid, so when somebody sorts the grid it should order the results correctly across pages. Currently, fetching the second page returns results that order-wise should've been in the first set of results.

Not using the `QueryFlag.PAGINATE` flag results in duplicate rows (and hence less than the full number once duplicates are removed), so the flag must be used.

**To Reproduce**

Reproduction in test form here: https://github.com/jamesmeneghello/mikro-orm/blob/bug/pagination-order/tests/issues/GH2095.test.ts

**Expected behavior**

Expected each page of results to be ordered correctly when retrieving them.

**Additional context**

This took a surprisingly long time to reproduce.

**Versions**

| Dependency | Version |
| - | - |
| node | 16.5 |

Other dependencies per mikro-orm.

jamesmeneghello:
Resolves #2095.

I have no idea if this is pure insanity or not, but it passes all tests (tests that were checking raw sql modified). The use of `min()` is required to sidestep group-by requirements.


packages/knex/src/query/QueryBuilder.ts
```diff
@@ -755,18 +755,32 @@ export class QueryBuilder<T extends AnyEntity<T> = AnyEntity> {
   }
 
   private wrapPaginateSubQuery(meta: EntityMetadata): void {
-    const pks = this.prepareFields(meta.primaryKeys, 'sub-query');
-    const subQuery = this.clone().limit(undefined).offset(undefined);
+    const pks = this.prepareFields(meta.primaryKeys, 'sub-query') as any;
```
jamesmeneghello, line 758: As much as I hate using `any` here, I'm not certain what the typing should be.

packages/knex/src/query/QueryBuilder.ts
```diff
@@ -755,18 +755,32 @@ export class QueryBuilder<T extends AnyEntity<T> = AnyEntity> {
...
+    const subQuery = this.clone().select(pks).groupBy(pks).limit(this._limit!);
+
+    if (this._offset) {
+      subQuery.offset(this._offset);
+    }
+
+    if (this._orderBy) {
```
jamesmeneghello, line 765: This section is required to make the query SQL/92 compatible, which is required for the default SQL strictness for both PG and MySQL.

B4nan:
Amazing, thank you! Will try to review it soon and ship it to 4.x branch once merged.

Btw I believe the paginate flag was also failing on mysql 8, so maybe those changes will help with that too.

jamesmeneghello:
Not sure why the tests are failing, they work locally. I'll take a look this weekend.

packages/knex/src/query/QueryBuilder.ts
```diff
@@ -755,18 +755,32 @@ export class QueryBuilder<T extends AnyEntity<T> = AnyEntity> {
...
     const knexQuery = subQuery.as(this.alias).clearSelect().select(pks);
 
     // 3 sub-queries are needed to get around mysql limitations with order by + limit + where in + group by (o.O)
     // https://stackoverflow.com/questions/17892762/mysql-this-version-of-mysql-doesnt-yet-support-limit-in-all-any-some-subqu
-    const subSubQuery = this.getKnex().select(pks).from(knexQuery).groupBy(pks).limit(this._limit!);
-
-    if (this._offset) {
-      subSubQuery.offset(this._offset);
-    }
+    const subSubQuery = this.getKnex().select(pks).from(knexQuery);
```
B4nan, line 783: Do we actually need this subquery, when it does not have groupby/limit/offset? Feels redundant this way.

tests/issues/GH2095.test.ts
```diff
@@ -0,0 +1,309 @@
+import { Collection, ConstraintViolationException, Entity, Enum, Filter, ManyToMany, ManyToOne, MikroORM, OneToMany, OneToOne, PrimaryKey, Property, QueryFlag, wrap } from '@mikro-orm/core';
+import { SchemaGenerator, SqliteDriver } from '@mikro-orm/sqlite';
+import { PostgreSqlDriver } from '../../packages/postgresql/src';
```
B4nan, line 3:
```suggestion
import { SchemaGenerator, PostgreSqlDriver } from '@mikro-orm/postgresql';
```

B4nan (changes requested):

packages/knex/src/query/QueryBuilder.ts
jamesmeneghello: Nope, totally redundant. Thanks!

packages/knex/src/query/QueryBuilder.ts
jamesmeneghello: Yep, that was it - the original branch was off 4.5.9 iirc. Fixed now.

packages/knex/src/query/QueryBuilder.ts
```diff
@@ -791,11 +797,7 @@ export class QueryBuilder<T extends AnyEntity<T> = AnyEntity> {
     subQuery.finalized = true;
     const knexQuery = subQuery.as(this.alias).clearSelect().select(pks);
 
-    // 3 sub-queries are needed to get around mysql limitations with order by + limit + where in + group by (o.O)
-    // https://stackoverflow.com/questions/17892762/mysql-this-version-of-mysql-doesnt-yet-support-limit-in-all-any-some-subqu
```
B4nan, line 777: lets keep this comment, we still use 3 subqueries because of this

packages/knex/src/query/QueryBuilder.ts
```diff
@@ -791,11 +797,7 @@ export class QueryBuilder<T extends AnyEntity<T> = AnyEntity> {
...
-    const subSubQuery = this.getKnex().select(pks).from(knexQuery);
-
-    const subSubSubQuery = this.getKnex().select(pks).from(subSubQuery.as(this.alias));
+    const subSubSubQuery = this.getKnex().select(pks).from(knexQuery);
```
B4nan, line 800: also the variable name can stay as `subSubQuery`, right?

packages/knex/src/query/QueryBuilder.ts
B4nan: what about just `as string[]` or `as unknown as string[]`?

packages/knex/src/query/QueryBuilder.ts
jamesmeneghello: 2 subqueries now, I think?

packages/knex/src/query/QueryBuilder.ts
jamesmeneghello: `string[]` works, thanks.

packages/knex/src/query/QueryBuilder.ts
jamesmeneghello: Sure can!

packages/knex/src/query/QueryBuilder.ts
B4nan: oh right, i though you added extra one and now you removed it, but i see its the full diff of the PR and we now have one less query in there, nice!

jamesmeneghello:
One last change, just removing the extra logic around pks for SQL/92, as they never match anyway (`e0.id` vs `id`), which is what CodeCov caught.

B4nan (approved):

PR #2402, opened 2021-11-11
fix(knex): quote version column

### LINKED ISSUE #2401: Auto-incrementing version feature does not escape field names
```
InvalidFieldNameException: update "table" set "foo" = 'bar', "versionId" = versionId + 1 where "id" = '2' - column "versionid" does not exist
```

unfortunately, postgres is case sensitive :(

parisholley:
Closes #2401


B4nan:
Thanks!

Issue #2364, opened 2021-11-06
Undetected Self Referencing Discriminators 


parisholley:
Given the following mapping:

```typescript
@Entity({ tableName: 'base', discriminatorColumn: 'type', abstract: true })
export default abstract class BaseMikro {
  @ManyToOne(() => BaseMikro, { fieldName: 'parentId', wrappedReference: true })
  parent: IdentifiedReference<ActivityMikro<any>>;
}
```

This is the compiled output function

```
  if (data.parent === null) {
    entity.parent = null;
  } else if (typeof data.parent !== 'undefined') {
    if (isPrimaryKey(data.parent, true)) {
      entity.parent = new Reference(factory.createReference('BaseMikro', data.parent, { merge: true, convertCustomTypes }));
    } else if (data.parent && typeof data.parent === 'object') {
      entity.parent = new Reference(factory.create('BaseMikro', data.parent, { initialized: true, merge: true, newEntity, convertCustomTypes }));
    }
  }
```

```Given the following data```

1: parent = null, type = 'Impl1'
2: parent = 1, type = 'Impl2'

When I run a query like the one below:

```typescript
const implementations = em.find(BaseMikro, {});

assert(implementations[0] instanceof Impl1); // pass
assert(implementations[1] instanceof Impl2); // fail (is BaseMikro)
```

The generated code above doesn't seem to understand it is a discriminated class.

**Versions**

```
    "@mikro-orm/cli": "^4.5.9",
    "@mikro-orm/core": "^4.5.9",
    "@mikro-orm/migrations": "^4.5.9",
    "@mikro-orm/nestjs": "^4.3.1",
    "@mikro-orm/postgresql": "^4.5.9",
    "@mikro-orm/reflection": "^4.5.9",
```

B4nan:
Please provide full reproduction for this, this is not enough to reproduce. For example [this test](https://github.com/mikro-orm/mikro-orm/blob/master/tests/features/single-table-inheritance.mysql.test.ts#L87) basicaly checks the same.

PR #2404, opened 2021-11-11
fix: reproduction test case for (#2364)

parisholley:
this reproduction is interesting as it only happens when you write a test case that effectively simulates multiple requests. i suspect there could be more undetected issues due to tests not adequate adding this level of variation. we may want to look into a test harness which can run the same test logic in multiple scenarios (transactions, forks, clear vs uncleared, etc)

### DIFF
--- a/tests/issues/GH2364.test.ts
+++ b/tests/issues/GH2364.test.ts
@@ -1,0 +1,88 @@
+import {
+  BigIntType,
+  Entity, Enum,
+  IdentifiedReference,
+  ManyToOne,
+  MikroORM,
+  PrimaryKey, QueryOrder,
+} from '@mikro-orm/core';
+import type { SqliteDriver } from '@mikro-orm/sqlite';
+
+export enum EntityType {
+  TYPE_1 = 'type_1',
+  TYPE_2 = 'type_2'
+}
+
+
+@Entity({ discriminatorColumn: 'type', abstract: true })
+abstract class BaseEntity {
+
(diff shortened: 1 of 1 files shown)


parisholley:
@B4nan this is a tough one

https://github.com/mikro-orm/mikro-orm/blob/d64d100b0ef6fd3335d234aeac1ffa9b34b8f7ea/packages/core/src/entity/EntityFactory.ts#L36-L51

the issue is that because it is an identity reference, we only have the id of the entity, not the discriminator value. a workaround for now, would be to eagerly join it vs a reference column.

as far as a solution, only thing i can think of is we have to make sure that the references are unmanaged or otherwise not kept in the identity map, so that the next time it is requested, it can be resolved correctly. or perhaps we mark the reference so that when we pull it out the map, we know to re-hydrate it

B4nan:
I see, that is indeed limitation of current implementation, I believe it's even documented in the STI section. 

The problem is that to support select-in population we need to have it in the identity map. And once it's there, we can't change the prototype dynamically, so we could only create new instance and replace it. Joining the relation and getting just the discriminator could be viable, we do similar hack with 1:1 inverse side relations, where we automatically join the owner to get its PK. 

https://github.com/mikro-orm/mikro-orm/blob/master/packages/knex/src/AbstractSqlDriver.ts#L527
https://github.com/mikro-orm/mikro-orm/blob/master/tests/QueryBuilder.test.ts#L211

B4nan:
Fixed via a0827f53c3

Issue #2406, opened 2021-11-11
Initialized connections when forceEntityConstructor = true


parisholley:
**Describe the bug**
This is either a documentation issue, or library issue. The docs say to initialize your model collections like so:

```typescript
procession = new Collection<JobProcessMikro>(this);
```

however the third parameter in that constructor is setting `Collection.initialized = true`, meaning any attempts to populate or `collection.loadItems()` will not do anything. When forceEntityConstructor is disabled, collections are added to the model with initialized = false

```
    "@mikro-orm/cli": "^4.5.9",
    "@mikro-orm/core": "^4.5.9",
    "@mikro-orm/migrations": "^4.5.9",
    "@mikro-orm/nestjs": "^4.3.1",
    "@mikro-orm/postgresql": "^4.5.9",
    "@mikro-orm/reflection": "^4.5.9",
```

B4nan:
Regular population probably works fine, right? Otherwise it would be quite weird noone would report this sooner. 

Anyway, definitely a bug.

parisholley:
populate doesn't work when doing em.findOne(Entity, id, {populate:...}); just runs a single query

parisholley:
believe i ran into the reason why it is true by default... if you create two new entities, and associate them, then run a populate (or loadItems) before a flush occurs, the collections will get reset since initialized = false, related to #2408

parisholley:
see my reproduction case here:

https://github.com/mikro-orm/mikro-orm/pull/2409

i tried to workaround by doing new Collection(this, undefined, false), but that opened up another can of worms

Issue #2408, opened 2021-11-11
Throw exception if adding item to unitialized collection


parisholley:
> When cascade persisting collections, keep in mind only fully initialized collections will be cascade persisted.

at the moment, the library will allow you to lose data without warning

```typescript
entity.collection.add(new Entity());

await entity.collection.loadItems();
```

It feels like this code should do one of two things:

a) if entity has been persisted, throw an exception when adding to an uninitialized collection
b) if loadItems() is invoked, persist any pending items in the collection

B4nan:
Adding to not initialized collection is a feature, so nothing i would like to explicitly disallow. 

Sounds like this would be covered with auto flush, or not? It uses em.find() under the hood too. 

I could even imagine that we just automatically merge the loaded items with those newly added. That way it would basically behave same way as if you swap those lines.

parisholley:
what is an example use case of adding to an unitialized collection for a persisted entity?

yep, could be covered by auto flush

merging could work too.

B4nan:
You dont need to load the items to add a new one, do you really want me to give you examples why thats good? :) collections can be huge. 

I would go with merging as autoflush could be disabled (or we might end up having it opt in due to performance penalties, lets see how it will behave when its there).

parisholley:
ah, good feature to have 👍

PR #2409, opened 2021-11-11
fix: reproduction (#2406)

parisholley:


parisholley:
https://github.com/mikro-orm/mikro-orm/blob/a5fb40f62cbddfd051aebf0c1f4b52e671e7d006/packages/core/src/entity/EntityFactory.ts#L142-L152

adding logic here to flip initialized = false, solves my issue, though it is protected and hard to say if this issue can occur other places

B4nan:
Yes, that is the right place, and it should be fired only if newEntity flag is false.

parisholley:
i'm also finding the same problem happens in reverse and is related to #2408

if you do new Entity(), the collections are initialed = true, if you do em.create() and DO NOT pass the collections as an empty array, initialized = false and similar problems occur

parisholley:
> Yes, that is the right place, and it should be fired only if newEntity flag is false.

you comfortable with a ts-ignore or is there a better way to do this?

parisholley:
https://github.com/mikro-orm/mikro-orm/blob/a5fb40f62cbddfd051aebf0c1f4b52e671e7d006/packages/core/src/EntityManager.ts#L513-L515

believe this PR would need the inverse here, loop through collections, always set initialized = true

B4nan:
> if you do new Entity(), the collections are initialed = true, if you do em.create() and DO NOT pass the collections as an empty array, initialized = false and similar problems occur

That needs to be addressed here, as with the flag the value is there:

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/hydration/ObjectHydrator.ts#L177

B4nan:
> you comfortable with a ts-ignore or is there a better way to do this?

Can't say I like it, but it's not a big deal.

B4nan:
Actually even better way to make the behaviour same is just deleting the collections and let the hydrator create new instances, when it's not a new entity. This way it will be closer to how it works without forcing the constructor usage.

Issue #2410, opened 2021-11-12
Cascading deletes break across transaction boundary


parisholley:
**Describe the bug**
If a new bidirectional relationship is flushed outside a transaction, then fetched inside a transaction, the cascading system doesn't issue delete commands properly if that relationship is destroyed.

Issue #2371, opened 2021-11-08
M:1 does not propagate to all matching 1:M


parisholley:
```typescript
@Entity({ tableName: 'vehicle', discriminatorColumn: 'type', abstract: true })
class Vehicle {
  @ManyToOne(() => Garage, { fieldName: 'garageId', wrappedReference: true })
  garage: IdentifiedReference<Garage>;
}

@Entity({ tableName: 'vehicle', discriminatorValue: 'car' })
class Car extends Vehicle {
}

@Entity({ tableName: 'vehicle', discriminatorValue: 'truck' })
class Truck extends Vehicle {
}

@Entity({ tableName: 'garage' })
class Garage {
	@OneToMany(() => Vehicle, (v) => v.garage)
	vehicles = new Collection<Vehicle>(this);

  	@OneToMany(() => Car, (v) => v.garage)
	cars = new Collection<Car>(this);
}

const garage = manager.create(Garage, {});
manager.persist(garage);

const car = manager.create(Car, { garage });
manager.persist(car);

assert(garage.cars.contains(car)); // pass
assert(garage.vehicles.contains(car)); // fail
```

Behavior is the same whether `abstract` is true or false on `Vehicle`. If i only have a single @OneToMany at a time, it will work either way just fine. Ultimately, I'll just refactor to not have conflicts, but figure it should either error if it detects overlapping mappings, or support propagating to all.

```
    "@mikro-orm/cli": "^4.5.9",
    "@mikro-orm/core": "^4.5.9",
    "@mikro-orm/migrations": "^4.5.9",
    "@mikro-orm/nestjs": "^4.3.1",
    "@mikro-orm/postgresql": "^4.5.9",
    "@mikro-orm/reflection": "^4.5.9",
```

lazharichir:
Got a similar issue when using `find` and `populate: true`, the `1:m` don't populate, but only the `1:1`.

PR #2411, opened 2021-11-12
fix: reproduction (#2410)

parisholley:


parisholley:
interesting to note about this test:

a) if you move the code out of the transaction and flush at the end, it passes without issue
b) if you leave the code in the transaction, but remove `member.users.removeAll();`, it runs as well

B4nan:
Would be good if you use sqlite for those reproductions that are not driver specific, so we can use in memory database that is much faster and does not require dropping/ensuring/etc.

This way it also passes just fine, which seems like how it would actually work in a real app:

```ts
    await orm.em.persistAndFlush(mu);
    orm.em.clear(); // <-- added clear to simulate new request

    await orm.em.transactional(async tx => {
      const member = await tx.findOne(MemberMikro, createdMember.id, {
        populate: ['ownedUsers', 'users'], // <-- populated users as the collection wouldn't be populated
      });

      if (member) {
        member.users.removeAll();
        tx.remove(member);
      }
    });
```

> b) if you leave the code in the transaction, but remove member.users.removeAll();, it runs as well

Yeah well, you are trying to be explicit about everything, and it can actually break things :D It also works fine if you first call `tx.remove(member)` and after that `member.users.removeAll()`. That `removeAll` call is kinda implicit due to the cascade delete enabled on the collection, maybe it's simply triggered twice if you do it this way.

The problem might be that the `tx.remove()` call cascades right ahead to the relation. Maybe we should let the cascading work only when flush is called, same way as persist works now.

parisholley:
> Would be good if you use sqlite for those reproductions that are not driver specific, so we can use in memory database that is much faster and does not require dropping/ensuring/etc

sure

> This way it also passes just fine, which seems like how it would actually work in a real app:

correct, this was an issue i produced in a test, the code under test would have been split into two requests. nonetheless, i think it is important to address test-only issues because it may reveal an underlying problem that could affect other code-paths

> It also works fine if you first call tx.remove(member) and after that member.users.removeAll()

i don't think anyone would intuitively do that given that if you were to write the SQL, you'd remove users first

> That removeAll call is kinda implicit due to the cascade delete enabled on the collection, maybe it's simply triggered twice if you do it this way

Yeah, this is what i've figured. I just removed the `removeAll` calls to work around

> The problem might be that the tx.remove() call cascades right ahead to the relation

I think this is fine, because you'd want those collections to propagate (i assume that logic is intertwined). perhaps a `removeAll` should mark the entity (`member` in this case) that the `users` have already been cascaded, and simply skip over that in its cascading.

B4nan:
The problem here is actually in the collection state, as its snapshot says it is empty, and that is the reason why the items are not cascaded. If you manally take the collection snapshot before opening the transaction, things work as expected. I believe this might be actually side effect of #2371 that is visible in the reproduction too on the `mu` entity. If I refactor to have a single `em.create()` and single `em.flush()` for the initial setup, the snapshot is taken properly:

```ts
    const mu = orm.em.create(MemberUser, {
      member: {},
      user: {},
    });
    const createdMember = mu.member.unwrap();
    await orm.em.persistAndFlush(mu);
```

PR #2417, opened 2021-11-14
refactor: respect `Loaded` type in `EntityDTO`

### LINKED ISSUE #2414: TypeScript type for EntityDTO with `ManyToOne` field is not correct
**Describe the bug**
When JSONify entities which has unpopulated ManyToOne fields, the return result has wrong TypeScript types.

**Stack trace**

**To Reproduce**
1. Given the following entities:
```ts
@Entity()
export class TeamMember extends BaseEntity<TeamMember, 'id'> {
  @PrimaryKey({ type: 'uuid', defaultRaw: 'uuid_generate_v4()' })
  id: string;

  @ManyToOne({ entity: () => Team, onDelete: 'CASCADE', onUpdateIntegrity: 'CASCADE' })
  team: IdentifiedReference<Team>;
}
```

```ts
@Entity()
export class Team extends BaseEntity<Team, 'id'> {
  @PrimaryKey({ type: 'uuid', defaultRaw: 'uuid_generate_v4()' })
  id: string;

  @OneToMany({ entity: () => TeamMember, mappedBy: (teamMember) => teamMember.team, orphanRemoval: true, cascade: [Cascade.ALL] })
  members = new Collection<TeamMember>(this);
}
```
2. Get the JSON verson of TeamMember record:
```ts
const teamMember = await orm.em.findOne(TeamMember, '{actual-team-member-id}');
const teamMemberJSON: EntityDTO<TeamMember> = teamMember.toJSON();
console.log(teamMemberJSON.team);
// 022b5aab-4a04-4c86-a9f0-e6c9748a7f74
```
`teamMemberJSON.team`'s type now should be `EntityDTO<Team>`, but actually, it's a string id. So it causes some type errors

**Expected behavior**
Either set unpopulatd ManyToOne fields of JSONified object to string, or something like `team: { id: '{actual-id}' }`;

**Additional context**

**Versions**

| Dependency | Version |
| - | - |
| node | 17 |
| typescript | 4.4.4 |
| mikro-orm | 5.0.0-dev.439 |
| your-driver | postgresql |

B4nan:
Closes #2414

### DIFF
--- a/packages/core/src/typings.ts
+++ b/packages/core/src/typings.ts
@@ -1,7 +1,7 @@
 import type { Transaction } from './connections';
 import type { Cascade, EventType, LoadStrategy, LockMode, QueryOrderMap } from './enums';
 import { ReferenceType } from './enums';
-import type { AssignOptions, Collection, EntityFactory, EntityIdentifier, EntityRepository, IdentifiedReference, Reference, SerializationContext } from './entity';
+import type { AssignOptions, BaseEntity, Collection, EntityFactory, EntityIdentifier, EntityRepository, IdentifiedReference, Reference, SerializationContext } from './entity';
 import type { EntitySchema, MetadataStorage } from './metadata';
 import type { Type } from './types';
 import type { Platform } from './platforms';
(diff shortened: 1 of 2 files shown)


B4nan:
@JanJakes could you take a look at this please? I managed to adjust the `Loaded` type so it is respected in `EntityDTO` (better say in the methods that return it, so `toObject()` etc), but now the `Loaded<T>` is no longer assignable to `T` when `T` is generic (it works fine with non-generic types). 

This problem in the codebase is workaroundable via type casting, user facing code (tests) are type checked correctly, so I guess it is good enough, but it you could find a better way, I am all ears.

PR #2444, opened 2021-11-21
feat(core): add support for custom property ordering

evantrimboli:
Related: https://github.com/mikro-orm/mikro-orm/discussions/2178

### DIFF
--- a/packages/core/src/platforms/Platform.ts
+++ b/packages/core/src/platforms/Platform.ts
@@ -381,4 +381,12 @@ export abstract class Platform {
     return;
   }
 
+  /**
+   * Generates a custom order by statement given a set of in order values, eg.
+   * ORDER BY (CASE WHEN priority = 'low' THEN 1 WHEN priority = 'medium' THEN 2 ELSE NULL END)
+   */
+  generateCustomOrder(escapedColumn: string, values: unknown[]) {
+    throw new Error('Not supported');
+  }
+
 }
--- a/packages/core/src/typings.ts
+++ b/packages/core/src/typings.ts
@@ -237,6 +237,7 @@ export interface EntityProperty<T extends AnyEntity<T> = any> {
   inversedBy: string;
   mappedBy: string;
   orderBy?: QueryOrderMap<T> | QueryOrderMap<T>[];
+  customOrder?: string[] | number[] | boolean[];
   fixedOrder?: boolean;
   fixedOrderColumn?: string;
   pivotTable: string;
--- a/packages/knex/src/AbstractSqlPlatform.ts
+++ b/packages/knex/src/AbstractSqlPlatform.ts
@@ -96,4 +96,13 @@ export abstract class AbstractSqlPlatform extends Platform {
     return false;
   }
 
+  /** @inheritDoc */
+  generateCustomOrder(escapedColumn: string, values: unknown[]): string {
+    let ret = '(case ';
+    values.forEach((v, i) => {
+      ret += `when ${escapedColumn} = ${this.quoteValue(v)} then ${i} `;
+    });
+    return ret + 'else null end)';
+  }
+
 }
--- a/packages/knex/src/query/QueryBuilderHelper.ts
+++ b/packages/knex/src/query/QueryBuilderHelper.ts
@@ -482,6 +482,7 @@ export class QueryBuilderHelper {
       // eslint-disable-next-line prefer-const
       let [alias, field] = this.splitField(k);
       alias = populate[alias] || alias;
+
       Utils.splitPrimaryKeys(field).forEach(f => {
         const prop = this.getProperty(f, alias);
         const noPrefix = (prop && prop.persist === false) || QueryBuilderHelper.isCustomExpression(f);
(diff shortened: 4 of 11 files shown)


evantrimboli:
@B4nan The test failures seem to do with the new entities, not sure where to correct that. Could you point me in the right direction?

packages/core/src/typings.ts
```diff
@@ -237,6 +238,7 @@ export interface EntityProperty<T extends AnyEntity<T> = any> {
   inversedBy: string;
   mappedBy: string;
   orderBy?: QueryOrderMap<T> | QueryOrderMap<T>[];
+  customOrder?: PropertyOptions<T>['customOrder'];
```
B4nan, line 241: maybe better to not import this at all and just inline the type (as the file where PropertyOptions is defined refereces typings.ts and vice versa)

tests/entities-sql/Task.ts
```diff
@@ -0,0 +1,27 @@
+import { Entity, PrimaryKey, Enum, ManyToOne, Property } from '@mikro-orm/core';
```
B4nan, line 1: instead of adding new entities here and dealing with all the snapshots, just create new test in features folder and define entities in the given file. that way there won't be any issues with the failed tests and the diff will be cleaner (no changed snapshots)

packages/knex/src/query/QueryBuilderHelper.ts
```diff
@@ -482,14 +482,23 @@ export class QueryBuilderHelper {
       // eslint-disable-next-line prefer-const
       let [alias, field] = this.splitField(k);
       alias = populate[alias] || alias;
+      const prop = this.getProperty(field, alias);
+      const customOrder = prop?.customOrder;
+
+      const genExpr = customOrder
```
B4nan, line 488: this looks weird, why dont you just inline it at the call site?

packages/knex/src/query/QueryBuilderHelper.ts
```diff
@@ -482,14 +482,23 @@ export class QueryBuilderHelper {
...
+        : null;
       Utils.splitPrimaryKeys(field).forEach(f => {
         const prop = this.getProperty(f, alias);
         const noPrefix = (prop && prop.persist === false) || QueryBuilderHelper.isCustomExpression(f);
         const column = this.mapper(noPrefix ? f : `${alias}.${f}`, type);
         /* istanbul ignore next */
         const rawColumn = Utils.isString(column) ? column.split('.').map(e => this.knex.ref(e)).join('.') : column;
 
-        ret.push(`${rawColumn} ${order.toLowerCase()}`);
+        const colPart = genExpr ? genExpr(rawColumn) : rawColumn;
```
B4nan, line 499: you can remove all from lines 485-491 and here do just

```ts
const prop = this.getProperty(field, alias);
const customOrder = prop?.customOrder;
const colPart = customOrder ? this.platform.generateCustomOrder(rawColumn, customOrder) : rawColumn;

ret.push(`${colPart} ${order.toLowerCase()}`);
```

B4nan (changes requested):
thanks!

this definitely needs more tests, not just asserting the QB output, but actually testing this works in all SQL drivers (for mongo we would have to use aggregations, so lets skip that now)

please create the tests as suggested in the `tests/features` folder and test it actually works in all of pg/mysql/sqlite

packages/core/src/platforms/Platform.ts
```diff
@@ -381,4 +381,10 @@ export abstract class Platform {
     return;
   }
 
+  generateCustomOrder(escapedColumn: string, values: unknown[]) {
+    // Generates a custom order by statement given a set of in order values, eg.
+    // ORDER BY (CASE WHEN priority = 'low' THEN 1 WHEN priority = 'medium' THEN 2) ... etc
```
B4nan, line 386: let's put this into jsdoc of the method

tests/features/custom-order/custom-order.mysql.test.ts
```diff
@@ -0,0 +1,274 @@
...
+  beforeEach(async () => {
+    await orm.em.nativeDelete(Task, {});
+    await orm.em.nativeDelete(User, {});
+  });
+  afterAll(async () => orm.close(true));
+
+  test('query string enum ASC', async () => {
+    const mock = jest.fn();
+    const logger = new Logger(mock, ['query']);
+    Object.assign(orm.config, { logger });
```
B4nan, line 111: after you rebase this with latest changes from master, you will need to change this in all the tests to use `mockLogger()` function from `bootstrap.ts`
```suggestion
    const mock = mockLogger(orm, ['query']);
```

(don't forget to remove the import for `Logger`)

B4nan (changes requested):
looking good, last two nitpicks!

B4nan:
Would be good to mention this in the docs somewhere, not just as a decorator option, but with an example. Maybe a recipe, or we could squeeze it on the entity manager page.

B4nan (approved):

B4nan:
good job, one last thing is the broken linter, then we are good to merge

evantrimboli:
The linting didn't pick up anything locally, does it not run for the test folder?

packages/knex/src/AbstractSqlPlatform.ts
```diff
@@ -96,4 +96,12 @@ export abstract class AbstractSqlPlatform extends Platform {
     return false;
   }
 
+  generateCustomOrder(escapedColumn: string, values: unknown[]): string {
```
B4nan, line 99:
```suggestion
  /** @inheritDoc */
  generateCustomOrder(escapedColumn: string, values: unknown[]): string {
```

packages/core/src/platforms/Platform.ts
```diff
@@ -381,4 +381,12 @@ export abstract class Platform {
     return;
   }
 
+  /**
+   * Generates a custom order by statement given a set of in order values, eg.
+   * ORDER BY (CASE WHEN priority = 'low' THEN 1 WHEN priority = 'medium' THEN 2 ELSE NULL END)
+   */
+  generateCustomOrder(escapedColumn: string, values: unknown[]) {
+    throw new Error('Not supported');
```
B4nan, line 389: this method is not covered by tests

B4nan:
thanks!

Issue #2455, opened 2021-11-23
Postgres: ASC_NULLS_FIRST is not translated to "ASC NULL FIRST" correctly 


woltob:
**Describe the bug**

This query should produce the following SQL:
```
em.find(
      Availability,
      { user: user },
      {
        orderBy: {
          startTime: 'ASC_NULLS_FIRST',
        },
      },
    );
```

```
select "e0".* from "availability" as "e0" where "e0"."user_id" = '188f50aa-428a-47a2-a389-677734e51f99' order by "e0"."start_time" asc nulls first 
```

However I am getting:

```
... "e0"."start_time" asc_nulls_first - syntax error at or near "asc_nulls_first"
```
The underscores in asc_nulls_first should not be there.

**Expected behavior**
See above.

**Additional context**
None.

**Versions**

| Dependency | Version |
| - | - |
| node | 16.10 |
| typescript | 4.4.2 |
    "@mikro-orm/cli": "^4.5.9",
    "@mikro-orm/core": "^4.5.9",
    "@mikro-orm/knex": "^4.5.9",
    "@mikro-orm/migrations": "^4.5.9",
    "@mikro-orm/nestjs": "^4.3.0",
    "@mikro-orm/postgresql": "^4.5.9",
    "@mikro-orm/reflection": "^4.5.9",
| your-driver | postgresql |

B4nan:
Use the enum instead of string, strings work only for `ASC/DESC/asc/desc`.

    orderBy: {
      startTime: QueryOrder.ASC_NULLS_FIRST,
    },

Or better say if you pass string, it needs to be that string you want to use (raw query fragment), not enum name, e.g. this works too:

    orderBy: {
      startTime: 'ASC NULLS FIRST', // or lower case, does not matter
    },

PR #2458, opened 2021-11-23
fix(core): normalize order keys

evantrimboli:
Fixes: https://github.com/mikro-orm/mikro-orm/issues/2455


B4nan:
Thanks, but this is not a bug, strings like that are not supported on purpose, it does not make sense to use them - either enum should be used, or the end query fragment we want, not something in the middle.

evantrimboli:
If that's the case, why does it explicitly support `keyof typeof QueryOrder` in the definition? It seems like it should either:
- Pick only those strings it can directly use as literals
- Have some code to transform them

PR #2477, opened 2021-11-29
docs(migrations): fix webpack context module api link

carmona44:
Couple of typos fixed:
- webpack extra 's' removed
- switched brackets from the link to the message and vice versa

### DIFF
--- a/docs/docs/migrations.md
+++ b/docs/docs/migrations.md
@@ -221,7 +221,7 @@ await MikroORM.init({
 });
 ```
 
-With the help of (webpacks context module api)[https://webpack.js.org/guides/dependency-management/#context-module-api]
+With the help of [webpack context module api](https://webpack.js.org/guides/dependency-management/#context-module-api)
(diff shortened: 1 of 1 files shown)


docs/docs/migrations.md
```diff
@@ -221,7 +221,7 @@ await MikroORM.init({
 });
 ```
 
-With the help of (webpacks context module api)[https://webpack.js.org/guides/dependency-management/#context-module-api]
+With the help of [webpack context module api](https://webpack.js.org/guides/dependency-management/#context-module-api)
```
B4nan, line 224: shouldn't this be `webpack's` instead?

carmona44: Sorry, yes of course

B4nan:
thanks! just so you know, the change was done only on next docs (so v5 basically), if you'd like to fix older versions too, it needs to be addressed in the `versioned_docs` folder

PR #2496, opened 2021-12-03
perf(core): reuse EntityComparator on fork()

fathyb:
I've noticed a memory leak on my server. After profiling it, I was able to see compiled functions being allocated and never released. It could be a bug in v8, but parsing and compiling code every request is expensive nonetheless.

This is caused by the `EntityManager` creating a new `EntityComparator` on each fork. Reusing `EntityComparator` shows a clear decrease in the `System Objects` report of the v8 profile.

Will post profiling screenshots later today.

A workaround I'm using in the meantime:
```ts
const comparator = orm.em.getComparator()
const em = orm.em.fork(true, true)
const patchEm: any = em

patchEm.comparator = comparator
patchEm.unitOfWork.comparator = comparator
patchEm.unitOfWork.changeSetComputer.comparator = comparator

OrmContext.set(em, { req, next })
```

### DIFF
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -28,7 +28,6 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
   private readonly validator = new EntityValidator(this.config.get('strict'));
   private readonly repositoryMap: Dictionary<EntityRepository<AnyEntity>> = {};
   private readonly entityLoader: EntityLoader = new EntityLoader(this);
-  private readonly comparator = new EntityComparator(this.metadata, this.driver.getPlatform());
   private readonly unitOfWork: UnitOfWork = new UnitOfWork(this);
   private readonly entityFactory: EntityFactory = new EntityFactory(this.unitOfWork, this);
   private readonly resultCache = this.config.getResultCacheAdapter();
@@ -43,7 +42,8 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
               private readonly driver: D,
               private readonly metadata: MetadataStorage,
               private readonly useContext = true,
-              private readonly eventManager = new EventManager(config.get('subscribers'))) { }
+              private readonly eventManager = new EventManager(config.get('subscribers')),
+              private readonly comparator = new EntityComparator(metadata, driver.getPlatform())) { }
 
   /**
    * Gets the Driver instance used by this EntityManager.
@@ -733,7 +733,7 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
     // we need to allow global context here as forking from global EM is fine
     const allowGlobalContext = this.config.get('allowGlobalContext');
     this.config.set('allowGlobalContext', true);
-    const em = new (this.constructor as typeof EntityManager)(this.config, this.driver, this.metadata, options.useContext, eventManager);
+    const em = new (this.constructor as typeof EntityManager)(this.config, this.driver, this.metadata, options.useContext, eventManager, this.comparator);
     this.config.set('allowGlobalContext', allowGlobalContext);
 
     em.filters = { ...this.filters };
--- a/packages/core/src/unit-of-work/ChangeSetComputer.ts
+++ b/packages/core/src/unit-of-work/ChangeSetComputer.ts
@@ -10,14 +10,14 @@ import { EntityComparator } from '../utils/EntityComparator';
 
 export class ChangeSetComputer {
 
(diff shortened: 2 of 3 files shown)


B4nan:
Thanks for reporting this! Agreed we should reuse the compiled functions, that kinda beats their purpose if we are creating new for each request.

B4nan:
Given that the comparator depends only on metadata storage and platform, we might wanna move it as a whole to the `Configuration` class (that kinda serves as internal DI container too), just like we do with hydrator.

packages/core/src/EntityManager.ts
```diff
@@ -43,7 +41,8 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
               private readonly driver: D,
               private readonly metadata: MetadataStorage,
               private readonly useContext = true,
-              private readonly eventManager = new EventManager(config.get('subscribers'))) { }
+              private readonly eventManager = new EventManager(config.get('subscribers')),
+              private readonly comparator = config.getComparator(metadata)) { }
```
B4nan, line 45: also does not need to be part of constructor

packages/core/src/entity/EntityHelper.ts
```diff
@@ -56,7 +56,7 @@ export class EntityHelper {
   }
 
   private static defineBaseProperties<T extends AnyEntity<T>>(meta: EntityMetadata<T>, prototype: T, em: EntityManager) {
-    const helperParams = meta.embeddable ? [] : [em.getComparator().getPkGetter(meta), em.getComparator().getPkSerializer(meta), em.getComparator().getPkGetterConverted(meta)];
+    const [getter, serializer, getterConverted] = [em.getComparator().getPkGetter(meta), em.getComparator().getPkSerializer(meta), em.getComparator().getPkGetterConverted(meta)];
```
B4nan, line 59: why are you changing this? is it related to the PR somehow?

embeddables do not have PKs, hence no getters are used for embeddables

packages/core/src/typings.ts
```diff
@@ -598,6 +599,20 @@ export interface HydratorConstructor {
   new (metadata: MetadataStorage, platform: Platform, config: Configuration): IHydrator;
 }
 
+export interface IComparator {
```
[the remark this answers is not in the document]
fathyb: Ah my bad, I thought I had to follow the same pattern as `Hydrator`. I think using `EntityComparator` directly is better and simpler.

Reverted!

packages/core/src/unit-of-work/ChangeSetComputer.ts
```diff
@@ -17,7 +17,7 @@ export class ChangeSetComputer {
               private readonly metadata: MetadataStorage,
               private readonly platform: Platform,
               private readonly config: Configuration,
-              private readonly comparator = new EntityComparator(metadata, platform)) { }
+              private readonly comparator = config.getComparator(metadata)) { }
```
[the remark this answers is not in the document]
fathyb: Reverted ✅

packages/core/src/EntityManager.ts
fathyb: Reverted!

packages/core/src/entity/EntityHelper.ts
fathyb: This was caused by the introduction of `ICompator`, TypeScript made the array non-constant somehow (`(e: T) => Primary<T> | string is not assignable to (e: T) => Primary<T>`).

Reverted as it's not needed now that `Configuration` uses `EntityComparator` instead of `IComparator`.

fathyb:
@B4nan thanks for the fast review! It's all addressed now.

B4nan:
Would be good if you sort out the broken tests locally.

packages/core/src/typings.ts
```diff
@@ -598,6 +598,10 @@ export interface HydratorConstructor {
   new (metadata: MetadataStorage, platform: Platform, config: Configuration): IHydrator;
 }
 
+export interface ComapratorConstructor {
```
B4nan, line 601: there is a typo. 

also note that if we are to keep it configurable, i would rather go with the previous approach (having separate interface) - or not have it configurable at all. this PR is currently somewhere in the middle.

fathyb: eh I just realized you meant just use `Configuration` for DI without making `comparator` configurable. that's much simpler. should be good now!

B4nan (approved):

B4nan:
cherry picked to 4.x branch too, so next dev build should include it too, will ship new stable later this week hopefully

thanks again!

PR #2552, opened 2021-12-18
refactor(migrations): prevent excessive extension stripping

MatthiasKunnen:
The original code stripped `test.migration.ts` to `test` instead of `test.migration`.
This resulted in migrations reverts (down) failing since the migration file `test.{ext}` does not exist.

Bug introduced in 40367166f9e74a042e2f6314f31877f27a15a14d.

### DIFF
--- a/packages/migrations/src/MigrationStorage.ts
+++ b/packages/migrations/src/MigrationStorage.ts
@@ -1,6 +1,7 @@
 import type { MigrationsOptions, Transaction } from '@mikro-orm/core';
 import { Utils } from '@mikro-orm/core';
 import type { AbstractSqlDriver, Table } from '@mikro-orm/knex';
+import * as path from 'path';
 import type { MigrationRow } from './typings';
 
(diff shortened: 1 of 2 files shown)


packages/migrations/src/MigrationStorage.ts
```diff
@@ -70,8 +71,13 @@ export class MigrationStorage {
   }
 
   protected getMigrationName(name: string) {
-    // strip extension
-    return name.replace(/\.\w+$/, '');
+    const parsedName = path.parse(name);
+    if (['.js', '.ts'].includes(parsedName.ext)) {
+      // strip extension
+      return parsedName.name;
+    }
```
MatthiasKunnen, line 78: An alternative should you want to avoid the usage of `path`.
```suggestion
    if (name.endsWith('.ts') || name.endsWith('.js')) {
      // strip extension
      return name.substring(0, name.length - 3);
    }
```

PR #2553, opened 2021-12-18
chore: fix commit message conventions link in CONTRIBUTING.md

MatthiasKunnen:


B4nan:
May I know where is this linked from (the broken link)? Have you verified it was not referenced from other places with the currently available id?

MatthiasKunnen:
It's referenced in the same file, here:
https://github.com/mikro-orm/mikro-orm/blob/227d5bce61e1e4259763fc849ab4c42200fd6bca/CONTRIBUTING.md#L64

B4nan:
Maybe it would be better to fix the link instead of changing the id?

MatthiasKunnen:
Will do. I just thought based on other links in the file and project that the anchor tag was the preferred approach.

B4nan:
Not sure, it probably does not matter much, but fixing links sounds like safer approach in general.

PR #2584, opened 2021-12-27
fix(postgres): fix runtime support for native pg enum arrays

### LINKED ISSUE #2583: Unable to unmarshall pg array
Seems like you forgot to implement unmarshallArray method for pg platform, so the [default split](https://github.com/silentroach/mikro-orm/blob/master/packages/core/src/platforms/Platform.ts#L259-L265) is used.

**Stack trace**
```
 ValidationError: Invalid enum array items provided in SomeEntity.values: [ '{first', 'second}' ]
    at EnumArrayType.convertToDatabaseValue (/code/node_modules/@mikro-orm/core/types/EnumArrayType.js:24:23)
    at /code/node_modules/@mikro-orm/core/hydration/ObjectHydrator.js:72:96
```

**To Reproduce**
```
enum SomeEnum {
  First = 'first',
  Second = 'second'
};

...

  @Enum({ items: () => SomeEnum, name: 'some', array: true, nullable: false })
  some!: SomeEnum[];
```

**Additional context**
PostgreSQL

**Versions**

| Dependency | Version |
| - | - |
| node | 16.13.1 |
| typescript | 4.5.4 |
| mikro-orm | 4.5.10 |
| your-driver | pg |

silentroach:
Adding support for pg native enum arrays, for example:

```ts
export enum WithEnumArrayValue {
    First = 'first',
    Second = 'second',
    Third = 'third'
}

@Entity()
export class WithEnumArray {
    @PrimaryKey()
    id!: number;

    @Enum({ name: 'values', nullable: false, items: () => WithEnumArrayValue, array: true })
    values!: WithEnumArrayValue[];
}
```

Looks like it works fine, but not sure it will work for all cases.

Closes #2583

### DIFF
--- a/packages/postgresql/src/PostgreSqlPlatform.ts
+++ b/packages/postgresql/src/PostgreSqlPlatform.ts
@@ -96,6 +96,14 @@ export class PostgreSqlPlatform extends AbstractSqlPlatform {
     return `{${values.join(',')}}`;
   }
 
+  unmarshallArray(value: string): string[] {
+    if (value === '{}') {
+      return [];
+    }
+
+    return value.substring(1, value.length - 1).split(',');
+  }
+
   getBlobDeclarationSQL(): string {
     return 'bytea';
   }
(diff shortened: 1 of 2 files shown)


B4nan:
Thanks! This will most probably require some changes in other tests (e.g. Entity generator), would be easier to add a separate test that does not alter the global bootstrap method that is reused in other tests.

silentroach:
moved to separate test, thanks

silentroach:
+ empty array case for codecov

B4nan:
chore: simplify test case

B4nan:
The test case could be greatly simplified and it also contains some code style issues, so I fixed that for you.

PR #2587, opened 2021-12-30
docs: add Ts.ED example usage

Romakita:


B4nan (changes requested):
few notes:
- missing link from sidebar
- please follow the code style this project uses (e.g. spaces in imports, trailing commas, single quotes ...)
- i would very much like to not suggest using `__dirname` in folder based discovery, it can easily break things
- i would rather have `orm` property than `connection` in that example, this "typeorm style" seems wrong, it is not a connection instance in the mikro orm context
- i dont like that you alias the `Property` decorator to `Column`, we are in the MikroORM docs, so please alias your property decorator and keep the ORM one as is to reduce confusion
- would be good to verify things work with v5, the release will come soon so if we should add things like this to the docs, they should all work with v5 already

Romakita:
Hello @B4nan 

thank for your review, I'll take a time to discuss with the author of this plugin to get your feedback.

---

> i dont like that you alias the Property decorator to Column, we are in the MikroORM docs, so please alias your property decorator and keep the ORM one as is to reduce confusion

I don't totally agree, because this decorator is also necessary to properly take advantage of the integration of our two frameworks. We are precisely on a page that talks about an integration between Ts.ED and mikro ORM.

It's as if on my documentation page, I presented our annotations without describing yours. (https://tsed.io/tutorials/mikroorm.html)

In our example, there is nothing confusing because the decorators are clearly imported via their original packages.

We can add split the example in two part, if it's fine for your:

> To begin, we need to define an Entity MikroORM like this:
```typescript
import {Entity, Property, PrimaryKey, Property as Column} from "@mikro-orm/core";

@Entity()
export class User {
  @PrimaryKey()
  id: number;
  
  @Column()
  firstName: string;
  
  @Column()
  lastName: string;
  
  @Column()
  age: number;
}
```

We can also add Ts.ED decorators in order to reuse the model at controller level (mapping, validation and generation of the OS3 Spec aka Swagger):

```typescript
import {Property, MaxLength, Required} from "@tsed/common";
import {Entity, Property, PrimaryKey, Property as Column} from "@mikro-orm/core";

@Entity()
export class User {
  @PrimaryKey()
  @Property()
  id: number;

  @Column()
  @MaxLength(100)
  @Required()
  firstName: string;

  @Column()
  @MaxLength(100)
  @Required()
  lastName: string;

  @Column()
  @Mininum(0)
  @Maximum(100)
  age: number;
}
```

---

> would be good to verify things work with v5, the release will come soon so if we should add things like this to the docs, they should all work with v5 already

I'm not a Mikro ORM user so I don't know what are the change about the v5. I let the plugin author to answer to this point. 
But I think it's not a blocker point to merge this PR (after fixing other points).

@derevnjuk I think the `@Connection` should be replaced by `@Orm` decorator. As @B4nan explained, it's not a connection in the sense that we mean it

I'll fix the other points :)

See you 
Romain

B4nan:
> I don't totally agree, because this decorator is also necessary to properly take advantage of the integration of our two frameworks. We are precisely on a page that talks about an integration between Ts.ED and mikro ORM.

I understand your point of view, but this is MikroORM docs we are talking about, so aliasing its decorators is a no-go for me - this would be the only place where we do such thing. My point is you should alias the Ts.ED decorator, as in the context of MikroORM docs, it is a 3rd party dependency. 

> In our example, there is nothing confusing because the decorators are clearly imported via their original packages.

If you are looking at MikroORM docs - where every single page except this one would use `@Property` decorator as something else - with different method signature and different options - to me that would be very confusing.

Romakita:
Ha ok you have a property decorator.
I didn't understand that you were following an identical decorator and why were talking about alias. and for information it’s not at all an alias to your decorator, its function is specific to Ts.ED to describe a jsonschema. ok now I get the point.

Required/MaxLength are a problem also?

B4nan:
>  and for information it’s not at all an alias to your decorator, its function is specific to Ts.ED to describe a jsonschema. ok now I get the point.

I know it is not alias for my decorator. I am talking about _aliasing_ my decorator when importing it. `import { Property as Column } from '@mikro-orm/core'` - that is what I don't like. You should alias your Ts.ED decorator this way as we are in the context of MikroORM docs here.

Romakita:
haaa ok, really sorry for this confusing things, I haven't written the original example, and I haven't seen that. 
My bad. I'm sorry for thank. Ok I fix the example.

Romakita:
Ok, I pushed all mandatory fixes:

- [x] missing link from sidebar
    please follow the code style this project uses (e.g. spaces in imports, trailing commas, single quotes ...)
- [x] i would very much like to not suggest using __dirname in folder based discovery, it can easily break things
- [x] i would rather have orm property than connection in that example, this "typeorm style" seems wrong, it is not a connection instance in the mikro orm context. (https://github.com/tsedio/tsed/pull/1742  is merged on our side)
- [x] i dont like that you alias the Property decorator to Column, we are in the MikroORM docs, so please alias your property decorator and keep the ORM one as is to reduce confusion
- [ ] would be good to verify things work with v5, the release will come soon so if we should add things like this to the docs, they should all work with v5 already

For the last point, there is migration note somewhere ? But regarding the Mikro ORM code used in this module, the risk seems to be limited. You can see the code here: https://github.com/tsedio/tsed/tree/production/packages/orm/mikro-orm/src/services . As I understand, the module create the Mikro ORM instance and provide a way to inject Orm/Em instance.

Tell me if you have more details for me or for @derevnjuk :)

See you
Romain

derevnjuk:
No issues have been encountered while testing this with v5.0.0-rc.0, especially when we speak about the main functionality. However, I have noticed you changed the `fork` method signature, which might lead to unexpected behavior while using the `@Transactional()` decorator. 

Before:

```ts
orm.em.fork(true, true);
```

After:
```ts
orm.em.fork({ clear: true, useContext: true });
```

I will try to figure out how to bypass this.

@Romakita @B4nan fyi

Romakita:
Hello @B4nan 
I hope your are fine. @derevnjuk and I, have done all requested changes on this PR :).

This PR will support mikro-orm v5 officially: https://github.com/tsedio/tsed/pull/1950

The documentation is up to date!

We should be good and waiting for your last ones if there are any :-)
See you
Romain

docs/docs/usage-with-tsed.md
```diff
@@ -0,0 +1,238 @@
+---
+title: Using MikroORM with Ts.ED framework
+sidebar_label: Usage with Ts.ED
+---
+
+## Installation
+
+Easiest way to integrate MikroORM to Ts.ED is via [`@tsed/mikro-orm` module](https://tsed.io/tutorials/mikroorm.html).
```
B4nan, line 8: there should be some clear note that there is possibly more up to date version of this recipe in the tsed docs

docs/docs/usage-with-tsed.md
```diff
@@ -0,0 +1,238 @@
...
+@Module({
+  imports: [MikroOrmModule],
+  mikroOrm: [
+    {
+      contextName: 'default',
+      type: 'postgres',
+      ...,
+
+      entities: [
+        './src/entity/*{.ts,.js}'
```
B4nan, line 44: trailing comma
```suggestion
        './src/entity/*{.ts,.js}',
```

docs/docs/usage-with-tsed.md
```diff
@@ -0,0 +1,238 @@
...
+}
+```
+
+## Use Entity with Controller
+
+To begin, we need to define an Entity MikroORM to define a model:
+
+```typescript
+import { Entity, Property, PrimaryKey, Property } from "@mikro-orm/core";
+import { Required } from '@tsed/schema'
```
B4nan, line 111:
```suggestion
import { Required } from '@tsed/schema';
```

docs/docs/usage-with-tsed.md
```diff
@@ -0,0 +1,238 @@
...
+  private readonly orm!: MikroORM;
+}
+```
+
+## Use Entity with Controller
+
+To begin, we need to define an Entity MikroORM to define a model:
+
+```typescript
+import { Entity, Property, PrimaryKey, Property } from "@mikro-orm/core";
```
B4nan, line 110:
```suggestion
import { Entity, Property, PrimaryKey, Property } from '@mikro-orm/core';
```

docs/docs/usage-with-tsed.md
```diff
@@ -0,0 +1,238 @@
...
+import { EntityManager } from '@mikro-orm/mysql'; // Import EntityManager from your driver package or `@mikro-orm/knex`
+
+@Injectable()
+export class UsersService {
+  @Em()
+  private readonly em!: EntityManager;
+
+  async create(user: User): Promise<User> {
+    await this.em.persistAndFlush(user);
+    console.log("Saved a new user with id: " + user.id);
```
B4nan, line 178:
```suggestion
    console.log('Saved a new user with id: ' + user.id);
```

docs/docs/usage-with-tsed.md
Romakita: Isn't necessary. The API have changed on our side. The latest changes is internal and your documentation point to the latest release. It shouldn't be a problem for you.

Romakita: @derevnjuk can you confirm this point ?

derevnjuk: @Romakita sure, you are right. However, I guess @B4nan  meant that we should point out somehow that source of truth is our docs.

Romakita: Ok. Do you have a formulation or an example of a sentence that would suit you in this case?

As I don't see anything equivalent on other pages such as Nest.js and Admin.js, I wouldn't want to put anything :)

Suggestion:

```
::: tip Note
The documentation presented on this page is taken from the Ts.ED site. The latest updates to this example can be found directly on the [`@tsed/mikro-orm` module](https://tsed.io/tutorials/mikroorm.html) page.
:::
```

B4nan: The nest.js guide has such link in the opposite direction, as the `@mikro-orm/nestjs` is our package. I'd imagine something like that. 

image

https://docs.nestjs.com/recipes/mikroorm

Romakita: done. Sorry for the very late answer @B4nan.

Romakita:
Hi @B4nan 

Sorry for the late answer. I fixed the latest comment, I hope it would be fine for your. Also the last commit added new documentation sections. We worked a lot on the MikroORM integration with Ts.ED to provide other features. Thanks to @derevnjuk for is great works on this module!.

I hope we will able to merge this PR ;)

See you!

B4nan:
Closing as stale. I wouldn't mind merging this to v7 docs, as long as v7 gets supported on your end, but let's create a new PR for that. I would also appreciate a smaller guide on our end, with a full featured docs on yours, so you can ensure they are up to date.

PR #2605, opened 2022-01-05
refactor(seeder): path like migrations

r3tard:
Hello

refactored seeder to be more consistent to migration package (`path` processing and default path): i think we shouldn't prepend `process.cwd()` to user provided path

deleted method `refreshDatabase`. `SchemaGenerator` exist only when user uses sql, but seeder for mongodb and custom drivers too, so user can explicit call `dropSchema` and `createSchema` methods on `SchemaGenerator`

BREAKING CHANGE:
- default seeders path: `./seeders` instead `./database/seeder` (like migrations `./migrations`). 
- removed `refreshDatabase` method on `SeedManager`

### DIFF
--- a/packages/mongodb/src/MongoDriver.ts
+++ b/packages/mongodb/src/MongoDriver.ts
@@ -122,6 +122,17 @@ export class MongoDriver extends DatabaseDriver<MongoConnection> {
     await this.rethrow(Promise.all(promises));
   }
 
+  async refreshCollections(options: RefreshCollectionsOptions = {}): Promise<void> {
+    options.ensureIndexes ??= true;
+
+    await this.dropCollections();
+    await this.createCollections();
+
+    if (options.ensureIndexes) {
+      await this.ensureIndexes();
+    }
+  }
+
   async ensureIndexes(): Promise<void> {
     await this.rethrow(this.createCollections());
     const promises: Promise<string>[] = [];
@@ -310,4 +321,9 @@ export class MongoDriver extends DatabaseDriver<MongoConnection> {
     return ret.length > 0 ? ret : undefined;
   }
 
+}
+
+export interface RefreshCollectionsOptions {
+  /** create indexes? defaults to true */
+  ensureIndexes?: boolean;
 }
--- a/docs/docs/seeding.md
+++ b/docs/docs/seeding.md
@@ -154,13 +154,13 @@ const books: Book[] = new BookFactory(orm.em).each(book => {
 }).make(5);
 ``` 
 
-## Default settings
+## Configuration
 The seeder has a few default settings that can be changed easily through the MikroORM config. Underneath you find the configuration options with their defaults.
 ```ts
 MikroORM.init({
     seeder: {
-        path: './database/seeders/', 
-        defaultSeeder: 'DatabaseSeeder'
+        path: './seeders', // path to the folder with seeders
+        defaultSeeder: 'DatabaseSeeder' // default seeder class name
     }
 });
 ```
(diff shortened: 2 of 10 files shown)


B4nan:
> i think we shouldn't prepend process.cwd() to user provided path

That's not how it works with migrations. You can provide absolute path, if you provide relative, `baseDir` is prepended, which defaults to `process.cwd()`. 

I also don't see a reason to remove the `refreshDatabase` method. But its location is indeed wrong, it should live on `SchemaGenerator`.

cc @Langstra

Langstra:
> > i think we shouldn't prepend process.cwd() to user provided path
> 
> That's not how it works with migrations. You can provide absolute path, if you provide relative, `baseDir` is prepended, which defaults to `process.cwd()`.
> 
> I also don't see a reason to remove the `refreshDatabase` method. But its location is indeed wrong, it should live on `SchemaGenerator`.
> 
> cc @Langstra

I'll also take some time tomorrow to review this.

Langstra:
I think most of these changes definitely are an improvement over te current code. Agree with B4nan that the `refreshDatabase` method is a useful one. 
I am unsure how the determination of the path works exactly with the migrations. The only thing I could find about providing a relative path is with the migration running itself, not with the generation of the migrations. So if you are okay with that, then I leave that up to you.

r3tard:
> Agree with B4nan that the `refreshDatabase` method is a useful one.

I added `refreshDatabase` method to `SchemaGenerator` and `refreshCollections` to `MongoDriver` (should it call `ensureIndexes` after `createCollections`?)

B4nan (changes requested):
> I added refreshDatabase method to SchemaGenerator and refreshCollections to MongoDriver (should it call ensureIndexes after createCollections?)

Let's add a parameter for that (let's use an options object right ahead, with a single property for now), I'd do it by default, but it should be configurable.

The changes are looking good to me, thanks a lot! 

I'd like to support relative paths as we do in migrations. The way it works there is basically this line, that normalizes the path to absolute, prepending the `baseDir` if the path is relative. 

https://github.com/mikro-orm/mikro-orm/blob/master/packages/migrations/src/Migrator.ts#L26

packages/mongodb/src/MongoDriver.ts
```diff
@@ -122,6 +122,17 @@ export class MongoDriver extends DatabaseDriver<MongoConnection> {
...
   }
 
+  async refreshCollections(options: RefreshCollectionsOptions = {}): Promise<void> {
+    options.ensureIndexes ??= true;
+
+    await this.rethrow(this.dropCollections());
+    await this.rethrow(this.createCollections());
+
+    if (options.ensureIndexes) {
+      await this.rethrow(this.ensureIndexes());
```
B4nan, line 132: no need to use `rethrow` from this method - all the methods it calls are using it internally
```suggestion
    await this.dropCollections();
    await this.createCollections();

    if (options.ensureIndexes) {
      await this.ensureIndexes();
```

B4nan (changes requested):
looking good, just one nitpick. if the test suite is green and we keep full branch coverage, we can merge

B4nan:
Something's wrong with the tests. They all passed, but did not finish the process. This was not happening before and is not happening in master.

r3tard:
> Something's wrong with the tests. They all passed, but did not finish the process. This was not happening before and is not happening in master.

Now it should finish correctly, my bad.

packages/mongodb/src/MongoDriver.ts
r3tard: removed `rethrow`

Issue #2598, opened 2022-01-03
Add enum support when generating entities from database schema


kevinh-canva:
**Is your feature request related to a problem? Please describe.**
I haven't tried with other databases, but for Postgres when generating entities `.ts` files from an existing database schema, the enum type is swallowed and the field has type `string` instead, so we lose enum type safety. We could manually add enum definition to the generated file, but the next time we need to re-generate the file due to a schema change, it will overwrite the enum definition.

**Describe the solution you'd like**
I will raise a PR to generate the enum definition in the same entity .ts file and use that for the field.

**Describe alternatives you've considered**
- Manually add enum definition to generated file -> risk the addition being overwritten the next time the file is re-generated

**Additional context**
@B4nan Would love to know if you have looked into this before and if it's feasible at all, since I'm not too familiar with knex which I guess is what we use to get the entity information? Would we be able to tell if a particular field is of type enum and if so what's the definition?

B4nan:
Are you talking about native pg enums? MikroORM does not use those, enums are represented via text fields with check constraints.

I am not opposed by introducing native pg enums support too, if you would contribut it, but it will be much more work than fixing the entity generator part. Note that from what I know, native pg enums should be supported on runtime - they are not supported in schema generator, so this would include:

- handling this new format when creating tables
- diffing of enums
- diffing of text format with check constraints vs native pg enums (in both way)
- all of this would have to work optionally (based on some flag on `@Enum()` decorator)
- enum arrays would have to address this somehow too i guess
- and then finally we can talk about entity generator :]

Overall quite a lot of work for rather small benefit, as using text fields with check constrains seems to be good enough solution. But I can understand that is not valid reason if you already have your existing database and you want to map it in the ORM.

> since I'm not too familiar with knex which I guess is what we use to get the entity information?

Most of the metadata is taken directly from information_schema, knex does not help much with advanced cases.

kevinh-canva:
> 

Yep, this is my first time contributing to an ORM tbh, so the above list seems a fair bit more than what a newbie like me would expect to be able to do :D 

I think it's beneficial to limit the scope of this issue. So right now when I generate the entities from schema using entity generator, even if we use check constraint, the `.ts` code property is generated with string type. So the main issue I want to fix is to generate this as an enum type. 

Let's say I have a `task` table, with `status` enum column on it with 3 values `OPEN, IN_PROGRESS, DONE`. It would be nice if the resulting entity file looks something like this:
```
@Entity()
[...]

PR #2608, opened 2022-01-06
feat(cli): add enum generation support

kevinh-canva:


B4nan:
no tests? :]

kevinh-canva:
> no tests? :]

haha sorry, it's still a draft version :P

B4nan:
So this PR will introduce support only for entity generator, or do you plan to include enum diffing in schema generator?

CLI is definitely not the right subsystem, CLI just works with other packages like entity generator and schema generator.

kevinh-canva:
> So this PR will introduce support only for entity generator, or do you plan to include enum diffing in schema generator?
> 
> CLI is definitely not the right subsystem, CLI just works with other packages like entity generator and schema generator.

Yep sorry about that, will update the commit to be more accurate. Yep for now this PR only contains entity generator-related changes to generate enum type definitions.

I'm still finding my way around the codebase though. Can you clarify what you mean by enum diffing?

B4nan:
If you change enum definition in the code and run `schema:update`, you need to compare what is in the information schema with the enum definition in code. That is enum diffing.

Btw no need to care about commit messages, I will be squash merging the PR, so changing the PR title is enough.

kevinh-canva:
Thanks for clarifying! Hmm if that's the case, wouldn't enum diffing already be supported? https://github.com/mikro-orm/mikro-orm/issues/2598 if you see my comment on this issue, I aim for this PR to mainly be about generating enum definition class in the `.ts` entity file. I might get rid of the changes to add pg native enum to keep things simple.

kevinh-canva:
@B4nan This PR is ready for review now, PTAL when you have time.

I got a problem with Code climate cognitive complexity though, any suggestion to get it resolved since I'm not too familiar with this tool?

B4nan:
> I got a problem with Code climate cognitive complexity though, any suggestion to get it resolved since I'm not too familiar with this tool?

Don't be too worried about that, it is quite strict.

But please add more tests so we have full (branch) coverage again.

kevinh-canva:
> > I got a problem with Code climate cognitive complexity though, any suggestion to get it resolved since I'm not too familiar with this tool?
> 
> Don't be too worried about that, it is quite strict.
> 
> But please add more tests so we have full (branch) coverage again.

I have modified the schema sql and snapshot which will resolve the coverage drop for `SourceFile.ts`. However for `EnumType.compareAsType()` I don't think it's really related to my change. My change just stops using `EnumType` and use a custom type for enum instead, so `EnumType.compareAsType()` is no longer called, that's probably why the coverage drop

B4nan:
> However for EnumType.compareAsType() I don't think it's really related to my change.

It is related to your change, as there is full branch coverage in master. And indeed it is caused by changes that made the method not needed/used/tested most probably.

It is not a "blocker", but if you won't do it, I will have to do it myself, so I would rather let you do that instead of adding more work to my plate :]

But we have a bigger problem - 12 tests are now failing. The fact that the schema diffing tests are failing kinda shows that this is now complete addition. We can't introduce support for enums that would make the schema diffing produce unwanted queries. Maybe this is caused by the parts of the PR you removed? I remember it was all green initially, but might be just because you haven't adjusted existing tests before. But that is definitely a blocker.

kevinh-canva:
> > However for EnumType.compareAsType() I don't think it's really related to my change.
> 
> It is related to your change, as there is full branch coverage in master. And indeed it is caused by changes that made the method not needed/used/tested most probably.
> 
> It is not a "blocker", but if you won't do it, I will have to do it myself, so I would rather let you do that instead of adding more work to my plate :]
> 
> But we have a bigger problem - 12 tests are now failing. The fact that the schema diffing tests are failing kinda shows that this is now complete addition. We can't introduce support for enums that would make the schema diffing produce unwanted queries. Maybe this is caused by the parts of the PR you removed? I remember it was all green initially, but might be just because you haven't adjusted existing tests before. But that is definitely a blocker.

Hmm, if you take a look at my change, you will see that the reason `EnumType.compareAsType()` no longer has full coverage is because it's no longer used by the enum generation process. For example, if you define an enum in your schema and generate the type, it use to map it to `EnumType` and call that method, but now since I will be generating a custom enum definition for that, it no longer picks it up. 
The fact that it has full coverage in master and not here mean that the _other parts_ of the codebase does not have enough extensive tests that cover the usage of this method, now that the EntityGenerator stops using it, that's why I say it's not related to my change. But I'm happy to spend some more time digging into other parts of the code that use this, it's just that it would take more time for me to do it than you for obvious reasons :D 

Yeah I just take a look at those failures, I believe they are all good, they are most probably because I update the SQL definition to include a default enum now which breaks the snapshot for the schema generator tests. I will fix them now!

packages/knex/src/schema/DatabaseTable.ts
```diff
@@ -278,6 +281,12 @@ export class DatabaseTable {
       const parts = fk.referencedTableName.split('.', 2);
       return namingStrategy.getClassName(parts.length > 1 ? parts[1] : parts[0], '_');
     }
+    // If this column is using an enum.
```
kevinh-canva, line 284: @B4nan so any enum would be handled by this if block instead. There's practically no way for me to trigger the else branch i.e `EnumType.compareAsType()` because if it's really an enum, it would not escape this `if` block.

B4nan:
> There's practically no way for me to trigger the else branch i.e EnumType.compareAsType() because if it's really an enum, it would not escape this if block.

Don't be too concerned with the way it was previously used and tested. `compareAsType` is a method needed for diffing the actual value, nothing schema related. It is used in `EntityComparator`. It sounds weird that it would not be used in that code, unless the `EnumType` would not be inferred - which would be wrong too.

I would expect this line to trigger it:

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/utils/EntityComparator.ts#L417

kevinh-canva:
@B4nan after further digging I think I know what the issue with the failing tests is. Looks like when converting db schema for a column like this
```
status text check (status IN ('OPEN', 'IN_PROGRESS', 'DONE')) NOT NULL DEFAULT 'OPEN',
```

B4nan:
Database defaults need to be defined in the decorator options, property initializer won't have any effect on the schema generator. Here I would say we want both the initializer and the `default` decorator option.

packages/knex/src/schema/SchemaComparator.ts
```diff
@@ -455,7 +455,10 @@ export class SchemaComparator {
     }
 
     if (from.default && to.default) {
-      return from.default.toString().toLowerCase() === to.default.toString().toLowerCase();
+      return from.default.toString().toLowerCase() === to.default.toString().toLowerCase() ||
```
B4nan, line 458: this is something to be taken care of elsewhere, if it is a string, it should be already quoted

tests/entities-sql/Publisher2.ts
```diff
@@ -35,10 +35,10 @@ export class Publisher2 extends BaseEntity2 {
   @ManyToMany({ entity: () => Test2, pivotTable: 'publisher2_tests', fixedOrder: true })
   tests!: Collection<Test2>;
 
-  @Enum(() => PublisherType)
+  @Enum({ items: () => PublisherType, default: 'local' })
```
B4nan, line 38: let's not change existing entity definition, this breaks other tests

tests/mysql-schema.sql
```diff
@@ -37,7 +37,7 @@ set foreign_key_checks = 0;
 
 create table `sandwich` (`id` int(10) unsigned not null auto_increment primary key, `name` varchar(255) not null, `price` int(11) not null) default character set utf8mb4 engine = InnoDB;
 
-create table `publisher2` (`id` int(10) unsigned not null auto_increment primary key, `name` varchar(255) not null, `type` enum('local', 'global') not null, `type2` enum('LOCAL', 'GLOBAL') not null, `enum1` tinyint null, `enum2` tinyint null, `enum3` tinyint null, `enum4` enum('a', 'b', 'c') null, `enum5` enum('a') null) default character set utf8mb4 engine = InnoDB;
+create table `publisher2` (`id` int(10) unsigned not null auto_increment primary key, `name` varchar(255) not null, `type` enum('local', 'global') not null default 'local', `type2` enum('LOCAL', 'GLOBAL') not null default 'LOCAL', `enum1` tinyint null, `enum2` tinyint null, `enum3` tinyint null, `enum4` enum('a', 'b', 'c') null, `enum5` enum('a') null) default character set utf8mb4 engine = InnoDB;
```
B4nan, line 40: please dont change this

tests/postgre-schema.sql
```diff
@@ -39,7 +39,7 @@ alter table "foo_bar2" add constraint "foo_bar2_foo_bar_id_unique" unique ("foo_
 create table "foo_param2" ("bar_id" int4 not null, "baz_id" int4 not null, "value" varchar(255) not null, "version" timestamptz(3) not null default current_timestamp(3));
 alter table "foo_param2" add constraint "foo_param2_pkey" primary key ("bar_id", "baz_id");
 
-create table "publisher2" ("id" serial primary key, "name" varchar(255) not null, "type" text check ("type" in ('local', 'global')) not null, "type2" text check ("type2" in ('LOCAL', 'GLOBAL')) not null, "enum1" int2 null, "enum2" int2 null, "enum3" int2 null, "enum4" text check ("enum4" in ('a', 'b', 'c')) null, "enum5" text check ("enum5" in ('a')) null);
+create table "publisher2" ("id" serial primary key, "name" varchar(255) not null, "type" text check ("type" in ('local', 'global')) not null default 'local', "type2" text check ("type2" in ('LOCAL', 'GLOBAL')) not null default 'LOCAL', "enum1" int2 null, "enum2" int2 null, "enum3" int2 null, "enum4" text check ("enum4" in ('a', 'b', 'c')) null, "enum5" text check ("enum5" in ('a')) null);
```
B4nan, line 42: same here

packages/knex/src/schema/DatabaseTable.ts
```diff
@@ -224,7 +224,8 @@ export class DatabaseTable {
     return !!this.getPrimaryKey();
   }
 
-  private getPropertyDeclaration(column: Column, namingStrategy: NamingStrategy, schemaHelper: SchemaHelper, compositeFkIndexes: Dictionary<{ keyName: string }>, compositeFkUniques: Dictionary<{ keyName: string }>) {
+  private getPropertyDeclaration(column: Column, namingStrategy: NamingStrategy, schemaHelper: SchemaHelper,
+    compositeFkIndexes: Dictionary<{ keyName: string }>, compositeFkUniques: Dictionary<{ keyName: string }>) {
```
B4nan, line 228: wrong formatting, either leave it on a single line, or refactor to have one parameter on one line

B4nan (changes requested):

packages/knex/src/schema/SchemaComparator.ts
kevinh-canva: @B4nan tbh I don't really understand why, but without this it will break.
From what I can investigate so far, the `toSchema` will have the default value as `''local''` (2 pairs of quote), whereas the `fromSchema` will have the default value as `'local'` (only a single pair of quote)

kevinh-canva: AFAIK we can't remove the double quote for `toSchema` either because seems like without it knex will generate the wrong default value for the enum i.e `local` instead of `'local'`.

tests/entities-sql/Publisher2.ts
kevinh-canva: I have run the test suites for schema generator again and it all passes, just wondering which tests are failing? I think it's necessary to add this in, because as per our previous conversation, if we have a default runtime value, we should also have a `default` property in the Enum decorator.
The reason is because when the entity generator runs, it will convert the SQL default into the runtime default, but the schema generator (the other direction) cannot infer the SQL default from the runtime default, hence the need for the decorator option, so that we don't generate a wrong migration.

packages/knex/src/schema/SchemaComparator.ts
B4nan: This will definitely need to go for the reasons I mentioned above. I am not saying the problem is invalid, I am saying you are trying to fix it on a wrong place and in a wrong way. You need to make sure the value is in the right shape when we create the column object. I believe the value should be quoted (but please ensure how it was working without your changes first), so instead of removing quotes, you should be quoting the other value.

tests/entities-sql/Publisher2.ts
B4nan: you changed things i dont want you to change - the shared SQL schema files and shared entities. i was referring more to the previous test failures, where the entity generator tests were failing because you added the db defaults to the schema files. that should all go, its not related to this PR at all. 

so please remove the db defaults for existing entities like this one, as well as all the changes from shared schema files. you should add new tests if you need to test this behaviour

also i would be quite surprised in the tests would be passing this way, as here you specify the db default for this property, it is also in the schema file, so it should be generated in the entity generator snapshot and there i cant see it. so if the tests are passing, it points to a wrong implementation. if there is a db default, it should be part of the property decorator options, as property initialzer will never produce db default (in other words entities generated this way would produce additional queries in schema diffing, as they are missing the default option but it is present in the db schema - the diffing would try to remove those db defaults as they are missing from entity definition).

B4nan: namely this is wrong: 

https://github.com/mikro-orm/mikro-orm/pull/2608/files#diff-4a7257bf4caeef31a7f90185abd2fc11b3aaf69953f8d62b0b73f92daea35f32R373

it wont produce the same schema as the one used to generate such entity

kevinh-canva: Ahh yep I meant the schema generator tests all passed, yeah I notice the problem with the entity generator now thanks.  Will fix

packages/knex/src/schema/SchemaComparator.ts
kevinh-canva: Can you explain the example with the `now()` and `"now()"` a bit more? Does this change result in false equality somehow?

Tbh I don't think I really understand the interaction between `default` and `defaultRaw`, when and why we need to quote things etc. This column object is constructed from `DatabaseTable.init`, which I believe is also used by the entity generator when constructing the `EntityProperty`. Is it safe to quote the default value here without introducing more changes elsewhere?

B4nan: > Can you explain the example with the now() and "now()" a bit more? Does this change result in false equality somehow?

kevinh-canva: Can you explain why in `MetadataDiscovery.initDefaultValue` we need to quote the default value when setting `prop.defaultRaw`? For the enum default, `prop.default = 'local'` for example, and this additional quoting makes it `''local''`.
Then in `DatabaseTable.addColumnFromProperty`, `prop.defaultRaw` is assigned to `column.default`, causing the mismatch between the `from` value `'local'` and the `to` value `''local''`.

B4nan: I already said that `defaultRaw` is a quoted form of `default`, that is exactly what `initDefaultValue` does. If the default value is a string, it should _not_ be quoted in `prop.default`, but only in `prop.defaultRaw`.

kevinh-canva: Got it. However, what I don't understand is why `column.default` is assigned the value of `prop.defaultRaw` in `DatabaseTable.addColumnFromProperty`?
That would make it `''local''` instead of `'local'` which I believe is what we get from the database during the entity generation process. 
In `DatabaseTable.addColumnFromProperty`, do you think it's ok if we first check that if `prop.default` is a string, use it for `column.default`, otherwise, use `prop.defaultRaw`? I think that would be the minimal change to fix this, but not sure if it has any other consequences.

B4nan: Because `column.default` should be quoted too. It is always a string (or null), unline `prop.default`.

kevinh-canva: Ok so we need to start from the database level first. I believe a default value for an enum is just a normal string, so if you declare the field as `enum('local', 'global')` and want to assign a default value to it it would be `default 'local'`. It's a string.
In the entity generation process, if the default matches `/^'.*'$/`, (which I believe signify it as a string default?) we would generate the entity file with `default` and not `defaultRaw`.

The consequence of this is when `MetadataDiscovery` reads the entity file, it will try to coerce this default value into another pair of quote i.e `''local''` and assign that to `prop.defaultRaw`. I don't think this is wrong, just want to point it out.

And then when we fetch the database value to set the `from` value, it would just be a single-quoted `'local'`. So now we are trying to compare a double quoted `''local''` with a single-quoted `'local'` and the schema comparator thought there's a change and generate a migration.

B4nan: > The consequence of this is when MetadataDiscovery reads the entity file, it will try to coerce this default value into another pair of quote i.e ''local'' and assign that to prop.defaultRaw. I don't think this is wrong, just want to point it out.

I still dont understand where you see that happening (in master - it is it specific to the changes in your PR?). Can you point me to where should I `console.log` to see that happen?

packages/entity-generator/src/SourceFile.ts
```diff
@@ -155,13 +188,17 @@ export class SourceFile {
       options.nullable = true;
     }
 
-    if (prop.default && typeof prop.default === 'string') {
-      if ([`''`, ''].includes(prop.default)) {
-        options.default = `''`;
-      } else if (prop.default.match(/^'.*'$/)) {
-        options.default = prop.default;
+    if (prop.default !== null && prop.default !== undefined) {
```
kevinh-canva, line 191: This is so that if prop.default is 0 or false, we still go into this if block.

B4nan: you can just do `prop.default != null`, its the same

packages/mysql-base/src/MySqlSchemaHelper.ts
```diff
@@ -169,7 +169,8 @@ export class MySqlSchemaHelper extends SchemaHelper {
   }
 
   protected wrap(val: string | undefined, type: Type<unknown>): string | undefined {
-    return typeof val === 'string' && val.length > 0 && (type instanceof StringType || type instanceof TextType) ? this.platform.quoteValue(val) : val;
+    return typeof val === 'string' && val.length > 0 && (type instanceof StringType
```
kevinh-canva, line 172: @B4nan I believe I have found the best way to fix this. 
Basically, the problem with double quoting I notice is only present for MySQL. To be exact, double quoting is the expected behaviour. The enum default from Postgres is quoted after being retrieved from DB, and the metadata-discovery value is also quoted thanks to `MetadataDiscovery.initDefaultValue` so they are equal for postgres.

For MySQL however, the enum is not quoted, hence the mismatch. LMK if this fix makes sense.

packages/entity-generator/src/SourceFile.ts
kevinh-canva: thanks for the neat trick!

B4nan:
Will try to look into this later this week, for now just one thing - please improve the coverage, as said I am fine with that one line about `EnumType` (fine in a sense that I will handle that myself, it should be still covered), but there are others problems too.

kevinh-canva:
> Will try to look into this later this week, for now just one thing - please improve the coverage, as said I am fine with that one line about `EnumType` (fine in a sense that I will handle that myself, it should be still covered), but there are others problems too.

@B4nan WDYM about increase code coverage sorry? All the checks seems to be passing, except some related to cognitive complexity?

B4nan:
> @B4nan WDYM about increase code coverage sorry? All the checks seems to be passing, except some related to cognitive complexity?

I am not tlaking about checks, I am talking about coverage, which has decreased.

```
----------------------------------|---------|----------|---------|---------|-------------------
File                              | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
----------------------------------|---------|----------|---------|---------|-------------------
All files                         |   99.96 |    99.86 |   99.96 |   99.95 |                   
 cli/src                          |     100 |      100 |     100 |     100 |                   
  CLIConfigurator.ts              |     100 |      100 |     100 |     100 |                   
  CLIHelper.ts                    |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 cli/src/commands                 |     100 |      100 |     100 |     100 |                   
  ClearCacheCommand.ts            |     100 |      100 |     100 |     100 |                   
  CreateDatabaseCommand.ts        |     100 |      100 |     100 |     100 |                   
  CreateSeederCommand.ts          |     100 |      100 |     100 |     100 |                   
  DatabaseSeedCommand.ts          |     100 |      100 |     100 |     100 |                   
  DebugCommand.ts                 |     100 |      100 |     100 |     100 |                   
  GenerateCacheCommand.ts         |     100 |      100 |     100 |     100 |                   
  GenerateEntitiesCommand.ts      |     100 |      100 |     100 |     100 |                   
  ImportCommand.ts                |     100 |      100 |     100 |     100 |                   
  MigrationCommandFactory.ts      |     100 |      100 |     100 |     100 |                   
  SchemaCommandFactory.ts         |     100 |      100 |     100 |     100 |                   
 core/src                         |     100 |      100 |     100 |     100 |                   
  EntityManager.ts                |     100 |      100 |     100 |     100 |                   
  MikroORM.ts                     |     100 |      100 |     100 |     100 |                   
  enums.ts                        |     100 |      100 |     100 |     100 |                   
  errors.ts                       |     100 |      100 |     100 |     100 |                   
  exceptions.ts                   |     100 |      100 |     100 |     100 |                   
  typings.ts                      |     100 |      100 |     100 |     100 |                   
 core/src/cache                   |     100 |      100 |     100 |     100 |                   
  FileCacheAdapter.ts             |     100 |      100 |     100 |     100 |                   
  MemoryCacheAdapter.ts           |     100 |      100 |     100 |     100 |                   
  NullCacheAdapter.ts             |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 core/src/connections             |     100 |      100 |     100 |     100 |                   
  Connection.ts                   |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 core/src/decorators              |     100 |      100 |     100 |     100 |                   
  Embeddable.ts                   |     100 |      100 |     100 |     100 |                   
  Embedded.ts                     |     100 |      100 |     100 |     100 |                   
  Entity.ts                       |     100 |      100 |     100 |     100 |                   
  Enum.ts                         |     100 |      100 |     100 |     100 |                   
  Filter.ts                       |     100 |      100 |     100 |     100 |                   
  Formula.ts                      |     100 |      100 |     100 |     100 |                   
  Indexed.ts                      |     100 |      100 |     100 |     100 |                   
  ManyToMany.ts                   |     100 |      100 |     100 |     100 |                   
  ManyToOne.ts                    |     100 |      100 |     100 |     100 |                   
  OneToMany.ts                    |     100 |      100 |     100 |     100 |                   
  OneToOne.ts                     |     100 |      100 |     100 |     100 |                   
  PrimaryKey.ts                   |     100 |      100 |     100 |     100 |                   
  Property.ts                     |     100 |      100 |     100 |     100 |                   
  Subscriber.ts                   |     100 |      100 |     100 |     100 |                   
  UseRequestContext.ts            |     100 |      100 |     100 |     100 |                   
  hooks.ts                        |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 core/src/drivers                 |     100 |      100 |     100 |     100 |                   
  DatabaseDriver.ts               |     100 |      100 |     100 |     100 |                   
  IDatabaseDriver.ts              |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 core/src/entity                  |     100 |      100 |     100 |     100 |                   
  ArrayCollection.ts              |     100 |      100 |     100 |     100 |                   
  BaseEntity.ts                   |     100 |      100 |     100 |     100 |                   
  Collection.ts                   |     100 |      100 |     100 |     100 |                   
  EntityAssigner.ts               |     100 |      100 |     100 |     100 |                   
  EntityFactory.ts                |     100 |      100 |     100 |     100 |                   
  EntityHelper.ts                 |     100 |      100 |     100 |     100 |                   
  EntityIdentifier.ts             |     100 |      100 |     100 |     100 |                   
  EntityLoader.ts                 |     100 |      100 |     100 |     100 |                   
  EntityRepository.ts             |     100 |      100 |     100 |     100 |                   
  EntityTransformer.ts            |     100 |      100 |     100 |     100 |                   
  EntityValidator.ts              |     100 |      100 |     100 |     100 |                   
  Reference.ts                    |     100 |      100 |     100 |     100 |                   
  WrappedEntity.ts                |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
  wrap.ts                         |     100 |      100 |     100 |     100 |                   
 core/src/events                  |     100 |      100 |     100 |     100 |                   
  EventManager.ts                 |     100 |      100 |     100 |     100 |                   
  TransactionEventBroadcaster.ts  |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 core/src/hydration               |     100 |      100 |     100 |     100 |                   
  Hydrator.ts                     |     100 |      100 |     100 |     100 |                   
  ObjectHydrator.ts               |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 core/src/logging                 |     100 |      100 |     100 |     100 |                   
  DefaultLogger.ts                |     100 |      100 |     100 |     100 |                   
  colors.ts                       |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 core/src/metadata                |     100 |      100 |     100 |     100 |                   
  EntitySchema.ts                 |     100 |      100 |     100 |     100 |                   
  JavaScriptMetadataProvider.ts   |     100 |      100 |     100 |     100 |                   
  MetadataDiscovery.ts            |     100 |      100 |     100 |     100 |                   
  MetadataProvider.ts             |     100 |      100 |     100 |     100 |                   
  MetadataStorage.ts              |     100 |      100 |     100 |     100 |                   
  MetadataValidator.ts            |     100 |      100 |     100 |     100 |                   
  ReflectMetadataProvider.ts      |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 core/src/naming-strategy         |     100 |      100 |     100 |     100 |                   
  AbstractNamingStrategy.ts       |     100 |      100 |     100 |     100 |                   
  EntityCaseNamingStrategy.ts     |     100 |      100 |     100 |     100 |                   
  MongoNamingStrategy.ts          |     100 |      100 |     100 |     100 |                   
  UnderscoreNamingStrategy.ts     |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 core/src/platforms               |     100 |      100 |     100 |     100 |                   
  ExceptionConverter.ts           |     100 |      100 |     100 |     100 |                   
  Platform.ts                     |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 core/src/types                   |    99.5 |      100 |   98.82 |   99.43 |                   
  ArrayType.ts                    |     100 |      100 |     100 |     100 |                   
  BigIntType.ts                   |     100 |      100 |     100 |     100 |                   
  BlobType.ts                     |     100 |      100 |     100 |     100 |                   
  BooleanType.ts                  |     100 |      100 |     100 |     100 |                   
  DateTimeType.ts                 |     100 |      100 |     100 |     100 |                   
  DateType.ts                     |     100 |      100 |     100 |     100 |                   
  DecimalType.ts                  |     100 |      100 |     100 |     100 |                   
  DoubleType.ts                   |     100 |      100 |     100 |     100 |                   
  EnumArrayType.ts                |     100 |      100 |     100 |     100 |                   
  EnumType.ts                     |      75 |      100 |      50 |      75 | 12                
  FloatType.ts                    |     100 |      100 |     100 |     100 |                   
  IntegerType.ts                  |     100 |      100 |     100 |     100 |                   
  JsonType.ts                     |     100 |      100 |     100 |     100 |                   
  SmallIntType.ts                 |     100 |      100 |     100 |     100 |                   
  StringType.ts                   |     100 |      100 |     100 |     100 |                   
  TextType.ts                     |     100 |      100 |     100 |     100 |                   
  TimeType.ts                     |     100 |      100 |     100 |     100 |                   
  TinyIntType.ts                  |     100 |      100 |     100 |     100 |                   
  Type.ts                         |     100 |      100 |     100 |     100 |                   
  UnknownType.ts                  |     100 |      100 |     100 |     100 |                   
  UuidType.ts                     |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 core/src/unit-of-work            |     100 |      100 |     100 |     100 |                   
  ChangeSet.ts                    |     100 |      100 |     100 |     100 |                   
  ChangeSetComputer.ts            |     100 |      100 |     100 |     100 |                   
  ChangeSetPersister.ts           |     100 |      100 |     100 |     100 |                   
  CommitOrderCalculator.ts        |     100 |      100 |     100 |     100 |                   
  IdentityMap.ts                  |     100 |      100 |     100 |     100 |                   
  UnitOfWork.ts                   |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 core/src/utils                   |     100 |      100 |     100 |     100 |                   
  Configuration.ts                |     100 |      100 |     100 |     100 |                   
  ConfigurationLoader.ts          |     100 |      100 |     100 |     100 |                   
  EntityComparator.ts             |     100 |      100 |     100 |     100 |                   
  NullHighlighter.ts              |     100 |      100 |     100 |     100 |                   
  QueryHelper.ts                  |     100 |      100 |     100 |     100 |                   
  RequestContext.ts               |     100 |      100 |     100 |     100 |                   
  TransactionContext.ts           |     100 |      100 |     100 |     100 |                   
  Utils.ts                        |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 entity-generator/src             |   98.36 |    92.64 |     100 |   98.27 |                   
  EntityGenerator.ts              |     100 |      100 |     100 |     100 |                   
  SourceFile.ts                   |   98.08 |    91.93 |     100 |   97.98 | 101-103           
  index.ts                        |     100 |      100 |     100 |     100 |                   
 knex/src                         |     100 |      100 |     100 |     100 |                   
  AbstractSqlConnection.ts        |     100 |      100 |     100 |     100 |                   
  AbstractSqlDriver.ts            |     100 |      100 |     100 |     100 |                   
  AbstractSqlPlatform.ts          |     100 |      100 |     100 |     100 |                   
  MonkeyPatchable.ts              |     100 |      100 |     100 |     100 |                   
  SqlEntityManager.ts             |     100 |      100 |     100 |     100 |                   
  SqlEntityRepository.ts          |     100 |      100 |     100 |     100 |                   
 knex/src/query                   |     100 |      100 |     100 |     100 |                   
  ArrayCriteriaNode.ts            |     100 |      100 |     100 |     100 |                   
  CriteriaNode.ts                 |     100 |      100 |     100 |     100 |                   
  CriteriaNodeFactory.ts          |     100 |      100 |     100 |     100 |                   
  ObjectCriteriaNode.ts           |     100 |      100 |     100 |     100 |                   
  QueryBuilder.ts                 |     100 |      100 |     100 |     100 |                   
  QueryBuilderHelper.ts           |     100 |      100 |     100 |     100 |                   
  ScalarCriteriaNode.ts           |     100 |      100 |     100 |     100 |                   
  enums.ts                        |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 knex/src/schema                  |     100 |      100 |     100 |     100 |                   
  DatabaseSchema.ts               |     100 |      100 |     100 |     100 |                   
  DatabaseTable.ts                |     100 |      100 |     100 |     100 |                   
  SchemaComparator.ts             |     100 |      100 |     100 |     100 |                   
  SchemaGenerator.ts              |     100 |      100 |     100 |     100 |                   
  SchemaHelper.ts                 |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 mariadb/src                      |     100 |      100 |     100 |     100 |                   
  MariaDbConnection.ts            |     100 |      100 |     100 |     100 |                   
  MariaDbDriver.ts                |     100 |      100 |     100 |     100 |                   
  MariaDbPlatform.ts              |     100 |      100 |     100 |     100 |                   
  MariaDbSchemaHelper.ts          |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 migrations/src                   |     100 |      100 |     100 |     100 |                   
  JSMigrationGenerator.ts         |     100 |      100 |     100 |     100 |                   
  Migration.ts                    |     100 |      100 |     100 |     100 |                   
  MigrationGenerator.ts           |     100 |      100 |     100 |     100 |                   
  MigrationRunner.ts              |     100 |      100 |     100 |     100 |                   
  MigrationStorage.ts             |     100 |      100 |     100 |     100 |                   
  Migrator.ts                     |     100 |      100 |     100 |     100 |                   
  TSMigrationGenerator.ts         |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 mongodb/src                      |     100 |      100 |     100 |     100 |                   
  MongoConnection.ts              |     100 |      100 |     100 |     100 |                   
  MongoDriver.ts                  |     100 |      100 |     100 |     100 |                   
  MongoEntityManager.ts           |     100 |      100 |     100 |     100 |                   
  MongoEntityRepository.ts        |     100 |      100 |     100 |     100 |                   
  MongoExceptionConverter.ts      |     100 |      100 |     100 |     100 |                   
  MongoPlatform.ts                |     100 |      100 |     100 |     100 |                   
 mysql-base/src                   |     100 |      100 |     100 |     100 |                   
  MySqlConnection.ts              |     100 |      100 |     100 |     100 |                   
  MySqlDriver.ts                  |     100 |      100 |     100 |     100 |                   
  MySqlExceptionConverter.ts      |     100 |      100 |     100 |     100 |                   
  MySqlPlatform.ts                |     100 |      100 |     100 |     100 |                   
  MySqlSchemaHelper.ts            |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 mysql/src                        |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 postgresql/src                   |     100 |     98.7 |     100 |     100 |                   
  PostgreSqlConnection.ts         |     100 |      100 |     100 |     100 |                   
  PostgreSqlDriver.ts             |     100 |      100 |     100 |     100 |                   
  PostgreSqlExceptionConverter.ts |     100 |      100 |     100 |     100 |                   
  PostgreSqlPlatform.ts           |     100 |      100 |     100 |     100 |                   
  PostgreSqlSchemaHelper.ts       |     100 |    97.56 |     100 |     100 | 105               
  index.ts                        |     100 |      100 |     100 |     100 |                   
 reflection/src                   |     100 |      100 |     100 |     100 |                   
  TsMorphMetadataProvider.ts      |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
 seeder/src                       |     100 |      100 |     100 |     100 |                   
  factory.ts                      |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
  seed-manager.ts                 |     100 |      100 |     100 |     100 |                   
  seeder.ts                       |     100 |      100 |     100 |     100 |                   
 sqlite/src                       |     100 |      100 |     100 |     100 |                   
  SqliteConnection.ts             |     100 |      100 |     100 |     100 |                   
  SqliteDriver.ts                 |     100 |      100 |     100 |     100 |                   
  SqliteExceptionConverter.ts     |     100 |      100 |     100 |     100 |                   
  SqlitePlatform.ts               |     100 |      100 |     100 |     100 |                   
  SqliteSchemaHelper.ts           |     100 |      100 |     100 |     100 |                   
  index.ts                        |     100 |      100 |     100 |     100 |                   
----------------------------------|---------|----------|---------|---------|-------------------
```

This means you added code that is not tested.

B4nan (approved):
I guess this looks good enough to be merged, I can handle the rest myself :] 

Thanks a lot!

edit: Checking things locally now, and I see you still did not get this 100% right. `prop.default` can't be a quoted string (hence the conditional unquoting does not make sense), and if the default is a string, `prop.default` can never be the same as `prop.defaultRaw`. You kept trying to solve it in wrong place, dealing with already invalid state in the metadata - it has to be normalized at the place where we extract the metadata from information_schema. I'll share the commit where I fix this so you can better understand what I was talking about.

edit2: the prolem was there even before your changes :]

kevinh-canva:
@B4nan just wondering which version will this change be in? I'm having some trouble using my local package (with this change) with a custom naming strategy, so looking forward for even a minor version release that I can get to get this change.

PR #2614, opened 2022-01-08
added graphql example

47vigen:
just a graphql example created by typescript + fastify + mikro-orm(postgresql) + mercurius(graphql adaptor) + type-graphql


B4nan:
Closing as stale, I'd be happy to include the example here, but it would need to be up to date (I think this was already created when v5 was out, and now we have v6).

Issue #2631, opened 2022-01-12
Running Migrations with es modules


PodaruDragos:
**Describe the bug**
There is no way of running migration using the CLI with es modules.

**Stack trace**
```
mikro-orm migration:create --initial

Create new migration with current schema diff

Options:
  -b, --blank    Create blank migration                                [boolean]
  -i, --initial  Create initial migration                              [boolean]
  -d, --dump     Dumps all queries to console                          [boolean]
  -p, --path     Sets path to directory where to save entities          [string]
  -v, --version  Show version number                                   [boolean]
  -h, --help     Show help                                             [boolean]

Error: No entities found, please use `entities` option
    at Configuration.validateOptions (/..../mikro-es-modules/node_modules/@mikro-orm/core/dist/utils/Configuration.js:169:19)
    at new Configuration (/...l/mikro-es-modules/node_modules/@mikro-orm/core/dist/utils/Configuration.js:24:18)
    at Function.getConfiguration (/.../mikro-es-modules/node_modules/@mikro-orm/core/dist/utils/ConfigurationLoader.js:55:20)
    at async Function.getORM (/.../mikro-es-modules/node_modules/@mikro-orm/cli/dist/CLIHelper.js:19:25)
    at async Function.handleMigrationCommand (/.../personal/mikro-es-modules/node_modules/@mikro-orm/cli/dist/commands/MigrationCommandFactory.js:88:21)
```

**To Reproduce**
1. Clone this repo [repo](https://github.com/PodaruDragos/mikro-es-modules)

**Expected behavior**
I would like to be able to run `migration-create --initial`.

**Versions**

| Dependency | Version |
| - | - |
| node | 17.3.0 |
| typescript | 4.5.4 |
| mikro-orm | 5.0.0-dev.604 |
| your-driver | postgres |

B4nan:
I believe you can't use folder based discovery if you want native ESM. In other words, you can't use env vars for the CLI as that implies folder based discovery - you will need a CLI config file where you use `entities: [A, B]` just like in your app.

PodaruDragos:
> I believe you can't use folder based discovery if you want native ESM. In other words, you can't use env vars for the CLI as that implies folder based discovery - you will need a CLI config file where you use `entities: [A, B]` just like in your app.

Will try it out. Thanks.
Didn't work tho, but I see i can do `await connection.getMigrator().createMigration('./src/migrations/', false, true); and that works.

[...]

PR #2641, opened 2022-01-16
refactor: fix path on windows for esmodules

PodaruDragos:
I was no able to remove those `eslint` things on line 609-610.

This closes https://github.com/mikro-orm/mikro-orm/issues/2631#issuecomment-1013506198.

Edit: I would really appreciate if you can merge this since I really can't work on windows without this. ( only with `wsl` and that is really slow ).

### DIFF
--- a/packages/core/src/utils/Utils.ts
+++ b/packages/core/src/utils/Utils.ts
@@ -3,6 +3,8 @@ import clone from 'clone';
 import type { GlobbyOptions } from 'globby';
 import globby from 'globby';
 import { extname, isAbsolute, join, normalize, relative, resolve } from 'path';
+import { platform } from 'os';
+import { pathToFileURL } from 'url'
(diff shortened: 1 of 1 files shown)


packages/core/src/utils/Utils.ts
```diff
@@ -762,6 +764,10 @@ export class Utils {
       return import(id);
     }
 
+    if (platform() === 'win32') {
```
B4nan, line 767:
```suggestion
    /* istanbul ignore next */
    if (platform() === 'win32') {
```

B4nan:
Would be great to compile all the knoledge you get from this into an article or something

PodaruDragos:
> Would be great to compile all the knoledge you get from this into an article or something

yeah. it should be: 

Esmodules in nodejs
Please don't even try. Thanks for reading 😆

PR #2650, opened 2022-01-19
fix(core): fix nested query with fk as pk

ml1nk:
Some tests are broken but at least the cases of #2649 are working as expected. Both changes break tests separately and i only have a rough idea why.

@B4nan I try to improve this fix tomorrow, but i'm missing insight.

### DIFF
--- a/packages/core/src/utils/QueryHelper.ts
+++ b/packages/core/src/utils/QueryHelper.ts
@@ -1,7 +1,7 @@
 import { Reference } from '../entity/Reference';
 import { Utils } from './Utils';
 import type { AnyEntity, Dictionary, EntityMetadata, EntityProperty, FilterDef, ObjectQuery, FilterQuery } from '../typings';
-import { ARRAY_OPERATORS, GroupOperator } from '../enums';
+import { ARRAY_OPERATORS, GroupOperator, ReferenceType } from '../enums';
 import type { Platform } from '../platforms';
 import type { MetadataStorage } from '../metadata/MetadataStorage';
 import { JsonType } from '../types';
@@ -56,7 +56,17 @@ export class QueryHelper {
     }
 
     if (meta.primaryKeys.every(pk => pk in where) && Utils.getObjectKeysSize(where) === meta.primaryKeys.length) {
-      return !GroupOperator[key as string] && Object.keys(where).every(k => !Utils.isPlainObject(where[k]) || Object.keys(where[k]).every(v => !Utils.isOperator(v, false)));
+      return !GroupOperator[key as string] && Object.keys(where).every(k => !Utils.isPlainObject(where[k]) || Object.keys(where[k]).every(v => {
+        if (Utils.isOperator(v, false)) {
+          return false;
+        }
+
+        if ([ReferenceType.ONE_TO_ONE, ReferenceType.MANY_TO_ONE].includes(meta.properties[k].reference)) {
+          return this.inlinePrimaryKeyObjects(where[k], meta.properties[k].targetMeta, metadata, v);
+        }
+
+        return true;
+      }));
     }
 
     Object.keys(where).forEach(k => {
--- a/packages/knex/src/query/ObjectCriteriaNode.ts
+++ b/packages/knex/src/query/ObjectCriteriaNode.ts
@@ -105,7 +105,16 @@ export class ObjectCriteriaNode extends CriteriaNode {
     const embeddable = this.prop.reference === ReferenceType.EMBEDDED;
     const knownKey = [ReferenceType.SCALAR, ReferenceType.MANY_TO_ONE, ReferenceType.EMBEDDED].includes(this.prop.reference) || (this.prop.reference === ReferenceType.ONE_TO_ONE && this.prop.owner);
     const operatorKeys = knownKey && Object.keys(this.payload).every(key => Utils.isOperator(key, false));
-    const primaryKeys = knownKey && Object.keys(this.payload).every(key => this.metadata.find(this.entityName)!.primaryKeys.includes(key));
+    const primaryKeys = knownKey && Object.keys(this.payload).every(key => {
+      const meta = this.metadata.find(this.entityName)!;
+      if (!meta.primaryKeys.includes(key)) {
+        return false;
+      }
+      if (!Utils.isPlainObject(this.payload[key].payload) || ![ReferenceType.ONE_TO_ONE, ReferenceType.MANY_TO_ONE].includes(meta.properties[key].reference)) {
+        return true;
+      }
+      return Object.keys(this.payload[key].payload).every(k => meta.properties[key].targetMeta!.primaryKeys.includes(k));
+    });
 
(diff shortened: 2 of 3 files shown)


packages/core/src/utils/QueryHelper.ts
```diff
@@ -56,7 +56,19 @@ export class QueryHelper {
...
     if (meta.primaryKeys.every(pk => pk in where) && Utils.getObjectKeysSize(where) === meta.primaryKeys.length) {
-      return !GroupOperator[key as string] && Object.keys(where).every(k => !Utils.isPlainObject(where[k]) || Object.keys(where[k]).every(v => !Utils.isOperator(v, false)));
+      return !GroupOperator[key as string] && Object.keys(where).every(k => !Utils.isPlainObject(where[k]) || Object.keys(where[k]).every(v => {
+        if (Utils.isOperator(v, false)) {
+          return false;
+        }
+
+        // Only the owning side of OneToOne can be a primary
+        if ([ReferenceType.ONE_TO_ONE, ReferenceType.MANY_TO_ONE].includes(meta.properties[k].reference)) {
+          const meta2 = metadata.find(meta.properties[k]?.type) || meta;
```
B4nan, line 66: you can just use `prop.targetMeta`, it will be set on all relation properties

i am quite sure the fallback to `meta` is never triggered, and if it would, it would be broken, you cant just swap them - that would work only in case of self referencing relation

packages/core/src/utils/QueryHelper.ts
```diff
@@ -56,7 +56,19 @@ export class QueryHelper {
     }
 
     if (meta.primaryKeys.every(pk => pk in where) && Utils.getObjectKeysSize(where) === meta.primaryKeys.length) {
-      return !GroupOperator[key as string] && Object.keys(where).every(k => !Utils.isPlainObject(where[k]) || Object.keys(where[k]).every(v => !Utils.isOperator(v, false)));
+      return !GroupOperator[key as string] && Object.keys(where).every(k => !Utils.isPlainObject(where[k]) || Object.keys(where[k]).every(v => {
+        if (Utils.isOperator(v, false)) {
+          return false;
+        }
+
+        // Only the owning side of OneToOne can be a primary
```
B4nan, line 64: i dont understand this comment - why are you not checking that in the condition then? is it ok that 1:1 inverse will go inside it the branch too? or why is the comment there?

packages/knex/src/query/ObjectCriteriaNode.ts
```diff
@@ -105,7 +105,18 @@ export class ObjectCriteriaNode extends CriteriaNode {
...
-    const primaryKeys = knownKey && Object.keys(this.payload).every(key => this.metadata.find(this.entityName)!.primaryKeys.includes(key));
+    const primaryKeys = knownKey && Object.keys(this.payload).every(key => {
+      const meta = this.metadata.find(this.entityName)!;
+      if (!meta.primaryKeys.includes(key)) {
+        return false;
+      }
+      if (!Utils.isPlainObject(this.payload[key].payload) || ![ReferenceType.ONE_TO_ONE, ReferenceType.MANY_TO_ONE].includes(meta.properties[key].reference)) {
+        return true;
+      }
+      const meta2 = this.metadata.find(meta.properties[key]?.type) || meta;
```
B4nan, line 116: same here, use `prop.targetMeta` and remove the fallback clause to `meta`

B4nan (changes requested):

B4nan:
let's not amend/force push after you get a review. this makes reviewing the changes almost impossible to me, as I only see the whole commit and not the changes you did after i reviewed it

packages/core/src/utils/QueryHelper.ts
ml1nk: 1:1 inverse never enters this branch, because it is never a primary. I removed the comment as it only leads to confusion.

ml1nk:
> let's not amend/force push after you get a review. this makes reviewing the changes almost impossible to me, as I only see the whole commit and not the changes you did after i reviewed it

I will stop amending, sorry for creating more work for you.

Any idea how to decrease the complexity to solve the codeclimate issue?

B4nan:
> Any idea how to decrease the complexity to solve the codeclimate issue?

You could move some of the logic to new method to distribute the complexity. 

But dont be too concerned with codeclimate, as long as we will have 100% branch coverage I will be happy to merge. (note that we dont have it, one line is not covered by tests currently)

tests/issues/GH2648.test.ts
```diff
@@ -49,39 +49,51 @@ export class B4 {
...
+interface Test {
+  t1: string;
+  t2: string;
+}
+
+@Entity()
+export class D {
+
+  @PrimaryKey({ type: JsonType })
+  id!: Test;
```
B4nan, line 86: JSON primary key? wow, is that really some actual use case you have or just a general testing of custom types? :] seems quite out of space

ml1nk: I tested it with some data and surprisingly it works, but i would never use it like this. I needed a primary key which allows a smart query like { id: { t1: 'a', t2: 'b' } } and isn't a FK to reach the last line for coverage,

ml1nk: Do you want me to move this test to another place or do you have any idea to cover the line otherwise?

B4nan: if you dont see any better location inside `tests/features` than let's keep it here, it's fine

ml1nk: I don't see any better location. Thanks.

packages/core/src/utils/QueryHelper.ts
```diff
@@ -56,7 +56,17 @@ export class QueryHelper {
     }
 
     if (meta.primaryKeys.every(pk => pk in where) && Utils.getObjectKeysSize(where) === meta.primaryKeys.length) {
-      return !GroupOperator[key as string] && Object.keys(where).every(k => !Utils.isPlainObject(where[k]) || Object.keys(where[k]).every(v => !Utils.isOperator(v, false)));
+      return !GroupOperator[key as string] && Object.keys(where).every(k => !Utils.isPlainObject(where[k]) || Object.keys(where[k]).every(v => {
+        if (Utils.isOperator(v, false)) {
```
B4nan, line 60: are you sure this is correct? previously it was doing `Object.keys(where[k]).every(v => !Utils.isOperator(v, false))`, but here you are not negating the return value

ml1nk: I'm still negating the value:
`if (Utils.isOperator(v, false)) {
   return false;
}`
If any v is an operator than inlinePrimaryKeyObjects is false, like before.

ml1nk: Effectively only line 64-66 should change the behavior.

B4nan (approved):

B4nan:
Thanks a lot!

PR #2664, opened 2022-01-24
fix(core): fix missing embedded columns on joined select query

### LINKED ISSUE #2663: Embeddables in populated entity returning undefined when using JOINED strategy
**Describe the bug**
I have the following entities and embeddable:
```ts
@Embeddable()
export class ContactPerson {
	@Property()
	name: string;
}

@Entity()
export class User extends Base {
	@Embedded({ entity: () => ContactPerson, object: true })
	contactPerson: ContactPerson;
}

@Entity()
export class Subscription extends Base {
	@Index()
	@ManyToOne({ entity: () => User, onDelete: 'cascade' })
	user: User;
}
```

When I try to get the `Subscription` with populated `User`, while using `LoadStrategy.JOINED`, I'd always get undefined for `contactUser`. However, if using `LoadStrategy.SELECT_IN`, it is working fine.

```ts
const subscription = await em.findOne(Subscription, {
	id: 'cc60b4dc-2cba-4009-bae9-41e81acd548a'
}, {
	populate: ['user'],
	strategy: LoadStrategy.SELECT_IN,
});
console.log(subscription.user.contactPerson); // Return object result

const subscription2 = await em.findOne(Subscription, {
	id: '046c343b-3842-44cb-bdcc-a3dabf89cb09'
}, {
	populate: ['user'],
	strategy: LoadStrategy.JOINED,
});
console.log(subscription2.user.contactPerson); // Return undefined
```

**Expected behavior**
By using `LoadStrategy.JOINED`, the embedded property should be included in the select query.

**Versions**
^5.0.0-dev.651

| Dependency | Version |
| - | - |
| node | v16.13.2 |
| typescript | 4.3.5 |
| mikro-orm | ^5.0.0-dev.651 |
| postgres | 14.1 |

adlanarifzr:
Closes #2663

### DIFF
--- a/packages/core/src/platforms/Platform.ts
+++ b/packages/core/src/platforms/Platform.ts
@@ -376,7 +376,7 @@ export abstract class Platform {
       return false;
     }
 
-    return [ReferenceType.SCALAR, ReferenceType.MANY_TO_ONE].includes(prop.reference) || (prop.reference === ReferenceType.ONE_TO_ONE && prop.owner);
+    return [ReferenceType.SCALAR, ReferenceType.MANY_TO_ONE, ReferenceType.EMBEDDED].includes(prop.reference) || (prop.reference === ReferenceType.ONE_TO_ONE && prop.owner);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


B4nan:
Thanks, but let's add a test case too, please.

adlanarifzr:
@B4nan Sorry, I forgot about that. I've added the unit test now.

Issue #2675, opened 2022-01-27
The populated relation query does not follow the given schema when using `LoadStrategy.SELECT_IN`


adlanarifzr:
**Describe the bug**
I have the following entities:
```ts
@Entity()
export class User extends Base {
	...
}

@Entity()
export class Trainer extends Base {
	@Index()
	@ManyToOne({ entity: () => User, onDelete: 'cascade' })
	user: User;
}
```

When I try to get the trainer with populated user data from specific schema, I would get the error `relation "user" does not exist`. The DB query logs are given below. I noticed on the second query, it does not use the given schema when querying.

**Stack trace**
```console
select "t0".* from "myschema"."trainer" as "t0" where "t0"."id" is not null limit 1
select "u0".* from "user" as "u0" where "u0"."id" in ('37179884-3741-4fb0-8155-644a1940b45a') order by "u0"."id" asc - relation "user" does not exist
```

**To Reproduce**
```ts
const trainer = await em.findOne(Trainer, {
	id: { $ne: null }
}, {
	schema: 'myschema',
	populate: ['user']
});
```

**Expected behavior**
I should return `trainer` data with `user` data included.

**Versions**

| Dependency | Version |
| - | - |
| node | v16.13.2 |
| typescript | 4.3.5 |
| mikro-orm | ^5.0.0-dev.669 |
| postgres | 14.1 |

PR #2676, opened 2022-01-27
fix(core): respect specified schema when populating (select-in)

adlanarifzr:
Closes [#2675](https://github.com/mikro-orm/mikro-orm/issues/2675) issue.


B4nan:
makes sense, thank you!

Issue #1711, opened 2021-04-22
Check Constraints


evantrimboli:
**Is your feature request related to a problem? Please describe.**
To keep data layer logic centralized, it would be nice to be able to specify check constraints on fields.

**Describe the solution you'd like**
```typescript
@Entity()
class Product {
  @Property({
    check: (productMeta) => `${productMeta.price.dbName} >= 0`,
  })
  price!: number
}
```

The sql generated should be something like:

```sql
CREATE TABLE products (
    price numeric CHECK (price >= 0)
);
```

The `productMeta` would be some sort of construct to access the underlying column names to be able to construct the constraint according to the naming strategy. There are a few ways this could go down, not sure what would be most idiomatic.

**Describe alternatives you've considered**
- They could be manually added to migrations
- Handle these at the API layer

**Additional context**
N/A

DASPRiD:
Check constraints can also be added to a table directly. This is actually how TypeORM did implement them. The resulting SQL from that usually looks something like this:

```sql
CREATE TABLE foo (
  price numeric NOT NULL,
  CONSTRAINT foo_bar CHECK (price >= 0)
)
```

This should be fairly easy to implement as a `@Check` annotation to the `@Entity`. We could basically allow the developer to write any SQL check expression, and then diff on the expression for migrations. This would also be the most flexible approach. For instance:

```typescript
@Entity()
@Check('price >= 0')
class Foo {
  @Property()
  price : number;
}
```

PR #2688, opened 2022-01-30
feat(core): add custom table check constraint support for postgres

DASPRiD:
Adds a new `@Check` decorator for entities, allowing to set up custom check
constraints. At the moment, this is only for Postgres, as MySQL only supports
check constraints as of version 8, which is above the minimum test version.

While SQLite does in theory support check constraints already, tests broke due
to some bug in Knex, which is why the implementation was skipped for now.

This partially addresses #1711

> **Note:**
> Due to already existing other check constraints (namely, enums in Postgres), I decided to suffix all custom check constraint names with `_custom`, so I can filter them out where necessary.

### DIFF
--- a/packages/knex/src/schema/SchemaComparator.ts
+++ b/packages/knex/src/schema/SchemaComparator.ts
@@ -1,6 +1,6 @@
 import type { Dictionary, EntityProperty } from '@mikro-orm/core';
 import { BooleanType, DateTimeType } from '@mikro-orm/core';
-import type { Column, ForeignKey, Index, SchemaDifference, TableDifference } from '../typings';
+import type { Check, Column, ForeignKey, Index, SchemaDifference, TableDifference } from '../typings';
 import type { DatabaseSchema } from './DatabaseSchema';
 import type { DatabaseTable } from './DatabaseTable';
 import type { AbstractSqlPlatform } from '../AbstractSqlPlatform';
@@ -116,12 +116,15 @@ export class SchemaComparator {
       addedColumns: {},
       addedForeignKeys: {},
       addedIndexes: {},
+      addedChecks: {},
       changedColumns: {},
       changedForeignKeys: {},
       changedIndexes: {},
+      changedChecks: {},
       removedColumns: {},
       removedForeignKeys: {},
       removedIndexes: {},
+      removedChecks: {},
       renamedColumns: {},
       renamedIndexes: {},
       fromTable,
@@ -204,6 +207,38 @@ export class SchemaComparator {
     }
 
     this.detectIndexRenamings(tableDifferences);
+
+    const fromTableChecks = fromTable.getChecks();
+    const toTableChecks = toTable.getChecks();
+
+    // See if all the checks in "from" table exist in "to" table
+    for (const check of toTableChecks) {
+      if (fromTable.hasCheck(check.name)) {
+        continue;
+      }
+
+      tableDifferences.addedChecks[check.name] = check;
+      changes++;
+    }
+
+    // See if there are any removed checks in "to" table
+    for (const check of fromTableChecks) {
+      if (!toTable.hasCheck(check.name)) {
+        tableDifferences.removedChecks[check.name] = check;
+        changes++;
+        continue;
+      }
+
+      // See if index has changed in "to" table
+      const toTableCheck = toTable.getCheck(check.name)!;
+
+      if (!this.diffCheck(check, toTableCheck)) {
+        continue;
+      }
+
+      tableDifferences.changedChecks[check.name] = toTableCheck;
+      changes++;
+    }
 
     const fromForeignKeys = { ...fromTable.getForeignKeys() };
     const toForeignKeys = { ...toTable.getForeignKeys() };
@@ -432,6 +467,10 @@ export class SchemaComparator {
     }
 
     return index1.primary === index2.primary && index1.unique === index2.unique;
+  }
+
+  diffCheck(check1: Check, check2: Check): boolean {
+    return check1.expression !== check2.expression;
   }
 
   hasSameDefaultValue(from: Column, to: Column): boolean {
--- a/packages/knex/src/schema/SchemaGenerator.ts
+++ b/packages/knex/src/schema/SchemaGenerator.ts
@@ -1,7 +1,7 @@
 import type { Knex } from 'knex';
 import type { Dictionary, EntityMetadata } from '@mikro-orm/core';
 import { AbstractSchemaGenerator } from '@mikro-orm/core';
-import type { Column, ForeignKey, Index, SchemaDifference, TableDifference } from '../typings';
+import type { Check, Column, ForeignKey, Index, SchemaDifference, TableDifference } from '../typings';
 import { DatabaseSchema } from './DatabaseSchema';
 import type { DatabaseTable } from './DatabaseTable';
 import type { AbstractSqlDriver } from '../AbstractSqlDriver';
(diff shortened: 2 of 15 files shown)


packages/knex/src/schema/SchemaGenerator.ts
```diff
@@ -480,6 +500,14 @@ export class SchemaGenerator extends AbstractSchemaGenerator<AbstractSqlDriver>
     }
   }
 
+  private createCheck(table: Knex.CreateTableBuilder, check: Check) {
+    table.check(check.expression, {}, `${check.name}_custom`);
```
B4nan, line 504: why the `_custom` suffix? can't say i like it, this should be fully configurable. is this kind of hack due to the missing support for property level check constraints? would be good to support those too, i guess it should be rather easy, or do you see problems with it? 

nit: not sure if the input type should be `CreateTableBuilder` - we use this for altering too, right?

packages/knex/src/schema/SchemaComparator.ts
```diff
@@ -205,6 +208,37 @@ export class SchemaComparator {
...
+    const fromTableChecks = fromTable.getChecks();
+    const toTableChecks = toTable.getChecks();
+
+    // See if all the checks in "from" table exist in "to" table
+    for (const check of toTableChecks) {
+      if (fromTable.hasCheck(check.name)) {
+        continue;
+      }
+
+      tableDifferences.addedChecks[check.name] = check;
```
B4nan, line 220: i guess we should also do `changes++` here

packages/postgresql/src/PostgreSqlSchemaHelper.ts
```diff
@@ -1,8 +1,9 @@
 import type { Dictionary } from '@mikro-orm/core';
 import { BigIntType, EnumType, Utils } from '@mikro-orm/core';
-import type { AbstractSqlConnection, Column, Index, DatabaseTable, TableDifference } from '@mikro-orm/knex';
+import type { AbstractSqlConnection, Column, Index, DatabaseTable, TableDifference, Check } from '@mikro-orm/knex';
 import { SchemaHelper } from '@mikro-orm/knex';
 import type { Knex } from 'knex';
+import { schema } from '../../../tests/entities';
```
B4nan, line 6: remove this

DASPRiD:
I gonna address the outstanding issues once I receive general sign-off or feedback on this PR :)

packages/postgresql/src/PostgreSqlSchemaHelper.ts
```diff
@@ -81,6 +82,16 @@ export class PostgreSqlSchemaHelper extends SchemaHelper {
     })));
   }
 
+  async getChecks(connection: AbstractSqlConnection, tableName: string, schemaName: string): Promise<Check[]> {
+    const sql = this.getChecksSQL(tableName, schemaName);
+    const checks = await connection.execute<any[]>(sql);
+
+    return checks.filter(check => check.name.endsWith('_custom')).map(check => ({
+      name: check.name.replace(/_custom$/, ''),
+      expression: check.expression.replace(/^CHECK \(\((.+)\)\)$/, '$1'),
```
B4nan, line 91: lets make this case insensitive

packages/sqlite/src/SqliteSchemaHelper.ts
```diff
@@ -1,5 +1,5 @@
 import type { Connection, Dictionary } from '@mikro-orm/core';
-import type { AbstractSqlConnection, Index } from '@mikro-orm/knex';
+import type { AbstractSqlConnection, Index ,Check } from '@mikro-orm/knex';
```
B4nan, line 2: bad formatting

tests/features/schema-generator/check-constraint-postgres.test.ts
```diff
@@ -0,0 +1,79 @@
+import { Check, Entity, EntitySchema, MikroORM, PrimaryKey, Property } from '@mikro-orm/core';
```
B4nan, line 1: the file name should be different, aligned with the rest (so `check-constraints.postgres.test.ts`)

tests/features/schema-generator/check-constraint-postgres.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    const diff = await orm.getSchemaGenerator().getCreateSchemaSQL({ wrap: false });
+    expect(diff).toMatchSnapshot('postgres-check-constraint-decorator');
+
+    await orm.close();
+  });
+
+  test('check constraint diff [postgres]', async () => {
+    const orm = await initORMPostgreSql();
+    const meta = orm.getMetadata();
+    const generator = new SchemaGenerator(orm.em);
```
B4nan, line 35: `orm.getSchemaGenerator()`

B4nan (changes requested):
thanks, overall looking great!

would be good to support the property level checks too, but we can merge this first and add it later. will be happy to give it a shot myself, as i would rather not ship that `_custom` thing to stable v5

B4nan:
> Due to already existing other check constraints (namely, enums in Postgres), I decided to suffix all custom check constraint names with _custom, so I can filter them out where necessary.

If that is the only reason, I would definitely like to refactor it before stable v5 is out, but I can do that myself after we merge.

DASPRiD:
Thanks a lot for your quick feedback!

Would you like to me to address the mentioned points, or would you rather prefer to just fork from my branch, since you mentioned that you'd like to introduce some changes and additions to it anyway?

B4nan:
Yes please, address it here before we merge it (as said I am fine merging the `_custom` thingy for now). Would be good to also test the uncovered lines.

packages/knex/src/schema/SchemaGenerator.ts
DASPRiD: It seems to work fine for altering as well, and `CreateTableBuilder` was also used for all other alterations (see `createIndex` and `dropIndex`).

B4nan:
please dont rebase and force push, so i can see what you actually changed _after_ the review

DASPRiD:
> please dont rebase and force push, so i can see what you actually changed _after_ the review

Sorry, force of habit :) – Will to separate commits on here from now on!

B4nan (approved):
perfect!

PodaruDragos:
Hello guys
Are we aware that this PR breaks projects since it needs to add checks in migrations ?
I had to create migration again so that it adds checks, checks that I don't need.

```
    this.addSql('alter table "user" drop constraint if exists "user_id_check";');
    this.addSql('alter table "user" alter column "id" type bigint using ("id"::bigint);');
    this.addSql('alter table "user" drop constraint if exists "user_created_at_check";');
```

Do I need to do anything to ignore this ?
Is this behavior by default since the merging of this PR ?

B4nan:
It works just fine in the tests, so please provide reproduction so we can see more details. Not sure what you mean by "since it needs to add checks in migrations", this PR adds support for diffing check constraints. If you were using some before, you will need to define them in the entity definition (via the `@Check()` decorator), otherwise the ORM will try to remove them. That is correct behaviour, you need to define all check constraints now at the metadata level.

PodaruDragos:
I have no idea what so ever. If this persist I am gonna make a repro.
what I can tell you for sure is that I didn't have any `checks` before.

B4nan:
Don't be too concerned with the `drop constraint if exists` queries - [those are there for each column update query](https://github.com/mikro-orm/mikro-orm/blob/master/packages/postgresql/src/PostgreSqlConnection.ts#L72), we do it blindly as we did not know before if there is a constrain or not. We could improve there finally, as we have this information. Will take a look if we can, I would very much like to get rid of this hack.

Try adding some logs to the schema comparator to see what caused the alter column query, there can be many reasons - obviously if it broke now, it might be because of the check constraints, but the ORM was using them only for enums in postgres, so it sounds weird to see a query for altering PK column (but I really can imagine you actually had such constraint before, PKs are unsigned and postgres does not have unsigned columns, it can be done only via check constraints, and maybe that is what you get from serial/bigserial meta types).

https://github.com/mikro-orm/mikro-orm/blob/master/packages/knex/src/schema/SchemaComparator.ts#L375

denisamirov:
> I have no idea what so ever. If this persist I am gonna make a repro. what I can tell you for sure is that I didn't have any `checks` before.

Hi. I have a similar problem. When setting constraints with the @Check, migrations are overwritten.
Have you found a solution?

PR #2725, opened 2022-02-08
fix(core): dynamically retrieve MikroORM constructor when constructing

Lioness100:
This change will allow extensions of the `MikroORM` to return instances of the new class when calling `MikroORM.init()`. Example:

```ts
class CustomMikroORM extends MikroORM {
  public test = 1;
}

// before change:

const orm = await CustomMikroORM.init(...);
console.log(orm.test); // undefined
console.log(orm instanceof CustomMikroORM); // false

// after change:

const orm = await CustomMikroORM.init(...);
console.log(orm.test); // 1
console.log(orm instanceof CustomMikroORM); // true
```

### DIFF
--- a/packages/core/src/MikroORM.ts
+++ b/packages/core/src/MikroORM.ts
@@ -34,7 +34,7 @@ export class MikroORM<D extends IDatabaseDriver = IDatabaseDriver> {
     }
 
     options = options instanceof Configuration ? options.getAll() : options;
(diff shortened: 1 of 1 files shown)


Lioness100:
Sorry, I messed up the types lol

Lioness100:
Oh, ok :(

PR #2745, opened 2022-02-11
fix(seeder): fix factory typing for entity with parameters in constru…

romainbrancourt:
Given an Entity with parameters in constructor and his factory :

```js
@Entity()
export class Project {

  [OptionalProps]?: 'createdAt';

  @PrimaryKey()
  id!: number;

  @Property()
  name!: string;

  constructor(name: string) {
    this.name = name;
  }
}

export class ProjectFactory extends Factory<Project> {
  model = Project;

  definition(faker: Faker): Partial<Project> {
    return {
      name: 'Money vault',
    };
  }
}
```

This will result in a typescript error:

Capture d’écran 2022-02-11 à 15 05 30

This PR fix the typing to allow any parameters in model constructor.


packages/seeder/src/Factory.ts
```diff
@@ -4,7 +4,7 @@ import type { RequiredEntityData, EntityManager } from '@mikro-orm/core';
 
 export abstract class Factory<T> {
 
-  abstract readonly model: { new(): T };
+  abstract readonly model: { new(...args: any[]): T };
```
B4nan, line 7: could you try replacing this with `Constructor<T>`? it's basically alias for the same thing we already use on various places

B4nan (approved):
thanks, looking good, just one idea for simplification

packages/seeder/src/Factory.ts
romainbrancourt: Done ! thx

Issue #2774, opened 2022-02-16
Embeddable with all fields set to `null` is hydrated as `undefined`


DASPRiD:
**Describe the bug**
When an embeddable has all its fields set to `null`, it will be hydrated as `undefined` upon retrieval.

**To Reproduce**
@Embeddable()
class Bar {
  @Property({type: t.string, nullable: true})
  public value : string | null;
}

@Entity()
class Foo {
    @PrimaryKey({type: t.uuid})
    public readonly id : string;

    @Embedded(() => Bar)
    public bar : Bar;
}

1. Create and persist `Foo` entity with `Bar` embeddable value set to `null`.
2. Clear the identity map.
3. Retrieve entity back from the entity manager.
4. `foo.bar.value === null` fails, because `foo.bar` is `undefined`.

**Expected behavior**
`foo.bar.value === null` should succeed.

**Versions**

| Dependency | Version |
| - | - |
| node | 1.16 |
| typescript | 4.4.3 |
| mikro-orm | 5.0.1 |
| your-driver | postgres |

B4nan:
Not reproducible. Maybe you are using object mode embeddables? In such case, you need to explicitly set the `null` value, otherwise it gets lots when the value is serialized to JSON.

DASPRiD:
Can you point me to the file where the hydration for embeddables takes place? I have a really hard time finding my way around the codebase. If I knew where to look, I could debug it myself :)

B4nan:
Are you using object embeddables or not?

DASPRiD:
I'm not.

B4nan:
Then please provide a failing repro. That will always be the first step, I need to see things failing, won't be accepting PRs without tests anyway :]

Hydration happens in ObjectHydrator, its done via compiled functions.

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/hydration/ObjectHydrator.ts#L208

DASPRiD:
Thanks, I'll have a look at this over the course of this week and will provide at least a failing test case :)

B4nan:
Note that the hydration works via the exact same method as for regular columns - in the end they are regular columns, just mapped to a nested object structure.

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/hydration/ObjectHydrator.ts#L267

Only case where we explicitly set things to `undefined` is when you use `forceUndefined` configuration:

[...]

PR #2785, opened 2022-02-17
test(core): supply failing test for issue #2774

DASPRiD:
Failing test case for #2774


DASPRiD:
#2774 was resolved and includes this test, thanks!

PR #2809, opened 2022-02-22
Update mikro-orm v5 nestjs docs

sergeyampo:


docs/versioned_docs/version-5.0/usage-with-nestjs.md
```diff
@@ -209,6 +209,27 @@ object.
 > still need CLI config with the full list of entities. On the other hand, we can
 > use globs there, as the CLI won't go through webpack.
 
+## Request context
+You need to fork entity manager for each request so their
```
B4nan, line 213: this section is not here for a purpose - this works automatically with nest adapter (you need to explicitly disable this behaviour if you want to register it like this)

B4nan:
We should probably add a section about this - I see its totally missing in there, so it might look confusing in a sense that people may think the `@UseRequestContext` or special context handling is required, which is pretty much the opposite of truth as things work out of box in most cases.

sergeyampo:
@B4nan Using GraphQL with Nest.js I am getting errors on every request about RequestContext. If I explicitly register express middleware after all body parsers as you described in the docs I'll get an application works a proper way.

B4nan:
So? Do I get it right that you are telling me that there is a limitation we already describe in the docs? :]

https://mikro-orm.io/docs/usage-with-nestjs#request-scoping-when-using-graphql

PR #2813, opened 2022-02-22
fix(core): fix property names for columns with dashes

123FLO321:
Dashes in property names would result in invalid generated typescript code.


B4nan:
Thanks, can you please add some tests? Unit tests could be enough, so just extend this one:

https://github.com/mikro-orm/mikro-orm/blob/master/tests/UnderscoreNamingStrategy.test.ts

PR #2818, opened 2022-02-24
fix(knex): respect explicit transaction in `em.count()`

asiFarran:
Looks like something got left behind during the refactoring to a single-options-object pattern


B4nan:
Are you sure this single change is enough? I would expect it will break usage as well as tests (let's see in few minutes)

theJoeBiz:
@asiFarran Thank you for this! Was pulling my hair out watching the debugger return [Array(1), 0] from `findAndCount`.

PR #2827, opened 2022-02-27
docs(Property Validation): add an explainer for validateRequired

hcharley:


docs/versioned_docs/version-5.0/property-validation.md
```diff
@@ -2,9 +2,10 @@
 title: Property Validation
 ---
 
-> Since v4.0.3 the validation needs to be explicitly enabled via `validate: true`.
+> Since v4.0.3 the validation needs to be explicitly enabled via `validate: true` in the [Configuration](configuration.md).
 > It has performance implications and usually should not be needed, as long as
-> you don't modify your entities via `Object.assign()`.
+> you don't modify your entities via `Object.assign()`. 
+> If validation is enabled, `validateRequired: true` in the [Configuration](configuration.md) will enable throwing `ValidationError` during `flush` events, while `validateRequired: false` will simply log a warning if validation fails
```
B4nan, line 8: `validate` and `validateRequired` options are not connected anyhow. `validate` is disabled by default, while `validateRequired` is enabled by default. If you disable `validateRequired` you won't get a warning instead (only a TS compile time error).

`validate: true` is for checking the property value types, e.g. if you want a validation in case you assign a number to a string property. `validateRequired` only checks if the value is present for a mandatory property.

We should have a new paragraph about this (exactly in this section), also describing the `OptionalProps` symbol and having some examples.

PR #2830, opened 2022-02-28
chore: use `preserveConstEnums: true`

asiFarran:
Currenly, since the `const` keyword is used in this export, no code is generated for the enum and attempts to reference it in application code would fail (`trying to read from undefined ....`)


B4nan:
Please use `preserveConstEnums` instead. This is indeed intentional and I do plan to rewrite the other enums to const enums too (probably in v6 as it might be semibreaking).

B4nan:
[Yes, ](https://github.com/mikro-orm/mikro-orm/blob/master/tsconfig.build.json)

PR #2838, opened 2022-03-01
feat(core): allow forcing write connection via `forceWriteConnection`

oliversalzburg:
This allows to bypass replica read attempts through the `FindOptions`, by setting the `forceWriteConnection` flag.

The intent is to allow users to force reads onto a master, if the user assumes that an entity has not been replicated yet.

### DIFF
--- a/docs/docs/read-connections.md
+++ b/docs/docs/read-connections.md
@@ -41,5 +41,5 @@ await em.transactional(async em => {
   a.name = 'test'; // will trigger update on write connection once flushed
 });
 
-const res4 = await em.findOne(Author, 1, { forceMasterRead: true }); // write connection
+const res4 = await em.findOne(Author, 1, { forceWriteConnection: true }); // write connection
 ```
--- a/docs/docs/read-connections.md
+++ b/docs/docs/read-connections.md
@@ -40,4 +40,6 @@ await em.transactional(async em => {
   const a = await em.findOne(Author, 1); // write connection
   a.name = 'test'; // will trigger update on write connection once flushed
 });
+
+const res4 = await em.findOne(Author, 1, { forceMasterRead: true }); // write connection
 ```
--- a/packages/core/src/drivers/IDatabaseDriver.ts
+++ b/packages/core/src/drivers/IDatabaseDriver.ts
@@ -113,6 +113,7 @@ export interface FindOptions<T, P extends string = never> {
   lockMode?: Exclude<LockMode, LockMode.OPTIMISTIC>;
   lockTableAliases?: string[];
   ctx?: Transaction;
+  forceWriteConnection?: boolean;
 }
 
 export interface FindOneOptions<T, P extends string = never> extends Omit<FindOptions<T, P>, 'limit' | 'offset' | 'lockMode'> {
--- a/packages/knex/src/AbstractSqlDriver.ts
+++ b/packages/knex/src/AbstractSqlDriver.ts
@@ -40,7 +40,7 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
     const meta = this.metadata.find<T>(entityName)!;
     const populate = this.autoJoinOneToOneOwner(meta, options.populate as unknown as PopulateOptions<T>[], options.fields);
     const joinedProps = this.joinedProps(meta, populate);
-    const qb = this.createQueryBuilder<T>(entityName, options.ctx, !!options.ctx, false);
+    const qb = this.createQueryBuilder<T>(entityName, options.ctx, !!options.ctx || !!options.forceWriteConnection, false);
     const fields = this.buildFields(meta, populate, joinedProps, qb, options.fields as Field<T>[]);
     const joinedPropsOrderBy = this.buildJoinedPropsOrderBy(entityName, qb, meta, joinedProps);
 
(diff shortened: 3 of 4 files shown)


tests/features/replica-selection.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  beforeAll(async () => {
+    orm = await MikroORM.init({
+      entities: [User],
+      dbName: 'replicaselection',
+      type: 'postgresql',
+      replicas:[ { host:'invalidhostname' } ],
+    });
+    await new SchemaGenerator(orm.em).ensureDatabase();
+    await new SchemaGenerator(orm.em).dropSchema();
+    await new SchemaGenerator(orm.em).createSchema();
```
B4nan, line 31:
```suggestion
    await orm.getSchemaGenerator().refreshDatabase();
```

docs/docs/read-connections.md
```diff
@@ -40,4 +40,6 @@ await em.transactional(async em => {
   const a = await em.findOne(Author, 1); // write connection
   a.name = 'test'; // will trigger update on write connection once flushed
 });
+
+const res4 = await em.findOne(Author, 1, { forceMasterRead: true }); // write connection
```
B4nan, line 44: let's call it `forceWriteConnection`

tests/features/replica-selection.test.ts
```diff
@@ -0,0 +1,46 @@
...
+
+describe('Replica selection', () => {
+  let orm: MikroORM<PostgreSqlDriver>;
+
+  beforeAll(async () => {
+    orm = await MikroORM.init({
+      entities: [User],
+      dbName: 'replicaselection',
+      type: 'postgresql',
+      replicas:[ { host:'invalidhostname' } ],
```
B4nan, line 27:
```suggestion
      replicas: [{ host: 'invalidhostname' }],
```

tests/features/replica-selection.test.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  test('uses master for reads if instructed', async () => {
+    const em = orm.em.fork();
+
+    const user = new User();
+    await em.persistAndFlush(user);
+
+    const readFromMaster = await em.findOne(User, user.id, { forceMasterRead:true });
+
+    expect(readFromMaster).toBe(user);
```
B4nan, line 44: No need for forking, we have the global context check disabled in tests.

But I am not sure what was this this test about. It would not fire any query, `findOne` with a PK condition will just take the entity from identity map. It would be passing even without that flag, and even if you would remove that entity from the database between the calls. 

You should check that it used the write connection, check this test: 

https://github.com/mikro-orm/mikro-orm/blob/master/tests/EntityManager.mysql.test.ts#L1943

Maybe you could just add this to the mysql test instead of creating new one.

B4nan (changes requested):

tests/features/replica-selection.test.ts
oliversalzburg: I missed those specific tests earlier. Good call. I updated the commit and am now waiting for checks to pass.

docs/docs/read-connections.md
B4nan: docs still using the old name

B4nan (changes requested):

B4nan:
please dont rebase after you get a review, so we maintain the history

oliversalzburg:
Are the commits ultimately squashed before merging or does every single commit actually have to make sense?

B4nan:
Yes, only one commit will go to master and the commit message is based on the PR title (and I can override it when merging).

B4nan:
Yeah not a big deal for such a small PR, but I started to say this out load when I see it, otherwise things never change :]

asiFarran:
@B4nan @oliversalzburg, in our application the vast majority of queries should run against the master in order to support immediate-read-after-write scenarios, but we do want to use read-replicas for a few specific use-cases (e.g. customer search or analytics queries).

Using the pattern above we'd have to add `forceWriteConnection : true` to almost all queries which is .... you get it.

I would like to propose the following (more generic) pattern:
1. Add a `useReadReplicasByDefault?: boolean`  to [Configuration](https://github.com/mikro-orm/mikro-orm/blob/ffe6bd886ddc0c97bbfde600c3bec26570f3ca33/packages/core/src/utils/Configuration.ts#L41). This defaults to `true` so no breaking change and existing behavior is maintained.
2. Replace  `forceWriteConnection` on `FindOptions` with a `useConnection?: 'read' | 'write'` property that lets the developer be specific in both directions
3. Use both the configurable default and the optional explicit setting to choose the connection.

This way the connection resolution strategy is completely configurable and can support all scenarios.

Does this sound reasonable to you ? I'm happy to open a PR for it

asiFarran:
@B4nan @oliversalzburg, in our application the vast majority of queries should run against the master in order to support immediate-read-after-write scenarios, but we do want to use read-replicas for a few specific use-cases (e.g. customer search or analytics queries).

Using the pattern above we'd have to add `forceWriteConnection : true` to almost all queries which is .... you get it.

I would like to propose the following (more generic) pattern:
1. Add a `useReadReplicasByDefault?: boolean`  to [Configuration](https://github.com/mikro-orm/mikro-orm/blob/ffe6bd886ddc0c97bbfde600c3bec26570f3ca33/packages/core/src/utils/Configuration.ts#L41). This defaults to `true` so no breaking change and existing behavior is maintained.
2. Replace  `forceWriteConnection` on `FindOptions` with a `useConnection?: 'read' | 'write'` property that lets the developer be specific in both directions
3. Use both the configurable default and the optional explicit setting to choose the connection.

This way the connection resolution strategy is completely configurable and can support all scenarios.

Does this sound reasonable to you ? I'm happy to open a PR for it

B4nan:
Sounds good to me, PR welcome. Given this PR is not yet shipped in stable, I don't mind replacing the newly added `forceWriteConnection` with `useConnection`.

oliversalzburg:
Awesome change btw :)

PR #2848, opened 2022-03-03
fix(knex): `order by` with a formula field should not include `as`

tsangste:
When doing an `order by` with a formula field it is generating it with an `as` which causes an invalid SQL error


B4nan:
Thanks!

PR #2855, opened 2022-03-04
Docs: Change removing to correct past tense: removed

jenssogaard:


B4nan:
Please change it in the next docs too

PR #2887, opened 2022-03-11
fix(core): allow onUpdate access to other onUpdate results from earlier properties

techmunk:
Allow onUpdate to get access to previous entity values returned via onUpdate,
which can be useful for ensuring created and updated dates have the exact same time
or basing the value of one property off the value of another.

This already worked for onCreate, as onCreate directly updates the entity.
The onUpdate hook however stores its values in a map to ensure values are not updated
unless they actually need to be, which means onUpdate would always get property values
not accounting for prior onUpdate hooks.

### DIFF
--- a/packages/core/src/unit-of-work/ChangeSetComputer.ts
+++ b/packages/core/src/unit-of-work/ChangeSetComputer.ts
@@ -73,7 +73,7 @@ export class ChangeSetComputer {
 
     if (prop.onUpdate && type === ChangeSetType.UPDATE) {
       const pairs = map.get(entity) ?? [];
-      pairs.push([prop.name, prop.onUpdate(entity)]);
+      pairs.push([prop.name, prop.onUpdate({ ...entity, ...Object.fromEntries(pairs) })]);
       map.set(entity, pairs);
     }
 
--- a/tests/EntityManager.mongo.test.ts
+++ b/tests/EntityManager.mongo.test.ts
@@ -9,6 +9,7 @@ import { AuthorRepository } from './repositories/AuthorRepository';
 import { initORMMongo, mockLogger } from './bootstrap';
 import FooBar from './entities/FooBar';
 import { FooBaz } from './entities/FooBaz';
+import { UpdateCreatePropAccess } from './entities/UpdateCreatePropAccess';
 
 describe('EntityManagerMongo', () => {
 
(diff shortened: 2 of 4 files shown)


techmunk:
I just realized, that as this no longer provides the whole entity, mutating the entity in an onUpdate is no longer possible which might be an issue? Be interested to hear your feedback on this, as I could just mutate the entity in my code, but then every flush call will make a change to the entity, even if no other property has changed, which onUpdate seems specifically designed around being able to only commit "onUpdate" changes when other non-auto update values change.

B4nan:
I am not interested in doing this, I don't see how it could work propertly. We can't modify the entity directly, as it would always produce queries even if nothing else would be updated. That is the whole point of using the `pairs` thingy. `onUpdate` on property level was never designed for this - its purpose is to allow property initializer for interface-only entities.

Use `@BeforeUpdate` hook instead.

techmunk:
@B4nan thanks for your quick response. Of course it is your call on what is merged in, but I feel from your response that you might have misunderstood what this pull request is doing, or I might not have expressed it clearly enough.

This pull request does not allow mutating the entity directly, and uses the pairs thingy so extra queries are not produced. In fact, this pull request means mutating the entity in an onUpdate hook will now actually not produce a query, where it currently would.

All I have done is pass through the results of onUpdate calls from other properties on the entity object provided to the onUpdate hook.

I feel my second comment might have confused the issue, as I'm advocating against mutating the entity directly, and would like my onUpdate hooks to be able to use values from other properties onUpdate hooks, but still only actually commit the results and run the query if other entities are actually changed.

If I'm mistaken on this, I apologize for the extra noise and thanks for an amazing project!

B4nan:
What I don't like is that you no longer have the entity instance in the argument. While it may sound like a feature to ignore any updates on it, for others it may easily be a breaking change.

As said above, the use case for `onUpdate` is to mimic property initializers, you should not need the entity reference for such. `onUpdate` was never meant to "replace BeforeUpdate hook".

What is the actual use case?

techmunk:
I agree that would be a breaking change, and was something I realized not long after I pushed this request.

The `beforeUpdate` hooks like exactly what I should be using. I saw it in the docs but for some reason though it was global, and not per entity.

One thing I'm struggling with though, is how to define this hook. I'm currently using entity schema with typescript and I can't seem to work out how to define the hook function and I have not been able to find a good example of it in the docs.

The use case I have is to add a microsecond epoch timestamp as a `BIGINT`, as well as a `DATETIME` of the same timestamp. If only javascript date objects support microseconds! haha

B4nan:
One example is this: 

https://github.com/mikro-orm/mikro-orm/blob/master/tests/entities-js/Author3.js#L116

You define the methods on entity class as usual, and just add those method names to `hooks` object in the schema definition.

techmunk:
Oh ok. So it requires a concrete class? I currently only have an interface. I think it should be ok to change to use a class.

B4nan:
I don't remember if its implemented, I remember I was planning to allow also passing function handles instead of function names for this use case. Can't find it in the code now, so probably still missing.

techmunk:
@B4nan just made pull request #2895, is this what you were thinking?

PR #2895, opened 2022-03-11
feat(core): allow using hooks for interface entities

techmunk:


techmunk:
I'm not sure the typing of hooks is 100% correct, but I'm not savvy enough with typescript to get it more accurate.

techmunk:
Thanks @B4nan. I had considered the subscribers, however I was building this functionality onto a base entity and was hoping to keep it contextualized to that entity. Does the `getSubscribedEntities` method of an entity subscriber work for base entities?

B4nan:
With a bit of metadata magic you could automatically list entity types that extends specific base entity. But the approach this PR allows seems to be much cleaner.

B4nan:
Speaking of base entities, it would be probably good idea to add a test that defines a hook on base entity this way too.

B4nan:
And we also need a mention about this new feature to the docs. I would add it to the entity schema docs, as that is the main use case for it.

techmunk:
@B4nan I've pushed an update with some docs written and some tests on a base entity. I did have one strange issue with the base entity that I could not get to the bottom of though....

When the base entity schema was just an object (and not an EntitySchema instance), the hook functions would be in the hooks array, but would be `null`. Changing the schema to use `new EntitySchema()` instead, fixed this issue.

I don't really understand why, but I assume there's something in the entity discovery that works  a bit differently in those two cases.

techmunk:
To be clear, only the functions would be null, the string method names were there and fine.

docs/docs/entity-schema.md
```diff
@@ -183,3 +183,54 @@ export const schema = new EntitySchema<BookTag>({
...
+  }
+}
+
+// Defined outside of entity class - this bound to entity instance
+function beforeUpdate() {
+  this.version++;
+}
+
+// Defined outsite, this available via args.
+const beforeUpdate2 = (args: EventArgs) => args.entity.version++
```
B4nan, line 221:
```suggestion
const beforeUpdate2 = (args: EventArgs) => args.entity.version++;
```

docs/docs/entity-schema.md
```diff
@@ -183,3 +183,54 @@ export const schema = new EntitySchema<BookTag>({
...
+
+export const schema = new EntitySchema<BookTag>({
```
B4nan, line 223: this is inferred from `class`

```suggestion
export const schema = new EntitySchema({
```

packages/core/src/events/EventManager.ts
```diff
@@ -27,12 +27,12 @@ export class EventManager {
   dispatchEvent<T extends AnyEntity<T>>(event: EventType.onInit, args: Partial<EventArgs<T>>): unknown;
   dispatchEvent<T extends AnyEntity<T>>(event: EventType, args: Partial<EventArgs<T> | FlushEventArgs>): Promise<unknown>;
   dispatchEvent<T extends AnyEntity<T>>(event: EventType, args: Partial<EventArgs<T> | FlushEventArgs | TransactionEventArgs>): Promise<unknown> | unknown {
-    const listeners: [EventType, EventSubscriber<T>][] = [];
+    const listeners: (readonly [PropertyKey | NonNullable<EventSubscriber<T>[keyof EventSubscriber<T>]>, EventSubscriber<T> | AnyEntity<T>])[] = [];
```
B4nan, line 30: please extract this into a new type, its too complex and reduces readabilty, same for the `runListener` usage

packages/core/src/events/EventManager.ts
```diff
@@ -42,11 +42,18 @@ export class EventManager {
       }
     }
 
+    const runListener = ([hook, subscriber]: readonly [PropertyKey | NonNullable<EventSubscriber<T>[keyof EventSubscriber<T>]>, EventSubscriber<T> | AnyEntity<T>]) => {
+      if (typeof hook !== 'function') {
+        return subscriber[hook]!(args as (EventArgs<T> & FlushEventArgs & TransactionEventArgs));
```
B4nan, line 47: let's extract `EventArgs<T> & FlushEventArgs & TransactionEventArgs` to a type too

B4nan (changes requested):

techmunk:
@B4nan Thanks. I was actually just sitting down to work on this and saw that you merged it! Cheers for all the hard work you put in!

B4nan:
FYI just finished with simplifying the implementation as well as types a bit: https://github.com/mikro-orm/mikro-orm/commit/f3599292515941d3bef365160b18685eb15227d3

techmunk:
That's a much cleaner implementation. :+1:

PR #2896, opened 2022-03-11
feat(core): Flexible connection resolution strategy when using read-replicas

asiFarran:
Allows developers working with read-replicas the flexibility to:

- Set the default resolution strategy for read operations (resolve to a read-replica or write connection)
- Be explicit on a per-query basis by extending `FindOptions` and `CountOptions` with a `connectionType` property


B4nan:
please remove all unrelated code style changes, and the yarn log file

B4nan:
Note that we need to keep supporting the string constants (otherwise this would be a breaking change on type level), so remove all of those changes too.

asiFarran:
> Note that we need to keep supporting the string constants (otherwise this would be a breaking change on type level), so remove all of those changes too.

I restored the string constants but did wrap them to a union-type so to make signatures shorter & 'safer'

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -620,8 +618,9 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
   }
 
   /** @internal */
-  createQueryBuilder<T extends AnyEntity<T>>(entityName: string, ctx?: Transaction<Knex.Transaction>, write?: boolean, convertCustomTypes?: boolean): QueryBuilder<T> {
-    const qb = new QueryBuilder<T>(entityName, this.metadata, this, ctx, undefined, write ? 'write' : 'read');
+  createQueryBuilder<T extends AnyEntity<T>>(entityName: string, ctx?: Transaction<Knex.Transaction>, prefferedConnectionType?: ConnectionType, convertCustomTypes?: boolean): QueryBuilder<T> {
+    const connectionType = this.resolveConnectionType({ ctx, connectionType: prefferedConnectionType });
```
B4nan, line 622: typo in `preferred`
```suggestion
    const connectionType = this.resolveConnectionType({ ctx, connectionType: preferredConnectionType });
```

tests/EntityManager.mysql.test.ts
```diff
@@ -1940,50 +1938,212 @@ describe('EntityManagerMySql', () => {
     process.env.FORCE_COLOR = '0';
   });
 
-  test('read replicas', async () => {
-    const mock = mockLogger(orm, ['query']);
+  describe('read-replicas',  () => {
```
B4nan, line 1941: let's move this to `tests/features/read-replicas`

B4nan (changes requested):
looking good, just two nits

tests/features/read-replicas.test.ts
```diff
@@ -0,0 +1,232 @@
...
+  afterEach(() => {
+    orm.config.set('debug', false);
+    Author2Subscriber.log.length = 0;
+    EverythingSubscriber.log.length = 0;
+    FlushSubscriber.log.length = 0;
+    Test2Subscriber.log.length = 0;
+  });
+
+  describe('when preferReadReplicas is true (default behaviour)', () => {
+      test('will prefer replicas for read operations outside a transaction', async () => {
```
B4nan, line 25: one last nit: bad indentation all over the new file

PR #2898, opened 2022-03-11
Allow create yargs commands manually

sashaaro:
```
const config = {}
const mikroOrmPromise = async => MikroORM.init(config)
const ormCommands = CLIConfigurator.createOrmCommands(mikroOrmPromise)

const argv: yargs.Argv = yargs(hideBin(process.argv))
argv.command(...)
ormCommands.forEach(cmd => argv.command(cmd))

```


B4nan:
I don't really want to make the CLI extensible, don't see any reason for it - what is your use case?

PR #2901, opened 2022-03-13
feat(core): allow specifying custom pivot table entity

B4nan:
```ts
  @ManyToMany({ entity: () => Product, pivotEntity: () => OrderItem })
  products = new Collection<Product>(this);
```

### DIFF
--- a/packages/core/src/decorators/ManyToMany.ts
+++ b/packages/core/src/decorators/ManyToMany.ts
@@ -29,6 +29,7 @@ export interface ManyToManyOptions<T, O> extends ReferenceOptions<T, O> {
   fixedOrder?: boolean;
   fixedOrderColumn?: string;
   pivotTable?: string;
+  pivotEntity?: string | (() => EntityName<any>);
   joinColumn?: string;
   joinColumns?: string[];
   inverseJoinColumn?: string;
--- a/packages/core/src/drivers/DatabaseDriver.ts
+++ b/packages/core/src/drivers/DatabaseDriver.ts
@@ -191,7 +191,7 @@ export abstract class DatabaseDriver<C extends Connection> implements IDatabaseD
     }
 
     if (prop.fixedOrder) {
-      return [{ [`${prop.pivotTable}.${prop.fixedOrderColumn}`]: QueryOrder.ASC } as QueryOrderMap<T>];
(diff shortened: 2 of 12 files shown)


yackinn:
Is this supposed to add another column specified in the OrderItem entity to the pivot table?
Can this be used for sorting purposes?

B4nan:
Full example is in the docs: https://mikro-orm.io/docs/collections#custom-pivot-table-entity

The `OrderItem` entity represents the pivot table, so yes, it allows you to create additional columns in that table. You can't sort by them from the M:N relation currently, we could allow defining default order by based on that, if you'd like to do it from the `FindOptions` API, you would need to use the old 1:m - m:1 approach instead as in https://mikro-orm.io/docs/composite-keys#use-case-3-join-table-with-metadata

yackinn:
If I add the LocationMedium entity to the entiies array it will error.
`Error: Please provide either 'type' or 'entity' attribute in LocationMedium.location`

If I don't add it to the entities array it doesn't change anything in the pivot table though.

I think that's how it's stated in the docs.

```ts
@Entity()
export class LocationMedium {

  @ManyToOne({ primary: true })
  location: Location;

  @ManyToOne({ primary: true })
  medium: Medium;

  @Property({ default: 1 })
  prop: number;

}


// location.entity.ts
@Entity()
export class Location extends OmputBaseEntity {
  @Property()
  title: string;
  ...

  @OneToMany(() => Event, courseDate => courseDate.location, { nullable: true })
  event?: Event[];

  @ManyToMany({ entity: () => Medium, nullable: true, pivotEntity: () => LocationMedium })
  media?: Medium[];
}
```


Bildschirmfoto 2022-03-14 um 12 47 25

B4nan:
Note that this is wrong too, you need to use `Collection` wrapper:

```
  @OneToMany(() => Event, courseDate => courseDate.location, { nullable: true })
  event?: Event[]; // should be `new Collection<Event>(this)`

  @ManyToMany({ entity: () => Medium, nullable: true, pivotEntity: () => LocationMedium })
  media?: Medium[]; // should be `new Collection<Medium>(this)`
```

yackinn:
Thank you. 
I should have tried that.

The version without the Collection type also works though. The table is update with the new column.

By the way I noticed the search functionality is broken on the 5.1 docs.

PR #2912, opened 2022-03-14
cli for schema generator should use orm config more

m11m:
There are a number of issues with [postgres and session_replication_role](https://github.com/mikro-orm/mikro-orm/search?q=session_replication_role&type=issues)

I think I found another one here, I could not use the recommended fixes (add disableForeignKeys to orm config) with success.

In my case, I needed the following code fix, then I could run cli command like `schema:create` without the `--fk-checks`  cli flag and it would use the config specified in orm configuration like:

```
{
  ...
  schemaGenerator: {
    disableForeignKeys: false,
  }
}
```  

We can see why we might want this to be 'tri state' at the place I made the change by looking what happens down in the [schema generator](https://github.com/mikro-orm/mikro-orm/blob/master/packages/knex/src/schema/SchemaGenerator.ts#L61)

### DIFF
--- a/packages/cli/src/commands/SchemaCommandFactory.ts
+++ b/packages/cli/src/commands/SchemaCommandFactory.ts
@@ -97,7 +97,8 @@ export class SchemaCommandFactory {
 
     const orm = await CLIHelper.getORM() as MikroORM<AbstractSqlDriver>;
     const generator = new SchemaGenerator(orm.em);
-    const params = { wrap: !args.fkChecks, ...args };
+    let params = args.fkChecks !== undefined ? {wrap: !args.fkChecks} : {};
+    params = {...params, ...args};
 
     if (args.dump) {
       const m = `get${method.substr(0, 1).toUpperCase()}${method.substr(1)}SchemaSQL` as 'getCreateSchemaSQL' | 'getUpdateSchemaSQL' | 'getDropSchemaSQL';


packages/cli/src/commands/SchemaCommandFactory.ts
```diff
@@ -97,7 +97,8 @@ export class SchemaCommandFactory {
 
     const orm = await CLIHelper.getORM() as MikroORM<AbstractSqlDriver>;
     const generator = new SchemaGenerator(orm.em);
-    const params = { wrap: !args.fkChecks, ...args };
+    let params = args.fkChecks !== undefined ? {wrap: !args.fkChecks} : {};
+    params = {...params, ...args};
```
B4nan, line 101: i dont understand how this can fix anything, it should do exactly the same thing is the previous version, but more verbose/less readable. 

more importantly, if there is a bug, we need a test case

B4nan (changes requested):

packages/cli/src/commands/SchemaCommandFactory.ts
m11m: Yes, I apologize for not knowing about the rules, such as needing a test case. I don't know much about this project and was only trying to learn a little about it when I found the issue.

PR #2929, opened 2022-03-18
fix(knex): `order by` with a formula field should not include `as` for sub-queries

tsangste:
follow on from #2848, noticed for sub-queries it was still adding `as` to formula field on `order by`

### DIFF
--- a/packages/knex/src/query/QueryBuilderHelper.ts
+++ b/packages/knex/src/query/QueryBuilderHelper.ts
@@ -505,7 +505,7 @@ export class QueryBuilderHelper {
 
       Utils.splitPrimaryKeys(field).forEach(f => {
         const prop = this.getProperty(f, alias);
-        const noPrefix = (prop && prop.persist === false) || QueryBuilderHelper.isCustomExpression(f);
+        const noPrefix = (prop && prop.persist === false && !prop.formula) || QueryBuilderHelper.isCustomExpression(f);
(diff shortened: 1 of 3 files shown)


packages/knex/src/query/QueryBuilderHelper.ts
```diff
@@ -505,7 +505,7 @@ export class QueryBuilderHelper {
 
       Utils.splitPrimaryKeys(field).forEach(f => {
         const prop = this.getProperty(f, alias);
-        const noPrefix = (prop && prop.persist === false) || QueryBuilderHelper.isCustomExpression(f);
+        const noPrefix = (prop && prop.persist === false && !prop.formula) || QueryBuilderHelper.isCustomExpression(f);
```
tsangste, line 508: need advice on this one, I added this because on the test it the order was coming out as `e0.priceTaxed` and not the formula and was able to get my test working by doing this

B4nan: formulas should be marked with `persist: false` already, i dont mind the explicit check

B4nan (approved):

Lemas97:
Btw, there is the same issue on group by

B4nan:
https://github.com/mikro-orm/mikro-orm/commit/e27e4b907154933feba985badd4f6a60dee06317 should help with that

PR #2943, opened 2022-03-23
docs: provide more details on `wrappedReference`

SpencerKaiser:
Updated several pages to add details and add emphasis around the use of the `wrappedReference` option for defining entities. I also added some extra details and emphasis on pages where the lack of this option might cause issues.

I know verbiage for docs is super important; please let me know if you'd like to modify/remove any of what I added 😉 

I also think I got the link right in `query-builder.md`, but double check me.

### DIFF
--- a/docs/versioned_docs/version-5.1/query-builder.md
+++ b/docs/versioned_docs/version-5.1/query-builder.md
@@ -160,8 +160,7 @@ console.log(qb.getQuery());
 // order by `e1`.`tags` asc
 ```
 
-This is currently available only for filtering (`where`) and sorting (`orderBy`), only the root entity will be selected. To populate its relationships, you can
-use [`em.populate()`](nested-populate.md).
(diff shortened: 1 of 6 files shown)


docs/versioned_docs/version-5.1/query-builder.md
```diff
@@ -160,8 +160,7 @@ console.log(qb.getQuery());
 // order by `e1`.`tags` asc
 ```
 
-This is currently available only for filtering (`where`) and sorting (`orderBy`), only the root entity will be selected. To populate its relationships, you can
-use [`em.populate()`](nested-populate.md).
```
SpencerKaiser, line 164: I didn't delete this line; there was a line break for some reason so I moved it up to the end of the line above.

B4nan:
looking good, please also update the next docs (`./docs/docs`)

SpencerKaiser:
> looking good, please also update the next docs (./docs/docs)
@B4nan ✅

PR #2979, opened 2022-04-01
fix(core): nativeInsert bug when using customTypes

asiFarran:
Trying to use `em.nativeInsert` with an entity that has a _custom type_ property results in errors:

When inserting we need to specify `convertCustomTypes: false` in the options or  `convertToDatabaseValue` will end up being fed its own output leading to errors.

When done through the `uow`, the `changePersister` (which talks directly to the driver) does it, but if we're using `em.nativeInsert` (from user-land) the `InsertOptions` type used in the method wont allow us to set it.

Changing the options type from `InsertOptions` to `NativeInsertUpdateOptions` helps.

Then, it appears that in a recent refactor, the logic when dealing with the  multiple parameter scenario was broken - `prepareEntity` was being applied to the wrong parameter!


B4nan (changes requested):
Thanks, looking good:

- lets fix this in the entity repository too
- lets remove the `InsertOptions` type as it wont be used anywhere

PR #3007, opened 2022-04-12
fix(core): use schemaGenerator option disableForeignKeys when dropping database

### LINKED ISSUE #3004: CLI command "fresh" fails to execute when disableForeignKeys is set to false
### Description
When trying to execute the command `npx mikro-orm migration:fresh` it fails and throws an error.
### Stack trace
```
DriverException: set session_replication_role = 'replica'; - permission denied to set parameter "session_replication_role"
    at PostgreSqlExceptionConverter.convertException (my-project/node_modules/@mikro-orm/core/platforms/ExceptionConverter.js:8:16)
    at PostgreSqlExceptionConverter.convertException (my-project/node_modules/@mikro-orm/postgresql/PostgreSqlExceptionConverter.js:42:22)
    at PostgreSqlDriver.convertException (my-project/node_modules/@mikro-orm/core/drivers/DatabaseDriver.js:173:54)
    at my-project/node_modules/@mikro-orm/core/drivers/DatabaseDriver.js:177:24
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async SchemaGenerator.execute (my-project/node_modules/@mikro-orm/knex/schema/SchemaGenerator.js:350:13)
    at async SchemaGenerator.dropSchema (my-project/node_modules/@mikro-orm/knex/schema/SchemaGenerator.js:77:9)
    at async Function.handleFreshCommand (my-project/node_modules/@mikro-orm/cli/commands/MigrationCommandFactory.js:164:9)
    at async Function.handleMigrationCommand (my-project/node_modules/@mikro-orm/cli/commands/MigrationCommandFactory.js:110:17)
```
### Configuration
```
const config: Options = {
  ...,
  migrations: {
    ...,
    disableForeignKeys: false,
  },
  schemaGenerator: {
    disableForeignKeys: false,
  },
  ...
};
```
### To reproduce
Execute command `npx mikro-orm migration:fresh` with any schema using PostgreSQL and setting `schemaGenerator.disableForeignKeys = true` and/or `migrations.disableForeignKeys = true`.
### Additional context
PostgreSQL
### Versions

| Dependancy | Version |
| --- | --- |
| node | 16.10.0 |
| typescript | 4.3.5 |
| mikro-orm | 5.1.2 |
| your-driver | pg |

luksiv:
closes #3004

### DIFF
--- a/packages/knex/src/schema/SchemaGenerator.ts
+++ b/packages/knex/src/schema/SchemaGenerator.ts
@@ -82,7 +82,7 @@ export class SchemaGenerator extends AbstractSchemaGenerator<AbstractSqlDriver> 
   }
 
   async dropSchema(options: { wrap?: boolean; dropMigrationsTable?: boolean; dropDb?: boolean; schema?: string } = {}): Promise<void> {
-    options.wrap ??= true;
+    options.wrap ??= this.options.disableForeignKeys ?? true;
 
     if (options.dropDb) {
       const name = this.config.get('dbName')!;
--- a/tests/issues/GH3004.test.ts
+++ b/tests/issues/GH3004.test.ts
@@ -1,0 +1,54 @@
+import { Entity, MikroORM, PrimaryKey } from '@mikro-orm/core';
+import { SchemaGenerator } from '@mikro-orm/knex';
+
+@Entity()
+export class User {
+
+  @PrimaryKey()
+  id: number;
+
+  constructor(id: number) {
+    this.id = id;
+  }
+
+}
+
+describe('GH issue 3004', () => {
+
+  let orm: MikroORM | undefined;
+
+  const getDropSchemaSQLSpy = jest.spyOn(SchemaGenerator.prototype, 'getDropSchemaSQL');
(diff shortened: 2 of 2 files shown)


luksiv:
@B4nan could you re-trigger tests, because yesterday Github Actions were down.

B4nan:
Well, the tests ran, and timed out after 6 hours because of one broken assertion that caused the tests to hang (as no `orm.close()` was called as the test failed). This is something to be fixed, rerunning the tests won't help. There are test failures in `tests/features/schema-generator/SchemaGenerator.mysql2.test.ts` because of this change, and maybe elsewhere too. 

The problem is in circular FK between author2 and book2 tables (and probably other similar issues in other relations), I believe we actually need to disable the FK checks for this use case in mysql (dropping tables). The alternative would be to first drop the FKs, which actually sounds quite reasonable, but should be rather a separate PR.

luksiv:
If you can try to resolve this, I would really appreciate it, since this is a little bit too advanced for me to try to do without pouring a lot of hours into trying to fix this 😄

B4nan:
Yeah sure, it should be rather easy, I can look into that myself when I find some spare time.

PR #3023, opened 2022-04-17
feat(mssql): working mssql driver

UTGuy:
All tests are passing

image


B4nan:
Just realiazed you are targetting my `mssql` branch, perfect, lets merge it right ahead so I can play with that easier. I just shipped 5.1.4 stable, so would like to look into this now and get it out finally in 5.2

Thanks a lot for this!

packages/mssql/src/MsSqlConnection.ts
```diff
@@ -28,10 +29,11 @@ export class MsSqlConnection extends AbstractSqlConnection {
     config.options = {
       ...(options as any),
       ...(config.options || {}),
+      database: config.database,
     };
 
-    // TODO is this ok? we should select the db afterwards?
-    delete (config as any).database;
+    // TODO is this ok? we should select the db afterwards? - No, don't do this - Michael
```
B4nan, line 35: I really do believe we need to do something like this. This change breaks tests locally on my end, as the database does not exist. Trying to run the tests locally after merge (made some small changes along the way but nothing related I believe), and I cant get pass `Login failed for user 'sa'. Reason: Failed to open the explicitly specified database 'mikro_orm_test'. [CLIENT: 172.20.0.1]`, because the database don't exist yet probably. 

How come it does work in the CI, is there something I am missing, or can it be something specific for the ARM build of mssql?

PR #3031, opened 2022-04-19
fix(core): respect mapToPk when expanding properties

asiFarran:
When using references that are mapped to Primary Key (`mapToPk`), the expanding will result in an error as we are actually dealing with a scalar value.

Adding this constraint into the expansion logic.


B4nan (approved):
thanks, lgtm!

PR #3052, opened 2022-04-22
fix(core): updating JSON properties should be shallow by default

jbmikk:
I'm not really sure this is a bug, but it started to happen to me after upgrading to V5.

This happens when updating a plain JSON property. I expect the full JSON to be replaced, but I'm getting merged results when reading the JSON back from the database. Look at the failing test, `t1` should not be defined.

I managed to fix this with `{ mergeObjects: false }` but it doesn't feel right. 

According to the V5 docs, the default should be `false` already, so I shouldn't have to do this.

Is this a bug or are the docs outdated?

### DIFF
--- a/tests/issues/GHx3.test.ts
+++ b/tests/issues/GHx3.test.ts
@@ -1,0 +1,71 @@
+import { Entity, IdentifiedReference, JsonType, ManyToOne, MikroORM, OneToOne, PrimaryKey, PrimaryKeyType, Property, wrap } from '@mikro-orm/core';
+import type { PostgreSqlDriver } from '@mikro-orm/postgresl';
+
+interface Test {
+  t1: string;
+  t2: {
+    t3: string | null;
+    t4?: {
+      t5: string;
+    };
+  };
+}
+
+@Entity()
+export class A {
+
+  @PrimaryKey({ type: Number })
+  id!: number;
+
+  @Property({ type: JsonType })
+  test: Test;
+
+}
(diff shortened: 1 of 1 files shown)


B4nan:
> According to the V5 docs, the default should be false already, so I shouldn't have to do this.

It was false in v4, this is one of the BCs from v5.

https://mikro-orm.io/docs/upgrading-v4-to-v5#changes-in-assign-helper

B4nan:
I believe the reason why I did it was assigning to embeddables, as there it feels like the right thing to do (as it behaves the same as if you assign to an entity). I am not sure if its a good idea to have different behaviour when assigning to an embedded or to a JSON property. Let's see if there will be more confusion, we can revert it. Will fix the docs.

jbmikk:
@B4nan I agree, for embedables it makes more sense. I'm not sure if it's possible to have the behavior be different depending on the property type maybe (JSON vs embedables).

Or maybe have some sort of decorator or helper to indicate intent?
```
em.assign({ columnName: literal({ ... }) })
```
It's kind of ugly, but then you can indicate what to do for each property independently.

jbmikk:
Maybe define the default merging strategy for each column on the `@Property()` decorator?

B4nan:
We can make the behaviour different based on metadata, that's not a problem. So we could have the current behaviour only for assigning to embedded properties and keep the v4 behaviour for regular JSON properties.

PR #3064, opened 2022-04-25
fix(seeder): fix type of Factory methods

### LINKED ISSUE #3062: Factory / Seeder: makeOne & createOne expect wrong type for IdentifiedReference (wrappedReference)
**Describe the bug**
We used the syntax described in [the docu: "Reference Wrapper"](https://mikro-orm.io/docs/entity-references#better-type-safety-with-referencet-wrapper) and now we cannot satisfy the typeScript types anymore when creating data in our seeder.

**To Reproduce**

```ts
@Entity()
export class Child {
    @ManyToOne({
        entity: () => Parent,
        wrappedReference: true, // new
    })
    parent: IdentifiedReference<Parent>; // before we had: `parend: Parent` and it worked with the Seeders
}

const child = new ChildFactory(orm.em).make({
    ...otherProps,
    parent: parent, // TS Error
});

// -> TS Error:
// Type 'Parent' is not assignable to type '{ id: number; } & Reference<Parent>'.
// Type 'Parent' is missing the following properties from type 'Reference<Parent>': entity, load, set, unwrap, and 5 more.
```

**The same with:**
- `Factory.create`
- `Factory.createOne`
- `Factory.make`
- `Factory.makeOne`

**Expected behaviour**
I want to be able to create an entity with a ManyToOne Relationship by simply passing the other entity or its id.
Just like it works when calling:

```ts
this.orm.em.create(Child, {...otherProps, parent});
```

**Suggestion**
We could give the methods in Factory<T> similar types as entityManager.create:

```diff
- makeOne(overrideParameters?: Partial<T>): T;
+ makeOne(overrideParameters?: RequiredEntityData<T>): T;
```

**Versions**

| Dependency | Version |
| - | - |
| mikro-orm | 5.1.1 |

tobiaskraus:
As asked by [B4nan](https://github.com/B4nan) in related issue, I opened this PR.

Closes #3062

### Description

See issue above

### Tests

Only types were changed, so no useful test which I can think of can cover that besides TypeScript itself (`yarn tsc-check-tests`). Existing tests are running successful.

### Docu

Only [here](https://mikro-orm.io/docs/seeding#entity-factories) the documentation could be updated, but I think for the majority of people a simple partial is enough - so let's keep it how it is instead of:

```diff
- definition(faker: Faker): Partial<Author>
+ definition(faker: Faker): Partial<RequiredEntityData<Author>>
```

### DIFF
--- a/packages/seeder/src/Factory.ts
+++ b/packages/seeder/src/Factory.ts
@@ -9,9 +9,9 @@ export abstract class Factory<T> {
 
   constructor(private readonly em: EntityManager) { }
 
-  protected abstract definition(faker: Faker): Partial<T>;
+  protected abstract definition(faker: Faker): Partial<RequiredEntityData<T>>;
 
-  private makeEntity(overrideParameters?: Partial<T>): T {
+  private makeEntity(overrideParameters?: Partial<RequiredEntityData<T>>): T {
     const entity = this.em.create(this.model, {
       ...this.definition(faker),
       ...overrideParameters,
@@ -28,7 +28,7 @@ export abstract class Factory<T> {
    * Make a single entity and persist (not flush)
    * @param overrideParameters Object specifying what default attributes of the entity factory should be overridden
    */
-  makeOne(overrideParameters?: Partial<T>): T {
+  makeOne(overrideParameters?: Partial<RequiredEntityData<T>>): T {
     const entity = this.makeEntity(overrideParameters);
     this.em.persist(entity);
     return entity;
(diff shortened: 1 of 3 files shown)


B4nan:
thanks!

packages/seeder/src/Factory.ts
```diff
@@ -9,9 +9,9 @@ export abstract class Factory<T> {
 
   constructor(private readonly em: EntityManager) { }
 
-  protected abstract definition(faker: Faker): Partial<T>;
+  protected abstract definition(faker: Faker): Partial<RequiredEntityData<T>>;
```
B4nan, line 12: oops, i got distracted and didnt really review this one :D 

we should use just `EntityData` instead of `Partial<RequiredEntityData>`, it should do exactly the same. will fix it in a minute

PR #3084, opened 2022-05-04
fix(core): subsribers should receive m to n collection changes

jbmikk:
This is just a faling test for now. I am using v5.1.

I have a subscriber I need to keep track of all entity changes and log them to a separate system.

The problem I have is on the `afterFlush` event, the payload in change sets don't contain the changes for the m to n collection.

Basically after doing this:

```
    await orm.em.transactional(em => {
      const user1 = em.create(User, { id: 1, name: 'User 1' });
      const user2 = em.create(User, { id: 2, name: 'User 2' });
      const project = em.create(Project, { id: 1, name: 'Project 1', users: [user1, user2] });
      em.persist(user1);
      em.persist(user2);
      em.persist(project);
    });
```

I get everything in the payloads, except the `users` collection. Also there is no separate change set for the collection, so it seems to be completely ignored.

I'm not sure if m to n collections should be reported in the parent entity payload or as a separate change set (because it's a separate table) but I think they should be reported somehow.

### DIFF
--- a/tests/issues/GHx3.test.ts
+++ b/tests/issues/GHx3.test.ts
@@ -1,0 +1,88 @@
+import type { ChangeSet, EventSubscriber, FlushEventArgs } from '@mikro-orm/core';
+import { Collection, Entity, ManyToMany, MikroORM, PrimaryKey, Property } from '@mikro-orm/core';
+import type { SqliteDriver } from '@mikro-orm/sqlite';
+
+@Entity()
+class User {
+
+  @PrimaryKey()
+  id: number;
+
+  @Property()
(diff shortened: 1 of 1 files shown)


B4nan:
This is by design, changes to collections are not part of change set, they are diffed separately. You can get the colletion updates from via `uow.getCollectionUpdates()`.

jbmikk:
Right, I'm closing this.

So the only changes I should get with `uow.getCollectionUpdates()` are `m:n` changes right?

By the way, what is `uow.getExtraUpdates()` for?

PR #3085, opened 2022-05-04
fix(core): improve validation of wrong entity references

hcharley:
I haven't created a bug ticket for this, but I did explain the problem in [this Slack thread](https://mikroorm.slack.com/archives/CFTFEK39D/p1651689602208229).

### DIFF
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -441,8 +441,7 @@ export class MetadataDiscovery {
 
       if (value instanceof Function) {
         const meta2 = this.metadata.get(prop.type);
-        
-        prop[type] = value(meta2.properties)?.name ?? null;
+        prop[type] = value(meta2.properties)?.name;
 
         if (prop[type] == null) {
           throw MetadataError.fromWrongReference(meta, prop, type as 'mappedBy' | 'inversedBy');
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -441,7 +441,7 @@ export class MetadataDiscovery {
 
       if (value instanceof Function) {
         const meta2 = this.metadata.get(prop.type);
-        
+        prop[type] = value(meta2.properties)?.name;
         const returnedValue = value(meta2.properties);
         
         if (returnedValue) {
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -441,14 +441,10 @@ export class MetadataDiscovery {
 
       if (value instanceof Function) {
         const meta2 = this.metadata.get(prop.type);
-        prop[type] = value(meta2.properties)?.name;
-        const returnedValue = value(meta2.properties);
         
-        if (returnedValue) {
-          prop[type] = returnedValue.name;
-        }
+        prop[type] = value(meta2.properties)?.name ?? null;
 
-        if (!returnedValue || prop[type] == null) {
+        if (prop[type] == null) {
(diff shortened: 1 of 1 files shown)


packages/core/src/metadata/MetadataDiscovery.ts
```diff
@@ -441,9 +441,14 @@ export class MetadataDiscovery {
 
       if (value instanceof Function) {
         const meta2 = this.metadata.get(prop.type);
-        prop[type] = value(meta2.properties).name;
+
```
B4nan, line 444: isn't this the exact same, but with literally "1 char addition" instead of 5 LoC diff? (note the optional chanining)
```suggestion
        prop[type] = value(meta2.properties)?.name;
```

B4nan (changes requested):
let's simplify and more importantly, add a test case - this is coming from wrong decorator usage, right? so can be just a unit test in https://github.com/mikro-orm/mikro-orm/blob/master/tests/decorators.test.ts

hcharley:
@B4nan I made your suggested change.

I'm going to hold off on writing a test for this until I can fix my own bug.

> this is coming from wrong decorator usage, right? 

I'm still running into the underlying error. So I do not know where this is coming from. Which is a part of the reason why I'm not in a position to write a test.

hcharley:
Dropping this in hopes this could possibly reproduce the issue later:

```ts
import { Field, ObjectType } from '@nestjs/graphql';

import {
  Collection,
  Entity,
  IdentifiedReference,
  ManyToOne,
  OneToMany,
} from '@mikro-orm/core';

import { MyBaseEntity } from '@my/entities';

@ObjectType()
@Entity()
export class OpsManagerService extends MyBaseEntity <OpsManagerService> {
  @OneToMany({
    entity: () => OpsManagerEnvironmentService,
    mappedBy: 'opsManagerService',
  })
  public opsManagerEnvironmentServices = new Collection<OpsManagerEnvironmentService>(
    this
  );
}

@ObjectType()
@Entity()
export class OpsManagerEnvironmentService extends YellowBrimBaseEntity<OpsManagerEnvironmentService> {
  @ManyToOne({
    entity: () => OpsManagerService,
    eager: true,
    wrappedReference: true,
    inversedBy: 'opsManagerEnvironmentServices',
  })
  @Field(() => OpsManagerService, { nullable: false })
  opsManagerService: IdentifiedReference<OpsManagerService>;
}
```

hcharley:
Hmmm.... seems that Mikro has some inconsistencies with its path-to-class attribute. I'm console logging out everything that comes into the validate bidirectional functions, and I'm seeing two different path types:

One entity's path is the bundle: `main.js`

The other's path is a webpack path: `webpack:/ops-manager/libs/domain/src/lib/ops-manager-environment-service/entities/ops-manager-environment-service.entity.ts`

packages/core/src/metadata/MetadataDiscovery.ts
```diff
@@ -441,7 +441,8 @@ export class MetadataDiscovery {
 
       if (value instanceof Function) {
         const meta2 = this.metadata.get(prop.type);
-        prop[type] = value(meta2.properties).name;
+        
+        prop[type] = value(meta2.properties)?.name ?? null;
```
B4nan, line 445: we really need just that one char change i proposed, there is no need for explicit `null` fallback

```suggestion
        prop[type] = value(meta2.properties)?.name;
```

B4nan:
I was asking for a unit test case, not sure what that (very incomplete) entity definition should reproduce... Also hearing about webpack, the path to entity is irrelevant in there.

Let's merge, the change is definitely not problematic, it can't break anything, so I am fine merging without tests.

hcharley:
@B4nan I understood your request, but I didn't have time yesterday to task switch, as I am still attempting to debug the underlying error that I ran into. I'm only seeing it in production, and so I was just dropping notes above on what I think might be causing it.

Glad it was okay enough to merge.

PR #3088, opened 2022-05-06
feat(core): add `strict` option to `em.findOneOrFail()`

### LINKED ISSUE #3051: Add a method for finding strictly 1 entity
Often when we query the database the condition that we would like to express is "does there exist a single entity that fits these criteria". Currently `EntityManager` has a `.findOneOrFail` method which solves half of the problem - it will throw in case there are no entities but will happily return the first one in case there are multiple, meaning that real name of the method is probably closer to `.findFirstOrFail`.

I would like to be able to use `em.findSingleOrFail` or even `em.single` method. Signature of the method should be the same as for the current `.findOneOrFail`. This method should throw in case there are 0 or 2+ entities instead of 1.

We have 2 major options:
1. Always use `.find(..., {limit: 2})` and check for `length === 0 || length > 1` in an external helper (this is our current approach)
2. Monkey-patch `em` with the method that we need that will call `.find` internally

**Additional context**
I think that documentation from similar C# methods can be useful to consider:
[Enumerable.Single](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.single?view=net-6.0)
[Enumerable.First](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.first?view=net-6.0)

williamwdoyle:
With `strict: true` we also throw in case there are more than one result found.

Closes #3051


packages/core/src/drivers/IDatabaseDriver.ts
```diff
@@ -123,6 +123,7 @@ export interface FindOneOptions<T, P extends string = never> extends Omit<FindOp
 
 export interface FindOneOrFailOptions<T, P extends string = never> extends FindOneOptions<T, P> {
   failHandler?: (entityName: string, where: Dictionary | IPrimaryKey | any) => Error;
+  single?: boolean;
```
B4nan, line 126: I'd call this `strict` instead of `single`, as `single` and `one` seems like synonyms to me.

packages/core/src/EntityManager.ts
```diff
@@ -396,14 +396,24 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
...
-  async findOneOrFail<T extends AnyEntity<T>, P extends string = never>(entityName: EntityName<T>, where: FilterQuery<T>, options: FindOneOrFailOptions<T, P> = {}): Promise<Loaded<T, P>> {
-    const entity = await this.findOne(entityName, where, options);
+   async findOneOrFail<T extends AnyEntity<T>, P extends string = never>(entityName: EntityName<T>, where: FilterQuery<T>, options: FindOneOrFailOptions<T, P> = {}): Promise<Loaded<T, P>> {
+    let entity;
+    let isSingleViolation = false;
 
-    if (!entity) {
-      options.failHandler = options.failHandler || this.config.get('findOneOrFailHandler');
+    if (options.single) {
+      const ret = await this.find(entityName, where, options as FindOptions<T, P>);
```
B4nan, line 408: we need to enforce `limit: 2` here

packages/core/src/EntityManager.ts
```diff
@@ -396,14 +396,24 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
...
   /**
    * Finds first entity matching your `where` query. If nothing found, it will throw an error.
+   * If the `single` option is specified and nothing is found or more than one matching entity is found, it will throw an error.
    * You can override the factory for creating this method via `options.failHandler` locally
-   * or via `Configuration.findOneOrFailHandler` globally.
+   * or via `Configuration.findOneOrFailHandler` (`findOneOrFailHandlerSingle` when specifying `single`) globally.
    */
-  async findOneOrFail<T extends AnyEntity<T>, P extends string = never>(entityName: EntityName<T>, where: FilterQuery<T>, options: FindOneOrFailOptions<T, P> = {}): Promise<Loaded<T, P>> {
-    const entity = await this.findOne(entityName, where, options);
+   async findOneOrFail<T extends AnyEntity<T>, P extends string = never>(entityName: EntityName<T>, where: FilterQuery<T>, options: FindOneOrFailOptions<T, P> = {}): Promise<Loaded<T, P>> {
```
B4nan, line 403: bad ident
```suggestion
  async findOneOrFail<T extends AnyEntity<T>, P extends string = never>(entityName: EntityName<T>, where: FilterQuery<T>, options: FindOneOrFailOptions<T, P> = {}): Promise<Loaded<T, P>> {
```

packages/core/src/errors.ts
```diff
@@ -233,4 +233,8 @@ export class NotFoundError<T extends AnyEntity = AnyEntity> extends ValidationEr
     return new NotFoundError(`${name} not found (${inspect(where)})`);
   }
 
+  static findOneFailedSingle(name: string, where: Dictionary | IPrimaryKey): NotFoundError {
```
B4nan, line 236: I'd call this `findExactlyOneFailed` instead

packages/core/src/errors.ts
```diff
@@ -233,4 +233,8 @@ export class NotFoundError<T extends AnyEntity = AnyEntity> extends ValidationEr
...
+    return new NotFoundError(`Multiple ${name} entities were found (${inspect(where)})`);
```
B4nan, line 237: the error message is wrong, it will be used also if no results were found, it needs to be more generic, like "Wrong number of ${name} entities found for query ${inspect(where)}, expected exactly one"

tests/issues/GH3051.test.ts
```diff
@@ -0,0 +1,66 @@
+import { BigIntType, Collection, Entity, ManyToOne, MikroORM, OneToMany, PrimaryKey, Property } from '@mikro-orm/core';
```
B4nan, line 1: this should not be an issue test, but rather feature test. issue tests are for reproducing bugs, not for testing new features

the test can stay as is, just move it to the `tests/features` directory and rename the file so we can tell what its testing

tests/issues/GH3051.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      dbName: ':memory:',
+      type: 'sqlite',
+    });
+
+    await orm.getSchemaGenerator().createSchema();
+    const author = orm.em.create(Author, {
+      name: 'Jon Snow',
+      books: [{ title: 'b1' }, { title: 'b2' }, { title: 'b3' }],
+    });
+    await orm.em.fork().persistAndFlush(author);
```
B4nan, line 47: this feels wrong, your `em.create` call will be creating references for the books, they will stay in the identity map. 
```suggestion
    const author = orm.em.create(Author, {
      name: 'Jon Snow',
      books: [{ title: 'b1' }, { title: 'b2' }, { title: 'b3' }],
    });
    await orm.em.persistAndFlush(author);
    orm.em.clear();
```

B4nan (changes requested):
thanks, overall it looks good, i'd just improve the naming

B4nan:
No need to squash things, the PR will be squash merged. It's actually the opposite, dont squash so its clear what changes you made after the review.

williamwdoyle:
Saw your comment after. I had squashed :(. I will fix that.

packages/core/src/EntityManager.ts
```diff
@@ -396,14 +396,24 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
...
+    if (options.strict) {
+      const ret = await this.find(entityName, where, { ...options as FindOptions<T, P>, limit: 2 });
+      isStrictViolation = ret.length !== 1;
+      entity = ret[0];
+    } else {
+      entity = await this.findOne(entityName, where, options);
+    }
+
+    if (!entity || isStrictViolation) {
+      options.failHandler ??= options.strict ? this.config.get('findExactlyOneOrFailHandler') : this.config.get('findOneOrFailHandler');
```
B4nan, line 416:
```suggestion
      const key = options.strict ? 'findExactlyOneOrFailHandler' : 'findOneOrFailHandler';
      options.failHandler ??= this.config.get(key);
```

packages/core/src/EntityManager.ts
```diff
@@ -396,14 +396,24 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
...
   /**
    * Finds first entity matching your `where` query. If nothing found, it will throw an error.
+   * If the `single` option is specified and nothing is found or more than one matching entity is found, it will throw an error.
    * You can override the factory for creating this method via `options.failHandler` locally
-   * or via `Configuration.findOneOrFailHandler` globally.
+   * or via `Configuration.findOneOrFailHandler` (`findOneOrFailHandlerSingle` when specifying `single`) globally.
    */
   async findOneOrFail<T extends AnyEntity<T>, P extends string = never>(entityName: EntityName<T>, where: FilterQuery<T>, options: FindOneOrFailOptions<T, P> = {}): Promise<Loaded<T, P>> {
-    const entity = await this.findOne(entityName, where, options);
+    let entity;
```
B4nan, line 404:
```suggestion
    let entity: T | null;
```

packages/core/src/EntityManager.ts
```diff
@@ -396,14 +396,24 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
 
   /**
    * Finds first entity matching your `where` query. If nothing found, it will throw an error.
+   * If the `single` option is specified and nothing is found or more than one matching entity is found, it will throw an error.
    * You can override the factory for creating this method via `options.failHandler` locally
-   * or via `Configuration.findOneOrFailHandler` globally.
+   * or via `Configuration.findOneOrFailHandler` (`findOneOrFailHandlerSingle` when specifying `single`) globally.
```
B4nan, line 401: wrong names (single -> strict)

packages/core/src/EntityManager.ts
```diff
@@ -396,14 +396,24 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
 
   /**
    * Finds first entity matching your `where` query. If nothing found, it will throw an error.
+   * If the `single` option is specified and nothing is found or more than one matching entity is found, it will throw an error.
```
B4nan, line 399: wrong names (single -> strict)

B4nan (changes requested):

packages/core/src/EntityManager.ts
williamwdoyle: Missed this one. Need to add it.

packages/core/src/EntityManager.ts
williamwdoyle: And this after unsquashing

packages/core/src/EntityManager.ts
williamwdoyle: Originally, I didn't type this because the TypeScript compiler didn't like the `null` return value without changing the method return type. 

Ive made this `let entity: Loaded<T, P> | null` and changed the return type to be `Promise<Loaded<T, P> | null>` to match what is done in the `findOne` method

packages/core/src/EntityManager.ts
```diff
@@ -396,14 +396,25 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
...
   /**
    * Finds first entity matching your `where` query. If nothing found, it will throw an error.
+   * If the `strict` option is specified and nothing is found or more than one matching entity is found, it will throw an error.
    * You can override the factory for creating this method via `options.failHandler` locally
-   * or via `Configuration.findOneOrFailHandler` globally.
+   * or via `Configuration.findOneOrFailHandler` (`findExactlyOneOrFailHandler` when specifying `strict`) globally.
    */
-  async findOneOrFail<T extends AnyEntity<T>, P extends string = never>(entityName: EntityName<T>, where: FilterQuery<T>, options: FindOneOrFailOptions<T, P> = {}): Promise<Loaded<T, P>> {
-    const entity = await this.findOne(entityName, where, options);
+  async findOneOrFail<T extends AnyEntity<T>, P extends string = never>(entityName: EntityName<T>, where: FilterQuery<T>, options: FindOneOrFailOptions<T, P> = {}): Promise<Loaded<T, P> | null> {
```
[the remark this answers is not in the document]
williamwdoyle: I reverted the return type. 

> you should cast the entity at the end when we return, not change the return type to something that is not true
here do you mean 
```
let entity: T | null;
// ...
return entity as Loaded<T, P>;
```
or is it fine to define entity as since that's the return value of the `find` and `findOne` methods.
```
let entity: Loaded<T, P> | null;
```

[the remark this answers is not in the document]
B4nan: both are fine, use whatever works, `Loaded` type is more accurate but the important thing is the method return type, not what we use inside its implementation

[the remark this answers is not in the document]
williamwdoyle: This is updated and pushed now. Thanks for your patience!

B4nan (approved):
looking good, thanks a lot!

Issue #3134, opened 2022-05-22
Every properties of embedded entity has value `undefined` when loaded twice


mung9:
**Describe the bug**
If an embedded entity loaded twice inside a UoW and the entity got updated inside the other UoW between the two load, loaded properties are filled with `undefined` except that got updated

**To Reproduce**
```ts
@Embeddable()
class Nested {

  @Property()
  field1: string;

  @Property()
  field2: string;

  @Property()
  field3: string;

  constructor(field1: string, field2: string, field3: string) {
    this.field1 = field1;
    this.field2 = field2;
    this.field3 = field3;
  }

}

@Entity()
class Parent {

  @PrimaryKey({ autoincrement: false })
  id: number;

  @Embedded(() => Nested)
  nested: Nested;

  constructor(id: number, nested: Nested) {
    this.id = id;
    this.nested = nested;
  }

}

// initialize
const nested = new Nested('A', 'B', 'C');
await orm.em.fork().persistAndFlush(new Parent(1, nested));

// first load
const em1 = orm.em.fork();
const p1 = await em1.findOneOrFail(Parent, 1);
expect(p1.nested.field1).toBe('A');
expect(p1.nested.field2).toBe('B');
expect(p1.nested.field3).toBe('C');

// update inside other UoW
const em2 = orm.em.fork();
const p = await em2.findOneOrFail(Parent, 1);
p.nested.field1 = 'Z';
await em2.persistAndFlush(p);

// reload
const p2 = await em1.findOneOrFail(Parent, 1);

console.log(p2.nested) // => `{ field1: 'Z' }`
```

**Expected behavior**
Expect `field2` and `field3` to be `B` and `C` respectively, but both are `undefined`

**Versions**

| Dependency            | Version |
| --------------------- | ------- |
| node                  | 16.13.1 |
| typescript            | 4.6.4   |
| @mikro-orm/core       | 5.1.4   |
| @mikro-orm/postgresql | 5.1.4 |
| driver | pg |

mung9:
**reproduction code**
https://github.com/mung9/mikro-orm/blob/master/tests/features/embeddables/GH3134.test.ts

B4nan:
Interesting, thanks for the repro!

Looks like the data are stored just fine, if you would reload that entity you would see the correct output.

mung9:
I've wrote some code to fix this issue. can you take a look?
[...]

PR #3135, opened 2022-05-22
fix(core): assign new embeddable entity only when it is null or undefined

mung9:
As proposed in https://github.com/mikro-orm/mikro-orm/issues/3134#issuecomment-1133862820

Updated snapshots using `yarn test -u`

### DIFF
--- a/tests/features/embeddables/GH3134.test.ts
+++ b/tests/features/embeddables/GH3134.test.ts
@@ -1,0 +1,87 @@
+
+import {
+  Embeddable,
+  Embedded,
+  Entity,
+  MikroORM,
+  PrimaryKey,
+  Property,
+} from '@mikro-orm/core';
+
+@Embeddable()
+class Nested {
+
+  @Property()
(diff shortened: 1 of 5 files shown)


tests/features/embeddables/GH3134.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+}
+
+let orm: MikroORM;
+
+beforeAll(async () => {
+  orm = await MikroORM.init({
+    entities: [Parent, Nested],
+    dbName: ':memory:',
+    type: 'postgresql',
```
B4nan, line 53:
```suggestion
    type: 'sqlite',
```

i dont think this does what you think -- `:memory:` is a special sqlite db name, postgres does not offer such thing (but allows you to have pretty much any database name as long as its correctly quoted)

mung9: Ah, I was careless. Thank you!

PR #3163, opened 2022-05-30
fix: typing detection with typescript 4.7 node16

ml1nk:
If module is 'node16' (typescript 4.7, see tsconfig.json) and module type is 'module' (see package.js) then typescript currently searches for a file named index.d.mts (as the base file is index.d.ts) and doesn't find it (see image).

I read about it [here](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#package-json-exports-imports-and-self-referencing):  "./commonjs/index.cjs" =>  "./commonjs/index.d.cts"

A solution is to declare 'index.d.ts' explicitly as the valid typing for 'index.mjs'.

~~~json
"import": "./dist/index.mjs"
~~~
=>
~~~json
 "import": {
        "types": "./dist/index.d.ts",
        "default": "./dist/index.mjs"
 },
~~~

_______

image

package.json
~~~json
{
    "type": "module"
}
~~~

tsconfig.json
~~~json
"compilerOptions": {
        "module": "node16"
}
~~~

### DIFF
--- a/packages/better-sqlite/package.json
+++ b/packages/better-sqlite/package.json
@@ -8,7 +8,10 @@
   "exports": {
     "./package.json": "./package.json",
     ".": {
-      "import": "./dist/index.mjs",
+      "import": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.mjs"
+      },
       "require": "./dist/index.js"
     }
   },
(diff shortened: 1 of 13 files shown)


B4nan:
Thanks!

Issue #732, opened 2020-08-11
High memory profile and very slow


marcj:
**Describe the bug**

I've integrated Mikro-ORM in a benchmark suite and noticed a memory leak and very slow execution times.

The benchmark basically creates 10k objects, persist() each, and then flushes. After that it reads all 10k from the database. What happens with Mikro ORM is that the first 10k insert takes 7 seconds, the second 10k insert 11 seconds, the third 21 seconds. Memory climbing as well rapidly 166mb, then 278mb, then 417mb.

Raw output:

```
round 1
0.13634226180139436 ops/s Mikro-ORM insert 7,334.482989996672 ms, 166.2109375 MB memory
7.202537523174704 ops/s Mikro-ORM fetch 138.83995699882507 ms, 175.12109375 MB memory
round 2
0.09214466685562421 ops/s Mikro-ORM insert 10,852.500032007694 ms, 278.3046875 MB memory
9.313704069546839 ops/s Mikro-ORM fetch 107.36866798996925 ms, 286.4140625 MB memory
round 3
0.04624159269974236 ops/s Mikro-ORM insert 21,625.552703022957 ms, 417.4453125 MB memory
10.529599470788153 ops/s Mikro-ORM fetch 94.97037401795387 ms, 419.08984375 MB memory
round 4
0.03031239307193595 ops/s Mikro-ORM insert 32,989.80709397793 ms, 481.515625 MB memory
15.424512427769912 ops/s Mikro-ORM fetch 64.83187100291252 ms, 482.24609375 MB memory
round 5
0.023591314642157822 ops/s Mikro-ORM insert 42,388.48131901026 ms, 548.6171875 MB memory
9.086378374941388 ops/s Mikro-ORM fetch 110.05484899878502 ms, 550.55859375 MB memory
```

**To Reproduce**

Execute the following Typescript code:

```typescript
import 'reflect-metadata';
import {Entity as MikroEntity, MikroORM, PrimaryKey, Property, ReflectMetadataProvider} from 'mikro-orm';
import {performance} from 'perf_hooks';

export async function bench(times: number, title: string, exec: () => void | Promise<void>) {
    const start = performance.now();
    for (let i = 0; i < times; i++) {
        await exec();
    }
    const took = performance.now() - start;

    process.stdout.write([
        (1000 / took) * times, 'ops/s',
        title,
        took.toLocaleString(undefined, {maximumFractionDigits: 17}), 'ms,',
[...]

PR #3211, opened 2022-06-11
Speeding up the creation of a large number of entities

mrmlnc:
Hello, @B4nan,

First, thanks for a really good ORM for the Node.js world.

## What's the problem?

Simplifying, we have two related entities in our project: `TestRunEntity` and `TestCaseEntity`. There is a 1:M relationship between these entities.

According to the specifics of our project, we can have up to 10_000 test cases in a test run.

The problem here is that creating such a large set of entities takes up to 12 seconds:

```ts
orm.em.create(TestRunEntity, {
    cases: Array(10_000).fill(undefined).map((_, index) => ({
        title: `Test Case #${index}`
    }))
});
```

I have a repository that makes it easy to reproduce the problem.

https://github.com/mrmlnc/mikro-orm-performance-issue

## Investigation

Most likely related issues

* #732

I investigated and found that most of the time it takes to clone a set of entities at each iteration in the code below:

https://github.com/mikro-orm/mikro-orm/blob/2b5cabc102103b8d2852ca3af5916af6f782d7d7/packages/core/src/entity/ArrayCollection.ts#L111-L116

Some screenshots obtained during investigation:

<details>

<summary>npx clinic flame and doctor</summary>

clinic-current

clinic-current-doctor

</details>

I believe that the `Object.assign` operation is not necessary here. Let me describe it to you as I see it.

The operations above do literally the following on each iteration:

1. Remove the current entity from the `Set`.
1. Remove the current entity from the context (`this`).
1. Converts a `Set` to an array and writes it to the context (`this`).

After steps 1 and 2, both of these sets are *synchronized*. Both of these sets refer to the same classes by reference.

In the third step, we simply overwrite all existing entries in the context (`this`) with the same ones from `Set` as `this['0']`, `this['1']`, …. These are all the same references to instantiated entity classes. This is the same context for all methods in the `ArrayCollection` class. So, after overwriting the context properties, nothing changes.

⚠️ Maybe I'm wrong. But I can't find one reason to do it. Please tell me what you think about it.

## What is the purpose of this pull request?

So, I propose to stop doing the third step, since both sets are synchronized by calling the corresponding methods earlier.

Time measurements for 10_000 entities:

* Time before making changes: **~12s**.
* Time after making changes: **~150ms**.

Some screenshots obtained after changes:

<details>

<summary>npx clinic flame and doctor</summary>

clinic-next

clinic-next-doctor

</details>

### DIFF
--- a/packages/core/src/entity/ArrayCollection.ts
+++ b/packages/core/src/entity/ArrayCollection.ts
@@ -111,7 +111,6 @@ export class ArrayCollection<T, O> {
       if (this.items.delete(entity)) {
         this.incrementCount(-1);
         delete this[this.items.size]; // remove last item
-        Object.assign(this, [...this.items]); // reassign array access
         this.propagate(entity, 'remove');
       }
     }
@@ -137,7 +136,6 @@ export class ArrayCollection<T, O> {
 
     this.incrementCount(-1);
     delete this[this.items.size];
-    Object.assign(this, [...this.items]);
   }
 
   contains(item: T | Reference<T>, check?: boolean): boolean {


mrmlnc:
Thanks for the quick response.

I appreciate your time, so I checked tests before and after changes and studied your codebase as much as time allowed.

I understand your concern. Perhaps we can consider an intermediate option? What about doing a full synchronization only once after the loop is done (move `Object.assign` after `for…of`)?

https://github.com/mikro-orm/mikro-orm/blob/2b5cabc102103b8d2852ca3af5916af6f782d7d7/packages/core/src/entity/ArrayCollection.ts#L104-L117

I also pushed the commit to the repository showing a comparison of elements between two sets.

https://github.com/mrmlnc/mikro-orm-performance-issue/commit/3d833b6d9985f3e1cc01f799ffbd97e27495fc77

B4nan:
Mmmm this is indeed weird, how come changes in those methods affect entity creation? You adjust only the sync after removal of items, I'd say this shouldn't be called at all.

Anyway, thanks for the repro and bringing this up.

mrmlnc:
> Mmmm this is indeed weird, how come changes in those methods affect entity creation?

Oh, yeap, you're right. I fixed the symptom, not the cause.

IMHO, there are two independent incorrect behaviors here.

**First, where we create a collection we doing some extra work:**

The [`Collection.create`](
https://github.com/mikro-orm/mikro-orm/blob/2b5cabc102103b8d2852ca3af5916af6f782d7d7/packages/core/src/entity/Collection.ts#L33-L41) method instantiates the [`ArrayCollection`](https://github.com/mikro-orm/mikro-orm/blob/2b5cabc102103b8d2852ca3af5916af6f782d7d7/packages/core/src/entity/ArrayCollection.ts#L16-L22) class that writes a complete set of items to itself (`this.items = new Set(items)`). Then the `Collection.create` method calls the [`coll.set`](https://github.com/mikro-orm/mikro-orm/blob/2b5cabc102103b8d2852ca3af5916af6f782d7d7/packages/core/src/entity/Collection.ts#L37-L39) method for the same set of elements. And [inside this method](https://github.com/mikro-orm/mikro-orm/blob/2b5cabc102103b8d2852ca3af5916af6f782d7d7/packages/core/src/entity/ArrayCollection.ts#L78-L81) we call the cleanup of the newly created set with the same items. Here we write the same items twice.

It seems to me that we should use the `.set` method instead of filling from the class constructor, because the `set` method does more operation on the each item (like `this.propagate` call).

**Second, we doing some extra work on `remove` call:**

In my opinion, it is enough to perform full synchronization only once, or trust the code about deleting from sets.

mrmlnc:
Hmmm, most likely Apple Silicon works better in this case. I am work on a remote virtual machine with Ubuntu (part of Xeon Gold 6230).

I made a [commit](https://github.com/mrmlnc/mikro-orm-performance-issue/commit/95f10a940771704a65265752481a87fc2bcfdb57) to my repository where I updated the dependencies and edited the code to be closer to your version in the comment above.

Just a few tests on the new version of dependencies on `node@18.3.0`:

* As Is — ~12s
* Only without [these lines](https://github.com/mikro-orm/mikro-orm/blob/2b5cabc102103b8d2852ca3af5916af6f782d7d7/packages/core/src/entity/ArrayCollection.ts#L17-L21) — ~100ms
* Only without [this line](https://github.com/mikro-orm/mikro-orm/blob/2b5cabc102103b8d2852ca3af5916af6f782d7d7/packages/core/src/entity/ArrayCollection.ts#L114) — ~105ms
* Only move `Object.assign` after `for…of` — ~107ms

B4nan:
Oh, this was so weird. Apparently the slowness was caused by the way class is being compiled in the ORM, as we don't have `useDefineForClassFields` enabled, hence there were no class properties allocated (with this enabled you have the fields set to `undefined` automatically even without providing the values). With that I see pretty much the same boost as with your repro even if I keep using jest/ts-jest (it's a bit slower of course, around 100ms). But with this flag the ORM tests fail to serialize things correctly and some are failing :/

> Second, we doing some extra work on remove call:
>
> In my opinion, it is enough to perform full synchronization only once, or trust the code about deleting from sets.

Can you elaborate on this? You mean full sync of the index access?

mrmlnc:
Thank you for the investigation and fix for the performance issue 🎉 

> Can you elaborate on this? You mean full sync of the index access?

Yeap, I mean full synchronization after removing all items in the `ArrayCollection.remove` method. I still think that synchronization at each iteration of the cycle is not necessary and we can safely move `Object.assign` after the loop.

In fact, this problem can be reproduced on [this commit](https://github.com/mrmlnc/mikro-orm-performance-issue/commit/c151b9937c4edf6db4ee2c55aec6ee8edcf5a976) with the `@mikro-orm/core` from the master branch. The `Collection.removeAll` method takes 11s.

B4nan:
Well, we could optimize the `removeAll` and `set` methods, but keep in mind the `remove` method is a public one - it needs to sync the index on its own.

PR #3249, opened 2022-06-24
Specify error type caught from findOneOrFail

chiubaka:
To avoid leakage of errors that might not be NotFoundError

### DIFF
--- a/docs/versioned_docs/version-5.2/entity-manager.md
+++ b/docs/versioned_docs/version-5.2/entity-manager.md
@@ -270,7 +270,11 @@ try {
   const author = await em.findOneOrFail(Author, { name: 'does-not-exist' });
   // author will be always found here
 } catch (e) {
-  console.error('Not found', e);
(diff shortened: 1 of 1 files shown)


B4nan:
the idea behind this is not forcing users to wrap every single call in try/catch, this should be handled by some middleware that distinguishes the error type (and you can plug in your own error class via the configuration as describer on the following lines).

so i would rather not have such example in the code, as people often like to copy paste things without thinking

chiubaka:
Even if handled in middleware feels like an example showing this specific error thrown when a record is not found (which is really the one the user is expecting on "fail") is less dangerous / confusing than an example which may catch / suppress _all_ errors, some of which may be unrelated. In a middleware people are still likely to want to do something specific in response to this error case or cases like it (e.g. process a 404 rather than a generic 500).

Just my two cents, since this was my instinct and I ended up needing to dig through the typings to find any reference to how to do this. Thanks for your work on MikroORM!

PR #3254, opened 2022-06-27
fix typo

luckspt:


Langstra:
Could you also change this in new 'next' version? Otherwise this typo will exist again in newer versions of the docs. 

I am unsure if @B4nan would also like you to change some older versions as well.

B4nan:
this is not a typo, you can use the mysql driver for mariadb as well

PR #3271, opened 2022-06-30
test: inconsistent results with entity manager and repositories

jbmikk:
I found what I believe is a problem, or at least an inconsistency.

I wrote two tests which are identical except for the implementation of the `requestCommonService` function.

The test basically changes an entity, flushes it, then calls the service which happens to `find` the same entity from the database. All of this happens in the same transaction but on different EM instances. Outside the service we use the EM returned by the `transactional` callback, inside the service we use the "request" EM instance.

The only difference between the  two service implementations is that one executes `find` on the entity manager, and the other uses `find` on a repository (created with the same entity manager).

The test using the repository works as expected, but the test using the entity manager seems to be resetting the outer entity back to it's original value before flushing in the outer transaction. This is unexpected, because an entity that was already flushed is getting reset back to an old value without the outer code being aware of it.

Is this a bug or expected behavior?

### DIFF
--- a/tests/EntityManager.postgre.test.ts
+++ b/tests/EntityManager.postgre.test.ts
@@ -322,6 +322,76 @@ describe('EntityManagerPostgre', () => {
     await expect(orm.em.rollback()).rejects.toThrowError('An open transaction is required for this operation');
   });
 
+  test('test nested find with repository', async () => {
+    const mock = mockLogger(orm, ['query', 'query-params']);
+
+    const book = new Book2('My Life on The Wall, part 1');
+
+    const author = new Author2('Bartleby', 'bartelby@writer.org');
+    author.books.add(book);
+
+    await orm.em.persistAndFlush(author);
+
+    orm.em.clear();
+
+    const reqEm = orm.em.fork();
+
+    async function requestCommonService() {
+      const [b] = await reqEm.getRepository(Book2).find({});
+      return b.title;
+    }
+
+    const { titleA, titleB, titleInner } = await reqEm.transactional(async em => {
+      const b = await em.findOneOrFail(Book2, book.uuid);
+      wrap(b).assign({
+        title: 'New title',
+      });
+      await em.flush();
+      const titleA = b.title;
+      const titleInner = await requestCommonService();
+      const titleB = b.title;
+      return { titleA, titleB, titleInner };
+    });
+
+    expect(titleA).toEqual(titleInner);
+    expect(titleA).toEqual(titleB);
+  });
+
+  test('test nested find with EM', async () => {
+    const mock = mockLogger(orm, ['query', 'query-params']);
+
+    const book = new Book2('My Life on The Wall, part 1');
+
+    const author = new Author2('Bartleby', 'bartelby@writer.org');
+    author.books.add(book);
+
+    await orm.em.persistAndFlush(author);
+
+    orm.em.clear();
+
+    const reqEm = orm.em.fork();
+
+    async function requestCommonService() {
+      const [b] = await reqEm.find(Book2, {});
+      return b.title;
+    }
+
+    const { titleA, titleB, titleInner } = await reqEm.transactional(async em => {
+      const b = await em.findOneOrFail(Book2, book.uuid);
+      wrap(b).assign({
+        title: 'New title',
+      });
+      await em.flush();
+      const titleA = b.title;
+      const titleInner = await requestCommonService();
+      const titleB = b.title;
+      return { titleA, titleB, titleInner };
+    });
+
+    expect(titleA).toEqual(titleInner);
+    expect(titleA).toEqual(titleB);
+  });
+
   test('should load entities', async () => {
     expect(orm).toBeInstanceOf(MikroORM);
     expect(orm.em).toBeInstanceOf(EntityManager);


B4nan:
> The test basically changes an entity, flushes it, then calls the service which happens to find the same entity from the database. All of this happens in the same transaction but on different EM instances. Outside the service we use the EM returned by the transactional callback, inside the service we use the "request" EM instance.

If there is a `em.transactional` wrapper above, they _don't_ use the same EM. This is expected behaviour, there is a transaction context inside that `em.transactional` handler.

jbmikk:
Sorry if didn't describe the problem well. I'll try again.

> If there is a em.transactional wrapper above, they don't use the same EM. This is expected behaviour, there is a transaction context inside that em.transactional handler.

That is fine, I don't expect them to use the same EM, I just find it weird that the internal EM reads the old value from the entity. I do expect the internal EM to affect the external one, I just don't expect it to discard the flushed changes. I expect it to preserve them.

I pushed some changes to the tests to make them more clear.

Both tests do this:

      const b = await em.findOneOrFail(Book2, book.uuid);
      wrap(b).assign({
        title: 'New title',
      });
      em.flush();
      const titleA = b.title;
      const titleInner = await requestCommonService();
      const titleB = b.title;
      return { titleA, titleB, titleInner };

And I expect the assertions to pass for both tests:

    expect(titleA).toEqual(titleInner);
    expect(titleA).toEqual(titleB);

Right now only one of them is passing, the one using the `getRepository()`.

Why would `b.title` and `titleInner` return to the previous value and have the flushed change discarded? I expect `titleA === titleInner` always because `requestCommonService` just did a `find` call. If `requestCommonService` queries the database it should be getting the new flushed value (because it's sharing the transaction). If it doesn't query the database I don't expect it to read stale data from the cache, I expected to get the updated value. 

> I am open for making this configurable at the call site (e.g. having a bool option in the em.transactional call to disable the tx context).

I don't think this is what I want. What I was really expecting is the internal EM to behave like the passing test.

This is how I think about it:
* I have two EM sharing transactional context (all good)
* EM1 changes an entity and flushes it (db updated)
* EM2 "finds" the same entity (query or cache, doesn't matter, I don't care).
* EM2 should have the same entity with all the same changes EM1 applied, because it's running in the same transaction, within that transaction the entity has new state.
* EM2 instead has the entity without the flushed changes (this can't be coming from the DB, as it was updated already)
* After the service call returns, EM1 for some reason also loses the flushed changes and returns to the old state.

The really weird part is this faling only for the test using `reqEm.find()`. The other one, the one using `reqEm.getRepository().find()`, works the way I expect it.

What I really want to know is if that sequence of steps makes sense or not. I don't really understand what is the expected behavior for two separate EM instances sharing the same transactional context. Maybe my expectations are wrong, but I expected both tests to behave consistently, I expect both to pass.

tests/EntityManager.postgre.test.ts
```diff
@@ -322,6 +322,76 @@ describe('EntityManagerPostgre', () => {
...
+      const [b] = await reqEm.find(Book2, {});
+      return b.title;
+    }
+
+    const { titleA, titleB, titleInner } = await reqEm.transactional(async em => {
+      const b = await em.findOneOrFail(Book2, book.uuid);
+      wrap(b).assign({
+        title: 'New title',
+      });
+      em.flush();
```
B4nan, line 384: missing await

tests/EntityManager.postgre.test.ts
```diff
@@ -322,6 +322,76 @@ describe('EntityManagerPostgre', () => {
...
+      const [b] = await reqEm.getRepository(Book2).find(Book2);
+      return b.title;
+    }
+
+    const { titleA, titleB, titleInner } = await reqEm.transactional(async em => {
+      const b = await em.findOneOrFail(Book2, book.uuid);
+      wrap(b).assign({
+        title: 'New title',
+      });
+      em.flush();
```
B4nan, line 349: missing await

B4nan (commented):
might not matter much here, but this is usually cause of quite weird things

B4nan:
One thing to try is `em.getContext()`, it should be effectively called in both ways, but with the repository its more explicit. We might be missing some place... 

Can't help much more now, I am burried with work related stuff.

jbmikk:
@B4nan no problem, I understand. Thanks anyway.

Just to confirm, to you think both tests should pass? Is that the expected behaviour?

tests/EntityManager.postgre.test.ts
```diff
@@ -322,6 +322,76 @@ describe('EntityManagerPostgre', () => {
...
+    author.books.add(book);
+
+    await orm.em.persistAndFlush(author);
+
+    orm.em.clear();
+
+    const reqEm = orm.em.fork();
+
+    async function requestCommonService() {
+      const [b] = await reqEm.getRepository(Book2).find(Book2);
```
B4nan, line 340: this is not valid, the parameter needs to be query, not class, something like this
```suggestion
      const [b] = await reqEm.getRepository(Book2).find(book);
```

B4nan: in the second test you are doing `em.find(Book2, {})`

B4nan:
Ok, I can indeed confirm this is a bug, in general using the EM should behave the same, as repositories are only passing down the commands down to the EM anyway.

B4nan:
Looking into this, and I got it fixed, but your repro is not 100% correct. Given you are forking explictly and using the `reqEm` instead of the global EM, you are basically disabling the automatic context getter inside. Right now it master it was always getting the EM from explicit transaction contexts but that was wrong (it would cause issues if you fork inside explicit tx). Once I removed that, your tests started to fail again, and that is because you try to read things from the `reqEm` while still inside the transactional handler - so _before_ it commits. There are indeed two EM instances, and the changes will be propagated, but _after_ the commit. 

```ts
const reqEm = orm.em.fork(); // explicit fork will never use the context, just itself

async function requestCommonService() {
   const [b] = await reqEm.find(Book2, {}); // we use the explicit fork
   return b.title;
}

const { titleA, titleB, titleInner } = await reqEm.transactional(async em => {
   const b = await em.findOneOrFail(Book2, book.uuid);
   wrap(b).assign({
     title: 'New title',
   });
   await em.flush();
   const titleA = b.title;
   const titleInner = await requestCommonService(); // we call this before tx resolves, hence before the propagation to parent EM happens
   const titleB = b.title;
   return { titleA, titleB, titleInner };
});
```

I added two more tests that use the global `orm.em` instead and those will work fine. Why? Because the global EM will always take the contextual fork, so it will resolve into the same instance as you use for `assign` call. 

```ts
async function requestCommonService() {
   const [b] = await orm.em.find(Book2, {}); // we use the global one, calls `em.getContext()` and gets the same fork from tx context
   return b.title;
}

const { titleA, titleB, titleInner } = await orm.em.transactional(async em => { // the `em` here is the one `requestCommonService` gets from the tx context, we could also use `orm.em` here, same behaviour
   const b = await em.findOneOrFail(Book2, book.uuid);
   wrap(b).assign({
     title: 'New title',
   });
   await em.flush();
   const titleA = b.title;
   const titleInner = await requestCommonService(); // now it does not matter if the tx resolved, because it will use the exact same fork from context
   const titleB = b.title;
   return { titleA, titleB, titleInner };
});
```

jbmikk:
@B4nan thanks for the fix and the detailed explanation, this is a lot more clear now.

I'm wondering if the docs should be updated, because I hadn't originally understood that the `fork()` behaves differently depending on how I fork it, also I wasn't aware of the `useContext: true` parameter. I thought the async context always took the right context by default.

I'll play with this and see how it works and maybe I'll send a PR for the docs later.

Issue #620, opened 2020-06-22
Full Text Search


thekevinbrown:
**Is your feature request related to a problem? Please describe.**
We would like to run full text searches on our database.

**Describe the solution you'd like**
A `$fulltext` operator would be nice for this, e.g.

```typescript
const = await em.count(User, { username: { $fulltext: '@whatever.com' } });
```

**Describe alternatives you've considered**
We could bypass the ORM for these queries, but would really like entities to be populated based on the results.

**Additional context**
In most cases this can be done by extending the `getOperatorReplacement` functionality in `QueryBuilderHelper`, but unfortunately for us, MySQL breaks all that with their:

```
where match(?) against ('?' in natural language mode);
```

So this is going to require a more complete swapping of the subclause when using a fulltext operator.

ReactReduxRouterDOM:
Is there any update regarding the implementation of `$fulltext` in MongoDB?

thekevinbrown:
Nope, I was focused on SQL and haven't finished implementing this yet.

service-paradis:
It would be really great to have this supported in MikroOrm!

> where match(?) against ('?' in natural language mode);

The `match` portion here can also received more than one parameters. Ex.

```sql
SELECT * FROM post
        WHERE MATCH (title, description)
        AGAINST ('javascript' IN NATURAL LANGUAGE MODE);
```

 This case might be more difficult to support.

BrunMartins:
Are there any advancements regarding this feature?

Thanks in advance!

thekevinbrown:
We ended up going with trigram indexes, not full text search.

BrunMartins:
Alright, thanks!

bozorgmehr96:
> We ended up going with trigram indexes, not full text search.

I want to use trigram too but I can't figure out how to query with find methods, did you use query builder?

thekevinbrown:
Depending on how you set the index up, this works:

```typescript
const filter = { $ilike: `${fullTextSearch}%` };
```

PR #621, opened 2020-06-22
Feature: Add $fulltext operator

thekevinbrown:
Add a $fulltext operator that can be used in queries on SQL databases. The syntax used varies per database and is not supported on MongoDB.

Closes: #620

PR #3317, opened 2022-07-19
feat: add support for full text searches

jsprw:
Add a $fulltext operator that can be used in queries on SQL databases. 

Based on #621 by @thekevinbrown. This version has also added support for creating required indexes with the decorator `@Index({ type: 'fulltext' })` and full text searching in MongoDB. Documentation could possibly be improved with more clear instructions.

Closes: #620


packages/core/src/utils/Utils.ts
```diff
@@ -822,6 +822,20 @@ export class Utils {
...
   }
 
+  static hasNestedKey(object: any, key: string): boolean {
+    if (!object) {
+      return false;
+    }
+
+    if (Array.isArray(object)) {
+      return object.some(o => this.hasNestedKey(o, key));
+    } else if (typeof object === 'object') {
```
B4nan, line 832:
```suggestion
    }

    if (typeof object === 'object') {
```

packages/core/src/utils/Utils.ts
```diff
@@ -822,6 +822,20 @@ export class Utils {
     return !!GroupOperator[key];
   }
 
+  static hasNestedKey(object: any, key: string): boolean {
```
B4nan, line 825: let's not use `any`, either its object, so `Dictionary`, or `unknown`
```suggestion
  static hasNestedKey(object: Dictionary, key: string): boolean {
```

packages/knex/src/query/QueryBuilderHelper.ts
```diff
@@ -409,7 +409,18 @@ export class QueryBuilderHelper {
       return void qb[m](this.knex.raw(`(${this.subQueries[key]})`), replacement, value[op]);
     }
 
-    qb[m](this.mapper(key, type, undefined, null), replacement, value[op]);
+    if (op === '$fulltext') {
+      const entityMetadata = this.metadata.get(this.entityName);
```
B4nan, line 413: lets use just `meta`, as in the whole codebase
```suggestion
      const meta = this.metadata.get(this.entityName);
```

packages/knex/src/query/QueryBuilderHelper.ts
```diff
@@ -409,7 +409,18 @@ export class QueryBuilderHelper {
       return void qb[m](this.knex.raw(`(${this.subQueries[key]})`), replacement, value[op]);
     }
 
-    qb[m](this.mapper(key, type, undefined, null), replacement, value[op]);
+    if (op === '$fulltext') {
```
B4nan, line 412: maybe this could be handled in the `getOperatorReplacement` like the other operators? just like the regexp operator

packages/knex/src/schema/SchemaComparator.ts
```diff
@@ -463,8 +463,8 @@ export class SchemaComparator {
    * Compares index1 with index2 and returns index2 if there are any differences or false in case there are no differences.
    */
   diffIndex(index1: Index, index2: Index): boolean {
-    // if one of them is a custom expression, compare only by name
-    if (index1.expression || index2.expression) {
+    // if one of them is a custom expression or full text index, compare only by name
+    if (index1.expression || index2.expression || index1.type === 'fulltext' || index2.type === 'fulltext') {
```
B4nan, line 467: is this safe? cant we compare full text indexes in better way?

packages/knex/src/schema/SchemaGenerator.ts
```diff
@@ -514,6 +514,17 @@ export class SchemaGenerator extends AbstractSchemaGenerator<AbstractSqlDriver>
       table.unique(index.columnNames, { indexName: index.keyName });
     } else if (index.expression) {
       this.helper.pushTableQuery(table, index.expression);
+    } else if (index.type === 'fulltext') {
+      if (index.columnNames.length !== 1) {
+        throw new Error('Full text index is only supported on one column.');
```
B4nan, line 519: is this artificial limitation of does it work this way on database level?

docs/docs/query-conditions.md
```diff
@@ -62,6 +62,7 @@ const res = await orm.em.find(Author, [1, 2, 7]);
 | `$nin`   | not contains       | Matches none of the values specified in an array. |
 | `$like`  | like               | Uses LIKE operator |
 | `$re`    | regexp             | Uses REGEXP operator |
+| `$fulltext` | full text | A driver specific full text search function. Requires an FTS5 table for SQLITE and an index with type 'fulltext' for other drivers.
```
B4nan, line 65:
```suggestion
| `$fulltext` | full text       | A driver specific full text search function. Requires an FTS5 table for SQLITE and an index with type 'fulltext' for other drivers. |
```

also whats FTS5? can we add short explanation to parens maybe?

packages/mongodb/src/MongoDriver.ts
```diff
@@ -112,6 +112,20 @@ export class MongoDriver extends DatabaseDriver<MongoConnection> {
       this.inlineEmbeddables(meta, data, where);
     }
 
+    // move search terms from data['$fulltext'] to mongo's structure: data['$text']['search']
+    if ('$fulltext' in data) {
+      // eslint-disable-next-line dot-notation
+      data['$text'] = { $search: data['$fulltext'] };
```
B4nan, line 118: any reason for using bracket syntax? maybe better use casting if its about that, instead of bracket syntax and ignore comments - that feels less confusing i guess

packages/mongodb/src/MongoPlatform.ts
```diff
@@ -79,4 +79,8 @@ export class MongoPlatform extends Platform {
     }
   }
 
+  isAllowedTopLevelOperator(operator: string) {
+    return operator === '$not' || operator === '$fulltext';
```
B4nan, line 83:
```suggestion
    return ['$not', '$fulltext'].includes(operator);
```

B4nan (changes requested):
Looking great, thanks a lot! Did a quick first pass and left some comments. Would be definitely good to describe this more in the docs, otherwise it could get lost.

packages/knex/src/schema/SchemaComparator.ts
jsprw: i can't really compare it as the returned columnNames from the database are an expression and the two objects don't have too much overlap:

Postgres:
```
{
  "columnNames": ["to_tsvector(simple::regconfig, title::text)"],
  "composite": false,
  "keyName": "book2_title_index",
  "unique": false,
  "primary": false
}
VS
{
  "keyName": "book2_title_index",
  "columnNames": ["title"],
  "composite": false,
  "primary": false,
  "unique": false,
  "type": "fulltext",
  "expression": <undefined>
}
```

MySQL/MariaDB:

```
{
  "columnNames": ["title"],
  "keyName": "book2_title_index",
  "unique": false,
  "primary": false
}
VS
{
  "keyName": "book2_title_index",
  "columnNames": ["title"],
  "composite": false,
  "primary": false,
  "unique": false,
  "type": "fulltext",
  "expression": undefined
}
```

packages/knex/src/query/QueryBuilderHelper.ts
jsprw: Maybe, but it's more than just replacing the operator. It executes quite different than other operators, it is not A operator B but different for the platforms, fe Postgres: to_tsvector('simple', :column:) @@ plainto_tsquery(:query)

packages/knex/src/schema/SchemaComparator.ts
jsprw: Yes, understandable. I will also write a comment what could be done/improved for later versions.

B4nan:
It would be also great to catch up with the code coverage. But I am open for merging it without full branch coverage, will be happy to fill that gap myself afterwards.

packages/knex/src/schema/SchemaGenerator.ts
jsprw: Resolved it by allowing many columns to be used for this index, see upcoming commit. I don't know what the behaviour will be when full text searching a column which has an index that consists of multiple comments.

packages/mongodb/src/MongoDriver.ts
jsprw: It was quite hard to do as the copy is made to the same argument of type T. And modifying T might break something or we had to cast to unknown. I solved it by creating a new variable where copied data is stored (which is more easily casted).

jsprw:
What to do in the future:

- Add more thorough index diffing (see https://github.com/mikro-orm/mikro-orm/pull/3317#discussion_r925316081)
- Add support for passing language in the `@Index()` decorator. Right now it takes the simple approach while English or any other language might be more suitable to find more matches. (see https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-PARSING-DOCUMENTS, https://dev.mysql.com/doc/refman/8.0/en/fulltext-natural-language.html).
- Add support for ranking (see https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-RANKING, https://dev.mysql.com/doc/internals/en/full-text-search.html, https://www.mongodb.com/docs/manual/core/text-search-operators/)
- Add support for selecting text search mode (https://dev.mysql.com/doc/refman/8.0/en/fulltext-search.html)
- All drivers have a different way to set up a query (while offering the same functionality, see links in `docs/docs/query-conditions.md`). We could include typed support for constructing queries that are converted to the platform's style dynamically. Like this for `{ title: { $fulltext: { $include: ['keyword1', 'keyword2'], $exclude: ['keyword3'] } } }` or even { title: { $fulltext: { $or: [{ $include: ['keyword1', 'keyword2'], $exclude: ['keyword3'] }, { $include: ['keyword4'], $exclude: ['keyword1'] ] } } }.

Do I forget any?

B4nan:
> For the second I tried to switch the column type to 'tsvector' by doing meta.get('Book2').properties.title.columnTypes[0] = 'tsvector' running generator.getUpdateSchemaSQL() + execute() but that didn't work

Oh right, I see what's happening, unknown types will get mapped to `UnknownType` that defaults to string basically, and ignores provided `columnType`, good catch! [Fixed here](https://github.com/mikro-orm/mikro-orm/commit/f0a20fafa1425ca20f4f6fb2977eae7773d6ac6a).

jsprw:
> > For the second I tried to switch the column type to 'tsvector' by doing meta.get('Book2').properties.title.columnTypes[0] = 'tsvector' running generator.getUpdateSchemaSQL() + execute() but that didn't work
> 
> Oh right, I see what's happening, unknown types will get mapped to `UnknownType` that defaults to string basically, and ignores provided `columnType`, good catch! [Fixed here](https://github.com/mikro-orm/mikro-orm/commit/f0a20fafa1425ca20f4f6fb2977eae7773d6ac6a).

Nice, I will merge the changes and rewrite the test. Thanks!

> Btw I had to remove the default value (empty string), otherwise the query would fail. Maybe that's something to handle as well, not sure what the semantics are, never user this feature myself.

You mean the column default? I will test it now to see what you mean. I added tsvector columns with a default of empty string and it worked fine?

packages/core/src/platforms/Platform.ts
```diff
@@ -291,6 +295,21 @@ export abstract class Platform {
     return path.join('.');
   }
 
+  /* istanbul ignore next */
```
B4nan, line 298: i dont the ignore comments much in here, but note we test similar things in here

https://github.com/mikro-orm/mikro-orm/blob/master/tests/DatabaseDriver.test.ts

it might also be better to throw instead of provide default that is not really a good default :] (at least it looks like you override this method in each supported driver)

packages/core/src/utils/Utils.ts
```diff
@@ -822,6 +822,22 @@ export class Utils {
     return !!GroupOperator[key];
   }
 
+  static hasNestedKey(object: Dictionary, key: string): boolean {
```
B4nan, line 825: this sounds more accurate i guess?

```suggestion
  static hasNestedKey(object: unknown, key: string): boolean {
```

packages/mariadb/src/MariaDbPlatform.ts
```diff
@@ -69,4 +69,16 @@ export class MariaDbPlatform extends AbstractSqlPlatform {
...
+  supportsCreatingFullTextIndex(): boolean {
+    return true;
+  }
+
+  getFullTextWhereClause(): string {
+    return `match(:column:) against (:query in boolean mode)`;
+  }
+
+  getFullTextIndexExpression(indexName: string, schemaName: string | undefined, tableName: string, columns: SimpleColumnMeta[]): string {
+    return `alter table ${this.quoteIdentifier(schemaName ? `${schemaName}.${tableName}` : tableName)} add fulltext index ${this.quoteIdentifier(indexName)}(${columns.map(c => this.quoteIdentifier(c.name)).join(',')})`;
```
B4nan, line 81: lets use some temp vars so we dont have such a noodle :]

packages/mongodb/src/MongoDriver.ts
```diff
@@ -104,21 +104,34 @@ export class MongoDriver extends DatabaseDriver<MongoConnection> {
   }
 
   private renameFields<T>(entityName: string, data: T, where = false): T {
-    data = Object.assign({}, data); // copy first
-    Utils.renameKey(data, 'id', '_id');
+    // copy to new variable to prevent changing the T type or doing as unkown casts
```
B4nan, line 107:
```suggestion
    // copy to new variable to prevent changing the T type or doing as unknown casts
```

packages/mongodb/src/MongoDriver.ts
```diff
@@ -104,21 +104,34 @@ export class MongoDriver extends DatabaseDriver<MongoConnection> {
...
 
-    Object.keys(data).forEach(k => {
+    // move search terms from data['$fulltext'] to mongo's structure: data['$text']['search']
+    if ('$fulltext' in copiedData) {
+      copiedData.$text = { $search: copiedData.$fulltext! };
+      delete copiedData.$fulltext;
+    }
+
+    // mongo only allows the $text operator in the root of the object and will
+    // seach all documents where the field has a text index.
```
B4nan, line 123:
```suggestion
    // search all documents where the field has a text index.
```

B4nan (approved):
Gave it another look, last piece I havent really reviewed is the docs. 

Generally looking pretty much ready to me, just few nits.

B4nan:
I'd like to include this in 5.3, ideally I would ship that today evening or something like that. Let me know if you have time to finish this, I'll be happy to merge it and do the final steps myself if not, in the end its pretty much done.

packages/postgresql/src/types/FullTextType.ts
```diff
@@ -0,0 +1,17 @@
+import { Type } from '@mikro-orm/core';
+
+export class FullTextType extends Type<string, string> {
+
+  compareAsType(): string {
+    return 'tsvector';
```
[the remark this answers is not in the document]
jsprw: The column type is tsvector and not string on the db side. This custom type can be used for having a more performant full text search on postgres.

packages/core/src/platforms/Platform.ts
jsprw: Ah I see. Good suggestion.

packages/core/src/utils/Utils.ts
jsprw: Well hasnestedkey only works on object types. If typescript compiles with unknown, it should be fine.

B4nan: It works with arrays too, you should not represent array via `Dictionary` (I know its assignable, that's not really changing anything).

jsprw:
I have applied your suggestions, I will not be home until later tonight. So if you could apply the last suggestions, that’d be great (especially if you ship it tonight). Else i can implement them around 22 or 23h.

B4nan:
Ok no worries, will do it myself if I feel like releasing today. Might as well spend some time developing other things and ship it tomorrow morning. 

Anyway, thanks a lot, nice to see a complete solution for this!

jsprw:
I am going to work on this now!

jsprw:
Implemented your suggestions, it's now awaiting tests.

packages/postgresql/src/types/FullTextType.ts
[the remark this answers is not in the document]
B4nan: totally missed this comment. the thing is - `compareAsType` is used only when diffing values, if its `string/number/boolean` it will be compared by value (via `===`), otherwise it gets compared generically, which is slow(er).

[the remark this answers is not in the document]
jsprw: Oh I see - misunderstood it indeed. I have changed it, will run tests locally before pushing.

packages/postgresql/src/types/FullTextType.ts
```diff
@@ -2,10 +2,6 @@ import { Type } from '@mikro-orm/core';
 
 export class FullTextType extends Type<string, string> {
 
-  compareAsType(): string {
-    return 'tsvector';
```
B4nan: we should keep the method and change the value to `string` instead, or extend the `StringType`, as right now it would default to `any` which is again generic comparison

jsprw: Fixed!

packages/postgresql/src/types/FullTextType.ts
```diff
@@ -0,0 +1,17 @@
+import { Type } from '@mikro-orm/core';
+
+export class FullTextType extends Type<string, string> {
```
B4nan, line 3: one last thing - lets add this type to the `types` map here too:

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/types/index.ts#L30

B4nan (approved):

packages/postgresql/src/types/FullTextType.ts
jsprw: I don't think that we should do that, it would reference `@mikro-orm/postgresql` from `@mikro-orm/core`. This type should only be used in PostgreSQL as it's the only driver where you need a different type to store full text data. Could move the file to `core` with a warning or whatsoever?

B4nan: oh right, its pg only type

jsprw:
```
import { t } from "@mikro-orm/core";
import { FullTextType } from "./FullTextType";

t.tsvector = FullTextType; // typescript error though

export { FullTextType };
```
We could add it lazily like this in `packages/postgresql/src/types/index.ts`?

B4nan: that wont help much i guess, lets keep it like it is and try it in the wild

jagabs:
anyone experiencing this in postgres?

Screen Shot 2022-10-16 at 11 28 26 AM

for migration
   ` this.addSql('create index "tbl_employee_employee_name_index" on "public"."tbl_employee" using gin(to_tsvector(\'simple\', "employee_name"));');`

entity
` @Index({ type: "fulltext" })
  @Property({ columnType: "text", type: FullTextType })
  public employeeName: string;`
  
  i tried using this
  `@Index({ type: 'fulltext' })
  @Property({ type: FullTextType, onUpdate: (book) => book.title })
  searchableTitle!: string;`
  
  but im not sure what values to add in db migration.
  
  Thanks

B4nan:
> `@Property({ columnType: "text", type: FullTextType })`

This seems wrong, I think you should not override the column type, it needs to be `tsvector`. Note that you say you tried to drop that, but your example code is for a different property (`employeeName` vs `searchableTitle`).

> but im not sure what values to add in db migration.

I dont understand this, you are not using schema generator?

jsprw:
The tabs on the documentation site don't show correctly (https://mikro-orm.io/docs/query-conditions#postgresql). I understand the confusion.

It should show two possible ways to do FTS on Postgres. 

1. You can use a column to store full text searching metadata. Below the title and description property is duplicated to the searchable column for the fts metadata. This works the best if you have multiple properties that should be searchable by one fts query. Note that you have to use the `FullTextType` (so, columntype = tsvector) for the searchable column.

```
@Property()
  title!: string;

@Property()
  description!: string;

  @Index({ type: 'fulltext' })
  @Property({ type: FullTextType, onUpdate: (book) => book.title + ' ' + book.description })
  searchable!: string;
```

You could also use a index on an existing column. Then you can just add @Index with type `fulltext`.

```
 @Index({ type: 'fulltext' })
  @Property()
  title!: string;
```

As noted above, you use `FullTextType` and try to override the type. You should just use:

```
@Index({ type: "fulltext" }) 
@Property({ type: "text" })
public employeeName: string;
```

Jedliu:
Thank you for all your efforts. I'm using PostgreSQL. 

Is there any update for the `setweight` and advanced `constructing queries` as below?

> What to do in the future:
> 
> * Add more thorough index diffing (see [feat: add support for full text searches #3317 (comment)](https://github.com/mikro-orm/mikro-orm/pull/3317#discussion_r925316081))
> * Add support for passing language in the `@Index()` decorator. Right now it takes the simple approach while English or any other language might be more suitable to find more matches. (see https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-PARSING-DOCUMENTS, https://dev.mysql.com/doc/refman/8.0/en/fulltext-natural-language.html).
> * Add support for ranking (see https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-RANKING, https://dev.mysql.com/doc/internals/en/full-text-search.html, https://www.mongodb.com/docs/manual/core/text-search-operators/)
> * Add support for selecting text search mode (https://dev.mysql.com/doc/refman/8.0/en/fulltext-search.html)
> * All drivers have a different way to set up a query (while offering the same functionality, see links in `docs/docs/query-conditions.md`). We could include typed support for constructing queries that are converted to the platform's style dynamically. Like this for `{ title: { $fulltext: { $include: ['keyword1', 'keyword2'], $exclude: ['keyword3'] } } }` or even `{ title: { $fulltext: { $or: [{ $include: ['keyword1', 'keyword2'], $exclude: ['keyword3'] }, { $include: ['keyword4'], $exclude: ['keyword1'] ] } } }`.
> 
> Do I forget any?

Jedliu:
Btw, possible to include the highlighting function? Thank you,

Jedliu:
Just reviewed your code and found the to_tsvector was written fixed in the code [https://github.com/mikro-orm/mikro-orm/blob/8b8f14071b92e91161a32aa272315a0ecce1bc0b/packages/postgresql/src/types/FullTextType.ts#L13](FullTextType) so it's not possible to use `setweight` or change another `regconfig` (e.g `to_tsvector('english', 'your content')` or use another segment tool to preprocess the content.

Here is a code snippet which might be an option.
```typescript
@Index({ type: 'fulltext' })
@Property({
  type: FullTextType,
  nullable: true,
  onCreate: (book: Book) => `setweight(to_tsvector('english', '${book.title}'),'A') || setweight(to_tsvector('english', '${book.description}'),'B')`,
  onUpdate: (book: Book) => `setweight(to_tsvector('english', '${book.title}'),'A') || setweight(to_tsvector('english', '${book.description}'),'B')`,
})
searchableContent!: string;
```
@jsprw  Let me know your comments and see if I can help. Thanks,

B4nan:
You can create your own `FullTextType` if you don't like how it's designed, this is fully under your own control - you can as well override just the `columnType` while using it. Note that your proposed API wouldn't work, `onCreate/onUpdate` need to return runtime values, not SQL fragments.

Jedliu:
That's right. Have to update [AbstractSqlDriver.ts](https://github.com/mikro-orm/mikro-orm/blob/89b4dab31bdac52e1a6a4a281dd3bbd50b3745ec/packages/knex/src/AbstractSqlDriver.ts#L344) to make the proposed API work. 
Do you have any better suggestions for running these raw SQL fragments? Thank you,

B4nan:
As said above, create your own custom type, where you have absolute control over both the value and the SQL fragment.

https://mikro-orm.io/docs/custom-types

Jedliu:
> You can create your own `FullTextType` if you don't like how it's designed, this is fully under your own control - you can as well override just the `columnType` while using it. Note that your proposed API wouldn't work, `onCreate/onUpdate` need to return runtime values, not SQL fragments.
Thank you. I will look into this.

@jsprw your idea looks much more promising. Thanks for taking the time to do this.

PR #3343, opened 2022-07-27
fix(sqlite): throwing ForeignKeyConstraintViolationException instead of the DriverException

vedun:
Two database drivers sqlite and better-sqlite were changed according to [this](https://github.com/doctrine/dbal/blob/4.0.x/src/Driver/API/SQLite/ExceptionConverter.php#L83) link, which i found in [source code documentation](https://github.com/mikro-orm/mikro-orm/blob/master/packages/sqlite/src/SqliteExceptionConverter.ts#L13).


B4nan:
Thanks!

PR #3368, opened 2022-08-05
docs: fix entity schema example with base entity

pihentagy:


B4nan:
Thanks! Please change it in the next version too, its in docs/docs folder.

pihentagy:
> Thanks! Please change it in the next version too, its in docs/docs folder.

Ok, done.

PR #3375, opened 2022-08-08
[Docs] Update defining-entities.md

ctfdavis:
defaultRaw('now') should be defaultRaw('now()')


B4nan:
not really, depends on the driver, e.g. sqlite requires `now`

ctfdavis:
You're right. It depends on the driver. I didn't know that sqlite uses `now`.

It would be better if the docs mentions that the examples are using sqlite though. Thanks.

PR #3378, opened 2022-08-09
feat(knex): allow changing `FROM` clause using `QueryBuilder`

### LINKED ISSUE #3374: Allow to change `FROM` clause using `QueryBuilder`
**Description**
As a user, I would like to have the ability to change/add `FROM` clause in which the entity's table select/update/delete will be executed.

**Motivation**

For example, implementing the seek pagination with the ability to navigate to the previous page, you have to reverse the ordering of the result set (`ASC` vs `DESC`) using the same predicate so that the results you want are to the right of the cursor:

```sql
SELECT * 
FROM (
  SELECT * 
  FROM users 
  WHERE createdAt < ? 
  ORDER BY createdAt DESC 
  LIMIT 10
) AS u 
ORDER BY createdAt ASC;
```

Currently, this is possible to do only using `knex.fromRaw()`, exposing the implementation details of `QueryBuilder`. What could be a issue from design perspective. As soon as the implementation of either `QueryBuilder` or `knex` is changed (e.g. `mikro-orm` starts using a different query builder from `knex`), the user will have to rework a significant part of his code.

Just by declaring a `from()` method, we are able to give the user a powerful mechanism to handle this without necessity to rely on `knex`:

```ts
const qb = orm.em.qb(User, 'user')
  .where({ createdAt: { $lt: new Date() } })
  .orderBy({ createdAt: 'DESC' })
  .limit(10);

const users1 = qb.from(qb.getKnexQuery())
  .orderBy({ createdAt: 'ASC' });

const users2 = qb.from(qb.clone())
  .orderBy({ createdAt: 'ASC' });
```

**Possible solution**
The new method(s) can be introduced to be used to change/add `FROM` clause, removing all previously set from-s:

```ts
export declare class QueryBuilder<T extends AnyEntity<T> = AnyEntity> {
    from<TFrom extends AnyEntity<TFrom> = T>(entityTarget: TFrom | Knex.QueryBuilder | QueryBuilder<TFrom>): QueryBuilder<TFrom>;
    addFrom<TFrom extends AnyEntity<TFrom> = T>(entityTarget: TFrom | Knex.QueryBuilder | QueryBuilder<TFrom>): QueryBuilder<TFrom>;
}
```

derevnjuk:
## Information

| Type                  | Breaking change |
| --------------------- | --------------- |
| Feat | No          |

## Usage example

You can specify the table used in the `FROM` clause, replacing the current table name if one has already been specified. This is typically used to specify a sub-query expression in SQL.

```ts
const qb = em.createQueryBuilder(Book2);
qb.select('*').from(Author2).where({ id: { $gt: 2 } });

console.log(qb.getQuery());
// select `e0`.* from `author2` as `e0` where `e0`.`id` > 2;
```

You can also use sub-queries in the `FROM` like this:

```ts
const qb1 = em.createQueryBuilder(Book2).where({ id: { $lte: new Date() } }).orderBy({ id: 'DESC' }).limit(10);
const qb2 = em.createQueryBuilder(qb1.clone())
qb2.select('*').orderBy({ id: 'ASC' });

console.log(qb2.getQuery());
// select `e1`.* from (select `e0`.* from `book2` as `e0` where `e0`.`id` <= ? order by `e0`.`id` desc limit ?) as `e1` order by `e1`.`id`;
```

To set up an alias to refer to a table in a `SELECT` statement, pass the second argument as follows:

```ts
const qb1 = em.createQueryBuilder(Book2, 'b1').where({ id: { $lte: new Date() } }).orderBy({ id: 'DESC' }).limit(10);
const qb2 = em.createQueryBuilder(qb1.clone(), 'b2')
qb2.select('*').orderBy({ id: 'ASC' });

console.log(qb2.getQuery());
// select `b2`.* from (select `b1`.* from `book2` as `b1` where `b1`.`id` <= ? order by `b1`.`id` desc limit ?) as `b2` order by `b2`.`id`;
```

## Further improvements

To give the ability to create a clean query builder without `FROM` clause, we can change the signature of the `createQueryBuilder` as follows:

```diff
- createQueryBuilder<T>(entityName: EntityName<T>, alias?: string, type?: ConnectionType): QueryBuilder<T> {
+ createQueryBuilder<T>(entityName?: EntityName<T>, alias?: string, type?: ConnectionType): QueryBuilder<T> {
```
Thus, the example above can look more elegant:
```ts
const qb1 = em.createQueryBuilder(Book2, 'b1').where({ id: { $lte: new Date() } }).orderBy({ id: 'DESC' }).limit(10);
const qb2 = em.createQueryBuilder().from(qb.clone(), 'b2')
qb.select('*').orderBy({ id: 'ASC' });
```

closes #3374

### DIFF
--- a/docs/docs/query-builder.md
+++ b/docs/docs/query-builder.md
@@ -296,6 +296,40 @@ const count = await qb.getCount();
 
 This will also remove any existing limit and offset from the query (the QB will be cloned under the hood, so calling `getCount()` does not mutate the original QB state).
 
+## Overriding FROM clause
+
+You can specify the table used in the `FROM` clause, replacing the current table name if one has already been specified. This is typically used to specify a sub-query expression in SQL. 
+
+```ts
+const qb = em.createQueryBuilder(Book2);
+qb.select('*').from(Author2).where({ id: { $gt: 2 } });
+
+console.log(qb.getQuery());
+// select `e0`.* from `author2` as `e0` where `e0`.`id` > 2;
+```
+
+You can also use sub-queries in the `FROM` like this:
+
+```ts
+const qb1 = em.createQueryBuilder(Book2).where({ id: { $lte: new Date() } }).orderBy({ id: 'DESC' }).limit(10);
+const qb2 = em.createQueryBuilder(qb1.clone())
+qb2.select('*').orderBy({ id: 'ASC' });
+
+console.log(qb2.getQuery());
+// select `e1`.* from (select `e0`.* from `book2` as `e0` where `e0`.`id` <= ? order by `e0`.`id` desc limit ?) as `e1` order by `e1`.`id`;
+```
+
+To set up an alias to refer to a table in a `SELECT` statement, pass the second argument as follows:
+
+```ts
+const qb1 = em.createQueryBuilder(Book2, 'b1').where({ id: { $lte: new Date() } }).orderBy({ id: 'DESC' }).limit(10);
+const qb2 = em.createQueryBuilder(qb1.clone(), 'b2')
+qb2.select('*').orderBy({ id: 'ASC' });
+
+console.log(qb2.getQuery());
+// select `b2`.* from (select `b1`.* from `book2` as `b1` where `b1`.`id` <= ? order by `b1`.`id` desc limit ?) as `b2` order by `b2`.`id`;
+```
+
 ## Using sub-queries
 
 You can filter using sub-queries in where conditions:
--- a/packages/knex/src/AbstractSqlDriver.ts
+++ b/packages/knex/src/AbstractSqlDriver.ts
@@ -1,17 +1,16 @@
 import type { Knex } from 'knex';
 import type {
   AnyEntity, Collection, ConnectionType, Configuration, Constructor, CountOptions, DeleteOptions, Dictionary,
-  DriverMethodOptions, EntityData, EntityDictionary, EntityField, EntityManager, EntityMetadata, EntityProperty, FilterQuery,
+  DriverMethodOptions, EntityData, EntityDictionary, EntityField, EntityManager, EntityMetadata, EntityName, EntityProperty, FilterQuery,
   FindOneOptions, FindOptions, IDatabaseDriver, LockOptions, NativeInsertUpdateManyOptions, NativeInsertUpdateOptions,
   PopulateOptions, Primary, QueryOrderMap, QueryResult, RequiredEntityData, Transaction,
 } from '@mikro-orm/core';
 import { DatabaseDriver, EntityManagerType, helper, LoadStrategy, QueryFlag, QueryHelper, ReferenceType, Utils } from '@mikro-orm/core';
 import type { AbstractSqlConnection } from './AbstractSqlConnection';
 import type { AbstractSqlPlatform } from './AbstractSqlPlatform';
-import { QueryBuilder } from './query/QueryBuilder';
+import { QueryBuilder, QueryType } from './query';
 import { SqlEntityManager } from './SqlEntityManager';
 import type { Field } from './typings';
-import { QueryType } from './query';
 
 export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = AbstractSqlConnection> extends DatabaseDriver<C> {
 
@@ -712,7 +711,7 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
   }
 
   /** @internal */
-  createQueryBuilder<T extends object>(entityName: string, ctx?: Transaction<Knex.Transaction>, preferredConnectionType?: ConnectionType, convertCustomTypes?: boolean): QueryBuilder<T> {
+  createQueryBuilder<T extends object>(entityName: EntityName<T> | QueryBuilder<T>, ctx?: Transaction<Knex.Transaction>, preferredConnectionType?: ConnectionType, convertCustomTypes?: boolean): QueryBuilder<T> {
     const connectionType = this.resolveConnectionType({ ctx, connectionType: preferredConnectionType });
     const qb = new QueryBuilder<T>(entityName, this.metadata, this, ctx, undefined, connectionType);
 
--- a/packages/knex/src/query/Alias.ts
+++ b/packages/knex/src/query/Alias.ts
@@ -1,0 +1,9 @@
+import type { Knex } from 'knex';
+import type { EntityMetadata } from '@mikro-orm/core';
+
+export interface Alias {
+  aliasName: string;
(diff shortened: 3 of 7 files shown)


docs/docs/query-builder.md
```diff
@@ -296,6 +296,34 @@ const count = await qb.getCount();
 
 This will also remove any existing limit and offset from the query (the QB will be cloned under the hood, so calling `getCount()` does not mutate the original QB state).
 
+## Overriding FROM clause
```
B4nan, line 299: let's also copy this to the 5.3 docs snapshot, i dont want to ship new minor version because of this feature (and i dont want to postpone merging this just for artificial reasons, so lets just ship it in next patch release)

derevnjuk:
@B4nan thanks for the feedback) 

> What is qb? Is that third QB instance? Why do you clone it?

I suppose this is just a typo. Anyway, it has been fixed by the last commit 3a00280

> Definitely, lets do that. We should also validate there is a FROM clause when building the query.

Should we handle this in the same PR? I believe it would be better to open a separate one. Wdyt?

> Let's add some tests for that to see how it even behaves now.

Yeah, I will try to describe a few more test cases to verify that.

B4nan:
I'd handle it in this one so we have a single item for this feature in the changelog, it should be rather easy. But it can be a separate PR that targets this one if you want, and we can merge that one first.

derevnjuk:
@B4nan I have added a few more tests (4e837db). Unfortunately, you are right. We should somehow override aliases once they has been used to build some clauses (e.g. join, where, order by, having, etc).

B4nan:
Or we could just validate this and throw, if you think it would be hard to remap them.

derevnjuk:
@B4nan this is how one of the options. Anyway, I will check out if we have another option that does not imply a significant reworking of `CriteriaNode`.

B4nan:
Any updates on this? As mentioned above, I am fine with validations for the problematic case, we dont need to be too smart in the initial implementation. Would be good to merge this before conflicts start to pop :]

derevnjuk:
@B4nan Sorry for the delay) It is not possible easily to remap/override aliases  (BTW neither `knex` nor `typeorm` does this). It may require rewriting the whole logic of `QueryBuilder` by building clauses at the final stage as follows:
Before:
```ts
orderBy(orderBy: QBQueryOrderMap<T> | QBQueryOrderMap<T>[]): this {
  this._orderBy = [];
  Utils.asArray(orderBy).forEach(o => {
    const processed = QueryHelper.processWhere({
      where: o as Dictionary,
      entityName: this.mainAlias.entityName,
      metadata: this.metadata,
      platform: this.platform,
      aliasMap: this.getAliasMap(),
      aliased: !this.type || [QueryType.SELECT, QueryType.COUNT].includes(this.type),
      convertCustomTypes: false,
    })!;
    this._orderBy.push(CriteriaNodeFactory.createNode(this.metadata, this.mainAlias.entityName, processed).process(this));
  });

  return this;
}

getKnexQuery(): Knex.QueryBuilder {
  // ...
  Utils.runIfNotEmpty(() => {
      const queryOrder = this.helper.getQueryOrder(this.type, this._orderBy as FlatQueryOrderMap[], this._populateMap);

      if (queryOrder) {
        return qb.orderByRaw(queryOrder);
      }
  }, this._orderBy);
  // ...
}

```
After:

```ts
orderBy(orderBy: QBQueryOrderMap<T> | QBQueryOrderMap<T>[]): this {
  this._orderBy = Utils.asArray(orderBy);

  return this;
}

// ...

getKnexQuery(): Knex.QueryBuilder {
  // ...
  Utils.runIfNotEmpty(() => {
    const nodes = this._orderBy.map((o) => {
      const processed = QueryHelper.processWhere({
        where: o as Dictionary,
        entityName: this.mainAlias.entityName,
        metadata: this.metadata,
        platform: this.platform,
        aliasMap: this.getAliasMap(),
        aliased: !this.type || [QueryType.SELECT, QueryType.COUNT].includes(this.type),
        convertCustomTypes: false,
      })!;
    
      return CriteriaNodeFactory.createNode(this.metadata, this.mainAlias.entityName, processed).process(this);
    });
    const queryOrder = this.helper.getQueryOrder(this.type, nodes as FlatQueryOrderMap[], this._populateMap);
    
    if (queryOrder) {
      return qb.orderByRaw(queryOrder);
    }
  }, this._orderBy);
  // ...
}
```

So, I propose to open a separate issue to refactor clause by clause sequentially.

docs/versioned_docs/version-5.3/query-builder.md
```diff
@@ -296,6 +296,40 @@ const count = await qb.getCount();
 
 This will also remove any existing limit and offset from the query (the QB will be cloned under the hood, so calling `getCount()` does not mutate the original QB state).
 
+## Overriding FROM clause
```
B4nan, line 299: 5.4 was already released, so this change should go only to 5.4 docs. i havent added 5.4 snapshot for docs yet, so adjusting only the docs/docs file is all what we need here, you can remove this one

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -4,14 +4,13 @@ import type {
   DriverMethodOptions, EntityData, EntityDictionary, EntityField, EntityManager, EntityMetadata, EntityProperty, FilterQuery,
   FindOneOptions, FindOptions, IDatabaseDriver, LockOptions, NativeInsertUpdateManyOptions, NativeInsertUpdateOptions,
   PopulateOptions, Primary, QueryOrderMap, QueryResult, RequiredEntityData, Transaction,
-} from '@mikro-orm/core';
+ EntityName } from '@mikro-orm/core';
```
B4nan, line 7: bad ident/line break

packages/knex/src/query/Alias.ts
```diff
@@ -0,0 +1,15 @@
+import type { Knex } from 'knex';
+import type { EntityMetadata } from '@mikro-orm/core';
+
+export class Alias {
```
B4nan, line 4: seeing how is this class used, maybe interface would be enough?

i dont really mind having a class, just a though

B4nan:
> So, I propose to open a separate issue to refactor clause by clause sequentially.

I was trying to implement that some time ago, its not that simple, as with everything... :] 

I am fine with keeping this edge case not supported. This is already a huge improvement, thanks a lot!

derevnjuk:
@B4nan your comments are resolved including LGTM.com reported issue

packages/knex/src/query/Alias.ts
derevnjuk: Yeah, I don't mind

PR #3396, opened 2022-08-15
docs: fix typos

Lioness100:
🙂

### DIFF
--- a/docs/versioned_docs/version-2.7/collections.md
+++ b/docs/versioned_docs/version-2.7/collections.md
@@ -100,7 +100,7 @@ marked by `inversedBy` attribute pointing to the inverse side:
 tags = new Collection<BookTag>(this);
 ```
 
-And on the inversed side we define it with `mappedBy` attribute poining back to the owner:
+And on the inversed side we define it with `mappedBy` attribute joining back to the owner:
 
 ```typescript
 @ManyToMany({ entity: () => Book, mappedBy: 'tags' })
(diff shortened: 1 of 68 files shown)


docs/versioned_docs/version-2.7/collections.md
```diff
@@ -100,7 +100,7 @@ marked by `inversedBy` attribute pointing to the inverse side:
 tags = new Collection<BookTag>(this);
 ```
 
-And on the inversed side we define it with `mappedBy` attribute poining back to the owner:
+And on the inversed side we define it with `mappedBy` attribute joining back to the owner:
```
B4nan, line 103: this one should be `pointing back`, not `joining back`. i see its actually fixed in latest versions, i usually dont care much about fixing typos in old docs versions, so thats why i guess :]

but given its such old version, i dont really care about wording, lets merge :]

B4nan (approved):
thanks a lot! is this from some automated tool or you did this manually? :]

Lioness100:
My pleasure! I cloned every repository in the organization and put it in a folder, then from the root I ran [`cpell`](https://www.npmjs.com/package/cspell) on every `README`, `src/**`, `docs/**`, and `tests/**` file from every repository. This output ~700 suspected typos, of which I manually filtered through to find the real ones. There was 100 here, 1 in nestjs, and a few in both docs and docs-v3 although I learned after that those weren't used anymore. 

I was bored :3 Thanks for making such a great orm :))

PR #3399, opened 2022-08-15
fix(query-builder): support more operators in join conditions

### LINKED ISSUE #3412: Unable to use `$in` operator in a join clause
**Describe the bug**
When trying to use $in on a join(), knex complains that `in` is not allowed.  Some googling indicates that one should use `onIn` however this does not work either.

**Stack trace**
```
Uncaught TypeError TypeError: The operator "in" is not permitted
    at operator (/node_modules/knex/lib/formatter/wrappingFormatter.js:132:11)
    at onBasic (/node_modules/knex/lib/query/querycompiler.js:902:7)
    at join (/node_modules/knex/lib/query/querycompiler.js:445:40)
    at <anonymous> (/node_modules/knex/lib/query/querycompiler.js:133:40)
    at select (/node_modules/knex/lib/query/querycompiler.js:132:16)
    at toSQL (/node_modules/knex/lib/query/querycompiler.js:73:29)
    at toSQL (/node_modules/knex/lib/query/querybuilder.js:83:44)
    at Target.toQuery (/node_modules/knex/lib/builder-interface-augmenter.js:9:21)
    at toString (/node_modules/knex/lib/query/querybuilder.js:78:17)
    at eval (repl:1:19)```
```
**To Reproduce**
```
const qb = MikroInstance.get()
            .em.fork()
            .qb(Stream)
            .join('event', 'st', {
                status: {
                    $in: ['foo', 'bar'],
                },
            });

qb.getKnexQuery().toString()
```

**Expected behavior**
Valid SQL is produced

**Additional context**
None

**Versions**

| Dependency | Version |
| - | - |
| node | v14.18.3 |
| typescript | 4.7.4 |
| mikro-orm | 5.3.1 |
| your-driver | Sqlite |

r3tard:
Small fix for joining table using operator `$ne`  (`$eq`) with null value.

### Example

```
em.createQueryBuilder(Table, 't').join('relation', 'r', { 'r.nullableField': { $ne: null } });
```

This should create sql ``` `r`.`nullableField` is not null ```, but created sql was  ``` `r`.`nullableField` != null ```

Closes #3412

### DIFF
--- a/packages/knex/src/query/QueryBuilderHelper.ts
+++ b/packages/knex/src/query/QueryBuilderHelper.ts
@@ -428,56 +428,75 @@ export class QueryBuilderHelper {
   }
 
   private appendJoinSubClause(clause: Knex.JoinClause, cond: Dictionary, key: string, operator?: '$and' | '$or'): void {
-    const m = operator === '$or' ? 'orOn' : 'andOn';
+    const c = operator === '$or' ? 'or' : 'and';
+    const m = `${c}On`;
 
-    if (cond[key] instanceof RegExp) {
+    if (this.isSimpleRegExp(cond[key])) {
       return void clause[m](this.mapper(key), 'like', this.knex.raw('?', this.getRegExpParam(cond[key])));
     }
 
-    if (Utils.isPlainObject(cond[key])) {
-      return this.processObjectSubClause(cond, key, clause, m);
+    if (Utils.isPlainObject(cond[key]) || cond[key] instanceof RegExp) {
+      return this.processObjectSubClause(cond, key, clause, c);
     }
 
     if (QueryBuilderHelper.isCustomExpression(key)) {
       return this.processCustomExpression(clause, m, key, cond);
     }
 
-    const op = cond[key] === null ? 'is' : '=';
-    clause[m](this.knex.raw(`${this.knex.ref(this.mapper(key, QueryType.SELECT, cond[key]))} ${op} ?`, cond[key]));
+    if (cond[key] === null) {
+      clause[`${c}OnNull`](this.mapper(key));
+    } else {
+      clause[m](this.knex.raw(`${this.knex.ref(this.mapper(key, QueryType.SELECT, cond[key]))} = ?`, cond[key]));
+    }
   }
 
-  private processObjectSubClause(cond: any, key: string, clause: Knex.JoinClause, m: 'andOn' | 'orOn'): void {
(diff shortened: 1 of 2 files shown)


packages/knex/src/query/QueryBuilderHelper.ts
```diff
@@ -459,6 +459,16 @@ export class QueryBuilderHelper {
         continue;
       }
 
+      if (cond[key][op] === null) {
+        if (replacement === QueryOperator.$ne) {
+          clause[m](this.knex.raw(`${this.knex.ref(this.mapper(key))} is not ?`, cond[key][op]));
+          continue;
+        } else if (replacement === QueryOperator.$eq) {
```
B4nan, line 466: there is no need for else branch
```suggestion
        }

        if (replacement === QueryOperator.$eq) {
```

packages/knex/src/query/QueryBuilderHelper.ts
```diff
@@ -459,6 +459,16 @@ export class QueryBuilderHelper {
         continue;
       }
 
+      if (cond[key][op] === null) {
+        if (replacement === QueryOperator.$ne) {
+          clause[m](this.knex.raw(`${this.knex.ref(this.mapper(key))} is not ?`, cond[key][op]));
```
B4nan, line 464: this seems like a wrong place (and wrong way) to fix it, we already have support for the `null` comparison, and it works fine without raw queries

r3tard: we have full support for the `null` only in `where` clause, in `join on` it works only when using the `null` without operator. if you run changed test on master branch, you'll get error

B4nan: but the join conditions are processed the very same way as where conditions, or not? 

i will need more time to check this out, maybe you are right

r3tard: no, function `processObjectSubClause` relates only to join. where conditions are processing by `processObjectSubCondition`

packages/knex/src/query/QueryBuilderHelper.ts
r3tard: changed it to similar code from `where` conditions

r3tard:
I realized that other operators also doesn't work (`$in` for example)

B4nan (approved):

Issue #3037, opened 2022-04-20
ArrayType throws a malformed array literal error when values contain brackets {}


sylv:
**Describe the bug**

When using `ArrayType` for text arrays with postgresql, inserting an array with an element that contains a bracket will throw an error about a malformed array. 

**Stack trace**
```
/demo/node_modules/.pnpm/@mikro-orm+core@5.1.2_d55acb51b3d624ca716f3ead5a1a6eac/node_modules/@mikro-orm/core/platforms/ExceptionConverter.js:8
        return new exceptions_1.DriverException(exception);
               ^
DriverException: insert into "owner" ("id", "tags") values (1, '{test {test}}') returning "id" - malformed array literal: "{test {test}}"
    at PostgreSqlExceptionConverter.convertException (/demo/node_modules/.pnpm/@mikro-orm+core@5.1.2_d55acb51b3d624ca716f3ead5a1a6eac/node_modules/@mikro-orm/core/platforms/ExceptionConverter.js:8:16)
    at PostgreSqlExceptionConverter.convertException (/demo/node_modules/.pnpm/@mikro-orm+postgresql@5.1.2_755b0c50a47b2e0e2c828f14e7a81037/node_modules/@mikro-orm/postgresql/PostgreSqlExceptionConverter.js:42:22)
    at PostgreSqlDriver.convertException (/demo/node_modules/.pnpm/@mikro-orm+core@5.1.2_d55acb51b3d624ca716f3ead5a1a6eac/node_modules/@mikro-orm/core/drivers/DatabaseDriver.js:173:54)
    at /demo/node_modules/.pnpm/@mikro-orm+core@5.1.2_d55acb51b3d624ca716f3ead5a1a6eac/node_modules/@mikro-orm/core/drivers/DatabaseDriver.js:177:24
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async PostgreSqlDriver.nativeInsert (/demo/node_modules/.pnpm/@mikro-orm+knex@5.1.2_504e3b747281b68637e5c979c82a6e0f/node_modules/@mikro-orm/knex/AbstractSqlDriver.js:161:21)
    at async ChangeSetPersister.persistNewEntity (/demo/node_modules/.pnpm/@mikro-orm+core@5.1.2_d55acb51b3d624ca716f3ead5a1a6eac/node_modules/@mikro-orm/core/unit-of-work/ChangeSetPersister.js:85:21)
[...]

PR #3405, opened 2022-08-21
fix(postgres): fix escaping of special chars in string arrays

biancazzurri:
The fix in #3037 did not cover using of double quotes which broke the insert of values


B4nan:
Thanks!

PR #3420, opened 2022-08-27
feat(entity-generator): add import extension for referenced entities

JSin:
In a generated entity, add the ability for referenced entities to be imported by a named import.

Most common case is a extension of '.js' but node does support '.cjs' and '.mjs'. Supported ESM_FILE_FORMAT can be found in the node docs https://nodejs.org/docs/latest-v16.x/api/esm.html#resolution-algorithm

Another possible approach is to check package.json for `"type": "module"` but that felt like I was being too smart as I wouldn't know for sure which supported file format the consumer would want.

### DIFF
--- a/packages/entity-generator/src/SourceFile.ts
+++ b/packages/entity-generator/src/SourceFile.ts
@@ -9,7 +9,7 @@ export class SourceFile {
   constructor(protected readonly meta: EntityMetadata,
               protected readonly namingStrategy: NamingStrategy,
               protected readonly platform: Platform,
-              protected readonly importExtension: '' | 'js' | 'mjs' | 'cjs') { }
+              protected readonly esmImport: boolean) { }
 
   generate(): string {
     this.coreImports.add('Entity');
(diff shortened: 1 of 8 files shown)


packages/entity-generator/src/SourceFile.ts
```diff
@@ -50,7 +51,10 @@ export class SourceFile {
     const imports = [`import { ${([...this.coreImports].sort().join(', '))} } from '@mikro-orm/core';`];
     const entityImports = [...this.entityImports].filter(e => e !== this.meta.className);
     entityImports.sort().forEach(entity => {
-      imports.push(`import { ${entity} } from './${entity}';`);
+      imports.push(
+        `import { ${entity} } from './${entity}${
+          this.importExtension ? `.${this.importExtension}` : ''
+        }';`);
```
B4nan, line 57: nit: This part looks like formatted from prettier, would be nice to spend a bit more time with human powers, lol :] e.g. the exact line that adds the extension to the source file code, I truly hate prettier for this behaviour (I would be probably using it myself otherwise, this is a deal breaker for me). It handles line breaks in such weird way. Instead we should use temp variables to make some sense of it, not break lines like robots :]

this expression is simply too complex, mixing ternaries with template strings is not always a good idea

B4nan (commented):
Thanks! I understand why would you like to have the `.js` in there, not so much for the other options. To me this is only about "use .js extension" for ES modules and "use no extension" for ther rest. I would drop the other options, there is no reason for them, this is not a general purpose code generator, you will always have a package.json with `type: module` for ES modules. I dont see a reason to use mjs/cjs for entities, but maybe I am missing something.

> Another possible approach is to check package.json for "type": "module" but that felt like I was being too smart as I wouldn't know for sure which supported file format the consumer would want.

That sounds actually quite interesting, I would be up for that. It would serve as a "smart" default, so overridable if its not what the user wants.

packages/entity-generator/src/SourceFile.ts
JSin: Funny enough, I woke up thinking the same thing that it was hard to read.

I changed it to a temporary variable. Let me know if that is fine or if you prefer switching to an if-else statement on esm or not instead to be more explicit.

JSin:
Took your advice. Added a default option to check if an esm is turned on through `package.json`. Removed all the other options (cjs/mjs) and instead turned into a boolean to be a esm import.

B4nan:
Thanks!

PR #3431, opened 2022-08-30
feat(knex): add options params to `create` + `assign` methods within EntityRepository

tsangste:
Planning to use the multi-schema for multi-tenant but want to be able to set the schema on `create` or `assign` when using a EntityRepository.

I have noticed that `merge` method exposes its options so added an optional param for `create` and `assign` to gain access to the option so I can set the schema.

Should I be adding tests for this?


B4nan (approved):

B4nan:
Thanks!

PR #3439, opened 2022-08-31
fix(core): compile with `module: 'Node16'` to have real dynamic imports

B4nan:


biancazzurri:
Hey! I think that something is not compatible with Node14 because when I use the cli and the config loads 'async_hooks', like this: 
`import { AsyncLocalStorage } from 'async_hooks';`
it gives error:
` Module '"async_hooks"' has no exported member 'AsyncLocalStorage'`

PR #3453, opened 2022-09-04
docs(decorators): fix broken table

mellanslag-de:


B4nan (changes requested):
looking good, but we need to port the fix also to the next docs (`docs/docs` folder), this change is only affecting the 5.3 docs snapshot.

mellanslag-de:
Indeed, fixed it, thanks for the feedback!

B4nan (approved):

PR #3464, opened 2022-09-06
fix(core): add missing `MIKRO_ORM_SCHEMA` env var

milesje:
Added MIKRO_ORM_SCHEMA to the environment variables being read from .env file.


B4nan (approved):
thanks!

PR #3475, opened 2022-09-08
feat(core): allow custom ORM prop name in `@UseRequestContext()`

### LINKED ISSUE #3474: Feature Request: Custom orm property name in UseRequestContext decorator
That's a good idea to support custom attribute names:

```ts
constructor(
    private readonly mikroORM: MikroORM) {

}

@UseRequestContext(type => type.mikroORM)
public someFunction(id: number) {
  // ...
}
```

Reference code:

https://github.com/mikro-orm/mikro-orm/blob/b7f479aebf2032e86eac85eb9f64d40fe7c9443e/packages/core/src/decorators/UseRequestContext.ts#L4-L22

hungtcs:
We can customize the parameter name of `MikroORM` in `@UseRequestContext` decorator

```ts

@Injectable()
export class SomeService {

  constructor(
      private readonly mikroORM: MikroORM) {

  }

  @UseRequestContext<SomeService>(type => type.mikroORM)
  public someFunction(id: number) {
    // ...
  }

}
```

---

Closes #3474

### DIFF
--- a/packages/core/src/decorators/UseRequestContext.ts
+++ b/packages/core/src/decorators/UseRequestContext.ts
@@ -1,7 +1,7 @@
 import { MikroORM } from '../MikroORM';
 import { RequestContext } from '../utils/RequestContext';
 
-export function UseRequestContext<T>(getContext?: MikroORM | ((type: T) => MikroORM)): MethodDecorator {
+export function UseRequestContext<T>(getContext?: MikroORM | ((type?: T) => MikroORM)): MethodDecorator {
   return function (target: any, propertyKey: string | symbol, descriptor: PropertyDescriptor) {
     const originalMethod = descriptor.value;
     descriptor.value = async function (this: T, ...args: any[]) {
--- a/packages/core/src/decorators/UseRequestContext.ts
+++ b/packages/core/src/decorators/UseRequestContext.ts
@@ -9,7 +9,7 @@ export function UseRequestContext<T>(getContext?: MikroORM | ((type?: T) => Mikr
       const orm = getContext instanceof MikroORM ? getContext : (getContext?.(this) ?? (this as any).orm);
 
       if (!(orm as unknown instanceof MikroORM)) {
-        throw new Error('@UseRequestContext() decorator can only be applied to methods of classes with `orm: MikroORM` property, or with a callback parameter like `@UseRequestContext(type => type.orm)`');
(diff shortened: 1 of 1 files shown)


packages/core/src/decorators/UseRequestContext.ts
```diff
@@ -1,15 +1,15 @@
 import { MikroORM } from '../MikroORM';
 import { RequestContext } from '../utils/RequestContext';
 
-export function UseRequestContext(getContext?: MikroORM | (() => MikroORM)) {
+export function UseRequestContext<T>(getContext?: MikroORM | ((type: T) => MikroORM)): MethodDecorator {
```
B4nan, line 4: this needs to be optional, the use case was doing `() => orm` and that needs to be working too
```suggestion
export function UseRequestContext<T>(getContext?: MikroORM | ((type?: T) => MikroORM)): MethodDecorator {
```

B4nan (changes requested):
thanks!

types could be improved but i guess it does not matter much in here :]

packages/core/src/decorators/UseRequestContext.ts
```diff
@@ -1,15 +1,15 @@
...
     const originalMethod = descriptor.value;
-    descriptor.value = async function (this: { orm: MikroORM }, ...args: any[]) {
+    descriptor.value = async function (this: T, ...args: any[]) {
       /* istanbul ignore next */
-      const orm = getContext instanceof MikroORM ? getContext : (getContext?.() ?? this.orm);
+      const orm = getContext instanceof MikroORM ? getContext : (getContext?.(this) ?? (this as any).orm);
 
       if (!(orm as unknown instanceof MikroORM)) {
-        throw new Error('@UseRequestContext() decorator can only be applied to methods of classes with `orm: MikroORM` property, or with a callback parameter like `@UseRequestContext(() => orm)`');
+        throw new Error('@UseRequestContext() decorator can only be applied to methods of classes with `orm: MikroORM` property, or with a callback parameter like `@UseRequestContext(type => type.orm)`');
```
B4nan, line 12: i would probably keep the old error message. this new use case is rather edge case, and the error now suggest it as it would be the go to way. it is not, the parameter signature is there to support providing the orm from outside, not from the class level.

```suggestion
        throw new Error('@UseRequestContext() decorator can only be applied to methods of classes with `orm: MikroORM` property, or with a callback parameter like `@UseRequestContext(() => orm)`');
```

or we would need to rephrase it more

B4nan (approved):

PR #3482, opened 2022-09-10
feat(entity-generator): generate `OptionalProps` symbols

rhyek:
…ies that define a default value

### DIFF
--- a/packages/entity-generator/src/SourceFile.ts
+++ b/packages/entity-generator/src/SourceFile.ts
@@ -52,10 +52,11 @@ export class SourceFile {
           optionalProperties.push(prop);
       }
     });
+
     if (optionalProperties.length > 0) {
         this.coreImports.add('OptionalProps');
         const optionalPropertyNames = optionalProperties.map(prop => `'${prop.name}'`).sort();
-        ret += `\n\n${' '.repeat(2)}[OptionalProps]: ${optionalPropertyNames.join(' | ')};`;
+        ret += `\n\n${' '.repeat(2)}[OptionalProps]?: ${optionalPropertyNames.join(' | ')};`;
     }
     ret += `${classBody}}\n`;
     const imports = [`import { ${([...this.coreImports].sort().join(', '))} } from '@mikro-orm/core';`];
--- a/packages/entity-generator/src/SourceFile.ts
+++ b/packages/entity-generator/src/SourceFile.ts
@@ -27,27 +27,37 @@ export class SourceFile {
       ret += `@Unique({ name: '${index.name}', properties: [${properties.join(', ')}] })\n`;
     });
 
-    ret += `export class ${this.meta.className} {\n`;
+    ret += `export class ${this.meta.className} {`;
     const enumDefinitions: string[] = [];
+    const optionalProperties: EntityProperty<any>[] = [];
+    let classBody = '\n';
     Object.values(this.meta.properties).forEach(prop => {
       const decorator = this.getPropertyDecorator(prop, 2);
       const definition = this.getPropertyDefinition(prop, 2);
 
-      if (!ret.endsWith('\n\n')) {
-        ret += '\n';
+      if (!classBody.endsWith('\n\n')) {
+        classBody += '\n';
       }
 
-      ret += decorator;
-      ret += definition;
-      ret += '\n';
+      classBody += decorator;
+      classBody += definition;
+      classBody += '\n';
 
       if (prop.enum) {
         const enumClassName = this.namingStrategy.getClassName(this.meta.collection + '_' + prop.fieldNames[0], '_');
         enumDefinitions.push(this.getEnumClassDefinition(enumClassName, prop.items as string[], 2));
       }
-    });
-    ret += '}\n';
 
+      if (!prop.nullable && typeof prop.default !== 'undefined') {
+          optionalProperties.push(prop);
+      }
+    });
+    if (optionalProperties.length > 0) {
+        this.coreImports.add('OptionalProps');
+        const optionalPropertyNames = optionalProperties.map(prop => `'${prop.name}'`).sort();
+        ret += `\n\n${' '.repeat(2)}[OptionalProps]: ${optionalPropertyNames.join(' | ')};`;
+    }
+    ret += `${classBody}}\n`;
     const imports = [`import { ${([...this.coreImports].sort().join(', '))} } from '@mikro-orm/core';`];
     const entityImportExtension = this.esmImport ? '.js' : '';
     const entityImports = [...this.entityImports].filter(e => e !== this.meta.className);
--- a/tests/features/entity-generator/EntityGenerator.test.ts
+++ b/tests/features/entity-generator/EntityGenerator.test.ts
@@ -199,4 +199,14 @@ describe('EntityGenerator', () => {
     await orm.close(true);
   });
 
+  test('generate OptionalProps and include properties for columns that are not nullable, but have defaults', async () => {
+    const orm = await initORMMySql('mysql', {}, true);
+    await orm.schema.dropSchema();
+    const schema = "create table if not exists `account` (`id` bigint(20) unsigned NOT NULL AUTO_INCREMENT, `active` tinyint(1) NOT NULL DEFAULT '0', `receive_email_notifications` tinyint(1) NOT NULL DEFAULT '1', PRIMARY KEY (`id`)) default character set utf8mb4 engine = InnoDB;";
+    await orm.schema.execute(schema);
+    const dump = await orm.entityGenerator.generate({ save: false, baseDir: './temp/entities' });
+    expect(dump).toMatchSnapshot('generate-OptionalProps');
+    await orm.schema.execute('drop table if exists `account`');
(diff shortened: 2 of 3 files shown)


packages/entity-generator/src/SourceFile.ts
```diff
@@ -27,27 +27,37 @@ export class SourceFile {
...
         enumDefinitions.push(this.getEnumClassDefinition(enumClassName, prop.items as string[], 2));
       }
-    });
-    ret += '}\n';
 
+      if (!prop.nullable && typeof prop.default !== 'undefined') {
+          optionalProperties.push(prop);
+      }
+    });
+    if (optionalProperties.length > 0) {
```
B4nan, line 55:
```suggestion

    if (optionalProperties.length > 0) {
```

B4nan (approved):
Just one nit, otherwise looking good, thanks!

Btw this PR [got me thinking](https://twitter.com/MikroOrm/status/1568534840370237441), maybe its about time to provide `@mikro-orm/client` package or something similar with schema first codegen approach.

B4nan (approved):

B4nan:
Btw I did some PoC of that generated client I mentioned in the tweet, and found few more things we first need to support in the generator. If you like this part of the codebase, I'll be happy to accept PRs for following:

- generate simple repositories next to the entities (in separate files, linking them via `EntityRepositoryType` symbol)
- allow to generate JS/D.TS, again in separate files - for code gen we would have to go with this approach probably
- in the generated repository, we should extend the driver package, for that we will need a way to get package name and driver class (there is no way currently)

Last step would be that "generated client". I have this simple script I used for the PoC, it generates entities, repositories and a client file which initializes the ORM and exports a map of services:

```ts
import { dirname } from 'node:path';
import fs from 'fs-extra';
import { fileURLToPath } from 'node:url';

// TODO
//  - generate repositories in separate files
//  - provide way to generate .js + .d.ts files
//  - generate experimental client
//  - allow detecting m:n with AI PK
//  - import EM/repo from driver package instead of core

const orm = await MikroORM.init();
const dirName = dirname(fileURLToPath(import.meta.url));
const driver = { package: '@mikro-orm/better-sqlite', className: 'BetterSqliteDriver' };

// clean up everything in the folder except dot files
await fs.emptyDir(dirName + '/entities');
await fs.emptyDir(dirName + '/repositories');

const entityGenerator = orm.config.get('entityGenerator');
entityGenerator.identifiedReferences = true;
entityGenerator.bidirectionalRelations = true;
// orm.config.set('entityGenerator', entityGenerator);
const ret = await orm.entityGenerator.generate({
  baseDir: dirName + '/entities',
  save: true,
});

const metadata = orm.getMetadata().getAll();
const entities = Object.values(metadata).filter(meta => !meta.pivotTable && !meta.embeddable && !meta.virtual);

for (const meta of entities) {
  const code = [
    `import { EntityRepository } from '${driver.package}';`,
    `import { ${meta.className} } from '../entities/${meta.className}.js';`,
    '',
    `export class ${meta.className}Repository extends EntityRepository<${meta.className}> { }`,
  ];
  await fs.writeFile(`${dirName}/repositories/${meta.className}Repository.ts`, code.join('\n'));
}

const client: string[] = [];
const coreImports: string[] = [];

client.push(`import { ${driver.className} } from '${driver.package}';`);

for (const meta of entities) {
  client.push(`import { ${meta.className} } from './entities/${meta.className}.js';`);
  client.push(`import { ${meta.className}Repository } from './repositories/${meta.className}Repository.js';`);
}

client.push('');
coreImports.push('MikroORM');
client.push(`const orm = await MikroORM.init<${driver.className}>();`);
// we will need something like this, but that itself won't allow extension, we would have to check if something was
// discovered, discover the generated entities only if they are not provided and discovered already. user would extend
// the generated entities, this means we might have to mark them as base entities somehow, to get around duplicity warnings
client.push(`await orm.discoverEntity([${entities.map(meta => meta.className).join(', ')}]);`);
client.push(`const client = {`);
client.push(`  orm,`);
client.push(`  em: orm.em,`);
client.push(`  schema: orm.schema,`);
client.push(`  seeder: orm.seeder,`);
client.push(`  migrator: orm.migrator,`);

function lcfirst(word: string) {
  return word[0].toLowerCase() + word.substring(1);
}

for (const meta of entities) {
  client.push(`  ${lcfirst(meta.className)}: orm.em.getRepository(${meta.className}) as ${meta.className}Repository,`);
}

client.push(`};`);
client.push(`export default client;`);
client.push('');

client.unshift(`import { ${coreImports.join(', ')} } from '@mikro-orm/core';`);

console.log(client.join('\n'));
await fs.writeFile(dirName + '/client.ts', client.join('\n'));

console.info('Database client generated');
await orm.close();
```

This would work only when used with the generated entities, important bit here is finding a way how to make this work dynamically with user provided entitites (that would extend the generated ones).

rhyek:
@B4nan thanks for merging. i am still getting familiar with this package's code. could potentially help with your idea, but would need to understand the use-case first. i saw you mentioned schema-first code generation like prisma. sounds interesting. i will say i've tried prisma briefly and am not completely sold on the idea of schema-first for an orm. in the case of prisma it is a new domain-specific language that they needed to support and a language-server to support it (not sure if first or third-party).

anyways. i might not have understood your idea, correctly. i'll read through your tweet later. kinda swamped with work atm, heh. i'm sure that is familiar territory for you.

rhyek:
i had another observation. i noticed that for nullable columns you are generating:
```ts
@Property({ nullable: true })
aProperty?: string;
```
should that be `aProperty?: string | null`, instead? i have strict null checks enabled for my projects and noticed this was an issue for me today.

B4nan:
> should that be aProperty?: string | null, instead? i have strict null checks enabled for my projects and noticed this was an issue for me today.

This really depends on the user, personally I dont like to define properties that way, it breaks reflect-metadata, and to be consistent, it kinda requires the property initializer. I dont mind having a config toggle for this, but I dont think we should change the defaults now. But maybe I just had some kind of aversion to this, in context of entity generator we can easily provide the type explicitly (so we dont have to care about reflect-metadata at all). I guess the form I dislike is `aProperty: string | null = null`, which some people use I believe, and is not 100% precise either. 

Btw I made an issue based on the tweet, with initial PoC, maybe that will help you understand better: https://github.com/mikro-orm/mikro-orm/issues/3484

The point is to have the very same ORM support, just without creating the entities manually. Kinda like the prisma client, we would generate the entities (probably via CLI), the question is where and how it would work with IDE support if we put it into `node_modules`. For users this would work as usually, with one exception, there would be new package they would import this "client" from, and they would just use it, no need to init the ORM, register DI, care about imports, etc. It would be another way to use the ORM, for simpler projects.

```ts
import { client } from '@mikro-orm/client';

const user = client.user.create({ name: '...', email: '...' });
await client.em.persist(user).flush();
```

Things like this would be possible, without much setup, just the ORM config so the CLI knows what to build. The `client` object would be fully typed, `client.user` being a `User` repository instance. For this to make sense, we would need to find a way to let users provide their own entities (and extend the ones generated by the ORM probably). I think it should be doable, the interface might not be great, but it will unlock new ways and that is important bit. Imagine you would just add something like `generator: { customEntity: { user: MyUser } }` or similar - which would map your own `MyUser` entity on top of what the ORM generates for `user` table. You could skip all the boring code, all properties would be automatically there.

rhyek:
@B4nan, for
```sql
`first_name` varchar(255) null
```
would this make the most sense?
```ts
[OptionalProps]?: 'firstName';

@Property({ type: 'varchar', length: 255, nullable: true })
firstName!: string | null;
```

B4nan:
If you want to use the `null` type explicitly, i'd also add a runtime default:

```ts
@Property({ type: 'varchar', length: 255, nullable: true })
firstName: string | null = null;
```

Personally I just go with optional properties and don't use `null` explicitly, I know its not 100% correct but its much more ergonomic.

rhyek:
My thinking is I want to represent more accurately what the orm will return once I run a select query. The `= null` can be useful when creating entities (and before persisting), but to be consistent I'd then want to also do stuff like:
```ts
@Property({ type: 'varchar', length: 255, nullable: true, default: 'hello' })
firstName: string | null ='hello';
```
Might be problematic for many cases.

B4nan:
Keep in mind that you can still end up with `undefined` anywhere if you use partial loading. Another option is the opposite direction, enforcing `undefined` instead of `null` via `forceUndefined: true`.

B4nan:
Oh we are talking about what the entity generator does? I thought its a general question. I dont mind supporthing this in the generor, but rather opt-in. It should already produce the correct database default too.

rhyek:
What do you mean by "It should already produce the correct database default too."? Can you give an example?

B4nan:
https://github.com/mikro-orm/mikro-orm/blob/master/tests/features/entity-generator/__snapshots__/EntityGenerator.test.ts.snap#L15-L16

But I guess we should also do the runtime default there for such properties. In latest version (maybe not yet released) its also possible to infer the runtime defaults (as long as the entity instance can be created without any arguments via constructor).

PR #3503, opened 2022-09-14
feat(schema): add ability to ignore specific column changes

### LINKED ISSUE #1904: Workaround for virtual columns
Although MicroORM does not explicitly support MySql virtual columns, I can create one quite easily using `columnType`, e.g.

```
@Entity
class Foo {
  @Property({ columnType: 'bool GENERATED ALWAYS AS (<SOME EXPRESSION>))' })
  generated_value?: boolean
}
```

This appears to work fine, except that each `mikro-orm migration:create` will produce an `alter table ... modify ...` for that column, because the generator expression is currently not read back from the information schema.

I understand that implementing virtual columns properly would be a major effort.
For now, I would be happy if I could tell MicroORM to just ignore that column when generating migrations.

If it helps, I would even be fine with the column not beeing part of the entity class, i.e. not having MikroORM be aware of it at all.

I'm using the virtual column to enforce a partial uniqueness constraint on the table, and I would have to find a completely different solution to doing that, e.g. triggers (which MikroORM also does not support at the moment, I think?) or just using `LOCK TABLES`.

PenguinToast:
```ts
@Property({
  columnType: 'timestamp',
  extra: 'VIRTUAL GENERATED',
  ignoreSchemaChanges: ['type', 'extra'],
})
changingField!: Date;
```

This is useful for situations such as #1904, where `knex` is unable to properly diff the column.

Fixes #1904

### DIFF
--- a/docs/docs/defining-entities.md
+++ b/docs/docs/defining-entities.md
@@ -1281,6 +1281,80 @@ export abstract class CustomBaseEntity {
 }
 ```
 
+## SQL Generated columns
+
+Knex currently does not support generated columns, so the schema generator
+cannot properly diff them. To work around this, we can set `ignoreSchemaChanges`
+on a property to avoid a perpetual diff from the schema generator
+
+<Tabs
+groupId="entity-def"
+defaultValue="reflect-metadata"
+values={[
+{label: 'reflect-metadata', value: 'reflect-metadata'},
+{label: 'ts-morph', value: 'ts-morph'},
+{label: 'EntitySchema', value: 'entity-schema'},
+]
+}>
+<TabItem value="reflect-metadata">
+
+```ts title="./entities/Book.ts"
+@Entity
+export class Book {
+
+  @Property()
+  title!: string;
+
+  @Property({
+    columnType: 'VARCHAR GENERATED ALWAYS AS (LOWER(`title`)) VIRTUAL',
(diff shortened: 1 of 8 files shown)


B4nan (approved):
Looking great, thanks! 

Two notes:

- are we sure ignoring `type` and `extra` props is enough? we can always add more, so not a big deal, and I guess the `type` here is the most important part
- let's document this somewhere, I would go with the https://mikro-orm.io/docs/schema-generator and https://mikro-orm.io/docs/defining-entities pages, cant be very brief note about this, but it should be there, with a small example (with the generated column, that's nice and probably the most important use case for this)

PenguinToast:
@B4nan wow, thanks for the quick review!

Added a commit w/ the documentation changes. As for your other comment, when I tested this change against our database, ignoring `type` and `extra` changes seemed to be enough (`mikro-orm schema:update --dump` no longer generated any statements).

B4nan:
FYI there are some broken links

https://github.com/mikro-orm/mikro-orm/actions/runs/3051149375/jobs/4919032117

PenguinToast:
> FYI there are some broken links
> 
> https://github.com/mikro-orm/mikro-orm/actions/runs/3051149375/jobs/4919032117

Ah, my bad -- thanks for the fix!

PR #3563, opened 2022-10-04
test: add failing test case for issue #3564

Langstra:


B4nan:
why did you close this and removed the repro? it was the one with orphan removal on owning side? if so, lets keep it here, i was to debug that first before we call it unsupported

Langstra:
Huh, weird I thought had created a seperate branch for this. And I must have locally did something with master and pushed that. I'll repush it.

PR #3584, opened 2022-10-11
feat(core): allow JsonType to use a stable JSON.stringify algorithm

### LINKED ISSUE #3583: Diffs when json key order changes
**Describe the bug**

If we update a json field with an identical value, only with a different key order, it's present in changeset (and thus triggers updates).

TLDR:
```ts
// pre-existing row.someJsonField = { a: 1, b: 2 }
row.someJsonField = { b: 2, a: 1 }
// the json didn't really changed, yet changeset and updates won't be empty
```

**To Reproduce**

Minimal full repro: https://stackblitz.com/edit/mikro-orm-repro-svx3gr?file=test.ts

**Expected behavior**

The JSON are identical, there should be nothing in changeset and no updates.
(or at least there should be an option to enable this behaviour)

I'm opening a PR with a suggested improvement right away.

**Versions**

| Dependency | Version |
| - | - |
| node | v16.14.2 |
| typescript | 4.8.3 |
| mikro-orm | 5.4.2 |
| your-driver | Postgres (repro in SQLite) |

gggdomi:
fixes #3583 

We introduced an option on JsonType to allow user to choose between speed or better serialization.

I'm not sure if the option is required, or if stable stringify should simply be the only option because it's more robust ?  
I'm not sure either of what you prefer as a default: slightly more speed or less diffs. In doubt, we kept the the old behaviour by default.

### DIFF
--- a/docs/docs/custom-types.md
+++ b/docs/docs/custom-types.md
@@ -313,6 +313,14 @@ and `stringify` only when needed).
 object?: { foo: string; bar: number };
 ```
 
+The default `stringify` will output diffs for changes in object key order.  
+If you want to minimize diffs, you can use a slower, more precise, serialization function that will serialize `{a: 1, b: 2}` and `{b: 2, a: 1}` in identical ways.
+
+```ts
+@Property({ type: new JsonType({ useStableStringify: true }) })
+object: { foo: string; bar: number };
+```
+
 ### DateType
 
 To store dates without time information, we can use `DateType`. It does use `date`
--- a/packages/core/src/types/JsonType.ts
+++ b/packages/core/src/types/JsonType.ts
@@ -2,14 +2,24 @@ import { Type } from './Type';
 import type { Platform } from '../platforms';
 import type { EntityProperty } from '../typings';
 import { Utils } from '../utils';
+import { stableStringify } from '../utils/stableStringify';
 
 export class JsonType extends Type<unknown, string | null> {
+
+  constructor(
+    private opts: { useStableStringify?: boolean } = {},
+  ) {
+    super();
+  }
 
   convertToDatabaseValue(value: unknown, platform: Platform): string | null {
     if (platform.convertsJsonAutomatically(true) || value === null) {
       return value as string;
     }
 
+    if (this.opts.useStableStringify) {
+      return stableStringify(value);
+    }
     return JSON.stringify(value);
   }
 
--- a/packages/core/src/utils/stableStringify.ts
+++ b/packages/core/src/utils/stableStringify.ts
@@ -1,0 +1,41 @@
+// extracted and modified from npm package "fast-json-stable-hash"
+export function stableStringify(obj: any): string {
+    const type = typeof obj;
+    if (type === 'string') {
+      return JSON.stringify(obj);
+    }
+    if (Array.isArray(obj)) {
+      let str = '[';
+      const al = obj.length - 1;
+      for (let i = 0; i < obj.length; i++) {
+        str += stableStringify(obj[i]);
+        if (i !== al) {
+          str += ',';
+        }
+      }
+      return `${str}]`;
+    }
+    if (type === 'object' && obj !== null) {
+      let str = '{';
+      const keys = Object.keys(obj).sort();
+      const kl = keys.length - 1;
+      for (let i = 0; i < keys.length; i++) {
+        const key = keys[i];
+        const val = (obj as any)[key];
+        if (val === undefined) {
+          continue;
+        }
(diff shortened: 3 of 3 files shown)


B4nan:
Can you prepare some benchmarks so we know how much slower this solution is? If its fast enough, I dont mind doing this by default, but it would have to be very close to the native implementation.

I am not sure if the type constructor is the right place for this (I do understand this is the "least modified code path"), a global config option would be definitely a good idea, and a property level option to have fine grained control. Because in generall you dont need to work with the type instance, a `type: 'json'` (or `type: t.json`) will work the same here, and I dont want to force users to instantiate the types for this.

Maybe something like this?

```ts
@Property({ type: 'json', stableStringify: true })
object: { foo: string; bar: number };
```

That way we could easily control it globally too, I guess if the user wants it stable, they will want to have it stable for all the JSON properties.

I also dont mind having the constructor option in `JsonType`, I'd like to have it available in other ways too.

rvion:
> Can you prepare some benchmarks so we know how much slower this solution is? If its fast enough, I dont mind doing this by default, but it would have to be very close to the native implementation.

Will do

> I am not sure if the type constructor is the right place for this
> a global config option would be definitely a good idea, and a property level option to have fine grained control. 
> That way we could easily control it globally too, I guess if the user wants it stable, they will want to have it stable for all the JSON properties.

yeah, makes sense 👍 

-----

to add some thoughs about json fields before doing  more:

- going though strings to check for diffs may not be the best way.
   - a custom `deepEqual` made for ORM would be better / faster / use less memory when most diffs ends-up without changes
     - lodash and other ones are both too complex (support dates) and wrong (they compare the number of keys, which incorectly deals with undefined `{a:1, b:undefined}`  and `{a:1}` should be equal under a comparaison diff.

- sometimes, I think I'd also prefer to use ref equality (===) instead of real diffs, because I know my objects will only be replaces, not modified.  this is unsafe, and should totally be behind a "I know what I do and will not complain when it turns out I was wrong in my assumptions", but it is massively faster.
  - should there be an option for json object to also `Object.freeze` them, especially when === comparaison is active
      (this would catch real-life bugs for specific needs, just had to quickly debug that in my project yesterday, I had one field that needed a stable stringify compare, and an other that only needed a === comparaison)
 
- should there a way to specify diff strategy on Types, i.e. `ReferenceEquality`(or 'StrictEquality') | `==`(or 'Equality') | 'ViaDBValue', are there mutually exclusive "axis" here that would deserve a different strategy configuration

- would it makes sense to allow use to use hints at the compare change time, something like `computeChangeSets({ strategy: { BookEntity: {SomeJsonField: 'StricEquality' }} )` or `computeChangeSets({ useStricEqualityFor: ['Entity.Field']] })`. This is probably not the lowest hanging fruit for perf / memory usage optimisation

rvion:
@B4nan the perf impact is massive for my use-case, if used everywhere ! 

🔴 this should not be merged as is 😨 

👉 One big problem is that registering managed entities seems to also serialize JSON values pre-emptively 😨 

image

also, in my random current use case, >90%of json fields of my loaded models would be perfectly fine if compared as references, which I don't know how to do with current type abstraction.

(side note: all those little perf problems really seem easilly fixable. I have a ~150seconds script, where most of the time is just MikroORM doing things =>  early profiling show quick win everywhere, despite the code already beeing very clean; I know you probably have many other things to do, but I'd really like to help optimize mikro-orm further ! Your code is really clean ! :)

gggdomi:
@B4nan your precisions have been noted. I'll need to take a couple days to dive deeper in the Mikro ORM codebase.
An approximative bench : https://jsbench.me/x1l94p0qvx/1 (TLDR: 60% slower)

In any case, @rvion profiling proves that more work is required. I'll get back to you.

rvion:
> Maybe for your use case, the go to way shouldn't be replacing references, but object mutation instead and there you can't just compare references.

yeah, especially since it's impossible to hack around field order, since postgres will not preserve ordering on update like node does. But since node preserves the field oder from postgres, I went with `Object.assign(entity.jsonField, newJsonField)` so `JSON.stringify()` is similar, since `entity.jsonField` field order will remain unchanged.

-------

your proposal to allow custom way to compare, without having to serialize first the fields would indeed solve all my problems.

for this specific entity, I'll go with a custom `compareObject` function, to skip the double json serialization (at fetch and at compare time)

And for for some of my other entities (with hundreds of thouthands of instances loaded in the EM in this very specific example), some of them have json fields known to be frozen and only assignable by reference, I'll go with an even simpler reference comparaison, thus saving seconds of runtime

PR #3586, opened 2022-10-12
test: add expect for performance claim

Langstra:
Interestingly enough the comment claims that the test should run within 50 to 100ms. However, adding an expect shows that this test fails.


B4nan:
We cant test this in CI, as CI is very unstable when it comes to performance. Also performance differs when you run a single test vs running everything.

Langstra:
Hmm... That is right yeah, but it is hard to have this as a test then, since you won't notice it if the performance is off by a lot, right?

By the way, on my machine with an intel i7 (9th gen I believe) I got around 160ms.

B4nan:
I run the tests locally very often, the time gets logged, so I can see the changes. Also if its significant, it will be visible on the total test run time too. It would be nice to automate this, but I am not interested in adding flaky tests (already got some, lol). Maybe if the perf tests would be executed outside of jest, it could be more predictable. 

And well, Intel being slow, that's not my problem :P Didn't expected such a difference, do you run just the single test?

Langstra:
Hehe ofc that is not your problem. Neither did I expect such differences. Ah well, maybe someone could create a benchmarking suite and run that on every pr

Issue #3588, opened 2022-10-12
Allow to flush without recomputing changeset


rvion:
For some large transaction I have, I show a human readable summary before flushing the context, so user can manually review and accept changes about to be commited.

for this, I have a `makeHumanReadableTextSummary` that itself need to `computeChangeSets()`

sadly, right after human confirmation, `em.flush()` recompute the changeset despite no modification having taken place.

<img width="947" alt="image" src="https://user-images.githubusercontent.com/2150990/195380569-0446c6f3-9152-43da-906d-5a9ef40a6e26.png">

**Describe the solution you'd like**

can we have a `em.flushWithoutRecomputingChangesets()` or `em.flush({computeChangeSets: false})` 

**Describe alternatives you've considered**

a possible way to do this would be to copy-past the whole code in `UnitOfWork.commit` and `UnitOfWork.doCommit`, comment out the single line `this.computeChangeSets();`, and call that function binding it to my unit of work.

it seems like a very bad idea that will not survive well across releases

B4nan:
Why dont you use a onFlush event instead?

rvion:
@B4nan I didn't think about it to be honest x)

This being said, it doesn't feel very practical the way I use it. 

I use those human readable summaries in different code branches
sometimes several times before actually commiting the transaction, 
sometimes in situation where I already know  beforehand that the changes won't be commited. 

if I just take the case where a I want to interractively request several human summaries with various focuses may be done in the onFlush, but it would feel a little brittle I think 🤔 
it would force to split other scripts in several pieces so it all runs within the onFlush event hook

I also understand why you could prefer to avoid leaking those abstractions, though x)

B4nan:
Can you share some code, I dont really understand what are you actually doing. Because that suggested PR seems weird, the changesets will be cleared once the `UoW.commit` resolves.

PR #3589, opened 2022-10-12
feat(core): allow to commit UnitOfWork without re-computing changeSet

rvion:
example code that would fix https://github.com/mikro-orm/mikro-orm/issues/3588


rvion:
@B4nan if you think this could be valuable, feel free to discard my PR and do it better 😄 I just submitted as an illustration

PR #3596, opened 2022-10-14
perf(core): do not run Entity constructors twice per field while discovering default values

rvion:
I was investigating why my Entity constructors were begin called, both 1 - at startup, and 2 - at creation via `em.getRepository(...).create()` time despite doc saying

> Internally, MikroORM never calls entity constructor, so you are free to use it as you wish. The constructor will be called only when you instantiate the class yourself via new operator, so it is a handy place to require your data when creating new entity.

---

 I found that :

1. the constructor was indeed called twice for every prop in metadata, which this PR attempt to improve by instanciating entities twice only during metadata instanciation, and reusing those two instances for every property

2. `Internally, MikroORM never calls entity constructor` only refers to entities fetched from the DB, since entity constructor, as suggested in other pages, will indeed be called for every em.create(). 

3. 

> The constructor will be called only when you instantiate the class yourself via new operator

this is wrong ?

### DIFF
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -446,11 +446,20 @@ export class MetadataDiscovery {
     meta.forceConstructor = this.shouldForceConstructorUsage(meta);
     this.validator.validateEntityDefinition(this.metadata, meta.name!);
 
+    let entity1: any | undefined;
+    let entity2: any | undefined;
+    try {
+      // try to create two entity instances to detect if initialized values are stable
+       entity1 = new meta.class();
+       entity2 = new meta.class();
+    } catch {
+      // ignore
+    }
     for (const prop of Object.values(meta.properties)) {
       this.initNullability(prop);
       this.applyNamingStrategy(meta, prop);
-      this.initDefaultValue(meta, prop);
-      this.initVersionProperty(meta, prop);
+      this.initDefaultValue(meta, prop, entity1, entity2);
+      this.initVersionProperty(meta, prop, entity1, entity2);
       this.initCustomType(meta, prop);
       await this.initColumnType(meta, prop, meta.path);
       this.initRelation(prop);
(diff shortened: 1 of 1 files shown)


rvion:
📝 : this may just be a useless minor micro-optimisation; I opened it just in case

B4nan:
I really doubt this would be even measurable.

> Internally, MikroORM never calls entity constructor only refers to entities fetched from the DB, since entity constructor, as suggested in other pages, will indeed be called for every em.create().

You can create managed entities this way too, so not _every_ call. That sentence seems correct to me - the word _internally_ is important, using `em.create()` is considered explicit usage, not internal. I don't mind improving the wording, but I woudn't call it _wrong_.

rvion:
> I really doubt this would be even measurable.

@B4nan yeah, I just had some logging happening in my constructor =>  that's what made me investigate this in the first place to understand why did I had those extra logs :)

> the word internally is important, using em.create() is considered explicit usage

yeah, I now understand what you mean. I just didnt understand when I first saw it. I guess a small tweak could help here :)

packages/core/src/metadata/MetadataDiscovery.ts
```diff
@@ -446,11 +446,20 @@ export class MetadataDiscovery {
...
 
+    let entity1: any | undefined;
+    let entity2: any | undefined;
+    try {
+      // try to create two entity instances to detect if initialized values are stable
+       entity1 = new meta.class();
+       entity2 = new meta.class();
+    } catch {
+      // ignore
+    }
```
B4nan, line 457: i dont want this here, this should stay in the method for default values or has its own method. i dont mind having cache around it, but i do mind putting it on a place without correct context, and moverover passing some "entity1 and entity2" parameters, thats quite confusing.

packages/core/src/metadata/MetadataDiscovery.ts
```diff
@@ -910,14 +919,10 @@ export class MetadataDiscovery {
...
-  private initDefaultValue(meta: EntityMetadata, prop: EntityProperty): void {
+  private initDefaultValue(meta: EntityMetadata, prop: EntityProperty, entity1: any, entity2: any): void {
     try {
-      // try to create two entity instances to detect the value is stable
-      const entity1 = new meta.class();
-      const entity2 = new meta.class();
-
       // we compare the two values by reference, this will discard things like `new Date()`
-      if (entity1[prop.name] != null && entity1[prop.name] === entity2[prop.name]) {
+      if (entity1 != null && entity2 != null && entity1[prop.name] != null && entity1[prop.name] === entity2[prop.name]) {
```
B4nan, line 925: dont use `any`, dont allow empty values at all. this is internal code, private method, its safe to assume the parameters are correct. i dont see a way how this could end up as being null or undefined

also the code was in try/catch block because of the entity instantiation, which is not here anymore

B4nan:
I'll close this one as well, I dont think the perf-gain is there, not a measurable one, not worth complicating the code more.

PR #3613, opened 2022-10-19
docs: add clarification on assigning collections

Langstra:


B4nan (commented):
i am not sure how this is better, the information was all there, including commented example. what part of it was confusing to you?

i'd prefer the old wording that this, i am fine with adding more examples, but this dont feel like an overall improvement

Langstra:
> i am not sure how this is better, the information was all there, including commented example. what part of it was confusing to you?
> 
> i'd prefer the old wording that this, i am fine with adding more examples, but this dont feel like an overall improvement

I guess that when you know about `orphanRemoval` it is very logical that the items do not get removed from the database. However, if you do not, then it is kinda confusing they still exist afterwards. That is why I wanted to make it more explicit.

B4nan:
i dont mind the note about orphan removal, that's fine. i'd just like to keep the previous structure - assigning an array of items is important, appending item to it is quite niche feature, as normally you'd just do `ent.coll.add()` instead of using `assign` (or use `assign` with something like `[...ent.coll, newItem]` i guess)

B4nan:
> However, if you do not, then it is kinda confusing they still exist afterwards.

Thats indeed valid, but the PR is not improving on this very much. I could imagine a paragraph about this, the difference of removing item from collection and removing entity from database, with better examples (not justt entity def, but also the queries it would produce)

Langstra:
Perhaps I could link to that paragraph? However, it might be better to explain it in this section of the docs as well.

B4nan:
i dont mind duplicating the content to some degree, e.g. what we have there rn could be copied and only the collections page would have a better example (which would be linked from the assign docs)

Langstra:
Alright, so I updated `relationships.md` to correct the `orphanRemoval` property, which should not be used on the owning side (right?). Then I reverted the `entity-helper.md` and added a little explanation on what happens to items that were in the collection before and then referring to `collections.md`. Lastly I updated `collections.md` to mention `em.assign()` and reference to the `cascading.md` page for orphan removal.

For now I only changed the current docs and if you think it is all good, then I'll add this also to the 5.3 docs (to prevent doing things twice).

B4nan:
> Alright, so I updated relationships.md to correct the orphanRemoval property, which should not be used on the owning side (right?).

I said it looks weird, but that I found existing test case that does it (although not in a single flush). Maybe it can work, lets first double check that before changing the docs. I said that mainly because you cant use it with m:1 relation currently, which is kinda the same as 1:1 owning side. I did some quick searech and doctrine probably allows orphan removal on 1:1 owner (but not on m:1). Might be just a buggy implementation at our end.

Langstra:
Ah okay, yeah I also check the docs and came too the conclusion that with 1:m relations the owning side could not have the orphanRemoval: true and thus this should be also for 1:1 relations. 

Would you like me to undo those changes to the docs?

PR #3626, opened 2022-10-20
test: add failing tests for m:n relations with custom pivot table, fieldName and discriminator

jansedlon:
Hi Martin, as discussed today and yesterday in MikroOrm slack [Thread](https://mikroorm.slack.com/archives/CFSAF8DEX/p1666189658734029), I'm sending two failing cases that causes my app not to work.

### DIFF
--- a/tests/features/composite-keys/custom-pivot-entity-with-field-name-and-discriminator.test.ts
+++ b/tests/features/composite-keys/custom-pivot-entity-with-field-name-and-discriminator.test.ts
@@ -1,0 +1,108 @@
+import {
+  Collection,
+  Entity,
+  EntityCaseNamingStrategy,
+  ManyToMany,
+  ManyToOne,
+  MikroORM,
+  PrimaryKey,
+  Property,
+} from '@mikro-orm/core';
+
+@Entity({ abstract: true })
+export abstract class AbstractEntity {
+
+  @PrimaryKey()
+  public id!: number;
+
+}
+
+@Entity({
+  discriminatorColumn: 'role',
+  discriminatorMap: {
+    CREATOR: 'Creator',
+  },
+})
+class User extends AbstractEntity {
+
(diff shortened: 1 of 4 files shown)


tests/features/composite-keys/custom-pivot-entity-with-field-name-and-discriminator.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+}
+
+@Entity({ tableName: 'CreatorsOnTasks' })
+class CreatorsOnTasks {
+
+  @ManyToOne({ primary: true, fieldName: 'creatorId', entity: () => Creator })
+  creator!: Creator;
+
+  @ManyToMany({ primary: true, fieldName: 'taskId', entity: () => Task })
```
B4nan, line 56: we should validate this, only m:1 and 1:1 owners can be a primary key

```suggestion
  @ManyToOne({ primary: true, fieldName: 'taskId', entity: () => Task })
```

tests/features/composite-keys/custom-pivot-entity-with-field-name-and-discriminator.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  @ManyToMany(() => Creator, c => c.tasks)
+  creators = new Collection<Creator>(this);
+
+}
+
+@Entity({ tableName: 'CreatorsOnTasks' })
+class CreatorsOnTasks {
+
+  @ManyToOne({ primary: true, fieldName: 'creatorId', entity: () => Creator })
```
B4nan, line 53: and your second problem is that you got the property order wrong, with pivot entities the order of PKs is important.

again something to improve, this limitation make sense only for the case where both properties target the same entity type, otherwise we can safely detect the right connection

PR #3633, opened 2022-10-22
feat(cli): migration:rollup command

mareksuscak:


packages/cli/src/commands/MigrationCommandFactory.ts
```diff
@@ -181,6 +199,35 @@ export class MigrationCommandFactory {
...
+    const options = orm.config.get('migrations');
+    const snapshot = options.snapshot;
+
+    // for snapshots, we always want to use the path based on `emit` option, regardless of whether we run in ts-node context
+    /* istanbul ignore next */
+    const snapshotDir = options.emit === 'ts' && options.pathTs ? options.pathTs : options.path!;
+    const absoluteSnapshotPath = Utils.absolutePath(snapshotDir, orm.config.get('baseDir'));
+    const dbName = basename(orm.config.get('dbName'));
+    const snapshotName = options.snapshotName ?? `.snapshot-${dbName}`;
+    const snapshotPath = Utils.normalizePath(absoluteSnapshotPath, `${snapshotName}.json`);
```
mareksuscak, line 215: My original intent was to use `migrator.snapshotPath`. However it seems to be a private property right now and is not a part of the `IMigrator` interface. Some refactoring may be necessary.

packages/cli/src/commands/MigrationCommandFactory.ts
```diff
@@ -181,6 +199,35 @@ export class MigrationCommandFactory {
...
+    // Delete snapshot if exists so we start from a clean slate
+    if (snapshot && await pathExists(snapshotPath)) {
+      await remove(snapshotPath);
+    }
+
+    // Create the rollup
+    await this.handleCreateCommand(migrator, args, orm.config);
+
+    // Execute the rollup
+    await this.handleUpDownCommand(args, migrator, 'up');
```
mareksuscak, line 226: I'm questioning if it's a good idea to use the other command handlers directly or if I should try to implement all logic directly at a lower level.

packages/cli/src/commands/MigrationCommandFactory.ts
mareksuscak: Alternatively, is there a way to bypass having to delete the snapshot in order to roll up multiple migrations that are reflected in the snapshot?

packages/cli/src/commands/MigrationCommandFactory.ts
mareksuscak: In addition, I'm questioning whether or not the rollup should be automatically executed. In order to stay consistent with the create command, it most likely should not.

tests/features/cli/RollupCommand.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    const cmd = MigrationCommandFactory.create('rollup');
+
+    await expect(cmd.handler({} as any)).resolves.toBeUndefined();
+    expect(down.mock.calls.length).toBe(1);
+    expect(down.mock.calls[0][0]).toEqual({});
+    expect(closeSpy).toBeCalledTimes(1);
+    await expect(cmd.handler({ to: 'a' } as any)).resolves.toBeUndefined();
+    expect(down.mock.calls.length).toBe(2);
+    expect(down.mock.calls[1][0]).toEqual({ to: 'a' });
+    expect(closeSpy).toBeCalledTimes(2);
```
mareksuscak, line 56: This test needs more work, it's mostly just a stub for now.

packages/cli/src/commands/MigrationCommandFactory.ts
```diff
@@ -181,6 +199,38 @@ export class MigrationCommandFactory {
...
+    const snapshotName = options.snapshotName ?? `.snapshot-${dbName}`;
+    const snapshotPath = Utils.normalizePath(absoluteSnapshotPath, `${snapshotName}.json`);
+
+    // Delete snapshot if exists so we start from a clean slate
+    if (snapshot && await pathExists(snapshotPath)) {
+      await remove(snapshotPath);
+    }
+
+    // Delete migrations
+    // TODO(marek)
```
mareksuscak, line 223: Since we need to delete migration files, it may make sense to not run the down command above but rather re-implement it so we get access to the list of migrations here.

PodaruDragos:
This is actually a very nice feature to have. "collapsing" 10 migrations into 1 is something really useful down the line.
Is there no interest in this being considered / finished ? @mareksuscak  @B4nan

B4nan:
The idea is good, I am not entirely sure about the implementation. I don't like that is first migrates all the way down and up again, it could mess up with the data easily, e.g. every migration that created a column will first drop it and recreate again - but without data. This could even potentically fail if its non null without a default, as that works only on tables without any rows.

Why don't we just merge the contents of the migrations, instead of messing up with the database?

mareksuscak:
> I don't like that it first migrates all the way down and up again, it could mess up with the data easily, e.g. every migration that created a column will first drop it and recreate it again - but without data.

Totally. This was the only way I was able to get the rollup to work without writing a lot of custom logic. Also, you technically need to roll back to the first migration included in the rollup so you can do the diff and generate the rolled-up migration that contains all changes. Obviously, this wouldn't work if you wanted to run it in a non-local environment. This particular implementation is only good for rolling up migrations in feature branches as part of the final cleanup. 

I was inspired by a similar feature in Doctrine (PHP world):
https://localheinz.com/articles/2020/06/10/rolling-up-database-migrations-with-doctrine/

I'd be happy to revisit this but was holding to get some feedback from the community and I also outlined some questions as I wasn't sure what's the best way to do certain things (SDK-wise)

B4nan:
> Also, you technically need to roll back

But why? You have the migrations, they contain the queries, why recompute everything instead of just concatenating?

PodaruDragos:
Concatenating comes with the same problems doesn it ? Alters are usually
hard to handle anyways. I am very in favour of this addition just to make
it clear.

vin., 3 feb. 2023, 21:59 Martin Adámek ***@***.***> a scris:

mareksuscak:
> But why? You have the migrations, they contain the queries, why recompute everything instead of just concatenating?

You're not wrong. But consider you adding a new table, then adding a few more columns in the following migrations. If we just combine all migrations, there will be extraneous statements whereas it could be optimized to become a single statement. I know it's an optimization but there are other cases where you're experimenting with different states and so some migrations become unnecessary - which again - you could edit manually after the rollup. I'm impartial here and would be happy to explore the other approach which I did consider but again - it requires writing more custom logic - the solution contained in this PR only relies on existing commands.

B4nan:
I'd go with concatenation, as it will be safer (no chance of data loss) as well as faster (no queries involved), and you don't need to touch the snapshot file or the schema state at all. I don't mind supporting the down/up way too, but rather as opt-in than default, could be a `--recreate` flag/option.

Btw don't be afraid of doing changes around the codebase, this does not have to be just a self-contained CLI command - if you need to modify the snapshot, you can add some methods to the migrator, we can mark them as internal via jsdoc if they need to be public.

Maybe better target the v6 branch nowadays, I would be up to merging something not fully complete in there too and polish it a bit myself before the final release (which will take a few more months for sure).

mareksuscak:
@B4nan thank you for your feedback, Martin. I'll take a stab at this. It seems that I should be able to rely on the methods available in the `Migration` class that store all SQL statements in a string array and makes them available through `getQueries` so here's what I'm thinking in order of execution:

1. Determine the list of migrations to be concatenated
2. Dynamically import them and instantiate
3. Run `up` method and extract queries using `getQueries` - aggregate queries across all migrations in a new array
4. Run `reset` and then `down` method and extract queries using `getQueries` - aggregate queries across all migrations in a new array
6. Create a new migration file with all `up` and `down` queries concatenated
7. Delete the old migration files

How does that sound?

B4nan:
Migrations can contain code that directly works with the database (e.g. `this.execute()` calls), so I would rather go with direct code manipulation instead. It might be a bit tricky to do right, though. But we need to preserve those calls.

mareksuscak:
That's actually a great callout. I was trying to avoid having to use a JS/TS parser but I'm now not sure it's avoidable. I'd rather not rely on regex matching. If not, I'll have to use one of the available parsers (see the list below) in order to extract `up` and `down` functions' body. This will introduce a new dependency. Can you think of a different way to do this?

CleanShot 2023-02-19 at 12 34 23@2x

B4nan:
Closing in favor of #7468

PR #3644, opened 2022-10-24
test: 1:1 swap with different owner definition

Langstra:
This should yield the same results as the GH3614.test.ts I think. The only thing that is different here is the entity definition.

### DIFF
--- a/tests/issues/GH3614-owner-defined-different.test.ts
+++ b/tests/issues/GH3614-owner-defined-different.test.ts
@@ -1,0 +1,124 @@
+import { Entity, IdentifiedReference, OneToOne, PrimaryKey, Property, Reference, wrap } from '@mikro-orm/core';
+import { MikroORM } from '@mikro-orm/sqlite';
+import { mockLogger } from '../helpers';
+
+@Entity()
+class Project {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property()
+  name!: string;
+
+  @OneToOne(() => User, u => u.project1, {
+    eager: true,
+    orphanRemoval: true,
+    wrappedReference: true,
+  })
+  owner?: IdentifiedReference<User>;
+
+  @OneToOne(() => User, u => u.project2, {
+    eager: true,
+    wrappedReference: true,
+  })
+  secondaryOwner?: IdentifiedReference<User>;
+
+}
+
+@Entity()
+class User {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property()
+  name!: string;
+
+  @OneToOne(() => Project, p => p.owner, {
+    owner: true,
+    wrappedReference: true,
+    nullable: true,
+  })
+  project1?: IdentifiedReference<Project>;
+
+  @OneToOne(() => Project, p => p.owner, {
+    owner: true,
+    wrappedReference: true,
+    nullable: true,
+  })
+  project2?: IdentifiedReference<Project>;
+
+}
+
+let orm: MikroORM;
+
+beforeAll(async () => {
+  orm = await MikroORM.init({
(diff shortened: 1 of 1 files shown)


tests/issues/GH3614-owner-defined-different.test.ts
```diff
@@ -0,0 +1,124 @@
...
+  name!: string;
+
+  @OneToOne(() => Project, p => p.owner, {
+    owner: true,
+    wrappedReference: true,
+    nullable: true,
+  })
+  project1?: IdentifiedReference<Project>;
+
+  @OneToOne(() => Project, p => p.owner, {
```
B4nan, line 45:
```suggestion
  @OneToOne(() => Project, p => p.secondaryOwner, {
```

B4nan:
works fine if you fix that

Langstra:
Hmm... seems I was sleeping. I'm trying to debug something in this area that it almost exactly similar to this code, but still gives me a `UNIQUE constraint failed`. However, I think I got the bug right now. It is related to the way I called `orm.em.create(User, {name: 'Johnny', project1: 1})`

B4nan:
Maybe you have the same problem in your app, multiple owning properties mapped by a single inverse? I guess we could validate that.

Langstra:
https://github.com/RiskChallenger/mikro-orm/commit/d356e96b3eb2582b08020051d16f05b35aa758a4 the only change compared to GH 3614 is on line 92 `const newOwner = orm.em.create(User, { name: 'Johnny', project1: 1 });` where I added `project1: 1`

B4nan:
When you do this, you are setting the `project1` value on `newOwner`, which does not know anything about `oldOwner`.

```ts
newOwner.project1 = Reference.create(project);
```

When you do this, you literally replace old owner with new one, so it triggers the cleanup of the old owner.

```ts
project.owner = Reference.create(newOwner);
```

When you do this, the second line is a no-op, the relation is already set up, so the cleanup of `oldOwner` is not triggered.

```ts
const newOwner = orm.em.create(User, { name: 'Johnny', project1: project.id });
project.owner = Reference.create(newOwner);
```

B4nan:
I don't feel like the fix I came up with is the correct solution, but it seems to be working and I am tired enough to just do it conditionally :D 

[389b4a2](https://github.com/mikro-orm/mikro-orm/commit/389b4a2a750a45ecb9806640613de016132d2dfa)

Langstra:
Thanks for fixing. Would you like me to create an issue so it won't be forgotten and when you feel like it (or someone else) can fix it properly/differently?

Langstra:
Hmm... sorry to bother you again on this one. Trying to switch the relation twice will fail. https://github.com/RiskChallenger/mikro-orm/blob/4fe88b130b384a49a81f158a53c06ece261b6c23/tests/issues/GH3614.test.ts#L106 Added a test case here on line 106.

B4nan:
You are mixing contexts in there. You load the entity on a fresh fork, but call em.create on the global EM.

Langstra:
I do not think I understand it. If I remove (comment) the entire part about the fresh fork the error still exists.

B4nan:
You will need to wait, i wont be available till next week, so cant event test it now.

Langstra:
Sorry to notify you, you get enough questions already and this one is in a closed PR. Would you like to have another look here or rather have me create a separate issue/PR with test case? Many thanks.

B4nan:
Let's create an issue for that, it looks valid. Unlike my comment about the mixed contexts - the fork is used only to verify the removal of the entity, so that should be irrelevant.

B4nan:
Ok, this should do it

PR #3686, opened 2022-11-04
test: add test showing that relation is unset in uow which was not so before v5.5.0

Langstra:


B4nan:
This is a won't fix. Changes to relations are now propagated when you make them, not after flushing.

Langstra:
Is there another way to get the old relation in any way?

Langstra:
Ah missed your comment on the issue. Thanks for the quick response!

PR #3698, opened 2022-11-05
fix(knex): do not require virtual properties when seeding

mlsm-trl:
When trying to seed database entity that has virtual properties, they are required and `InvalidFieldNameException` is thrown.

Here is the reproduction and contextualization of the bug:
[mlsm-trl/mikro-orm-bug-virtual-properties-required-when-seeding](https://github.com/mlsm-trl/mikro-orm-bug-virtual-properties-required-when-seeding)

It appears to ignore `persist` property on verification, thus requiring virtual property if it met condition defined `primary || defaultRaw`.

This PR is a possible fix after a quick look at code. I only tested on my use case, so I don't know if this bug happens on other cases, like updates, neither if this verification should be replicated to other parts.

### DIFF
--- a/packages/knex/src/AbstractSqlDriver.ts
+++ b/packages/knex/src/AbstractSqlDriver.ts
@@ -356,7 +356,7 @@ export abstract class AbstractSqlDriver<C extends AbstractSqlConnection = Abstra
 
     if (this.platform.usesReturningStatement()) {
       /* istanbul ignore next */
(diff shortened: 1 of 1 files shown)


B4nan:
This should be actually already fixed by using the `hydrateProps`, can you reproduce with v5.5.1?

B4nan:
Hmm, I fixed it only for the getters, not for regular properties. Please adjust the test as in https://github.com/mikro-orm/mikro-orm/commit/35d51fecff9479f3d704bfcffa90d4fd5dcaf21a

Issue #3741, opened 2022-11-15
Commit .yarn folder


parisholley:
The old 1.x yarn clients can auto-switch to v3 if `yarnPath: .yarn/releases/yarn-3.x.x.cjs` exists and is reference in .yarnrc.yml like so: 

```yarnPath: .yarn/releases/yarn-x.x.x.cjs```

B4nan:
I don't want that. Use can use corepack instead of commiting the binary to every repository.

B4nan:
(at least not now, lets see how things will work, but I really dont like the idea of a huge versioned binary, for git repository it's almost like a memory leak)

PR #3742, opened 2022-11-15
fix(#3741): add yarn binary to repo

parisholley:
can close if you prefer user use corepack (maybe just update contributing docs with expected toolset)


B4nan:
The contributing guide it not updated, that's a problem for sure.

Are you using this approach yourself, on some long-lived repository? What I am afraid of is that if I update yarn 10 times, I will end up with 10 MBs of additional yarn binaries in the `.git` folder. Had the exact same issue with growing footprint of the docs, which is now resolved by auto-squashing the gh-pages branch.

parisholley:
i don't have a strong feeling on this because i've never used corepack day-to-day. i use `nvm` and thus keep a `.nvmrc` in all my projects but a lot of developers still use global installs for `node`/`yarn`, so I may just similarly be out the loop. with regards to yarn, outside of major version updates (eg: 3 <> 4), i don't see a reason to keep upgrading yarn unless there is a build related issue forcing you to do so.

B4nan:
Take a look at [`fnm`](https://github.com/Schniz/fnm) if you care about performance. `nvm` was unusable for me :]

True about the upgrading only major versions, maybe I was thinking about this the wrong way. But I guess I will wait for yarn 4 before adding it to the repo, should be behind the corner afaik.

parisholley:
@B4nan i didn't dig further, but an issue i just ran into was with the commit hooks. I can't prefix `git commit` with corepack and since my global yarn isn't on 3, i can't get around it. given that most projects aren't running yarn 3, i'm hesitant to make that change. unless you know a better way?

```
 git commit -a -m "feat(#3743): replace direct require calls with util to help with inlining"
yarn run v1.22.19
error @mikro-orm/root@5.0.0: The engine "yarn" is incompatible with this module. Expected version ">=3.2.0". Got "1.22.19"
```

> opening terminal lags

nice, i'll take a look at fnm

PR #3746, opened 2022-11-15
fix(core): allow quering against parent abstract classes

### LINKED ISSUE #3745: Unable to run queries against multi-level abstract classes
**Describe the bug**
Given the following model:

```typescript
abstract class BaseMikro {
  @PrimaryKey({ type: 'bigint' })
  id!: string;
}

@Entity({ discriminatorColumn: 'type', abstract: true })
abstract class LevelOne extends BaseMikro {
  @Property({ type: 'int' })
  type!: number;
}

@Entity({ abstract: true })
abstract class LevelTwo extends LevelOne {
  @PrimaryKey({ type: 'bigint' })
  id!: string;

  @Property()
  name!: string;
}

@Entity({ discriminatorValue: 1 })
export class LevelThree extends LevelTwo {
  @PrimaryKey({ type: 'bigint' })
  id!: string;

  @Property()
  name!: string;
}
```

Doing a find like so

```typescript
orm.em.find(LevelTwo, {});
```

**Stack trace**
```
    MetadataError: Metadata for entity LevelTwo not found

      211 |
      212 |   static missingMetadata(entity: string): MetadataError {
    > 213 |     return new MetadataError(`Metadata for entity ${entity} not found`);
          |            ^
      214 |   }
      215 |
      216 |   static conflictingPropertyName(className: string, name: string, embeddedName: string): MetadataError {

      at Function.missingMetadata (packages/core/src/errors.ts:213:12)
      at MetadataStorage.get (packages/core/src/metadata/MetadataStorage.ts:78:27)
      at SqlEntityManager.tryFlush (packages/core/src/EntityManager.ts:991:30)
      at SqlEntityManager.find (packages/core/src/EntityManager.ts:115:14)
      at Object.<anonymous> (tests/issues/GH3743.test.ts:49:16)
```

parisholley:
Closes #3745

### DIFF
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -58,7 +58,7 @@ export class MetadataDiscovery {
     }
 
     // ignore base entities (not annotated with @Entity)
-    const filtered = discovered.filter(meta => meta.name);
+    const filtered = discovered.filter(meta => meta.name || meta.abstract);
     // sort so we discover entities first to get around issues with nested embeddables
     filtered.sort((a, b) => !a.embeddable === !b.embeddable ? 0 : (a.embeddable ? 1 : -1));
     filtered.forEach(meta => this.initSingleTableInheritance(meta, filtered));
--- a/tests/issues/GH3745.test.ts
+++ b/tests/issues/GH3745.test.ts
@@ -1,0 +1,59 @@
+import { Entity, MikroORM, PrimaryKey, Property, t } from '@mikro-orm/core';
+
+abstract class BaseMikro {
+
+  @PrimaryKey({ type: 'bigint' })
+  id!: string;
+
+}
+
+@Entity({ discriminatorColumn: 'type', abstract: true })
+abstract class LevelOne extends BaseMikro {
+
+  @Property({ type: 'int' })
+  type!: number;
+
+}
+
+@Entity({ abstract: true })
+abstract class LevelTwo extends LevelOne {
+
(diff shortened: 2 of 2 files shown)


packages/core/src/metadata/MetadataDiscovery.ts
```diff
@@ -58,7 +58,7 @@ export class MetadataDiscovery {
     }
 
     // ignore base entities (not annotated with @Entity)
-    const filtered = discovered.filter(meta => meta.name);
+    const filtered = discovered.filter(meta => meta.name || meta.abstract);
```
B4nan, line 61: the point of this line was to do the exact opposite. abstract === no name. now with your change it basically does `[...].filter(a => true || false)`

B4nan (changes requested):
i dont think this is the correct fix.

you need to distinguish a regular parent class (which will be inlined, not representing any database table) and parent class in STI (which is a valid entity on runtime, represents specific database table). your fix is affecting both, we should allow this only for the parent classes that are representing a database table.

maybe check if the entity has a `meta.tableName` or `meta.discriminatorColumn` would do it (as this is really just about the STI)

parisholley:
my issue is that in my modal example as pointed out in the issue, the intermediary class wouldn't have a table name or discriminator definition unless it was duplicated (which wouldn't be an intuitive solution to someone new to project). 

 > the point of this line was to do the exact opposite. abstract === no name.

perhaps instead of blindly looking for abstract, we do abstract + (any parent with name)?

B4nan:
resolved via 0c56118603c25a471a0c636912b976a4291290b3

PR #3756, opened 2022-11-15
feat(core): return loaded type when populating a reference

### LINKED ISSUE #3755: Reference.load() should return Loaded<> when populating
When calling `reference.load()`, the return type is `Promise<T>` instead of `Promise<Loaded<T,P>>`

Add more overloads to the load type

parisholley:
Closes #3755

### DIFF
--- a/packages/core/src/entity/Reference.ts
+++ b/packages/core/src/entity/Reference.ts
@@ -5,7 +5,7 @@ import type {
   Constructor,
   Dictionary,
   EntityProperty,
-  IsUnknown,
+  IsUnknown, Loaded,
   Populate,
   Primary,
   PrimaryProperty,
@@ -89,7 +89,13 @@ export class Reference<T> {
    * Ensures the underlying entity is loaded first (without reloading it if it already is loaded).
    * Returns the entity.
    */
-  async load<K extends keyof T = never, P extends string = never>(options?: LoadReferenceOptions<T, P>): Promise<T>;
+  async load<K extends keyof T = never, P extends string = never>(options?: LoadReferenceOptions): Promise<T>;
+
+  /**
+   * Ensures the underlying entity is loaded first (without reloading it if it already is loaded).
+   * Returns the entity.
+   */
+  async load<K extends keyof T = never, P extends string = never>(options: LoadAndPopulateReferenceOptions<T, P>): Promise<Loaded<T, P>>;
 
   /**
    * Ensures the underlying entity is loaded first (without reloading it if it already is loaded).
@@ -101,7 +107,7 @@ export class Reference<T> {
    * Ensures the underlying entity is loaded first (without reloading it if it already is loaded).
    * Returns either the whole entity, or the requested property.
    */
-  async load<K extends keyof T = never, P extends string = never>(options?: LoadReferenceOptions<T, P> | K): Promise<T | T[K]> {
+  async load<K extends keyof T = never, P extends string = never>(options?: LoadAndPopulateReferenceOptions<T, P> | K): Promise<T | T[K]> {
     const opts: Dictionary = typeof options === 'object' ? options : { prop: options };
 
     if (!this.isInitialized()) {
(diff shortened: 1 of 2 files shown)


tests/issues/GH3755.test.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+class Child {
+
+  @ManyToOne(() => Parent, { inversedBy: 'children' })
+  parent!: IdentifiedReference<Parent>;
+
+}
+
+async function load() {
```
B4nan, line 17: this is not really a test case. if its supposed to be testing only the type safety, lets add it here instead:

https://github.com/mikro-orm/mikro-orm/blob/master/tests/types.test.ts

packages/core/src/entity/Reference.ts
```diff
@@ -5,7 +5,7 @@ import type {
   Constructor,
   Dictionary,
   EntityProperty,
-  IsUnknown,
+  IsUnknown, Loaded,
```
B4nan, line 8:
```suggestion
  IsUnknown,
  Loaded,
```

packages/core/src/entity/Reference.ts
```diff
@@ -173,8 +179,11 @@ Object.defineProperties(Reference.prototype, {
...
 });
 
-export interface LoadReferenceOptions<T, P extends string = never> {
-  populate?: Populate<T, P>;
+export interface LoadReferenceOptions {
   lockMode?: Exclude<LockMode, LockMode.OPTIMISTIC>;
   connectionType?: ConnectionType;
 }
+
+export interface LoadAndPopulateReferenceOptions<T, P extends string = never> extends LoadReferenceOptions{
```
B4nan, line 187: do we really need a separate signature/type/etc? why not just always return the `Loaded` type? if no populate hint, it will end up with `Loaded<T, never>`, just like the `em.find/...` methods

parisholley: that is fine, i did that at first but wasn't sure if that would cause issues always returning it

B4nan:
Hmm looking into this now, and I see the change triggers some TS cycle error. It appears to be working fine otherwise.

```
Run yarn tsc-check-tests
Error: tests/features/embeddables/GH3690.test.ts(67,14): error TS2615: Type of property 'bp' circularly references itself in mapped type '{ id: number; bp: ({ id?: number | undefined; } & Reference<BPE> & LoadedReference<Loaded<BPE | null, string>>) | null; }'.
Error: tests/features/embeddables/GH3690.test.ts(67,14): error TS2615: Type of property 'p' circularly references itself in mapped type '{ [K in keyof BPE]: K extends any ? LoadedLoadable<BPE[K], Loaded<ExtractType<BPE[K]>, string>> : BPE[K]; }'.
Error: tests/issues/GH269.test.ts(62,16): error TS2615: Type of property 'a' circularly references itself in mapped type '{ id: number; a?: ({ id: number; } & Reference<A> & LoadedReference<Loaded<A | undefined, string>>) | undefined; name: string; }'.
Error: tests/issues/GH269.test.ts(62,16): error TS2615: Type of property 'b' circularly references itself in mapped type '{ [K in keyof A]: K extends any ? LoadedLoadable<A[K], Loaded<ExtractType<A[K]>, string>> : A[K]; }'.
Error: tests/issues/GH535.test.ts(65,22): error TS2615: Type of property 'a' circularly references itself in mapped type '{ [K in keyof B]: K extends any ? LoadedLoadable<B[K], Loaded<ExtractType<B[K]>, string>> : B[K]; }'.
Error: tests/issues/GH535.test.ts(65,22): error TS2615: Type of property 'b' circularly references itself in mapped type '{ id: number; b: { id: number; } & Reference<B> & LoadedReference<Loaded<B, string>>; readonly calcProp: string; }'.
```

Spent a few hours debugging this but have no idea what's happening, probably a TS bug: https://github.com/microsoft/TypeScript/issues/37597

So we are kinda blocked by that.

PR #3762, opened 2022-11-16
docs: note about usage with NestJS for lifecycle subscriber

SDemonUA:
Added a note since other variants didn't work in NestJS
Ref: https://github.com/mikro-orm/nestjs/issues/14

### DIFF
--- a/docs/versioned_docs/version-4.5/lifecycle-hooks.md
+++ b/docs/versioned_docs/version-4.5/lifecycle-hooks.md
@@ -92,34 +92,6 @@ export class AuthorSubscriber implements EventSubscriber<Author> {
 }
 ```
 
-Note: for NestJS you should use `@Injectable` and register subscriber manually instead of `@Subscriber` decorator.
-
-```typescript
-import { Injectable } from '@nestjs/common';
-import { EntityName, EventArgs, EventSubscriber } from '@mikro-orm/core';
-
-@Injectable()
-export class AuthorSubscriber implements EventSubscriber<Author> {
-
-  constructor(em: EntityManager) {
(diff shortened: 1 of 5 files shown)


docs/versioned_docs/version-4.5/lifecycle-hooks.md
```diff
@@ -92,6 +92,34 @@ export class AuthorSubscriber implements EventSubscriber<Author> {
 }
 ```
 
+Note: for NestJS you should use `@Injectable` and register subscriber manually instead of `@Subscriber` decorator.
```
B4nan, line 95: this should go in here: https://mikro-orm.io/docs/next/usage-with-nestjs

also note you updated v4 docs, you should update the latest v5 docs snapshot + the next docs instead:

https://github.com/mikro-orm/mikro-orm/blob/master/docs/docs/usage-with-nestjs.md
https://github.com/mikro-orm/mikro-orm/blob/master/docs/versioned_docs/version-5.5/usage-with-nestjs.md

SDemonUA: Updated til 4.5 version - as it version I use right now

SDemonUA:
@B4nan hey, I found that this whole thing with `constructor` is needed only if there are more than 1 subscriber classes with the same name. In other cases it's enough to just 
```typescript
// --- module.ts ---

@Module({
  providers: [SomeSubscriber]
})
export class SomeModule { /* ... */ }

// --- subscriber.ts ---
@Subscriber()
export class SomeSubscriber implements EventSubscriber { /* ... */ }
```

Issue #3749, opened 2022-11-15
docs: How to format code?


parisholley:
Is there a prettier process in the project so I can avoid unrelated diffs due to IDE formatting preferences?

PR #3765, opened 2022-11-17
Format with Prettier

parisholley:
an example for #3749 

seems most formatters nowadays don't offer must customization but setting a large line width reduces the amount of new lines that happen


parisholley:
@B4nan was hoping you could highlight in the changes, what elements of prettier you are unhappy with and we can explore if plugins or other tweaks could solve that.

B4nan:
just look at the typings.ts file, it removed every single line break in type aliases.

B4nan:
Or here, it keeps line breaks but removes indents. better, but still, i dont want uglier and less readable code. 

image

Here is [a recent article](https://antfu.me/posts/why-not-prettier), I am quite aligned with what it says. 

So if you want to contribute here, just disable prettier on save, that's all 🤷 Coding style is defined in eslint, but remember, we started this discussion because you used prettier on a project that does not expect it - this is not about how to format code, this is about not using tools that the project does not expect.

parisholley:
it is a fair counter, to each their own. i find the changes to function args, object expansion, etc a welcomed advantage to readability in the face of some the mangling of types, but I'd argue that typescript types are inherently hard to grok anyway, like trying to follow the path of how Loaded<> works for the first time 😄. in your example, you can't understand what the type really is unless you click into all the extra types it references.

and everyone's diff flow is also going to be different, leading to a difference in opinion. i use Kaleidoscope locally on mac which has a `ignore whitespace` option (and i think a lot of difftools do as well), I usually only rely on github PRs for a quick review, but i can see that not being as practical for open source.

all in all, I can just tweak formatting in intellij to customize it to this project and its done, but gotta at least try to advocate for something first :) at least in future there is a thread here to reference when the next person brings it up

PR #3766, opened 2022-11-17
Format with Rome

parisholley:
an example for https://github.com/mikro-orm/mikro-orm/issues/3749

seems most formatters nowadays don't offer must customization but setting a large line width reduces the amount of new lines that happen


B4nan:
I tried both recently, as I said, I like the idea, but the current implementations are too weak for me, rome seems even worse than prettier actually.

PR #3771, opened 2022-11-17
test: show error on inserting an inline embeddable as null

Langstra:


Langstra:
Fixed with https://github.com/mikro-orm/mikro-orm/commit/6e0bedf5610b6e423605d250ed2ac1a15a63381e

PR #3786, opened 2022-11-23
test: serialization of json during upsert

boredland:
we discovered that during upsert a json property is being serialized into a string. this test demonstrates that behaviour.


B4nan:
Fixed via 434d417b6a38a4cd8293882061067cf6d89ba003

PR #3805, opened 2022-11-30
feat(postgres): add support for weighted tsvectors and a custom regconfig

jsprw:
Adds support for a custom regconfig and weighted tsvectors as requested by https://github.com/mikro-orm/mikro-orm/pull/3317#issuecomment-1330035868. 

For more context, see https://github.com/mikro-orm/mikro-orm/pull/3317#issuecomment-1330555809.

Docs: also (hopefully) fixed the tabs which caused the confusion as seen here: https://github.com/mikro-orm/mikro-orm/pull/3317#issuecomment-1279882693.

### DIFF
--- a/packages/postgresql/src/PostgreSqlPlatform.ts
+++ b/packages/postgresql/src/PostgreSqlPlatform.ts
@@ -1,9 +1,10 @@
 import { Client } from 'pg';
-import type { EntityProperty, Type, SimpleColumnMeta, Dictionary } from '@mikro-orm/core';
-import { ALIAS_REPLACEMENT, JsonProperty, raw, Utils } from '@mikro-orm/core';
+import type { EntityProperty, SimpleColumnMeta, Dictionary } from '@mikro-orm/core';
+import { ALIAS_REPLACEMENT, JsonProperty, raw, Utils, Type } from '@mikro-orm/core';
 import { AbstractSqlPlatform } from '@mikro-orm/knex';
 import { PostgreSqlSchemaHelper } from './PostgreSqlSchemaHelper';
 import { PostgreSqlExceptionConverter } from './PostgreSqlExceptionConverter';
+import { FullTextType } from './types/FullTextType';
 
 export class PostgreSqlPlatform extends AbstractSqlPlatform {
 
@@ -78,6 +79,10 @@ export class PostgreSqlPlatform extends AbstractSqlPlatform {
   }
 
   override getFullTextWhereClause(prop: EntityProperty): string {
+    if (prop.customType instanceof FullTextType) {
+      return `:column: @@ plainto_tsquery('${prop.customType.regconfig}', :query)`;
+    }
+
     if (prop.columnTypes[0] === 'tsvector') {
       return `:column: @@ plainto_tsquery('simple', :query)`;
     }
@@ -100,6 +105,13 @@ export class PostgreSqlPlatform extends AbstractSqlPlatform {
     }
 
     return `create index ${quotedIndexName} on ${quotedTableName} using gin(to_tsvector('simple', ${quotedColumnNames.join(` || ' ' || `)}))`;
+  }
+
+  override getMappedType(type: string): Type<unknown> {
+    switch (this.extractSimpleType(type)) {
+      case 'tsvector': return Type.getType(FullTextType);
+      default: return super.getMappedType(type);
+    }
   }
 
   override getRegExpOperator(val?: unknown, flags?: string): string {
--- a/packages/postgresql/src/types/FullTextType.ts
+++ b/packages/postgresql/src/types/FullTextType.ts
@@ -1,6 +1,22 @@
-import { Type } from '@mikro-orm/core';
+import { raw, Type } from '@mikro-orm/core';
+import type { TransformContext } from '@mikro-orm/core';
+import type { PostgreSqlPlatform } from '../PostgreSqlPlatform';
+import type { Knex } from '@mikro-orm/knex';
 
-export class FullTextType extends Type<string, string> {
+// Postgres has four levels of full text weights
+// https://www.postgresql.org/docs/current/textsearch-controls.html
+type FullTextWeight = 'A' | 'B' | 'C' | 'D';
+
+// Null types are allowed, mainly for readability so the user does not have to do null checks
+// themselves and keep using `onUpdate: (book: Book) => { A: book.title, B: book.description }`
+// with nullable properties.
+export type WeightedFullTextValue = { [K in FullTextWeight]?: string | null };
+
+export class FullTextType extends Type<string | WeightedFullTextValue, string | null | Knex.Raw> {
+
+  constructor(public regconfig: string = 'simple') {
+    super();
+  }
 
   override compareAsType(): string {
     return 'string';
@@ -10,8 +26,57 @@ export class FullTextType extends Type<string, string> {
     return 'tsvector';
   }
 
-  override convertToDatabaseValueSQL(key: string) {
-    return `to_tsvector('simple', ${key})`;
+  // Use convertToDatabaseValue to prepare insert queries as this method has
+  // access to the raw JS value. Return Knex#raw to prevent QueryBuilderHelper#mapData
+  // from sanitizing the returned chaing of SQL functions.
+  override convertToDatabaseValue(value: string | WeightedFullTextValue, platform: PostgreSqlPlatform, context?: TransformContext | boolean): string | null | Knex.Raw {
+    // Don't convert to values from select queries to the to_tsvector notation
+    // these should be compared as string using a special oparator or function
+    // this behaviour is defined in Platform#getFullTextWhereClause.
+    // This is always a string.
+    if (typeof context === 'object' && context.fromQuery) {
+      return value as string;
+    }
+
+    // Null values should not be processed
+    if (!value) {
+      return null;
+    }
+
+    // the object from that looks like { A: 'test data', B: 'test data2' ... }
+    // must be converted to
+    // setweight(to_tsvector(regconfig, value), A) || setweight(to_tsvector(regconfig, value), B)... etc
+    // use Knex#raw to do binding of the values sanitization of the boundvalues
+    // as we return a raw string which should not be sanitzed anymore
+    if (typeof value === 'object') {
+      const bindings: string[] = [];
+      const sqlParts: string[] = [];
+
+      for (const [weight, data] of Object.entries(value)) {
+        // Check whether the weight is valid according to Postgres,
+        // Postgres allows the weight to be upper and lowercase.
+        if (!['A', 'B', 'C', 'D'].includes(weight.toUpperCase())) {
+          throw new Error('Weight should be one of A, B, C, D.');
+        }
+
+        // Ignore all values that are not a string
+        if (typeof data === 'string') {
+          sqlParts.push('setweight(to_tsvector(?, ?), ?)');
+          bindings.push(this.regconfig, data, weight);
+        }
+      }
+
+      // Return null if the object has no valid strings
+      if (sqlParts.length === 0) {
+        return null;
+      }
+
+      // Join all the `setweight` parts using the PostgreSQL tsvector `||` concatenation operator
+      return raw(sqlParts.join(' || '), bindings);
+    }
+
+    // if it's not an object, it is expected to be string which does not have to be wrapped in setweight.
+    return raw('to_tsvector(?, ?)', [this.regconfig, value]);
   }
 
 }
--- a/packages/postgresql/src/types/FullTextType.ts
+++ b/packages/postgresql/src/types/FullTextType.ts
@@ -1,7 +1,6 @@
 import { raw, Type } from '@mikro-orm/core';
-import type { TransformContext } from '@mikro-orm/core';
+import type { TransformContext, RawQueryFragment } from '@mikro-orm/core';
 import type { PostgreSqlPlatform } from '../PostgreSqlPlatform';
-import type { Knex } from '@mikro-orm/knex';
 
 // Postgres has four levels of full text weights
 // https://www.postgresql.org/docs/current/textsearch-controls.html
@@ -12,7 +11,7 @@ type FullTextWeight = 'A' | 'B' | 'C' | 'D';
 // with nullable properties.
 export type WeightedFullTextValue = { [K in FullTextWeight]?: string | null };
 
-export class FullTextType extends Type<string | WeightedFullTextValue, string | null | Knex.Raw> {
+export class FullTextType extends Type<string | WeightedFullTextValue, string | null | RawQueryFragment> {
 
   constructor(public regconfig: string = 'simple') {
     super();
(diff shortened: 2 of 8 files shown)


packages/postgresql/src/types/FullTextType.ts
```diff
@@ -10,8 +27,38 @@ export class FullTextType extends Type<string, string> {
...
+    if (typeof context === 'object' && context.fromQuery) {
+      return value as string;
+    }
+
+    if (typeof value === 'object') {
+      const bindings: string[] = [];
+      const sqlParts: string[] = [];
+
+      for (const [weight, data] of Object.entries(value)) {
+        if (Utils.valueInObject(FullTextWeight, weight) && typeof data !== 'undefined') {
```
B4nan, line 43: feels like this can be also simplified if you get rid of the enum (i didn't like the Utils.valueInObject method even before getting to this place... :])

packages/postgresql/src/PostgreSqlPlatform.ts
```diff
@@ -74,8 +75,8 @@ export class PostgreSqlPlatform extends AbstractSqlPlatform {
   }
 
   getFullTextWhereClause(prop: EntityProperty): string {
-    if (prop.columnTypes[0] === 'tsvector') {
-      return `:column: @@ plainto_tsquery('simple', :query)`;
+    if (prop.customType instanceof FullTextType) {
```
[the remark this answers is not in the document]
jsprw: I added the tsvector type to the getMappedType function, see https://github.com/mikro-orm/mikro-orm/blob/fe5e7861570b2fe06ad18e2ceec5635577b4ed45/packages/postgresql/src/PostgreSqlPlatform.ts#L101. After this change, the removed if-branch was not called anymore from any test, so I assumed that this mapper worked properly.

packages/postgresql/src/types/FullTextType.ts
jsprw: Alright, I will remove it. Just added a check whether the key is in A, B, C or D. We could remove the check altogether? But a wrong key causes a incorrect query to be generated..

packages/postgresql/src/PostgreSqlPlatform.ts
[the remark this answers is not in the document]
B4nan: > After this change, the removed if-branch was not called anymore from any test, so I assumed that this mapper worked properly.

yes, but we dont have any `@Property({ type: 'tsvector' })` or `@Property({ columnType: 'tsvector' })` in the tests, right? please verify it works with those too

packages/postgresql/src/types/FullTextType.ts
```diff
@@ -10,8 +27,38 @@ export class FullTextType extends Type<string, string> {
...
+      }
+
+      return this.createRawValue(sqlParts.join(' || '), bindings, platform) as unknown as string;
+    }
+
+    return this.createRawValue('to_tsvector(?, ?)', [this.regconfig, value], platform) as unknown as string;
+  }
+
+  private createRawValue(sql: string, bindings: string[], platform: PostgreSqlPlatform) {
+    const knex = (platform.getConfig().getDriver() as PostgreSqlDriver).getConnection().getKnex();
```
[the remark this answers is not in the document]
jsprw: Yeah it's really ugly. However I need the raw function to process an object. When I use the method  convertToDatabaseValueSQL, the value is always string which doesnt work with the weighted object. 

When I use convertToDatabaseValue, the result is always quoted, unless it has the property __raw. But before I could make it raw, I had to get access to Knex.raw. And because I surpassed the quote, i had to sanitize the values which can be done in Knex.raw. If you have a better solution for

In your pointtype, the `convertToDatabaseValue` produces a result which could be passed as arguments to a SQL function (it produces the string 'point(y,x)' which is passed to 'st_fromtext(string)').

However in this case, it is not possible to pass arguments to a function as it should be conditionally wrapped with setweight() and repeated some times.

With no weights, the result is to_tsvector(regconfig, value). With weights it's setweight(to_tsvecor(regconfig, value), A) || setweight(to_tsvecor(regconfig, value), B) || etc...

See: https://github.com/mikro-orm/mikro-orm/blob/master/packages/knex/src/query/QueryBuilderHelper.ts#L774-780

The raw function is https://github.com/mikro-orm/mikro-orm/blob/master/packages/knex/src/query/QueryBuilder.ts#L392.

packages/postgresql/src/types/FullTextType.ts
B4nan: i dont mind a runtime validation, i do mind added complexity and perf overhead :] in this case even if we keep the enum, you dont need any method to check its values:

https://www.typescriptlang.org/play?#code/KYOwrgtgBAYmA28AqwAeAXA6sAlgcwAt0oBvAKCigEEoBeKAcioYBoKoAhOxj19gYW4N+fSgBEhYvgF8yZAMYB7EAGdF8YADp4ivAAomDKDhCwEyNFlyF0ASgDcC5Wo3bdBgKJGTZxCgzY+EQOQA

and then there is the thing that enums in TS require enum usage, which is often causing troubles more than solving anything, e.g. i can see people want to put there `A/B/C/D` string literal, instead of prefixing it with some enum name.

> But a wrong key causes a incorrect query to be generated..

wrong user input causing wrong query sounds like correct behavior. or is there a valid/expected fallback for different letters than ABCD?

packages/postgresql/src/PostgreSqlPlatform.ts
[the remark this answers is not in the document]
jsprw: You're right, re-added it and added a quick test for it. It indeed does not map the type correctly. Do you have a clue why?

packages/postgresql/src/types/FullTextType.ts
jsprw: I don't think anything should be expected, however when the advised/most used usage is not type checked, I think it would be nice that the ORM gives a clear message as the postgres message is quite cryptic. What do you think?

select setweight(to_tsvector('simple', 'test test'), 'K'); -> ERROR:  unrecognized weight: 75
select setweight(to_tsvector('simple', 'test test'), 'abc'); -> 'test':1A,2A (silently takes the first letter)

packages/postgresql/src/types/FullTextType.ts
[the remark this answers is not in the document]
jsprw: > As I said, this is rather hacking the ORM.

Agreed, especially with all the getter-calls. However it is required to work around the limitations of the custom type. 

> I am sure we can find a way to do it without knex.raw, e.g. we can add more context to the conversion methods.

I think we should make sanitizers available in the type and that there is a way to construct raw sql function chains. Eg add a function like `convertToDatabaseValueUnsafe`.

> Maybe the issue here is that you are trying to fit too many things into a single type? As it sounds like the shape of the value can be quite different.
> the value is already escaped SQL literal. The result needs to be SQL fragment, and that is a string too.

Splitting the type in two different types, eg WeightedFullTextType and the normal FullTextType, does help readability, but most of this monstrosity will then be copied to the `WeightedFullTextType` as we will still come across this limitation: setweight must be added 1-4 times.

WeightedFullTextType#convertToDatabaseValueSQL should return (this is the notation from the postgres documentation):

`setweight(to_tsvecor(regconfig, value), A)`

or

`setweight(to_tsvecor(regconfig, value), A) || setweight(to_tsvecor(regconfig, value), B)`

or

`setweight(to_tsvecor(regconfig, value), A) || setweight(to_tsvecor(regconfig, value), B) || setweight(to_tsvecor(regconfig, value), C)`

or

`setweight(to_tsvecor(regconfig, value), A) || setweight(to_tsvecor(regconfig, value), B) || setweight(to_tsvecor(regconfig, value), C) || setweight(to_tsvecor(regconfig, value), D)`

When using fragments, it is currently not possible to do how it's designed. The only possible thing is parsing (unquoting?) the JSON value which feels hacky too.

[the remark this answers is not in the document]
jsprw: I have added comments to the latest comit.

packages/postgresql/src/types/FullTextType.ts
```diff
@@ -26,29 +26,55 @@ export class FullTextType extends Type<string | WeightedFullTextValue, string> {
...
+    // don't convert to values from select queries to the to_tsvector notation
+    // these should be compared as string using a special oparator or function
+    // this behaviour is defined in Platform#getFullTextWhereClause.
     if (typeof context === 'object' && context.fromQuery) {
       return value as string;
     }
 
+    // null values should not be processed
+    if (!value) {
+      return null as unknown as string;
```
B4nan, line 42: feels like a lie in the method return type?

also i believe you can do just `null!`. ugly as well, but not so much :D

jsprw: Hehe, yeah. But else the override wouldn't match the function definition the parent class :/. Do I need to replace all casts with `null!`?

B4nan: i dont follow, what parent class? most of the ORM-provided types are explicitly allowing both null and undefined, this type should do it too, or not?

jsprw: My bad, didn't look well enough. I looked at the return type of convertToDatabaseValueSQL. Fixed it via the generic parameter.

packages/postgresql/src/types/FullTextType.ts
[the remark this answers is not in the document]
PodaruDragos: > i am still thinking about getting rid of knex

O yes !! 👍

B4nan:
What's the state in here, I kinda lost track.

jsprw:
> What's the state in here, I kinda lost track.

Excuse me, I missed your comment. The state is: you found the modifications a bit hacky, mainly due to the limited ability to inject SQL from a custom type. This comment describes the situation pretty well: https://github.com/mikro-orm/mikro-orm/pull/3805#discussion_r1035942764.

B4nan:
So apart from me not liking the implementation much, do you think it's ready/complete? 

Since it's a postgres specific thing, I guess we don't have to be so strict. I'd also maybe target the v6 branch with this instead. I'd rather merge this than make it fade away :]

jsprw:
Yeah, that's is basically it. I could rewrite it once there is a possibility of returning a full sql string from a custom type.

Should I target it to the v6 branch?

B4nan:
Yeah, let's move it there please.

>  I could rewrite it once there is a possibility of returning a full sql string from a custom type.

I'll think about how to do this in v6, we should have some native way for raw queries too so we don't have to be so tightly coupled to knex.

jsprw:
Alright 👍 . Any ETA for v6 (as my time is limited this week)? 

I would also like to make a PR for a fulltext query builder. Something like mentioned here: https://github.com/mikro-orm/mikro-orm/pull/3317#issuecomment-1190170191.

`{ title: { $fulltext: { $include: ['keyword1', 'keyword2'], $exclude: ['keyword3'] } } }` or even `{ title: { $fulltext: { $or: [{ $include: ['keyword1', 'keyword2'], $exclude: ['keyword3'] }, { $include: ['keyword4'], $exclude: ['keyword1'] ] } } }`

B4nan:
No ETA, it will take a few more months for sure, hopefully somewhere during spring. But the builds are already published to NPM on each commit.

jsprw:
Oh, perfect 👍 . Will have a look at rebasing this PR in a few weeks.

B4nan:
> I could rewrite it once there is a possibility of returning a full sql string from a custom type.

FYI this will be added via #4094, it just needs some internal improvements.

jsprw:
> > I could rewrite it once there is a possibility of returning a full sql string from a custom type.
> 
> FYI this will be added via #4094, it just needs some internal improvements.

Good-looking feature! I see it already supports value-bindings (?). Does it also support identifier-bindings (??) like knex#raw (https://knexjs.org/guide/raw.html#raw-parameter-binding)?

B4nan:
I think it should, as all it does is using the `platform.formatQuery()` which does so:

https://github.com/mikro-orm/mikro-orm/blob/master/packages/knex/src/AbstractSqlPlatform.ts#L88-L91

jsprw:
I don't know what happened here, but i fixed it. I have rewritten this pull request by using the `raw` function from #4094 and updated the tests.

packages/postgresql/src/types/FullTextType.ts
```diff
@@ -1,6 +1,22 @@
-import { Type } from '@mikro-orm/core';
+import { raw, Type } from '@mikro-orm/core';
+import type { TransformContext } from '@mikro-orm/core';
+import type { PostgreSqlPlatform } from '../PostgreSqlPlatform';
+import type { Knex } from '@mikro-orm/knex';
```
B4nan, line 4: I guess this is a left over from some previous implementation? raw values are no longer represented by `Knex.Raw`, but rather with `RawQueryFragment` instances. Technically, there is still some `Knex.Raw` here and there, but I'd rather not mention that.

tests/features/fulltext/full-text-search-tsvector.postgres.test.ts
```diff
@@ -10,10 +13,25 @@ export class Book {
...
+  @Property({ type: 'tsvector', nullable: true, onUpdate: (book: Book) => book.title, onCreate: (book: Book) => book.title })
+  searchableTitleNoType!: string;
+
+  @Index({ type: 'fulltext' })
+  @Property({ type: new FullTextType('english'), nullable: true, onUpdate: (book: Book) => book.title, onCreate: (book: Book) => book.title })
+  searchableTitleEnglish!: string;
+
+  @Index({ type: 'fulltext' })
+  @Property({ type: FullTextType, nullable: true, onUpdate: createWeightedValue, onCreate: createWeightedValue })
+  searchableTitleWeighted!: WeightedFullTextValue | string;
```
B4nan, line 33: Why is it also typed as `| string`? I guess so you can do typesafe queries over it, e.g. https://github.com/mikro-orm/mikro-orm/pull/3805/files#diff-613b82f629b1d46002394ecdff328d65c0591ca4c4d0bba6ae096f691c4360e2R183?

If so, maybe we should adjust the example in the docs, or improve the types for `$fulltext` operator so it's not needed.

B4nan (approved):
Thanks!

tests/features/fulltext/full-text-search-tsvector.postgres.test.ts
jsprw: I typed it as `WeightedFullTextValue | string` because the `FullTextType` now supports both of those types. So for correctness we could type all `FullTextType` properties in Postgres `WeightedFullTextValue | string`, however the enduser will probably use one or the other for this specific query.

As for improving the `$fulltext` operator, would you be interested in me making it smarter for all platforms? Personally I would use it and would be a great addition, because it is quite dumb now (it will try to match all words fully). Postgres and other platforms support queries where you can partially match words or exclude certain words.

I was thinking of doing it the same way like your smart queries and support the following operators: `$any`, `$all`, `$order`, `$not`, `$partial`, `$partialEnd`, `$partialStart`, and also support weights if using `WeightedFullTextValue`. 

Example of a complex query: `{searchableTitle: {$fulltext: {$exclude: ['search3'], $all: ['search1', {$partial: 'search2'}]}}`

I will open a different merge request for this feature. Should I create a new class for this like `QueryBuilderHelper`? Or should I expand `Platform#getFullTextWhereClause`?

B4nan: Yeah, any kind of improvements for this are welcome, the feature is a community addition, I haven't used this myself so I don't know how usable it actually is and what parts might be missing.

jsprw: Alright, will do! But before I start, could you provide me a hint where you'd like it to be implemented?

> I will open a different merge request for this feature. Should I create a new class for this like QueryBuilderHelper? Or should I expand Platform#getFullTextWhereClause?

B4nan: `QueryBuilderHelper` is a helper for SQL drivers (for the QB itself, it's not used anywhere else), this should work with mongo too I guess? It depends, you could combine this with the platform method too, I really don't know at this point as I am not entirely sure what changes you want to make :]

If it's simple enough, adjusting the platform method would be probably better but I don't mind having a new helper class to encapsulate this, if you think it's gonna be more code than just few lines. It can as well be just a few private methods on the platform level. I'd say we can refactor things once we see how they look.

PR #3825, opened 2022-12-05
feat(core): add `em.upsertMany`

B4nan:
Like `em.upsert()` but with batch support.

```ts
const [author1, author2, author3] = await em.upsertMany(Author, [
  { email: 'a1', age: 41 },
  { email: 'a2', age: 42 },
  { email: 'a3', age: 43 },
]);
```


moioo91120:
Hello @B4nan  , it seems like a really good functionnality, is it possible to expose it on EntityRepository ? 
thanks :)

Issue #3614, opened 2022-10-19
Swapping OneToOne relations results in am `UNIQUE constraint failed` error


Langstra:
**Describe the bug**
When replacing an entity on a OneToOne relation the ORM first tries to insert the new entity. This results in an `UNIQUE constraint failed` error. You'd expect the ORM either to first set the foreign key to `NULL` or to delete the old entity.

**Stack trace**
```
UniqueConstraintViolationException: insert into `user` (`name`, `project_id`) values ('Johnny', 1) returning `id` - SQLITE_CONSTRAINT: UNIQUE constraint failed: user.project_id
        at SqliteExceptionConverter.convertException (/home/langstra/code/mikro-orm/packages/sqlite/src/SqliteExceptionConverter.ts:26:14)
        at SqliteDriver.convertException (/home/langstra/code/mikro-orm/packages/core/src/drivers/DatabaseDriver.ts:262:50)
        at /home/langstra/code/mikro-orm/packages/core/src/drivers/DatabaseDriver.ts:267:18
        at async SqliteDriver.nativeInsertMany (/home/langstra/code/mikro-orm/packages/knex/src/AbstractSqlDriver.ts:363:17)
        at async ChangeSetPersister.persistNewEntity (/home/langstra/code/mikro-orm/packages/core/src/unit-of-work/ChangeSetPersister.ts:107:17)
        at async ChangeSetPersister.executeInserts (/home/langstra/code/mikro-orm/packages/core/src/unit-of-work/ChangeSetPersister.ts:38:7)
        at async ChangeSetPersister.runForEachSchema (/home/langstra/code/mikro-orm/packages/core/src/unit-of-work/ChangeSetPersister.ts:86:7)
        at async UnitOfWork.commitCreateChangeSets (/home/langstra/code/mikro-orm/packages/core/src/unit-of-work/UnitOfWork.ts:872:5)
        at async UnitOfWork.persistToDatabase (/home/langstra/code/mikro-orm/packages/core/src/unit-of-work/UnitOfWork.ts:811:7)
        at async SqliteConnection.transactional (/home/langstra/code/mikro-orm/packages/knex/src/AbstractSqlConnection.ts:51:19)
        at async UnitOfWork.doCommit (/home/langstra/code/mikro-orm/packages/core/src/unit-of-work/UnitOfWork.ts:356:9)
[...]

PR #3851, opened 2022-12-15
fix(core): propagation with nullable 1:1 relation

### LINKED ISSUE #3850: Nullable 1-to-1 entity propagation not working
**Describe the bug**
When assigning an entity to another in a 1-to-1 relationship (either way round) with an auto-generated primary key, the foreign key on the owner side does not get set at all when flushed.

After inspecting the queries generated, the foreign key is always being set to `NULL` after the new entity is assigned and flushed. Is this because the ID doesn't exist yet on the new entity (as ID is auto-generated)? In fact the 2 queries needed are:
- Insert new entity
- Assign new entity ID to foreign key
but it looks like the update is query is attempted first (so the ID does not exist yet), then the new entity is inserted. This is probably the issue.

I'm just trying to upgrade to v5, and this definitely used to work in v4 as the tests used to pass, but now in this fairly simple case, entities are not linked at all. Maybe I'm doing something wrong. I did check the docs but may have missed something.

**To Reproduce**
1. See reproduction: https://github.com/alisd23/mikro-orm-issues/tree/1-to-1-propagation-issue
2. Run `yarn` to install (docker-compose also needed)
3. Run `yarn test:watch`

**Expected behavior**
Expect the entities to be assigned in the database via the foreign key.

**Additional context**

**Versions**

| Dependency | Version |
| - | - |
| node | 16 |
| typescript | 4.9 |
| mikro-orm | 5.6.0 |
| your-driver | mysql |

alisd23:
Fixes #3850

I'm not very familar with the code, so I'm not confident with the solution, but I think this is what is needed.

Tests added and related tests for #3614 also still pass

### DIFF
--- a/packages/core/src/unit-of-work/UnitOfWork.ts
+++ b/packages/core/src/unit-of-work/UnitOfWork.ts
@@ -908,11 +908,14 @@ export class UnitOfWork {
 
     for (const prop of props) {
       const insert = inserts.find(c => Utils.equals(c.payload[prop.name], changeSet.originalEntity![prop.name as string]));
+      const propEmpty = changeSet.payload[prop.name] === null || changeSet.payload[prop.name] === undefined;
 
       if (
         prop.name in changeSet.payload &&
         insert &&
-        insert.payload[prop.name]
+        // We only want to update early if the unique property on the changeset is going to be empty, so that
+        // the previous unique value can be set on a different entity without constraint issues
+        propEmpty
       ) {
         changeSet.type = ChangeSetType.UPDATE_EARLY;
       }
--- a/packages/core/src/unit-of-work/UnitOfWork.ts
+++ b/packages/core/src/unit-of-work/UnitOfWork.ts
@@ -907,7 +907,13 @@ export class UnitOfWork {
     const props = changeSet.meta.uniqueProps;
 
     for (const prop of props) {
-      if (prop.name in changeSet.payload && inserts.find(c => Utils.equals(c.payload[prop.name], changeSet.originalEntity![prop.name as string]))) {
+      const insert = inserts.find(c => Utils.equals(c.payload[prop.name], changeSet.originalEntity![prop.name as string]));
+
+      if (
+        prop.name in changeSet.payload &&
+        insert &&
+        insert.payload[prop.name]
+      ) {
         changeSet.type = ChangeSetType.UPDATE_EARLY;
       }
     }
--- a/tests/issues/GH3850-required-relation.test.ts
+++ b/tests/issues/GH3850-required-relation.test.ts
@@ -1,0 +1,156 @@
+import { Entity, OneToOne, PrimaryKey, Property, RequiredEntityData } from '@mikro-orm/core';
+import { MikroORM } from '@mikro-orm/sqlite';
+
+@Entity()
+class User {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property({ nullable: true })
+  name!: string;
+
+  @OneToOne(() => Project, project => project.owner, {
+    owner: true,
+    nullable: false,
+  })
+  public project!: Project | null;
+
+}
+
+@Entity()
+class Project {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property({ nullable: true })
+  name!: string;
+
+  @OneToOne(() => User, user => user.project, { nullable: false })
+  public owner?: User;
+
+}
+
+let orm: MikroORM;
+
+beforeAll(async () => {
+  orm = await MikroORM.init({
+    entities: [Project, User],
+    dbName: ':memory:',
+  });
+});
+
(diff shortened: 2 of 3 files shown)


packages/core/src/unit-of-work/UnitOfWork.ts
```diff
@@ -907,7 +907,13 @@ export class UnitOfWork {
     const props = changeSet.meta.uniqueProps;
 
     for (const prop of props) {
-      if (prop.name in changeSet.payload && inserts.find(c => Utils.equals(c.payload[prop.name], changeSet.originalEntity![prop.name as string]))) {
+      const insert = inserts.find(c => Utils.equals(c.payload[prop.name], changeSet.originalEntity![prop.name as string]));
+
+      if (
+        prop.name in changeSet.payload &&
+        insert &&
+        insert.payload[prop.name]
```
B4nan, line 915:
```suggestion
        insert?.payload[prop.name]
```

B4nan (approved):
Amazing, thank you! I just briefly checked the repros but didn't have time to actually look into it just yet. The change looks fine to me, so as long as the test suite is passing, we are good to merge.

alisd23:
No worries - just give me a few hours so I can add more test cases as I'm not quite sure this is the right solution.

Let me know if you've had time to check whether you think it's right/wrong 👍

alisd23:
Added a few more test cases, so now we cover:
- Adding a new entity to the relation (either side)
- Swapping an entity on the relation (either side)
- Unlinking an entity - but not deleting either entity (either side)
- Deleting an entity - (either side)

All seems to work as expected - hopefully nothing broken downstream

packages/core/src/unit-of-work/UnitOfWork.ts
alisd23: I can add this if you like, but shouldn't be needed due to the previous `&& insert` check on line 914.

Could instead replace as `insert.payload?[prop.name]` if you want to be extra careful about the payload being null/undefined.

alisd23: I've tried to have more of a think, and I believe this may be working by chance, though the tests work - I don't think the logic actually directly relates to the issue.

The original solution only breaks on my first test case: creating a new inverse entity with a auto-generated primary key.

I think the actual issue is specifically checking for if a primary key exists on the new entity (the insert), and if it doesn't then we *need* to insert that entity first (i.e. no early update)

alisd23: Added a new but similar solution (plus extra tests)
- Let me know if you would like the new test file renamed or moved elsewhere. I couldn't think how else to add it in the current test structure
- Also let me know what you think of the newest solution, as I'm not clear on all the cases for `UPDATE_EARLY` changes when they were added initially

B4nan: > I can add this if you like, but shouldn't be needed due to the previous && insert check on line 914.

FYI I proposed to replace those two lines with just one, not to use optional chaining for no reason :] 

> Also let me know what you think of the newest solution, as I'm not clear on all the cases for UPDATE_EARLY changes when they were added initially

Whatever makes the test suite green is fine with me. Early updates are there to handle edge cases mainly with unique column updates, I have been literally just fixing things and had to introduce it, so its not a thorough concept or something :]

tests/issues/GH3850-required-relation.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  @OneToOne(() => User, user => user.project, { nullable: false })
+  public owner?: User;
+
+}
+
+let orm: MikroORM;
+
+beforeAll(async () => {
+  orm = await MikroORM.init({
```
B4nan, line 38: this test hangs because its not calling `orm.close`

also please don't use `public` modified, its implicit

alisd23:
@B4nan All comments are addressed now. I'm happier with the latest solution as I think it's more specific and makes more sense in my head. There's a hell of a lot of tests which is good, so hopefully it's not broken anything related to 1-M or M-M relations.

Feel free to squash all commits into the first when merging - thanks!

tests/issues/GH3850.test.ts
```diff
@@ -47,6 +47,8 @@ beforeEach(async () => {
   await orm.schema.refreshDatabase();
 });
 
+afterAll(() => orm.close(true));
```
B4nan, line 50: this test already calls `close` on line 74/76, calling it twice will probably fail

alisd23: My bad - fixed

alisd23:
All looks good now on CI 👍 

Any idea when we might be able to release the fix? This is blocking the upgrade on my project unfortunately

B4nan:
every commit to master is published to NPM as dev version, so you can try it in few minutes

i will probably do a patch release on sunday

Issue #2703, opened 2022-02-03
Skip update queries that will be handled by DB cascading


evantrimboli:
**Is your feature request related to a problem? Please describe.**
Using the following code, when removing the `User` object it generates a SQL query to set the underlying values to `null`. Since the relation is not nullable, this causes an error. Because it's using `onDelete: 'cascade'`, it could probably be inferred that the query can be skipped.

```pgsql
update "order" set "user_id" = case when ("id" = 12) then NULL when ("id" = 11) then NULL when ("id" = 10) then NULL else "user_id" end where "id" in (12, 11, 10)
```

```typescript
@Entity()
export class User {
  @PrimaryKey()
  id!: number;

  @OneToMany(() => Order, ({ user }) => user)
  orders: Collection<Order> = new Collection<Order>(this);
}

@Entity()
export class Order {
  @PrimaryKey()
  id!: number;

  @Property()
  key!: string;

  @ManyToOne({ onDelete: 'cascade' })
  user!: User;
}

// Later
const u = (await em.find(User, {}, { populate: ['orders'] }))[0];
await em.removeAndFlush(u);
```

**Describe the solution you'd like**
The queries to set the values to `null` should not be issued.

**Describe alternatives you've considered**
1. This can be fixed by specifying `cascade: [Cascade.REMOVE, Cascade.PERSIST]` on the `OneToMany`, however this alternative issues a delete query for the `Order` entities when the `User` is deleted. Rather, it should not issue that query and let the DB handle it.
2. Using `Cascade.REMOVE` and allowing the FK to be nullable, however this isn't preferred.

PR #3854, opened 2022-12-15
test: show onDelete cascade causes a bug

Langstra:
This somehow breaks the tests, but I guess it should not.

### DIFF
--- a/tests/issues/GH3564.test.ts
+++ b/tests/issues/GH3564.test.ts
@@ -13,6 +13,7 @@ class Part {
   @ManyToOne(() => Car, {
     wrappedReference: true,
     nullable: true,
+    onDelete: 'cascade',
   })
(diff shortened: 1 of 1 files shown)


B4nan:
Thanks probably coming from here:

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/unit-of-work/ChangeSetComputer.ts#L130

Trying to run the test suite without it, maybe its no longer required.

B4nan:
So the test for #2703 will break, but maybe the solution was wrong, and it might make more sense to not propagate the change if the target property is not nullable.

B4nan:
https://github.com/mikro-orm/mikro-orm/commit/69a7f9469250f2951e8e19b66bb92b444f2dd7fe should help

PR #3856, opened 2022-12-16
test: name

productdevbook:


B4nan:
that property is a reference wrapper, it cant have any other properties than the PK - it's not the entity instance.

PR #3874, opened 2022-12-26
feat(cli): test database connection

### LINKED ISSUE #3855: Ability to test connection to database when running `mikro-orm debug`
Adding a call to verify connection settings would be ideal through the debug command. 

Adding a call to `MikroORM.connect()` in the the CLI debug command and verify output, or something similar.

liang799:
Closes #3855

### DIFF
--- a/packages/cli/src/commands/DebugCommand.ts
+++ b/packages/cli/src/commands/DebugCommand.ts
@@ -1,5 +1,5 @@
 import type { CommandModule } from 'yargs';
-import type { IDatabaseDriver } from '@mikro-orm/core';
+import type { Configuration } from '@mikro-orm/core';
 import { colors, ConfigurationLoader, MikroORM, Utils } from '@mikro-orm/core';
 
 import { CLIHelper } from '../CLIHelper';
@@ -16,7 +16,6 @@ export class DebugCommand implements CommandModule {
     CLIHelper.dump(`Current ${colors.cyan('MikroORM')} CLI configuration`);
     await CLIHelper.dumpDependencies();
     const settings = await ConfigurationLoader.getSettings();
-    let isConnected = false;
 
     if (settings.useTsNode) {
       CLIHelper.dump(' - ts-node ' + colors.green('enabled'));
@@ -29,10 +28,12 @@ export class DebugCommand implements CommandModule {
     try {
       const config = await CLIHelper.getConfiguration();
 
-      const mikroOrm = new MikroORM(config);
-      await DebugCommand.connect(mikroOrm);
-      CLIHelper.dump(` - Database Connection ${colors.green('connected')}`);
-      isConnected = true;
+      const isConnected = await DebugCommand.isConnected(config);
+      if (isConnected) {
+        CLIHelper.dump(` - Database Connection ${colors.green('connected')}`);
+      } else {
+        CLIHelper.dump(` - Database Connection ${colors.green('failed')}`);
+      }
 
       CLIHelper.dump(` - configuration ${colors.green('found')}`);
       const tsNode = config.get('tsNode');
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/DebugCommand.ts
```diff
@@ -80,4 +93,8 @@ export class DebugCommand implements CommandModule {
     }
   }
 
+  static connect(mikroOrm: MikroORM): Promise<IDatabaseDriver> {
+    return mikroOrm.connect();
```
erie0210, line 97: How about using `isConnected()` if you only want to check if database is connected or not.

packages/cli/src/commands/DebugCommand.ts
```diff
@@ -93,8 +89,8 @@ export class DebugCommand implements CommandModule {
     }
   }
 
-  static connect(mikroOrm: MikroORM): Promise<IDatabaseDriver> {
-    return mikroOrm.connect();
+  static async isConnected(config: Configuration): Promise<boolean> {
+    return new MikroORM(config).isConnected();
```
B4nan, line 93: This can never work, you never connect, so `isConnected` will be always false. I would suggest you first try to develop the changes with real app so you are sure it does what you think.

Also, there is no point in having such static method, this code can be directly in the handler. If the reason was mocking, you can mock anything without it too, via `jest.spyOn`.

packages/cli/src/commands/DebugCommand.ts
```diff
@@ -29,10 +28,12 @@ export class DebugCommand implements CommandModule {
     try {
       const config = await CLIHelper.getConfiguration();
 
-      const mikroOrm = new MikroORM(config);
-      await DebugCommand.connect(mikroOrm);
-      CLIHelper.dump(` - Database Connection ${colors.green('connected')}`);
-      isConnected = true;
+      const isConnected = await DebugCommand.isConnected(config);
+      if (isConnected) {
+        CLIHelper.dump(` - Database Connection ${colors.green('connected')}`);
```
B4nan, line 33: it should say something like `database connection successful/failed` (lower-case) and this log needs to go _after_ the `configuration found` one, it does not make sense the other way as you first need to be able to load config before you can try to connect.

B4nan (commented):
this is a bit overcomplicated, both implementation and the added test case. all the PR should be about is replicating what `MikroORM.connect` method already does - so basically just calling `await this.driver.connect()` and checking `this.driver.getConnection().isConnected()`. 

the test should be in `tests/features/cli`, the `tests/issues` is for regression tests, for bug reproductions if you want.

liang799:
> this is a bit overcomplicated, both implementation and the added test case. all the PR should be about is replicating what `MikroORM.connect` method already does - so basically just calling `await this.driver.connect()` and checking `this.driver.getConnection().isConnected()`.
> 
> the test should be in `tests/features/cli`, the `tests/issues` is for regression tests, for bug reproductions if you want.

Thanks, are there any good books to read to brush up on my testing skills? I have read a few books on Test Driven Development, one of them being written by Kent Beck. However the books I have read seem to no cover mocking, stubbing and spies, only the basics of TDD.

packages/cli/src/commands/DebugCommand.ts
liang799: Sorry, I do not understand the last part of your sentence. May I ask what do you mean by mock anything without it?

Do you mean:
```ts
jest.spyOn(MikroORM, 'connect')
```

If so, I having difficulties. Here is my current workaround that I have planned:
```ts
jest.mock('../../../packages/core/src/MikroORM');
MikroORM.connect.mockImplementation(() => resolveStubPromise());
```

liang799:
@B4nan Should I close this pull request?

B4nan:
Yeah, lets close it as its implemented via #3875.

Can't help much with books, I don't read books very much :D

PR #3905, opened 2023-01-05
docs: add information about `populateWhere` and loading strategies

thekevinbrown:
The populate section didn't mention how to use `populateWhere` or that there were multiple loading strategies supported, so this PR adds a bit of information on `populateWhere` then links people over to the loading strategies documentation from populate in case they're wondering how MikroORM will actually do the populate action in detail.


B4nan (changes requested):
looking great, thanks!

only one thing, please put the same changes also to the next docs, this is just a version snapshot and the changes would stay only in that version

https://github.com/mikro-orm/mikro-orm/tree/master/docs

thekevinbrown:
Done, thanks!

B4nan (approved):

PR #3916, opened 2023-01-06
fix(entity-generator): use table name instead of class name in `EntitySchema`

### LINKED ISSUE #3915: Incorrect table mapping when using `EntityGenerator` with `entitySchema`
**Describe the bug**
When using the `EntityGenerator` helper to generate the entities from existing database schema, the generated entity is mapped to wrong or non-existing table name **_especially_** when the table name has underscore (`_`). 

This may be happening for other special characters as well. 

**Stack trace**
```
...
```

**To Reproduce**
1. Create a table `Address_lines` 
  ```sql
  CREATE TABLE `Address_lines` (
  `nodeId` char(25) CHARACTER SET utf8 NOT NULL,
  `position` int(4) NOT NULL,
  `value` mediumtext COLLATE utf8mb4_unicode_ci NOT NULL,
  PRIMARY KEY (`nodeId`,`position`),
  KEY `value` (`value`(191))
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
  ```
2. setup the entity generation
  ```ts
  const orm = await init({
    // ... other config
    entitiesTs: [`./src/entities`],
    entityGenerator: {
      entitySchema: true,
      identifiedReferences: true,
    },
    type: 'mysql',
  })

  await orm.getEntityGenerator().generate({
    baseDir: path.join(__dirname, `./src/entities`),
    save: true,
  })

  await orm.close(true)
  ```
3. the following is the generated after the execusting the generation
  ```ts
    export class AddressLines {
      nodeId!: string
      position!: number
      value!: unknown
    }

  export const AddressLinesSchema = new EntitySchema({
    class: AddressLines,
    tableName: 'AddressLines', // This is wrong and should be inferred as `Address_lines`
    properties: {
      nodeId: {
        primary: true,
        fieldName: 'nodeId',
        columnType: 'char(25)',
        length: 25 
      },
      position: { primary: true, type: 'number' },
      value: {
        type: 'unknown',
        columnType: 'mediumtext',
        length: 16777215,
        index: 'value',
      },
    },
  })
  ```
**Expected behavior**
Entity class and schema files generated with corresponding table name. 
```ts
tableName: 'Address_lines',
```
**Additional context**
A possible issue could be in this line https://github.com/mikro-orm/mikro-orm/blob/master/packages/entity-generator/src/EntitySchemaSourceFile.ts#L45

**Versions**

| Dependency | Version |
| - | - |
| node | 16.x |
| typescript | 4.9.4 |
| mikro-orm | 5.6.3 |
| your-driver | mysql |

koneru9999:
Closes #3915

### DIFF
--- a/packages/entity-generator/src/EntitySchemaSourceFile.ts
+++ b/packages/entity-generator/src/EntitySchemaSourceFile.ts
@@ -41,7 +41,7 @@ export class EntitySchemaSourceFile extends SourceFile {
     ret += `export const ${this.meta.className}Schema = new EntitySchema({\n`;
     ret += `  class: ${this.meta.className},\n`;
 
-    if (this.meta.collection !== this.namingStrategy.classToTableName(this.meta.className)) {
+    if (this.meta.tableName !== this.namingStrategy.classToTableName(this.meta.className)) {
       ret += `  tableName: ${this.quote(this.meta.tableName)},\n`;
     }
 
--- a/packages/entity-generator/src/EntitySchemaSourceFile.ts
+++ b/packages/entity-generator/src/EntitySchemaSourceFile.ts
@@ -42,7 +42,7 @@ export class EntitySchemaSourceFile extends SourceFile {
     ret += `  class: ${this.meta.className},\n`;
 
     if (this.meta.collection !== this.namingStrategy.classToTableName(this.meta.className)) {
-      ret += `  tableName: ${this.quote(this.meta.className)},\n`;
+      ret += `  tableName: ${this.quote(this.meta.tableName)},\n`;
     }
 
     /* istanbul ignore next */
--- a/tests/features/entity-generator/EntityGenerator.test.ts
+++ b/tests/features/entity-generator/EntityGenerator.test.ts
@@ -151,6 +151,23 @@ describe('EntityGenerator', () => {
     await orm.close(true);
   });
 
+  test('table name with underscore using entitySchema [mysql]', async () => {
+    const orm = await initORMMySql('mysql', {entityGenerator: {
+        entitySchema: true,
+        identifiedReferences: true,
+      }}, true);
+    await orm.schema.dropSchema();
+    await orm.schema.execute(`
+      create table if not exists \`123_table_name\` (\`id\` int(10) unsigned not null auto_increment primary key) default character set utf8mb4 engine = InnoDB;
+    `);
+    const dump = await orm.entityGenerator.generate({ save: false, baseDir: './temp/entities' });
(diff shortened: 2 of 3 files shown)


packages/entity-generator/src/EntitySchemaSourceFile.ts
```diff
@@ -42,7 +42,7 @@ export class EntitySchemaSourceFile extends SourceFile {
     ret += `  class: ${this.meta.className},\n`;
 
     if (this.meta.collection !== this.namingStrategy.classToTableName(this.meta.className)) {
-      ret += `  tableName: ${this.quote(this.meta.className)},\n`;
+      ret += `  tableName: ${this.quote(this.meta.tableName)},\n`;
```
B4nan, line 45: `meta.tableName` is an alias for `meta.collection`, i'd like to use the same in the condition on line 44 as well, to reduce confusion. i dont mind using `meta.tableName` on both places

tests/features/entity-generator/EntityGenerator.test.ts
```diff
@@ -151,6 +151,23 @@ describe('EntityGenerator', () => {
     await orm.close(true);
   });
 
+  test('table name with underscore using entitySchema [mysql]', async () => {
+    const orm = await initORMMySql('mysql', {entityGenerator: {
```
B4nan, line 155: is this passing the linter? :] i dont think so, the indents on following lines are also wrong

```suggestion
    const orm = await initORMMySql('mysql', { entityGenerator: {
```

B4nan (approved):

packages/entity-generator/src/EntitySchemaSourceFile.ts
koneru9999: thanks. applied the same check at L44 now.

tests/features/entity-generator/EntityGenerator.test.ts
koneru9999: true that. i didn't run lint from my local

tests/features/entity-generator/EntityGenerator.test.ts
```diff
@@ -151,6 +151,25 @@ describe('EntityGenerator', () => {
     await orm.close(true);
   });
 
+  test('table name with underscore using entitySchema [mysql]', async () => {
+    const orm = await initORMMySql('mysql', {
+      entityGenerator: {
+        entitySchema: true,
+        identifiedReferences: true,
+      }
```
B4nan, line 159:
```suggestion
      },
```

B4nan:
Thanks!

Issue #3893, opened 2023-01-03
docs: PropertyOptions


xenoterracide:
There's a ton of options on `@Property` but it's hard to understand what many of them do, and what the difference between some of them is. It would be nice if these options were comprehensively documented using jsdoc, although I'd also just accept docs on the website

B4nan:
https://mikro-orm.io/docs/decorators#property

B4nan:
it's true this should be also part of the interface jsdoc, PR welcome for that, maybe @erie0210 want to take a peek?

erie0210:
Of course. Let me give it a try :)

PR #3918, opened 2023-01-07
docs: add jsdoc for property

erie0210:
Hi, 

I addded js-doc for `Property()`, but  I'm not sure it's what you meant in [the issue](https://github.com/mikro-orm/mikro-orm/issues/3893) Want to check if it is in line with your intention. 🫠

BTW, some of the options of `Property()` are [in docs](https://mikro-orm.io/docs/decorators#property) while others like `name`, `onCreation` are not. Did you selected only some of them on purpose? 

Related issues: https://github.com/mikro-orm/mikro-orm/issues/3893

### DIFF
--- a/packages/core/src/decorators/Property.ts
+++ b/packages/core/src/decorators/Property.ts
@@ -76,14 +76,20 @@ export type PropertyOptions<T> = {
    */
   type?: keyof typeof types | 'ObjectId' | Date | Constructor<AnyEntity> | Constructor<Type<any>> | Type<any> | (() => unknown) | ColumnType | AnyString;
   /**
-   * Length of database column, used for datetime/timestamp/varchar column types for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
+   * Set length of database column, used for datetime/timestamp/varchar column types for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
    */
   length?: number;
   /**
-   * Precision of database column, used for datetime/timestamp/varchar column types for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
+   * Set precision of database column to represent the number of significant digits. (SQL only)
    */
   precision?: number;
+  /**
+   * Set scale of database column to represents the number of digits after the decimal point. (SQL only)
+   */
   scale?: number;
+  /**
+   * Explicitly specify the auto increment of the primary key.
+   */
   autoincrement?: boolean;
   /**
    * Automatically set the property value when entity gets created, executed during flush operation.
@@ -161,9 +167,43 @@ export type PropertyOptions<T> = {
    * @see https://mikro-orm.io/docs/defining-entities#lazy-scalar-properties
    */
   lazy?: boolean;
+  /**
+   * Set true to define entity's unique primary key identifier.
+   * Alias for `@PrimaryKey()` decorator
+   *
+   * @see https://mikro-orm.io/docs/decorators#primarykey
+   */
   primary?: boolean;
+  /**
+   * Set true to define the properties as setter. (virtual)
+   *
+   * @example
+   * ```
+   * @Property({ setter: true })
+   * set address(value: string) {
+   *     this._address = value.toLocaleLowerCase();
+   * }
+   * ```
+   */
   setter?: boolean;
+  /**
+   * Set true to define the properties as getter. (virtual)
+   *
+   * @example
+   * ```
+   * @Property({ getter: true })
+   * get fullName() {
+   *   return this.firstName + this.lastName;
+   * }
+   * ```
+   */
   getter?: boolean;
+  /**
+   * Set to define serialized primary key for MongoDB. (virtual)
+   * Alias for `@SerializedPrimaryKey()` decorator.
+   *
+   * @see https://mikro-orm.io/docs/decorators#serializedprimarykey
+   */
   serializedPrimaryKey?: boolean;
   /**
    * Set to use serialize property. Allow to specify a callback that will be used when serializing a property.
--- a/packages/core/src/decorators/Property.ts
+++ b/packages/core/src/decorators/Property.ts
@@ -41,42 +41,155 @@ export function Property<T>(options: PropertyOptions<T> = {}) {
 }
 
 export type PropertyOptions<T> = {
+  /**
+   * Alias for `fieldName`.
+   */
   name?: string;
+  /**
+   * Specify database column name for this property.
+   *
+   * @see https://mikro-orm.io/docs/naming-strategy
+   */
   fieldName?: string;
+  /**
+   * Specify database column names for this property.
+   * Same as `fieldName` but for composite FKs.
+   *
+   * @see https://mikro-orm.io/docs/naming-strategy
+   */
   fieldNames?: string[];
+  /**
+   * Explicitly specify the mapped type instance for this property.
+   *
+   * @see https://mikro-orm.io/docs/custom-types
+   */
   customType?: Type<any>;
+  /**
+   * Specify exact database column type for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
+   */
   columnType?: ColumnType | AnyString;
+  /**
+   * Explicitly specify the runtime type.
+   *
+   * @see https://mikro-orm.io/docs/metadata-providers
+   * @see https://mikro-orm.io/docs/custom-types
+   */
   type?: keyof typeof types | 'ObjectId' | Date | Constructor<AnyEntity> | Constructor<Type<any>> | Type<any> | (() => unknown) | ColumnType | AnyString;
+  /**
+   * Length of database column, used for datetime/timestamp/varchar column types for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
+   */
   length?: number;
+  /**
+   * Precision of database column, used for datetime/timestamp/varchar column types for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
+   */
   precision?: number;
   scale?: number;
   autoincrement?: boolean;
+  /**
+   * Automatically set the property value when entity gets created, executed during flush operation.
+   * @param entity
+   */
   onCreate?: (entity: T) => any;
+  /**
+   * Automatically update the property value every time entity gets updated, executed during flush operation.
+   * @param entity
+   */
   onUpdate?: (entity: T) => any;
+  /**
+   * Specify default column value for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}.
+   * This is a runtime value, assignable to the entity property. (SQL only)
+   */
   default?: string | string[] | number | number[] | boolean | null;
+  /**
+   * Specify SQL functions for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
+   * Since v4 you should use defaultRaw for SQL functions. e.g. now()
+   */
   defaultRaw?: string;
+  /**
+   * Set to map some SQL snippet for the entity.
+   *
+   * @see https://mikro-orm.io/docs/defining-entities#formulas Formulas}
+   */
   formula?: string | ((alias: string) => string);
+  /**
+   * Set column as nullable for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}.
+   */
   nullable?: boolean;
+  /**
+   * Set column as unsigned for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
+   */
   unsigned?: boolean;
+  /**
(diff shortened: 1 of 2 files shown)


packages/core/src/decorators/Property.ts
```diff
@@ -5,6 +5,12 @@ import { ReferenceType } from '../enums';
 import type { EntityName, EntityProperty, Constructor, CheckCallback, Dictionary, AnyString, AnyEntity } from '../typings';
 import type { Type, types } from '../types';
 
+/**
+ * @function
```
B4nan, line 9: i dont think we need things like `@function` or `@template`, not inside TS, this is all inferred.

instead of this i would just add jsdoc comments above each property from the `PropertyOptions`

```ts
export type PropertyOptions<T> = {
  /** foo bar 1 */
  name?: string;
  /** foo bar 2 */
  fieldName?: string;
  ...
```

B4nan (changes requested):
> BTW, some of the options of Property() are [in docs](https://mikro-orm.io/docs/decorators#property) while others like name, onCreation are not. Did you selected only some of them on purpose?

property `name` is not configurable, the one from the decorator options is an alias for `fieldName`. some things might be indeed missing like the `onCreate` callback

packages/core/src/decorators/Property.ts
erie0210: OMG. I got it. Let me retry 🥸

packages/core/src/decorators/Property.ts
```diff
@@ -42,28 +42,78 @@ export function Property<T>(options: PropertyOptions<T> = {}) {
 
 export type PropertyOptions<T> = {
   name?: string;
+  /**
+   * Override default property name.
```
B4nan, line 46: this is field name, or a column name if you want, not property name. field/column name is in the database, property name is on the entity. as said above, `PropertyOptions.name` is an alias for `fieldName`.

i would also document the plural variants - those are there as one entity property can be represented by multiple database columns - that happens for composite FKs

packages/core/src/decorators/Property.ts
```diff
@@ -42,28 +42,78 @@ export function Property<T>(options: PropertyOptions<T> = {}) {
...
   scale?: number;
   autoincrement?: boolean;
   onCreate?: (entity: T) => any;
+  /**
+   * Automatically update the property value every time entity gets updated.
+   * @param entity
+   */
   onUpdate?: (entity: T) => any;
+  /**
+   * 	Specify default column value for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
```
B4nan, line 83:
```suggestion
   * Specify default column value for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
```

B4nan (changes requested):

packages/core/src/decorators/Property.ts
```diff
@@ -42,28 +42,78 @@ export function Property<T>(options: PropertyOptions<T> = {}) {
...
+   */
   type?: keyof typeof types | 'ObjectId' | Date | Constructor<AnyEntity> | Constructor<Type<any>> | Type<any> | (() => unknown) | ColumnType | AnyString;
+  /**
+   * Length/precision of database column, used for datetime/timestamp/varchar column types for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
+   */
   length?: number;
   precision?: number;
   scale?: number;
   autoincrement?: boolean;
   onCreate?: (entity: T) => any;
```
B4nan, line 76: lets document those too, and also the `defaultRaw`. i hope its kind of self explanatory, let me know if i should provide some of the descriptions

erie0210: Thanks! But before asking, I searched and found these lines from [this document](https://mikro-orm.io/docs/defining-entities#default-values). Thought it might be something close to your explanation. :)

erie0210: I have a quick question, what does it mean by `SQL only` and how can I know?

B4nan: some of the options affect how the SchemaGenerator works, and those are SQL only, meaning they affect only SQL drivers - as in mongo there is no schema. maybe this should be explained somewhere on the top of that page.

B4nan: but speaking of the `default`, we should probably make this work with `em.create` too, as we know that will be a scalar value assignable to the entity property. as opposed to `defaultRaw` which is an SQL fragment, e.g. `now()`, or anything really, but if its a string, it needs to be quoted. that is what will be used directly in the create/alter table statement.

packages/core/src/decorators/Property.ts
```diff
@@ -41,29 +41,100 @@ export function Property<T>(options: PropertyOptions<T> = {}) {
...
+   */
   length?: number;
+  /**
+   * Precision of database column, used for datetime/timestamp/varchar column types for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
+   */
   precision?: number;
   scale?: number;
   autoincrement?: boolean;
+  /**
+   * Automatically set the property value when entity gets created.
```
B4nan, line 90:
```suggestion
   * Automatically set the property value when entity gets created, executed during flush operation.
```

packages/core/src/decorators/Property.ts
```diff
@@ -41,29 +41,100 @@ export function Property<T>(options: PropertyOptions<T> = {}) {
...
+   * @param entity
+   */
   onCreate?: (entity: T) => any;
+  /**
+   * Automatically update the property value every time entity gets updated.
```
B4nan, line 95:
```suggestion
   * Automatically update the property value every time entity gets updated, executed during flush operation.
```

packages/core/src/decorators/Property.ts
```diff
@@ -41,29 +41,100 @@ export function Property<T>(options: PropertyOptions<T> = {}) {
...
+   * @param entity
+   */
   onUpdate?: (entity: T) => any;
+  /**
+   * Specify default column value for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
```
B4nan, line 100:
```suggestion
   * Specify default column value for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. This is a runtime value, assignable to the entity property. (SQL only)
```

packages/core/src/decorators/Property.ts
```diff
@@ -41,29 +41,100 @@ export function Property<T>(options: PropertyOptions<T> = {}) {
...
+   */
   default?: string | string[] | number | number[] | boolean | null;
+  /**
+   * Specify SQL functions for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
+   * Since v4 you should use defaultRaw for SQL functions, as default with string values will be automatically quoted.
+   */
   defaultRaw?: string;
   formula?: string | ((alias: string) => string);
```
B4nan, line 108: this one should point to https://mikro-orm.io/docs/defining-entities#formulas

packages/core/src/decorators/Property.ts
```diff
@@ -41,29 +41,100 @@ export function Property<T>(options: PropertyOptions<T> = {}) {
...
+  /**
+   * Set column as nullable for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
```
B4nan, line 110: nowadays this flag also has an effect on how the hydration functions look like as well as some other internals (e.g. required fields validation, commit order calculation...), so its not just for SQL drivers.

packages/core/src/decorators/Property.ts
```diff
@@ -41,29 +41,100 @@ export function Property<T>(options: PropertyOptions<T> = {}) {
...
   nullable?: boolean;
+  /**
+   * Set column as unsigned for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
+   */
   unsigned?: boolean;
+  /**
+   * Set false to define {@link https://mikro-orm.io/docs/serializing#shadow-properties Shadow Property}.
+   */
   persist?: boolean;
   trackChanges?: boolean;
```
B4nan, line 121: https://mikro-orm.io/docs/unit-of-work#change-tracking-and-performance-considerations

packages/core/src/decorators/Property.ts
```diff
@@ -41,29 +41,100 @@ export function Property<T>(options: PropertyOptions<T> = {}) {
...
+  /**
+   * Set to true to omit the property when {@link https://mikro-orm.io/docs/serializing Serializing}.
+   */
   hidden?: boolean;
+  /**
+   * Set to true to enable {@link https://mikro-orm.io/docs/transactions#optimistic-locking Optimistic Locking} via version field. (SQL only)
+   */
   version?: boolean;
+  /**
+   * Set to true to enable {@link https://mikro-orm.io/docs/transactions#concurrency-checks Concurrency Check} via concurrency fields.
```
B4nan, line 131: i'd still call this optimistic locking. "enable concurrency check via concurrency fields" sounds weird...

packages/core/src/decorators/Property.ts
```diff
@@ -41,29 +41,100 @@ export function Property<T>(options: PropertyOptions<T> = {}) {
...
+   */
   concurrencyCheck?: boolean;
   index?: boolean | string;
```
B4nan, line 134: this is an alternative to explicit `@Index` decorator on a property

packages/core/src/decorators/Property.ts
```diff
@@ -41,29 +41,100 @@ export function Property<T>(options: PropertyOptions<T> = {}) {
...
+  /**
+   * Set column as unique for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
+   */
   unique?: boolean | string;
   check?: string | CheckCallback<T>;
```
B4nan, line 139: https://mikro-orm.io/docs/defining-entities#check-constraints

packages/core/src/decorators/Property.ts
```diff
@@ -41,29 +41,100 @@ export function Property<T>(options: PropertyOptions<T> = {}) {
...
   lazy?: boolean;
```
B4nan, line 140: https://mikro-orm.io/docs/defining-entities#lazy-scalar-properties

packages/core/src/decorators/Property.ts
```diff
@@ -73,7 +144,13 @@ export type PropertyOptions<T> = {
   serializedPrimaryKey?: boolean;
   serializer?: (value: any) => any;
   serializedName?: string;
+  /**
+   * Specify a custom order for the column. (SQL only)
```
B4nan, line 148:
```suggestion
   * Specify a custom order based on the values. (SQL only)
```

packages/core/src/decorators/Property.ts
```diff
@@ -73,7 +144,13 @@ export type PropertyOptions<T> = {
...
+   */
   customOrder?: string[] | number[] | boolean[];
+  /**
+   * Specify comment of column for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}. (SQL only)
+   */
   comment?: string;
   /** mysql only */
   extra?: string;
```
B4nan, line 156: dumb github wont let me commet on the line 157, but that is for https://mikro-orm.io/docs/defining-entities#sql-generated-columns

packages/core/src/decorators/Property.ts
```diff
@@ -73,7 +144,13 @@ export type PropertyOptions<T> = {
   serializedPrimaryKey?: boolean;
   serializer?: (value: any) => any;
   serializedName?: string;
```
B4nan, line 146: those are about https://mikro-orm.io/docs/serializing#property-serializers

B4nan (changes requested):

erie0210:
I'm working on it, but a little bit busy with my work. 👾

B4nan:
No worries 👍

packages/core/src/decorators/Property.ts
erie0210: haha thanks!

erie0210:
Thanks for waiting !! 🙂 Amended PR contents based on your reviews. Added `@see + link` for the links you attached.

erie0210:
While searching I found [property options description](https://mikro-orm.io/api/core#PropertyOptions), and there's `mysql only`. Can I ask you what it means?

erie0210:
Added explanations for the options not mentioned above, like `precision`, `scale`, `setter/getter`, `serializedPrimaryKey` etc. I haven't used some them, so I wrote them by searching which makes me not sure about them. So it will be appreciated if you check them. 🤓 Thanks!

erie0210:
@B4nan It would be grateful if you spare some time to check this PR.  (quick reminder in case you forgot!) 🙌

B4nan (approved):
looking good, thanks!

PR #3923, opened 2023-01-09
feat(cli): add check for migrations command

kpervin:
Allow checking for migrations. This would be useful for adding to git hooks to ensure that any changes to ORM entities are captured in migrations.

### DIFF
--- a/docs/docs/migrations.md
+++ b/docs/docs/migrations.md
@@ -161,10 +161,10 @@ You can use it via CLI:
 
 ```sh
 npx mikro-orm migration:create   # Create new migration with current schema diff
-npx mikro-orm migration:check    # Check if schema is up to date
 npx mikro-orm migration:up       # Migrate up to the latest version
 npx mikro-orm migration:down     # Migrate one step down
 npx mikro-orm migration:list     # List all executed migrations
+npx mikro-orm migration:check    # Check if schema is up to date
 npx mikro-orm migration:pending  # List all pending migrations
 npx mikro-orm migration:fresh    # Drop the database and migrate up to the latest version
 ```
--- a/docs/docs/migrations.md
+++ b/docs/docs/migrations.md
@@ -161,6 +161,7 @@ You can use it via CLI:
 
 ```sh
 npx mikro-orm migration:create   # Create new migration with current schema diff
+npx mikro-orm migration:check    # Check if schema is up to date
 npx mikro-orm migration:up       # Migrate up to the latest version
 npx mikro-orm migration:down     # Migrate one step down
 npx mikro-orm migration:list     # List all executed migrations
--- a/packages/cli/src/commands/MigrationCommandFactory.ts
+++ b/packages/cli/src/commands/MigrationCommandFactory.ts
@@ -169,13 +169,12 @@ export class MigrationCommandFactory {
   }
 
   private static async handleCheckCommand(migrator: IMigrator, orm: MikroORM): Promise<void> {
-    const diff = await migrator.checkMigrationNeeded();
-
-    if (diff.up.length === 0) {
+    if (!await migrator.checkMigrationNeeded()) {
       return CLIHelper.dump(colors.green(`No changes required, schema is up-to-date`));
     }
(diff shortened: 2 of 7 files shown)


packages/cli/src/commands/MigrationCommandFactory.ts
```diff
@@ -164,6 +168,16 @@ export class MigrationCommandFactory {
     CLIHelper.dump(colors.green(`${ret.fileName} successfully created`));
   }
 
+  private static async handleCheckCommand(migrator: IMigrator, orm: MikroORM): Promise<void> {
+    const diff = await migrator.checkMigrationNeeded();
```
B4nan, line 172: that method should return boolean i guess

packages/cli/src/commands/MigrationCommandFactory.ts
```diff
@@ -164,6 +168,16 @@ export class MigrationCommandFactory {
...
+
+    if (diff.up.length === 0) {
+      return CLIHelper.dump(colors.green(`No changes required, schema is up-to-date`));
+    }
+    await orm.close(true);
+    throw new Error(`Changes detected. Please create migrations to update schema.`);
```
B4nan, line 178: i am not entirely sure i like the idea of throwing here. maybe `process.exit(1)` would be better

kpervin: That's funny, because originally I was going to use `process.exit(1)` but thought it might be better to throw an error lol

B4nan: i dont like the idea of showing a stack trace for this, sounds quite confusing. and for automation you only care about the exit code

packages/cli/src/commands/MigrationCommandFactory.ts
```diff
@@ -8,6 +8,7 @@ export class MigrationCommandFactory {
 
   static readonly DESCRIPTIONS = {
     create: 'Create new migration with current schema diff',
+    check: 'Check if migrations are needed. Useful for bash scripts.',
```
B4nan, line 11: lets move this to less prominent place, maybe after `list` command

docs/docs/migrations.md
```diff
@@ -161,6 +161,7 @@ You can use it via CLI:
 
 ```sh
 npx mikro-orm migration:create   # Create new migration with current schema diff
+npx mikro-orm migration:check    # Check if schema is up to date
```
B4nan, line 164: lets move this to less prominent place, maybe after list command

packages/cli/src/commands/MigrationCommandFactory.ts
```diff
@@ -164,6 +168,15 @@ export class MigrationCommandFactory {
     CLIHelper.dump(colors.green(`${ret.fileName} successfully created`));
   }
 
+  private static async handleCheckCommand(migrator: IMigrator, orm: MikroORM): Promise<void> {
+    if (!await migrator.checkMigrationNeeded()) {
+      return CLIHelper.dump(colors.green(`No changes required, schema is up-to-date`));
+    }
+    await orm.close(true);
+    CLIHelper.dump(colors.yellow(`Changes detected. Please create migrations to update schema.`));
```
B4nan, line 176:
```suggestion
    CLIHelper.dump(colors.yellow(`Changes detected. Please create migration to update schema.`));
```

packages/migrations/src/Migrator.ts
```diff
@@ -51,7 +51,6 @@ export class Migrator implements IMigrator {
     if (initial) {
       return this.createInitialMigration(path);
     }
-
```
B4nan: unrelated change, the line break was there on purpose

B4nan (commented):
looking good, just few nits, thanks!

B4nan:
and please dont rebase after you get a review, i cant see what you changed. will be squash merging the PR so it does not really matter

kpervin:
>

B4nan (approved):

PR #3940, opened 2023-01-13
fix(core): propagate ManyToOne only to matching collections (STI)

### LINKED ISSUE #3939: Adding entity to collection also adds it to collection of _siblings_ if these entities use Single Table Inheritance
**Describe the bug**
When entities use Single Table Inheritance adding entity to a collection puts it to all collections of _siblings_.

**To Reproduce**
Suppose we have an abstract entity **Vehicle** and two descendants **Car** and **Truck** and they [share same table](https://mikro-orm.io/docs/inheritance-mapping#single-table-inheritance) 
**Garage** entity has collections for both `cars` and `trucks`, then adding a _car_ to _garage_ results in _car_ being added to `trucks` collection as well as `cars`.

Declaring entitites
```
@Entity({ tableName: 'vehicle', discriminatorColumn: 'type', abstract: true })
class Vehicle {

  @PrimaryKey()
  id!: number;

  @ManyToOne(() => Garage, { wrappedReference: true })
  garage!: IdentifiedReference<Garage>;

}

@Entity({ discriminatorValue: 'car' })
class Car extends Vehicle {}

@Entity({ discriminatorValue: 'truck' })
class Truck extends Vehicle {}

@Entity({ tableName: 'garage' })
class Garage {

  @PrimaryKey()
  id!: number;

  @OneToMany(() => Truck, v => v.garage)
  trucks = new Collection<Truck>(this);

  @OneToMany(() => Car, v => v.garage)
  cars = new Collection<Car>(this);
}
```
Create _garage_ and put a _car_ into it
```
const garage = orm.em.create(Garage, {});
const car = orm.em.create(Car, { garage });

expect(garage.cars.length).toBe(1);
expect(garage.trucks.length).toBe(0);
```

**Expected behavior**
Entity should be added to collections of exactly this type or parent type, not siblings.

**Versions**

| Dependency | Version |
| - | - |
| node | 18.13 |
| typescript | 4.9.4 |
| mikro-orm | 5.5.3 |
| your-driver | mongo |

funduck:
Closes #3939

### DIFF
--- a/packages/core/src/entity/EntityHelper.ts
+++ b/packages/core/src/entity/EntityHelper.ts
@@ -152,7 +152,7 @@ export class EntityHelper {
   }
 
   private static propagate<T extends object, O extends object>(meta: EntityMetadata<O>, entity: T, owner: O, prop: EntityProperty<O>, value?: T[keyof T & string], old?: object): void {
-    const inverseProps = prop.targetMeta!.relations.filter(prop2 => (prop2.inversedBy || prop2.mappedBy) === prop.name && prop2.targetMeta!.root.className === meta.root.className);
+    const inverseProps = prop.targetMeta!.relations.filter(prop2 => (prop2.inversedBy || prop2.mappedBy) === prop.name && (prop2.targetMeta!.abstract ? prop2.targetMeta!.root.className === meta.root.className : prop2.targetMeta!.className === meta.className));
 
     for (const prop2 of inverseProps) {
       const inverse = value?.[prop2.name as string];
--- a/packages/core/src/entity/EntityHelper.ts
+++ b/packages/core/src/entity/EntityHelper.ts
@@ -152,7 +152,14 @@ export class EntityHelper {
   }
 
   private static propagate<T extends object, O extends object>(meta: EntityMetadata<O>, entity: T, owner: O, prop: EntityProperty<O>, value?: T[keyof T & string], old?: object): void {
-    const inverseProps = prop.targetMeta!.relations.filter(prop2 => (prop2.inversedBy || prop2.mappedBy) === prop.name && (prop2.targetMeta!.abstract ? prop2.targetMeta!.root.className === meta.root.className : prop2.targetMeta!.className === meta.className));
+    const inverseProps = prop.targetMeta!.relations.filter(prop2 =>
+      (prop2.inversedBy || prop2.mappedBy) === prop.name
+      && (
+        prop2.targetMeta!.abstract ?
+          prop2.targetMeta!.root.className === meta.root.className :
+          prop2.targetMeta!.className === meta.className
+      ),
+    );
 
     for (const prop2 of inverseProps) {
       const inverse = value?.[prop2.name as string];
--- a/tests/issues/GH2371.test.ts
+++ b/tests/issues/GH2371.test.ts
@@ -60,5 +60,74 @@ describe('GH issue 2371', () => {
     expect(g.cars.contains(c)).toBe(true);
     expect(g.vehicles.contains(c)).toBe(true);
   });
+});
 
+
+class BaseEntity {
+
+  constructor(data = {}) {
+    Object.assign(this, data);
+  }
+
+  @PrimaryKey()
+  _id = +new Date() + Math.random();
+
+}
+
+@Entity({ tableName: 'basket' })
+class Basket extends BaseEntity {
+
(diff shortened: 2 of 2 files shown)


funduck:
Locally all tests pass but [this fails](https://github.com/mikro-orm/mikro-orm/actions/runs/3910092891/jobs/6681904214)
```
npx jest --runInBand 3294
 PASS  tests/features/result-cache/GH3294.test.ts
  hidden properties are still included when cached (GH 3294)
    ✓ single entity (findOne) (28 ms)
    ✓ multiple entities (find) (9 ms)

Tests:       2 passed, 2 total
Snapshots:   0 total
Time:        1.078 s
Ran all test suites matching /3294/i.
```

B4nan:
don't worry about that, it's just flaky because of slow CI, the test is asserting some time-constrained caching

could you try to refactor that line a bit so it's more readable? it's too complex for a one-liner now

tests/issues/GH2371.test.ts
```diff
@@ -60,5 +60,74 @@ describe('GH issue 2371', () => {
...
+
+}
+
+@Entity({ discriminatorValue: 'apple' })
+class Apple extends Fruit {}
 
+@Entity({ discriminatorValue: 'banana' })
+class Banana extends Fruit {}
+
+describe('GH issue 2371 (M)', () => {
```
B4nan, line 102: if this test dont share anything with the one for #2371, i'd put it into separate file. or reuse the entities and have just a separate `test` block, not a separate `describe` block

tests/issues/GH2371.test.ts
```diff
@@ -60,5 +60,74 @@ describe('GH issue 2371', () => {
...
     expect(g.vehicles.contains(c)).toBe(true);
   });
+});
+
+
+class BaseEntity {
+
+  constructor(data = {}) {
+    Object.assign(this, data);
+  }
```
B4nan, line 70: FYI this is quite an antipattern, you can do this only for scalar properties, no relations

funduck: Ok, I'll apply your advices :)

tests/issues/GH2371.test.ts
funduck: Sure. By the way, why not using prettier?

B4nan: because it produces crap much more often than i would be ok with. it does terrible job namely at formatting complex TS types.

i would very much like to use it if they add more configuration to, namely to working with blank lines and line breaks, which they say is a non-goal 🤷

packages/core/src/entity/EntityHelper.ts
```diff
@@ -152,7 +152,14 @@ export class EntityHelper {
   }
 
   private static propagate<T extends object, O extends object>(meta: EntityMetadata<O>, entity: T, owner: O, prop: EntityProperty<O>, value?: T[keyof T & string], old?: object): void {
-    const inverseProps = prop.targetMeta!.relations.filter(prop2 => (prop2.inversedBy || prop2.mappedBy) === prop.name && prop2.targetMeta!.root.className === meta.root.className);
+    const inverseProps = prop.targetMeta!.relations.filter(prop2 =>
+      (prop2.inversedBy || prop2.mappedBy) === prop.name
+      && (
+        prop2.targetMeta!.abstract ?
+          prop2.targetMeta!.root.className === meta.root.className :
+          prop2.targetMeta!.className === meta.className
```
B4nan, line 160: hmm interesting, there might be more places doing the same thing, as the invariant is "use `meta.root` almost always, as that points to the actual metadata representing the table. but when it comes to hydration, we need to care about the target being a leaf or not...

B4nan (approved):

B4nan:
thanks a lot for the fix!

packages/core/src/entity/EntityHelper.ts
funduck: Well, I'll use mikro-orm in production and STI is in its heart, so if there are more bugs related to this topic, they will probably come :)

tests/issues/GH2371.test.ts
B4nan: use `assign` helper from the ORM (there is a static export, its the same functioin as `em.assign`), it will throw when you try to assign relations by PK (as that requires EM instance), but i am planing to get around that soon (v6)

or `em.create()` / `em.assign()` from outside

PR #3942, opened 2023-01-13
test: add failing test for unexpected sql updates

Tlepel:


tests/issues/GH3941.test.ts
```diff
@@ -0,0 +1,111 @@
...
+  public organization!: IdentifiedReference<Organization>;
+
+}
+
+let orm: MikroORM;
+
+beforeAll(async () => {
+  orm = await MikroORM.init({
+    entities: [Organization, License, Workspace],
+    dbName: ':memory:',
```
B4nan, line 63: FYI only sqlite supports this. i see so many people trying to use it with other drivers. it just creates a database with such name...

Tlepel: Good to know! I wasn't really familiar with writing tests for this project, so I copied it from other tests and changed it to use MySQL to reflect our use-case.

PR #3943, opened 2023-01-13
doc: add usage for Rel wrapper

rubiin:
This adds the documentation for usage of `Rel` wrapper for avoiding circular dependencies mostly on esm projects

### DIFF
--- a/docs/docs/relationships.md
+++ b/docs/docs/relationships.md
@@ -170,3 +170,52 @@ export class BookTag {
 ```
 
 Again, more information about how collections work can be found on [collections page](collections.md). 
+
+## Relations in ESM projects
+If you use ESM in your TypeScript project, you should use the `Rel` wrapper type in relation properties to avoid circular dependency issues. Let's modify our entities:
+
+```ts
+import { Rel } from "@mikro-orm/core";
+
+@Entity()
+export class Book {
+
+  @ManyToOne() // plain decorator is enough, type will be sniffer via reflection!
+  author1!: Rel<Author>;
+
+  @ManyToOne(() => Author) // you can specify type manually as a callback
+  author2!: Rel<Author>;
+
+  @ManyToOne('Author') // or as a string
+  author3!: Rel<Author>;
+
+  @ManyToOne({ entity: () => Author }) // or use options object
+  author4!: Rel<Author>;
+
+}
+```
+
+
+```ts
+import { Rel } from "@mikro-orm/core";
+
+@Entity()
+export class User {
+
+  // when none of `owner/inverseBy/mappedBy` is provided, it will be considered owning side
+  @OneToOne()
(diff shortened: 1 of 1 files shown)


docs/docs/relationships.md
```diff
@@ -170,3 +170,52 @@ export class BookTag {
 ```
 
 Again, more information about how collections work can be found on [collections page](collections.md). 
+
+## Relations in ESM projects
+If you use ESM in your TypeScript project, you should use the `Rel` wrapper type in relation properties to avoid circular dependency issues. Let's modify our entities:
+
+```ts
+import { Rel } from "@mikro-orm/core";
```
B4nan, line 178:
```suggestion
import { Rel } from '@mikro-orm/core';
```

docs/docs/relationships.md
```diff
@@ -170,3 +170,52 @@ export class BookTag {
...
+  author1!: Rel<Author>;
+
+  @ManyToOne(() => Author) // you can specify type manually as a callback
+  author2!: Rel<Author>;
+
+  @ManyToOne('Author') // or as a string
+  author3!: Rel<Author>;
+
+  @ManyToOne({ entity: () => Author }) // or use options object
+  author4!: Rel<Author>;
```
B4nan, line 193: lets simplify this, we dont need to repeat the same example completely 

```suggestion
  @ManyToOne(() => Author)
  author!: Rel<Author>;
```

docs/docs/relationships.md
```diff
@@ -170,3 +170,52 @@ export class BookTag {
...
+  bestFriend2!: Rel<User>;
+
+  // when defining it like this, you need to specifically mark the owning side with `owner: true`
+  @OneToOne(() => User, user => user.bestFriend2, { owner: true })
+  bestFriend3!: Rel<User>;
+
+}
+
+```
+
```
B4nan, line 220: and i think we dont need the other side at all

```suggestion
```

docs/docs/relationships.md
```diff
@@ -170,3 +170,52 @@ export class BookTag {
 ```
 
 Again, more information about how collections work can be found on [collections page](collections.md). 
+
+## Relations in ESM projects
+If you use ESM in your TypeScript project, you should use the `Rel` wrapper type in relation properties to avoid circular dependency issues. Let's modify our entities:
```
B4nan, line 175:
```suggestion
## Relations in ESM projects

If you use ESM in your TypeScript project with `reflect-metadata`, you might fall into issues with circular dependencies, seeing errors like this:

    ReferenceError: Cannot access 'Author' before initialization

To get around them, use the `Rel` mapped type. It is an identity type, which disables the problematic inference from `reflect-metadata`, that causes ESM projects to fail.
```

B4nan (approved):

PR #3958, opened 2023-01-18
feat(core): allow filtering in `Collection.loadCount()`

### LINKED ISSUE #3527: Allowing pagination on entity collections
I've seen [#1502](https://github.com/mikro-orm/mikro-orm/pull/1502), however paginated results in the project I'm working on follow this structure:
```ts
export abstract class QueryResult<T> {
  data: T[];
  totalCount: number;
  pageCount: number;
  hasPrevious: boolean;
  hasNext: boolean;
}
```
I needed to create a function which fetches data from a collection considering **filters and pagination**, this is my working implementation:
```ts
public async getBookings(
    @Parent() endUser: EndUser,
    @Args({ name: 'pagination', type: () => PaginationInput })
    pagination: PaginationInput,
    @Args({ name: 'filter', type: () => BookingFilter, nullable: true })
    filter?: BookingFilter,
  ) {
    const queryFilter = {};
    if (filter) {
      filter.byEvent && (queryFilter['byEvent'] = filter.byEvent);
    }
    const items = await endUser.bookings.matching({
      limit: pagination.per_page,
      offset: (pagination.page - 1) * pagination.per_page,
      filters: queryFilter,
    });
    const totalCount = await this.em
      .getRepository(Booking)
      .count({ endUser: endUser.id, event: filter.byEvent?.eventId });
    return new BookingResult(
      items, // data
      pagination.per_page + (pagination.page - 1) * pagination.per_page <
        totalCount, // hasNext
      (pagination.page - 1) * pagination.per_page > 0, // hasPrevious
      items.length, // pageCount
      totalCount, // totalCount
    );
  }
```
`totalCount` is the total length of the filtered collection items (thus **not** paginated) is there a better way to get this value? Otherwise I would be happy to contribute by implementing something like `Collection.matchingCount` if that could be a great idea. Which would lead to something cleaner like this:
```ts
public async getBookings(
    @Parent() endUser: EndUser,
    @Args({ name: 'pagination', type: () => PaginationInput })
    pagination: PaginationInput,
    @Args({ name: 'filter', type: () => BookingFilter, nullable: true })
    filter?: BookingFilter,
  ) {
    const queryFilter = {};
    if (filter) {
      filter.byEvent && (queryFilter['byEvent'] = filter.byEvent);
    }
    const items = await endUser.bookings.matching({
      limit: pagination.per_page,
      offset: (pagination.page - 1) * pagination.per_page,
      filters: queryFilter,
    });
    const totalCount = await endUser.bookings.matchingCount({filters: queryFilter});
    return new BookingResult(
      items,
      pagination.per_page + (pagination.page - 1) * pagination.per_page <
        totalCount,
      (pagination.page - 1) * pagination.per_page > 0,
      items.length,
      totalCount,
    );
  }
```

manu-urba:
This should close #3527.

### DIFF
--- a/packages/core/src/entity/Collection.ts
+++ b/packages/core/src/entity/Collection.ts
@@ -76,13 +76,13 @@ export class Collection<T extends object, O extends object = object> extends Arr
         this._count = count;
       }
       return count;
-    }
+    } else {
       const count = await em.count(this.property.type, this.createLoadCountCondition(where ?? {} as FilterQuery<T>, pivotMeta));
       if (!where) {
         this._count = count;
       }
-      return count;
-
+      return count; 
+    }
   }
 
   async matching<P extends string = never>(options: MatchingOptions<T, P>): Promise<Loaded<T, P>[]> {
--- a/packages/core/src/entity/Collection.ts
+++ b/packages/core/src/entity/Collection.ts
@@ -57,8 +57,8 @@ export class Collection<T extends object, O extends object = object> extends Arr
    * Gets the count of collection items from database instead of counting loaded items.
    * The value is cached, use `refresh = true` to force reload it.
    */
-  async loadCount(refresh = false): Promise<number> {
-    if (!refresh && Utils.isDefined(this._count)) {
+  async loadCount(refresh = false, cond?: FilterQuery<T>): Promise<number> {
+    if (!refresh && !cond && Utils.isDefined(this._count)) {
       return this._count!;
     }
 
@@ -67,13 +67,18 @@ export class Collection<T extends object, O extends object = object> extends Arr
 
     if (!em.getPlatform().usesPivotTable() && this.property.reference === ReferenceType.MANY_TO_MANY) {
       this._count = this.length;
+      return this._count;
     } else if (this.property.pivotTable && !(this.property.inversedBy || this.property.mappedBy)) {
-      this._count = await em.count(this.property.type, this.createLoadCountCondition({} as FilterQuery<T>, pivotMeta), { populate: [{ field: this.property.pivotEntity }] });
-    } else {
-      this._count = await em.count(this.property.type, this.createLoadCountCondition({} as FilterQuery<T>, pivotMeta));
+      const count = await em.count(this.property.type, this.createLoadCountCondition(cond ?? {} as FilterQuery<T>, pivotMeta), { populate: [{ field: this.property.pivotEntity }] });
+      if (!cond) {
+        this._count = count;
+      }
+      return count;
     }
+      const count = await em.count(this.property.type, this.createLoadCountCondition(cond ?? {} as FilterQuery<T>, pivotMeta));
+      if (!cond) { this._count = count; }
+      return count;
 
-    return this._count!;
   }
 
   async matching<P extends string = never>(options: MatchingOptions<T, P>): Promise<Loaded<T, P>[]> {
--- a/packages/core/src/entity/Collection.ts
+++ b/packages/core/src/entity/Collection.ts
@@ -75,12 +75,13 @@ export class Collection<T extends object, O extends object = object> extends Arr
         this._count = count;
       }
       return count;
+      /* eslint-disable no-else-return */
     } else {
       const count = await em.count(this.property.type, this.createLoadCountCondition(options.where ?? {} as FilterQuery<T>, pivotMeta));
       if (!options.where) {
         this._count = count;
       }
-      return count; 
+      return count;
     }
   }
 
--- a/packages/core/src/entity/Collection.ts
+++ b/packages/core/src/entity/Collection.ts
@@ -57,11 +57,10 @@ export class Collection<T extends object, O extends object = object> extends Arr
    * Gets the count of collection items from database instead of counting loaded items.
    * The value is cached, use `refresh = true` to force reload it.
    */
-  async loadCount(options: LoadCountOptions<T> | boolean = false): Promise<number> {
-    const refresh = typeof options === 'boolean' ? options : typeof options === 'object' ? options.refresh ?? false : false;
-    const { where } = options as LoadCountOptions<T>;
+  async loadCount(options: LoadCountOptions<T> | boolean = {}): Promise<number> {
+    options = typeof options === 'boolean' ? { refresh: options } : options;
 
-    if (!refresh && !where && Utils.isDefined(this._count)) {
+    if (!options.refresh && !options.where && Utils.isDefined(this._count)) {
       return this._count!;
     }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/entity/Collection.ts
```diff
@@ -57,8 +57,8 @@ export class Collection<T extends object, O extends object = object> extends Arr
    * Gets the count of collection items from database instead of counting loaded items.
    * The value is cached, use `refresh = true` to force reload it.
    */
-  async loadCount(refresh = false): Promise<number> {
-    if (!refresh && Utils.isDefined(this._count)) {
+  async loadCount(refresh = false, cond?: FilterQuery<T>): Promise<number> {
```
B4nan, line 60: let's turn this into single options parameter, keep the bool form supported as well and I will remove that in v6 branch. and lets call this option `where`

tests/issues/GH3527.test.ts
```diff
@@ -0,0 +1,93 @@
+import {
```
B4nan, line 1: issue tests are for bugs/reproductions, lets put this somewhere to `tests/features`

tests/issues/GH3527.test.ts
```diff
@@ -0,0 +1,93 @@
...
+export class EndUser {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property()
+  name!: string;
+
+  // eslint-disable-next-line @typescript-eslint/no-use-before-define
+  @OneToMany(() => Booking, booking => booking.endUser, { default: [] })
```
B4nan, line 20: FYI that `default` option does not make much sense, its not doing anything

packages/core/src/entity/Collection.ts
```diff
@@ -67,13 +67,18 @@ export class Collection<T extends object, O extends object = object> extends Arr
...
-      this._count = await em.count(this.property.type, this.createLoadCountCondition({} as FilterQuery<T>, pivotMeta), { populate: [{ field: this.property.pivotEntity }] });
-    } else {
-      this._count = await em.count(this.property.type, this.createLoadCountCondition({} as FilterQuery<T>, pivotMeta));
+      const count = await em.count(this.property.type, this.createLoadCountCondition(cond ?? {} as FilterQuery<T>, pivotMeta), { populate: [{ field: this.property.pivotEntity }] });
+      if (!cond) {
+        this._count = count;
+      }
+      return count;
     }
+      const count = await em.count(this.property.type, this.createLoadCountCondition(cond ?? {} as FilterQuery<T>, pivotMeta));
```
B4nan, line 78: this is wrongly indented

packages/core/src/entity/Collection.ts
```diff
@@ -67,13 +67,18 @@ export class Collection<T extends object, O extends object = object> extends Arr
 
     if (!em.getPlatform().usesPivotTable() && this.property.reference === ReferenceType.MANY_TO_MANY) {
       this._count = this.length;
+      return this._count;
```
B4nan, line 70:
```suggestion
      return this._count = this.length;
```

packages/core/src/entity/Collection.ts
B4nan, line 78: and i would maybe rather keep the previous structure with the final else block, it felt more readable

B4nan (changes requested):
thanks!

packages/core/src/entity/Collection.ts
manu-urba: I'm not sure if you meant this:
```ts
async loadCount(options: LoadCountOptions<T> = {}): Promise<number>

export interface LoadCountOptions<T> {
  refresh?: boolean;
  where?: FilterQuery<T>;
}
```
Or this
```ts
async loadCount(refresh = false, options: LoadCountOptions<T> = {}): Promise<number>

export interface LoadCountOptions<T> {
  where?: FilterQuery<T>;
}
```

I thought the first option would be a breaking change so please let me know if it's ok now

packages/core/src/entity/Collection.ts
```diff
@@ -57,23 +57,29 @@ export class Collection<T extends object, O extends object = object> extends Arr
    * Gets the count of collection items from database instead of counting loaded items.
    * The value is cached, use `refresh = true` to force reload it.
    */
-  async loadCount(refresh = false): Promise<number> {
-    if (!refresh && Utils.isDefined(this._count)) {
+  async loadCount(refresh = false, options: LoadCountOptions<T> = {}): Promise<number> {
```
B4nan, line 60: i meant something like this:
```suggestion
  async loadCount(options: LoadCountOptions<T> | boolean = {}): Promise<number> {
    options = typeof options === 'boolean' ? { refresh: options } : options;
```

B4nan: and i will remove the boolean variant in v6 as it will be a breaking change

packages/core/src/entity/Collection.ts
```diff
@@ -435,3 +444,7 @@ export interface InitOptions<T, P extends string = never> {
   lockMode?: Exclude<LockMode, LockMode.OPTIMISTIC>;
   connectionType?: ConnectionType;
 }
+
+export interface LoadCountOptions<T> {
+  where?: FilterQuery<T>;
```
B4nan, line 449: refresh needs to be an option too

manu-urba:
Sorry for the confusion in these commits and force pushes, something really weird kept happening with git and husky. Everything should be done now though.

packages/core/src/entity/Collection.ts
```diff
@@ -58,10 +58,10 @@ export class Collection<T extends object, O extends object = object> extends Arr
    * The value is cached, use `refresh = true` to force reload it.
    */
   async loadCount(options: LoadCountOptions<T> | boolean = false): Promise<number> {
-    const refresh = typeof options === 'boolean' ? options : typeof options === 'object';
+    const refresh = typeof options === 'boolean' ? options : typeof options === 'object' ? options.refresh ?? false : false;
```
B4nan, line 61: please do it the way i proposed, this is very overcomplicated, you dont need 3 ternaries (and nesting ternaries on its own is quite unreadable on its own)

```suggestion
    options = typeof options === 'boolean' ? { refresh: options } : options;
```

and use the `options` directly, no need to destruct anything

packages/core/src/entity/Collection.ts
```diff
@@ -57,23 +57,31 @@ export class Collection<T extends object, O extends object = object> extends Arr
    * Gets the count of collection items from database instead of counting loaded items.
    * The value is cached, use `refresh = true` to force reload it.
```
B4nan, line 58: we should mention that the value is cached only if you don't provide the `where` option

B4nan (approved):
two final nits and we can merge! 👍

packages/core/src/entity/Collection.ts
```diff
@@ -75,12 +76,13 @@ export class Collection<T extends object, O extends object = object> extends Arr
...
       }
       return count;
+      /* eslint-disable no-else-return */
     } else {
       const count = await em.count(this.property.type, this.createLoadCountCondition(options.where ?? {} as FilterQuery<T>, pivotMeta));
       if (!options.where) {
         this._count = count;
       }
-      return count; 
+      return count;
```
B4nan, line 85: the linter is correct, we should do one final refactor - since we return from all the if/elseif branches, we dont need the else branch at all.

```ts
    }

    const count = await em.count(this.property.type, this.createLoadCountCondition(options.where ?? {} as FilterQuery<T>, pivotMeta));
    if (!options.where) {
      this._count = count;
    }

    return count;
```

packages/core/src/entity/Collection.ts
```diff
@@ -56,6 +56,7 @@ export class Collection<T extends object, O extends object = object> extends Arr
   /**
    * Gets the count of collection items from database instead of counting loaded items.
    * The value is cached, use `refresh = true` to force reload it.
+   * The value CAN'T be cached if `where` option is provided.
```
B4nan, line 59:
```suggestion
   * The value is cached (unless you use the `where` option), use `refresh = true` to force reload it.
```

packages/core/src/entity/Collection.ts
manu-urba: I made that because [you suggested it](https://github.com/mikro-orm/mikro-orm/pull/3958#discussion_r1083496060), but no worries we're reverting changes 💪🏻

manu-urba:
There you go, everything should be done now. By the way thank you for detailed review and for explaining me every single thing that wasn't right.

PR #3975, opened 2023-01-27
feat(core): add cursor-based pagination via `em.findByCursor()`

B4nan:
As an alternative to the offset based pagination with `limit` and `offset`, we can paginate based on a cursor. A cursor is an opaque string that defines specific place in ordered entity graph. You can use `em.findByCursor()` to access those options. Under the hood, it will call `em.find()` and `em.count()` just like the `em.findAndCount()` method, but will use the cursor options instead.

Supports `before`, `after`, `first` and `last` options while disallowing `limit` and `offset`. Explicit `orderBy` option is required.

Use `first` and `after` for forward pagination, or `last` and `before` for backward pagination.

- `first` and `last` are numbers and serve as an alternative to `offset`, those options are mutually exclusive, use only one at a time
- `before` and `after` specify the previous cursor value, it can be one of the:
	- `Cursor` instance
	- opaque string provided by `startCursor/endCursor` properties
	- POJO/entity instance

```ts
const currentCursor = await em.findByCursor(User, {}, {
  first: 10,
  after: previousCursor, // cursor instance
  orderBy: { id: 'desc' },
});

// to fetch next page
const nextCursor = await em.findByCursor(User, {}, {
  first: 10,
  after: currentCursor.endCursor, // opaque string
  orderBy: { id: 'desc' },
});

// to fetch next page
const nextCursor2 = await em.findByCursor(User, {}, {
  first: 10,
  after: { id: lastSeenId }, // entity-like POJO
  orderBy: { id: 'desc' },
});
```

The `Cursor` object provides following interface:

```ts
Cursor<User> {
  items: [
    User { ... },
    User { ... },
    User { ... },
    ...
  ],
  totalCount: 50,
  length: 10,
  startCursor: 'WzRd',
  endCursor: 'WzZd',
  hasPrevPage: true,
  hasNextPage: true,
}
```


packages/core/src/EntityManager.ts
```diff
@@ -398,6 +398,34 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
...
+  async findByCursor<
+    Entity extends object,
+    Hint extends string = never,
+  >(entityName: EntityName<Entity>, where: FilterQuery<Entity>, options: FindOptions<Entity, Hint> = {}): Promise<Cursor<Entity, Hint>> {
+    const em = this.getContext(false);
+    entityName = Utils.className(entityName);
+    const meta = em.metadata.get<Entity>(entityName);
+
+    // ordering is required, default to order by PK
+    if (Utils.isEmpty(options.orderBy)) {
```
winsontam, line 413: how about throw error instead of help user to choose PK as default?
because if PK is uuid, the result will be look weird.

packages/core/src/utils/Cursor.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  * [Symbol.iterator](): IterableIterator<Loaded<Entity, Hint>> {
+    for (const item of this.items) {
+      yield item;
+    }
+  }
+
+  static encode<Data extends unknown[]>(value: Data): string {
+    return Buffer.from(JSON.stringify(value)).toString('base64').replace(/\+/g, '-').replace(/\//g, '_');
```
winsontam, line 63: i find that in node native way would be better
return Buffer.from(JSON.stringify(value)).toString('base64url');

packages/core/src/utils/Cursor.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  static decode<Data extends unknown[]>(value: string): Data {
+    return JSON.parse(Buffer.from(value.replace(/-/g, '+').replace(/_/g, '/'), 'base64').toString('utf8'));
```
winsontam, line 67: return JSON.parse(Buffer.from(value, 'base64url').toString('utf8'));

packages/core/src/utils/Cursor.ts
B4nan: thanks, that's nice! 

FYI your script was wrong in handling this, the [`decode` function](https://github.com/winsontam/knex-cursor/blob/main/cursor.js#L12) here replaces `+/-` too late, and ends up replacing `-` inside the JSON string, not in the base64 string

packages/core/src/EntityManager.ts
B4nan: I guess you are right, that's another good point with the uuid. I did this initially as I wanted to have defaults rather than throw, but then I realized the defaults (even if the value would be sequential) would have to work dynamically based on the cursor options probably, things started to look weird when I wrote the initial tests. 

Sidenote: maybe we should enforce the same for `em.findAndCount`?

packages/core/src/utils/Cursor.ts
B4nan: lucky that I also dropped node < 16 in v6, as this `base64url` was added in node 15 and only backported to v14 later on :]

packages/core/src/EntityManager.ts
B4nan: Maybe we could use the PK default only if it's auto-increment?

productdevbook:
https://opensource.huntersofbook.com/global/relay-cursor

Great, now I can use it directly with graphql and the library I wrote.

Thank you Martin

bakura10:
This is a wonderful feature. Just a quick feedback @B4nan : in most API I have used with cursor based, the totalCount is not returned (what we are interested is only knowing if there is a next or prev page), which avoids a potentially costly SQL count. What about removing the count by default, and add an optional `{includeCount: true}` option to the cursor based pagination, for users who need the count?

B4nan:
I am not sure if this should be disabled by default, but I am open to having an option to control it.

PR #4016, opened 2023-02-06
feat(migrations): add `cjs` option for emit

### LINKED ISSUE #4005: Add option to emit js migrations with `.cjs` extension
**Problem:**
[Certain environments](https://nodejs.org/api/packages.html#determining-module-system) will interpret `.js` files as ES6 modules by default. This means that the migrations generated by Mikro ORM in CommonJS form won't work out of the box, and will either require a package.json with `"type": "commonjs"` in the migrations folder (to convert that folder into cjs), or to rename the files so that they have the `.cjs` extension.

**Solution:**
It would be ideal if there were an option to set the extension that Mikro ORM assigns to the emitted migrations. Either there could be a `cjs` option for the `migrations.emit` configuration property, or there could just be a new property like `migrations.fileExtension` that can be overridden to any arbitrary value.

I tried adding a `migrations.fileName` function to do this renaming, but it still appends `.js` to the end.

**Alternatives:**
If users of Mikro ORM in such an environment just need to have a package.json file in the migrations folder, it would be nice if the tooling generated that file for you when creating the first migration.

**Additional context:**
This came up for me when using Mikro ORM with [SvelteKit](https://kit.svelte.dev/).

erie0210:
In case @stabai doesn't reply, I'd like to submit my PR.

Closes #4005

### DIFF
--- a/packages/core/src/utils/Configuration.ts
+++ b/packages/core/src/utils/Configuration.ts
@@ -88,7 +88,7 @@ export class Configuration<D extends IDatabaseDriver = IDatabaseDriver> {
     migrations: {
       tableName: 'mikro_orm_migrations',
       path: './migrations',
-      glob: '!(*.d).{js,ts}',
+      glob: '!(*.d).{js,ts,cjs}',
       transactional: true,
       disableForeignKeys: true,
       allOrNothing: true,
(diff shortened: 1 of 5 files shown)


B4nan (approved):
Thanks! 

Let's adjust the default `glob` too, so it respects `cjs` files automatically, otherwise looking good.

B4nan:
and actually another one - do the same for the `migrations-mongodb` package, the code is unfortunately duplicated (otherwise we would have to have some internal packages, had it this way before and it was causing another sort of issues)

B4nan (approved):

Issue #3277, opened 2022-07-05
More flexible Loaded<> typing


parisholley:
Say I have the following model:

```typescript
class Parent {
  @ManyToOne(() => Child)
  child: IdentifiedReference<Child>;
}

class Child {
  @ManyToOne(() => Grandchild)
  grandchild: IdentifiedReference<Grandchild>;
}

class Grandchild {

}
```

If I have a function that looks like the following:

```typescript
function preloaded(parent:Loaded<Parent, 'child'>){

}

const parent1 = manager.findOne(Parent, 1, {populate: ['child']});

// works
preloaded(parent1);

const parent2 = manager.findOne(Parent, 1, {populate: ['child.grandchild']});

// typescript complains
preloaded(parent2);
```

I would expect to be able to pass both instances of parent since they at-least have the child populated. 
**Additional context**
Typescript: 4.5.5

B4nan:
Not reproducible, I dont see any TS errors in that code. At least on master, not sure what version you are on.

<img width="451" alt="image" src="https://user-images.githubusercontent.com/615580/177942320-234cc1cf-591d-496d-ab3d-d279ad899c25.png">

Let's reopen with a failing repro, this sounds like a bug to me, but clearly its not as easy as you described it.

parisholley:
added PR to reproduce this issue

B4nan:
That's weird, it complains about assignability in the wrong direction.

B4nan:
This is quite tricky, spent the better half of the day trying to solve it, but I feel desperate. One of the issues I haven't found a way around: the serialized forms of the relations will be different - those are also strictly typed and respect the `Loaded` type. So the `Collection.toJSON` method will have a different return type for `Loaded<E, 'col'>` and `Loaded<E, never>`, those are not assignable to each other in any way.

```ts
const foo: Collection<Loaded<Vehicle>> = {} as Collection<Loaded<Vehicle, 'type'>>;

TS2322: Type 'Collection<Loaded<Vehicle, "type">, object>' is not assignable to type 'Collection<Loaded<Vehicle, never>, object>'.
   The types returned by 'toJSON()' are incompatible between these types.
     Type 'EntityDTO<Loaded<Vehicle, "type">>[]' is not assignable to type 'EntityDTO<Loaded<Vehicle, never>>[]'.
[...]

PR #4037, opened 2023-02-10
V6: Broken loaded use-case #3277

parisholley:
The fixes in V6 for #3277 solve a portion of my population conflicts but I've run into this particular problem in most of my instances. When a property with the same name exists at the same "dot level", the compiler cannot handle it. As you can see in my PR reproduction, `event5` compiles just fine as `calendar.foobar` is not defined, but because both `calendar.owner.foobar` and `order.customer.foobar` both exist, `event1` / `event2` does not work.

### DIFF
--- a/tests/types.test.ts
+++ b/tests/types.test.ts
@@ -595,7 +595,7 @@ describe('check typings', () => {
     q = { [OptionalProps]: 'bar' };
   });
 
-  test('GH #3277', async () => {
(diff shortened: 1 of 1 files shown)


B4nan:
Looks like this is actually easy fix, the `Reference.set` method needs the same `TT` type arg extension as the others.

B4nan:
Fixed via 936f21f

PR #4066, opened 2023-02-23
fix(mongo): move $fulltext from $and to top level

### LINKED ISSUE #4065: Fulltext search breaks in mongodb when filters are enabled.
* With **mongodb** driver
* if query contains `$fulltext`
* and EntityManager applies filters

then error is thrown.

**Stack trace**
```
Error: Full text search is only supported on the top level of the query object.',
        '    at MongoDriver.renameFields (/app/node_modules/@mikro-orm/mongodb/MongoDriver.js:118:19)',
        '    at MongoDriver.find (/app/node_modules/@mikro-orm/mongodb/MongoDriver.js:23:22)',
        '    at MongoEntityManager.find (/app/node_modules/@mikro-orm/core/EntityManager.js:112:41)',
        '    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)',
```

**To Reproduce**
A simple test case
```
const god = new Author('God', 'hello@heaven.god');
const bible = new Book('Bible', god);
await orm.em.persistAndFlush(bible);
orm.em.clear();

const booksRepository = orm.em.getRepository(Book);
orm.em.addFilter('BreakFulltext', {title: {$re: '.*'}}, Book, true)
const books = await booksRepository.find({ $fulltext: 'Bible' })
```
**Expected behavior**
It should be possible to use full text search and use filters with Mongo driver.
Working example with Postgre:
```
const god = new Author2('God', 'hello@heaven.god');
const bible = new Book2('Bible', god);
await orm.em.persistAndFlush(bible);
orm.em.clear();

const booksRepository = orm.em.getRepository(Book2);
orm.em.addFilter('BreakFulltext', {title: {$re: '.*'}}, Book2, true);
const books = await booksRepository.find({ title: {$fulltext: 'Bible'} });
```

I propose following: 
* if query has `$and` in top level and `$fulltext` in any of its elements, move it to top level
* if there are several `$fulltext` in `$and` or top level already has one throw an error

funduck:
Closes #4065

### DIFF
--- a/packages/mongodb/src/MongoDriver.ts
+++ b/packages/mongodb/src/MongoDriver.ts
@@ -138,7 +138,7 @@ export class MongoDriver extends DatabaseDriver<MongoConnection> {
 
   private renameFields<T>(entityName: string, data: T, where = false): T {
     // copy to new variable to prevent changing the T type or doing as unknown casts
-    const copiedData: T & { $fulltext?: string; $text?: { $search: string }; $and?: {$fulltext?: string}[] } = Object.assign({}, data); // copy first
+    const copiedData: Dictionary = Object.assign({}, data); // copy first
     Utils.renameKey(copiedData, 'id', '_id');
     const meta = this.metadata.find(entityName);
 
--- a/packages/mongodb/src/MongoDriver.ts
+++ b/packages/mongodb/src/MongoDriver.ts
@@ -138,12 +138,27 @@ export class MongoDriver extends DatabaseDriver<MongoConnection> {
 
   private renameFields<T>(entityName: string, data: T, where = false): T {
     // copy to new variable to prevent changing the T type or doing as unknown casts
-    const copiedData: T & { $fulltext?: string; $text?: { $search: string } } = Object.assign({}, data); // copy first
+    const copiedData: T & { $fulltext?: string; $text?: { $search: string }; $and?: {$fulltext?: string}[] } = Object.assign({}, data); // copy first
     Utils.renameKey(copiedData, 'id', '_id');
     const meta = this.metadata.find(entityName);
 
     if (meta) {
       this.inlineEmbeddables(meta, copiedData, where);
+    }
+
+    // If we had a query with $fulltext and some filter we end up with $and with $fulltext in it.
+    // We will try to move $fulltext to top level.
+    if (copiedData.$and) {
+      for (let i = 0; i < copiedData.$and.length; i++) {
+        const and = copiedData.$and[i];
+        if ('$fulltext' in and) {
+          if ('$fulltext' in copiedData) {
+            throw new Error('Cannot merge multiple full text searches to top level of the query object.');
+          }
+          copiedData.$fulltext = and.$fulltext!;
+          delete and.$fulltext;
+        }
+      }
     }
 
     // move search terms from data['$fulltext'] to mongo's structure: data['$text']['search']
--- a/packages/mongodb/src/MongoDriver.ts
+++ b/packages/mongodb/src/MongoDriver.ts
@@ -153,7 +153,7 @@ export class MongoDriver extends DatabaseDriver<MongoConnection> {
         const and = copiedData.$and[i];
         if ('$fulltext' in and) {
           if ('$fulltext' in copiedData) {
-            throw new Error('Cannot merge multiple full text searches to top level of the query object.');
+            throw new Error('Cannot merge multiple $fulltext conditions to top level of the query object.');
           }
           copiedData.$fulltext = and.$fulltext!;
           delete and.$fulltext;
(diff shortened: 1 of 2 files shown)


packages/mongodb/src/MongoDriver.ts
```diff
@@ -138,14 +138,29 @@ export class MongoDriver extends DatabaseDriver<MongoConnection> {
 
   private renameFields<T>(entityName: string, data: T, where = false): T {
     // copy to new variable to prevent changing the T type or doing as unknown casts
-    const copiedData: T & { $fulltext?: string; $text?: { $search: string } } = Object.assign({}, data); // copy first
+    const copiedData: T & { $fulltext?: string; $text?: { $search: string }; $and?: {$fulltext?: string}[] } = Object.assign({}, data); // copy first
```
B4nan, line 141: this would introduce merge conficts with v6, lets do it the same [as there](https://github.com/mikro-orm/mikro-orm/blob/v6/packages/mongodb/src/MongoDriver.ts#L170) right ahead

```suggestion
    const copiedData: Dictionary = Object.assign({}, data); // copy first
```

packages/mongodb/src/MongoDriver.ts
```diff
@@ -138,14 +138,29 @@ export class MongoDriver extends DatabaseDriver<MongoConnection> {
...
     }
 
+    // If we had a query with $fulltext and some filter we end up with $and with $fulltext in it.
+    // We will try to move $fulltext to top level.
+    if (copiedData.$and) {
+      for (let i = 0; i < copiedData.$and.length; i++) {
+        const and = copiedData.$and[i];
+        if ('$fulltext' in and) {
+          if ('$fulltext' in copiedData) {
+            throw new Error('Cannot merge multiple full text searches to top level of the query object.');
```
B4nan, line 156: lets mention the `$fulltext` operator directly here instead of "full text searches"

packages/mongodb/src/MongoDriver.ts
```diff
@@ -138,14 +138,29 @@ export class MongoDriver extends DatabaseDriver<MongoConnection> {
...
+          }
+          copiedData.$fulltext = and.$fulltext!;
+          delete and.$fulltext;
+        }
+      }
+    }
```
B4nan, line 162: can't say i like the solution, but i don't know any better right now. note that you introduced 4 levels of branching, you will need ~4 tests to cover what this does, not just one. 

also lets move this into a private method (we should do the same with the rest of the fulltext code, this method is crazy long and complex)

B4nan (changes requested):
thanks!

packages/mongodb/src/MongoDriver.ts
```diff
@@ -138,14 +138,29 @@ export class MongoDriver extends DatabaseDriver<MongoConnection> {
...
     }
 
+    // If we had a query with $fulltext and some filter we end up with $and with $fulltext in it.
+    // We will try to move $fulltext to top level.
+    if (copiedData.$and) {
+      for (let i = 0; i < copiedData.$and.length; i++) {
+        const and = copiedData.$and[i];
+        if ('$fulltext' in and) {
+          if ('$fulltext' in copiedData) {
+            throw new Error('Cannot merge multiple full text searches to top level of the query object.');
```
B4nan, line 156:
```suggestion
            throw new Error('Cannot merge multiple $fulltext conditions to top level of the query object.');
```

packages/mongodb/src/MongoDriver.ts
funduck: Oh, I see you fixed those things by yourself. I was going to, but then started thinking about that limitation "Full text search is only supported on the top level of the query object", why is it so?
And it looks like mongodb (I checked 4.2.6 and latest) allows `$text` to appear in nested conditions. You can check it with these tests
```
  test('positive $text in nested conditions', async () => {
    const booksRepository = orm.em.getRepository(Book);
    const books = await booksRepository.find({ $or: [
      {$text: {$search: 'Bible'} },
      {title: {$regex: 'Something'} },
    ]} as unknown as FilterQuery<Book>);
    expect(books.length).toBe(1);
  })

  test('negative $text in nested conditions', async () => {
    const booksRepository = orm.em.getRepository(Book);
    const books = await booksRepository.find({ $or: [
      {$text: {$search: 'Something'} },
      {title: {$regex: 'Bible'} },
    ]} as unknown as FilterQuery<Book>);
    expect(books.length).toBe(1);
  })
```
So I thought is it really necessary to forbid `$fulltext` in nested objects?
Of course there are some limitations to use of `$text`, for example in `$or` mongo won't allow non-indexed fields together with `$text` and this kind of query will fail
```
{$or: [{$text:{$search: 'some full text request'}}, {'someNonIndexedField': 42}]}
``` 
If you think it's a good idea, I could work on that

B4nan: This came in as a community PR, feel free to improve it, the limitation indeed sounds quite artificial.

PR #4079, opened 2023-02-27
fix(postgres): use quoted schema+table name when dropping constraints

patricksmith:
I noticed that when I was updating a column with an Enum type using an entity with a schema defined, the migration generated attempted to drop the existing constraint but failed due to a quoting issue. The migration was generated with a line like:

```sql
alter table "my_schema.test" drop constraint if exists "test_column_check";
```

instead of 

```sql
alter table "my_schema"."test" drop constraint if exists "test_column_check";
```

I've added a test that I believe exercises the functionality.

### DIFF
--- a/packages/postgresql/src/PostgreSqlSchemaHelper.ts
+++ b/packages/postgresql/src/PostgreSqlSchemaHelper.ts
@@ -266,26 +266,28 @@ export class PostgreSqlSchemaHelper extends SchemaHelper {
 
   getPreAlterTable(tableDiff: TableDifference, safe: boolean): string {
     const ret: string[] = [];
+    const quoted = (val: string) => this.platform.quoteIdentifier(val);
 
     const parts = tableDiff.name.split('.');
     const tableName = parts.pop()!;
     const schemaName = parts.pop();
     /* istanbul ignore next */
     const name = (schemaName && schemaName !== this.platform.getDefaultSchemaName() ? schemaName + '.' : '') + tableName;
+    const quotedName = quoted(name);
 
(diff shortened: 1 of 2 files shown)


B4nan (approved):
thanks!

packages/postgresql/src/PostgreSqlSchemaHelper.ts
```diff
@@ -266,26 +266,28 @@ export class PostgreSqlSchemaHelper extends SchemaHelper {
 
   getPreAlterTable(tableDiff: TableDifference, safe: boolean): string {
     const ret: string[] = [];
+    const quoted = (val: string) => this.platform.quoteIdentifier(val);
```
B4nan, line 269: we can inline this
```suggestion
```

packages/postgresql/src/PostgreSqlSchemaHelper.ts
```diff
@@ -266,26 +266,28 @@ export class PostgreSqlSchemaHelper extends SchemaHelper {
...
 
     const parts = tableDiff.name.split('.');
     const tableName = parts.pop()!;
     const schemaName = parts.pop();
     /* istanbul ignore next */
     const name = (schemaName && schemaName !== this.platform.getDefaultSchemaName() ? schemaName + '.' : '') + tableName;
+    const quotedName = quoted(name);
```
B4nan, line 276:
```suggestion
    const quotedName = this.platform.quoteIdentifier(name);
```

B4nan:
How can such change decrease the code coverage, weird...

Issue #4060, opened 2023-02-20
Infer property type from its default value if not set explicitly


snigdha920:
**Describe the bug**

When we define a field with a default value, without explicitly specifying the type - either in TypeScript or in the `@Property` decorator, we get the following error while to create an entity:

Entity:

```
@Entity()
export class Author {
  @PrimaryKey()
  _id!: ObjectId;

  @SerializedPrimaryKey()
  public id!: string;

  @Property()
  name: string;

  @Property({ default: 0 })
  bestsellers = 0;

  constructor(name: string) {
    this.name = name;
  }
}
```

Error: 

```
"Trying to set Author.bestsellers of type 'string' to '0' of type 'number'"
```

**Stack trace**
```
"ValidationError: Trying to set Author.bestsellers of type 'string' to '0' of type 'number'",
            "    at Function.fromWrongPropertyType (/Users/snigdhasingh/nestjs-mikro-subscriptions-context/node_modules/.pnpm/@mikro-orm+core@5.6.11_@mikro-orm+mongodb@5.6.11/node_modules/@mikro-orm/core/errors.js:22:16)",
            "    at EntityValidator.validateProperty (/Users/snigdhasingh/nestjs-mikro-subscriptions-context/node_modules/.pnpm/@mikro-orm+core@5.6.11_@mikro-orm+mongodb@5.6.11/node_modules/@mikro-orm/core/entity/EntityValidator.js:65:44)",
            "    at /Users/snigdhasingh/nestjs-mikro-subscriptions-context/node_modules/.pnpm/@mikro-orm+core@5.6.11_@mikro-orm+mongodb@5.6.11/node_modules/@mikro-orm/core/entity/EntityValidator.js:24:35",
            "    at Array.forEach (<anonymous>)",
            "    at EntityValidator.validate (/Users/snigdhasingh/nestjs-mikro-subscriptions-context/node_modules/.pnpm/@mikro-orm+core@5.6.11_@mikro-orm+mongodb@5.6.11/node_modules/@mikro-orm/core/entity/EntityValidator.js:13:20)",
            "    at ChangeSetComputer.computeChangeSet (/Users/snigdhasingh/nestjs-mikro-subscriptions-context/node_modules/.pnpm/@mikro-orm+core@5.6.11_@mikro-orm+mongodb@5.6.11/node_modules/@mikro-orm/core/unit-of-work/ChangeSetComputer.js:38:28)",
[...]

PR #4081, opened 2023-03-01
feat(core): infer property type from its default value

erie0210:
Related to https://github.com/mikro-orm/mikro-orm/issues/4060

We talked about setting the type value in `initDefaultValue` before, but I think it's better to infer the type of the default value from the part that sets the type, so I implemented it this way. Can you give me some feedback on this PR?

### DIFF
--- a/packages/core/src/metadata/ReflectMetadataProvider.ts
+++ b/packages/core/src/metadata/ReflectMetadataProvider.ts
@@ -20,13 +20,26 @@ export class ReflectMetadataProvider extends MetadataProvider {
     // as we were mapping it to UnknownType which is a string. This is to prevent defaulting to JSON
     // column type, which can be often hard to revert and cause hard to understand issues with PKs.
     if (prop.kind === ReferenceKind.SCALAR && type === Object) {
-      type = String;
+      type = this.getTypeFromDefault(prop);
     }
 
     prop.type = type.name;
 
     if (prop.type && ['string', 'number', 'boolean', 'array', 'object'].includes(prop.type.toLowerCase())) {
       prop.type = prop.type.toLowerCase();
+    }
+  }
+
+  private getTypeFromDefault(prop: EntityProperty): any {
+    switch (typeof prop.default) {
+      case 'number':
+        return Number;
+      case 'boolean':
+        return Boolean;
+      case 'object':
+      case 'string':
+      default:
+        return String;
     }
   }
 
--- a/tests/features/schema-generator/type-infering-from-default.test.ts
+++ b/tests/features/schema-generator/type-infering-from-default.test.ts
@@ -1,0 +1,36 @@
+import { Entity, ManyToOne, PrimaryKey, Property } from '@mikro-orm/core';
+import { MikroORM } from '@mikro-orm/better-sqlite';
+
+@Entity()
+export class Author {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property({ default: 0 })
+  age=0;
+
+}
+
+test('infer property type from its default value when type is not set', async () => {
+  // given
+  const orm = await MikroORM.init({
+    entities: [Author],
+    dbName: `:memory:`,
+  });
+  await orm.schema.refreshDatabase();
+
+  const author = new Author();
+  const expected = 'number';
+
+  // when
+  await orm.em.persistAndFlush(author);
+  orm.em.clear();
+
+  // then
+  const result = await orm.em.findOneOrFail(Author, author.id);
+  const actual =Object.getPrototypeOf(result).__meta.properties.age.type;
+  // expect(actual).toBe(expected);
+
+  await orm.close(true);
+});


tests/issues/GH4060.test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+describe('GH4060', () => {
+
+  let orm: MikroORM<MongoDriver>;
+
+  beforeAll(async () => {
+    orm = await MikroORM.init({
+      entities: [Author],
+      clientUrl: await initMongoReplSet('mikro-orm-test'),
+      driver: MongoDriver,
```
B4nan, line 36: please do not use mongo for this, it is the slowest driver to start, especially if you use replica sets (which you use here for no reason, you don't test anything transaction related)

any test that is not driver specific should prefer sqlite in-memory database as it is the fastest and does not require any cleanup.

tests/issues/GH4060.test.ts
```diff
@@ -0,0 +1,60 @@
...
+import { MongoDriver, ObjectId } from '@mikro-orm/mongodb';
+import { closeReplSets, initMongoReplSet } from '../bootstrap';
+
+@Entity()
+class Author {
+
+  @PrimaryKey()
+  _id!: ObjectId;
+
+  @Unique()
```
B4nan, line 11: drop the indexes/uniques, that is also irrelevant

tests/issues/GH4060.test.ts
```diff
@@ -0,0 +1,60 @@
+import { Entity, MikroORM, PrimaryKey, Property, Unique } from '@mikro-orm/core';
```
B4nan, line 1: this is a new feature, the test should go to features folder (and have meaningful name), issues folder is for reproductions - the fact that this has been reported as a bug is irrelevant, it was never meant to be working before.

packages/core/src/metadata/EntitySchema.ts
```diff
@@ -335,6 +335,10 @@ export class EntitySchema<T = any, U = never> {
       type = type.toLowerCase();
     }
 
+    if (type === 'string' && options.default !== undefined) {
```
B4nan, line 338: this is wrong, we dont want to override the type if it is `string`, we want to infer the value when it is not set. `string` is the default, but your code does not distinguish default property type (string) from a user-provided one (which can be also a string).

B4nan (changes requested):

B4nan:
> We talked about setting the type value in initDefaultValue before, but I think it's better to infer the type of the default value from the part that sets the type, so I implemented it this way. Can you give me some feedback on this PR?

Why do you think it's better that way?

B4nan:
Also this should go to v6 branch, not master.

erie0210:
> Why do you think it's better that way?

I thought it as "infering and setting type" and wondered if a typifying function would be more appropriate.

B4nan:
what is "a typifying function"? :]

packages/core/src/metadata/MetadataDiscovery.ts
```diff
@@ -949,6 +949,10 @@ export class MetadataDiscovery {
     }
 
     prop.defaultRaw = typeof val === 'string' ? `'${val}'` : '' + val;
+
+    if (prop.default !== undefined && prop.type !== typeof prop.default) {
```
B4nan, line 953: In the end I don't think this is a good idea either, the type might be already something more specific that uses string runtime values. 

Best would be to handle this in the reflect-metadata provider, as that is the place where the default string type is used when we can't infer anything.

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/metadata/ReflectMetadataProvider.ts#L22-L24

Please check if it does not happen too early, e.g. before we infer the defaults from property initializers (so before this method is executed).

erie0210: Thanks, I was looking for the part where type is set to string as the default value, I'll work on that and get back to you!

packages/core/src/metadata/ReflectMetadataProvider.ts
```diff
@@ -20,7 +20,11 @@ export class ReflectMetadataProvider extends MetadataProvider {
     // as we were mapping it to UnknownType which is a string. This is to prevent defaulting to JSON
     // column type, which can be often hard to revert and cause hard to understand issues with PKs.
     if (prop.kind === ReferenceKind.SCALAR && type === Object) {
-      type = String;
+      if (prop.default!==undefined) {
```
B4nan: `null` is also problematic + spacing
```suggestion
      if (prop.default != null) {
```

and i'd move this logic into the private method + rename it

packages/core/src/metadata/ReflectMetadataProvider.ts
```diff
@@ -20,14 +20,27 @@ export class ReflectMetadataProvider extends MetadataProvider {
...
+      case 'number':
+        return Number;
+      case 'boolean':
+        return Boolean;
+      case 'object':
+      case 'string':
+      default:
+        return String;
+    }
+  }
```
B4nan, line 44: this can be greatly simplified and we should be smarter here, e.g. support at least buffers and dates - there is `UtilsgetObjectType()` you can use to get the type from such objects

```suggestion
  private getWrapperTypeObject(value: unknown): string {
    const type = Utils.getObjectType(value);
    return type === 'object' ? 'string' : type;
  }
```

please add more tests for this (can be just more properties in the test you have

tests/features/schema-generator/type-infering-from-default.test.ts
```diff
@@ -0,0 +1,36 @@
...
+import { MikroORM } from '@mikro-orm/better-sqlite';
+
+@Entity()
+export class Author {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property({ default: 0 })
+  age=0;
```
B4nan, line 11:
```suggestion
  age = 0;
```

please follow this convention, there is more places with this 

and please dont rebase once you get a review, i cant see what changes you make

erie0210: Current lint didn't work in my environment, and it looks like adding this rule works.
Or is there other rule that should have worked?

```
"space-infix-ops": ["error", {"int32Hint": true}]
```

B4nan: yes, but without the `int32Hint` option i guess

B4nan: https://github.com/mikro-orm/mikro-orm/commit/428e41f8b69155cb41c1ba6ebd93358653d78ed9

B4nan: dont forget to test this works with the inferred defaults:

```ts
  @Property()
  age = 0;
```

this needs to work too. i'd test numbers, strings, dates, buffers, booleans, and also some objects to test the fallback, as i am still not sure about that, if the default is object type, we might want to respect that and infer JSON property type

erie0210:
Thanks for the review @B4nan 🙂. How long do you think we should do this for? I was wondering if you had a deadline in mind.

B4nan:
Not sure I follow?

erie0210:
I asked if there was a deadline for this task, since I work a 9-6 job during the week and would only be able to work on it on the weekends! @B4nan

B4nan:
Not really, I might rebase the branch so worst case you will have some git issues, but v6 is still months ahead of us, so no rush.

B4nan:
Do you need some help with this? I keep getting notifications about rebases but I see no actual changes being made. 

FYI I will probably rebase the v6 branch during the week, as there are many new bug fixes in master. It will probably break your PR, I am kinda postponing it because of it.

erie0210:
I understood about the rebase and its aftermath. Thanks for letting me know in advance.
Made some changes base on your review but still working on it. 
Think I won't be available until this weekend. If so, should this PR better be a draft?

packages/core/src/metadata/ReflectMetadataProvider.ts
erie0210: I want to know more about `null` being problemtic and why you used `prop.default != null`? 🤓
I thought if there's no default value, `prop` won't have `default` field and will be `undefined`.
I considered using `if(!prop.default)` instead of `undefined` or `null`

tests/features/schema-generator/type-infering-from-default.test.ts
erie0210: Sure, I will include types you mentioned. Thanks :)

packages/core/src/metadata/ReflectMetadataProvider.ts
B4nan: you can get `default === null` and that needs the same treatment as `undefined`, doing `!= null` will catch both `null` and `undefined`. 

you dont want `!prop.default` as that would also catch empty string or zero, both valid default values.

tests/features/schema-generator/type-infering-from-default.test.ts
erie0210: The reflect-metadata library is not inferring the default value from the line below like ts-morph does.
```let type = Reflect.getMetadata('design:type', meta.prototype, prop.name);```
Is there a way to get default value from reflect-metadata, or do you have any idea on this?

B4nan: The whole point of this PR is to support that, because `reflect-metadata` is not capable of that.

erie0210: Okay. So we have two problems. One is to infer type from `PropertyOptions`, which we can do it in `ReflectMetadataProvider`. The Other one is to infer type when assign default value with `=`. It seem a little tricky. Will it be appropriate to infer when creating `MetadataStorage.metadata`?

B4nan: Not really, I'd say there is just one problem. I guess I will just take over if you don't mind, maybe the problem is harder than I envisioned.

erie0210: Okay. Thank you.

B4nan: It's indeed more complicated than I thought. The problem itself was rather easy (the trick is to move the `initDefaultValue` call before the metadata provider so the inference happens early, before the default string type is added), but when I started adjusting existing entities to test this there too, many tests were failing because of connected issues, like the `ref` helper not working with null values, causing constructors to fail.

PR #4114, opened 2023-03-10
fix: use schema-qualified table names for auto-increment and comment statements in postgresql

### LINKED ISSUE #4108: @Property comment with PostgreSQL driver generates invalid SQL
**Describe the bug**

When using `comment` in `@Property` annotations for a postgresql database, the generated SQL only uses the table name, and not the fully qualified schema name and table name (e.g. `comment on "table_name"."column_name" is \'hello world\';` instead of `comment on "schema_name"."table_name"."column_name" is \'hello world\';`). This causes migrations to fail when the schema is not public

**Stack trace**
example of a migration when `comment` is used on an entity property.
```
MigrationError: Migration Migration20230307174746 (up) failed: Original error: comment on column "table_name"."column_name" is 'Hello world'; - relation "table_name" does not exist
    at /Users/yeedle/project/node_modules/umzug/src/umzug.ts:259:12
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Umzug.runCommand (/Users/yeedle/project/node_modules/umzug/src/umzug.ts:208:11)
    at async Migrator.runInTransaction (/Users/yeedle/project/node_modules/@mikro-orm/migrations/Migrator.js:299:21)
    at async PostgreSqlConnection.transactional (/Users/yeedle/project/node_modules/@mikro-orm/knex/AbstractSqlConnection.js:36:25)
    at async Function.handleFreshCommand (/Users/yeedle/project/node_modules/@mikro-orm/cli/commands/MigrationCommandFactory.js:153:9)
    at async Function.handleMigrationCommand (/Users/yeedle/project/node_modules/@mikro-orm/cli/commands/MigrationCommandFactory.js:88:17) {
  cause: TableNotFoundException: comment on column "table_name"."column_name" is 'Hello world'; - relation "table_name" does not exist
```

**To Reproduce**
1. Add a `comment` to an entity `@Property`
2. run `migration:create`
3. run `migration:up`

**Expected behavior**
Generated SQL should include the table schema so that migrations succeed

**Additional context**
I'm happy to make a PR fixing this issue for Postgresql.

**Versions**

| Dependency | Version |
| - | - |
| node | 16.14.2 |
| typescript | 1.22.15 |
| mikro-orm | 5.6.8 |
| your-driver | PostgreSQL |

yeedle:
When using comment in @Property annotations for a postgresql database, the generated SQL only uses the table name, and not the fully qualified schema name and table name (e.g. `comment on "table_name"."column_name" is \'hello world\';`). This PR changes this to instead generate fully qualified table names (`comment on "schema_name"."table_name"."column_name" is \'hello world\';`). 

Closes #4108

### DIFF
--- a/packages/knex/src/schema/SchemaGenerator.ts
+++ b/packages/knex/src/schema/SchemaGenerator.ts
@@ -388,7 +388,7 @@ export class SchemaGenerator extends AbstractSchemaGenerator<AbstractSqlDriver> 
           continue; // will be handled via knex
         }
 
-        this.helper.pushTableQuery(table, this.helper.getChangeColumnCommentSQL(tableName, column));
+        this.helper.pushTableQuery(table, this.helper.getChangeColumnCommentSQL(tableName, column, schemaName));
       }
 
       for (const [oldColumnName, column] of Object.entries(diff.renamedColumns)) {
--- a/packages/postgresql/src/PostgreSqlSchemaHelper.ts
+++ b/packages/postgresql/src/PostgreSqlSchemaHelper.ts
@@ -292,26 +292,30 @@ export class PostgreSqlSchemaHelper extends SchemaHelper {
     return ret.join(';\n');
   }
 
-  getAlterColumnAutoincrement(tableName: string, column: Column): string {
+  getAlterColumnAutoincrement(tableName: string, column: Column, schemaName?: string): string {
     const ret: string[] = [];
     const quoted = (val: string) => this.platform.quoteIdentifier(val);
+    const schemaReference = (schemaName !== undefined && schemaName !== 'public') ? ('"' + schemaName + '".') : '';
+    const tableReference = schemaReference + tableName;
 
     /* istanbul ignore else */
     if (column.autoincrement) {
       const seqName = this.platform.getIndexName(tableName, [column.name], 'sequence');
       ret.push(`create sequence if not exists ${quoted(seqName)}`);
-      ret.push(`select setval('${seqName}', (select max(${quoted(column.name)}) from ${quoted(tableName)}))`);
-      ret.push(`alter table ${quoted(tableName)} alter column ${quoted(column.name)} set default nextval('${seqName}')`);
+      ret.push(`select setval('${seqName}', (select max(${quoted(column.name)}) from ${quoted(tableReference)}))`);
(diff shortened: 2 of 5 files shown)


packages/postgresql/src/PostgreSqlSchemaHelper.ts
```diff
@@ -292,26 +292,30 @@ export class PostgreSqlSchemaHelper extends SchemaHelper {
     return ret.join(';\n');
   }
 
-  getAlterColumnAutoincrement(tableName: string, column: Column): string {
+  getAlterColumnAutoincrement(tableName: string, column: Column, schemaName?: string): string {
     const ret: string[] = [];
     const quoted = (val: string) => this.platform.quoteIdentifier(val);
+    const schemaReference = (schemaName !== undefined && schemaName !== 'public') ? ('"' + schemaName + '".') : '';
```
B4nan, line 298: the `quoted` method right above this line is what you should use for quoting, no need to manually wrap things with `"`. The `quoteIdentifier` even supports dot paths, so you can feed in the whole `schema.table` string and get `"schema"."table"` back.

packages/knex/src/schema/SchemaGenerator.ts
```diff
@@ -388,7 +388,7 @@ export class SchemaGenerator extends AbstractSchemaGenerator<AbstractSqlDriver>
           continue; // will be handled via knex
         }
 
-        this.helper.pushTableQuery(table, this.helper.getChangeColumnCommentSQL(tableName, column));
+        this.helper.pushTableQuery(table, this.helper.getChangeColumnCommentSQL(tableName, column, schemaName));
```
B4nan, line 391: dont forget to pass the argument also to the `getAlterColumnAutoincrement` method on line 383

and looking at the failed tests, you will need to adjust some of the existing assertions, as they did not expect the quoting.

B4nan:
Also looking at the failed tests, the quoting is apparently not working as expected, the failed assertions are because of the missing quotes.

```
    - "comment on column "book"."id" is 'this is primary''s key';
    + "comment on column book."id" is 'this is primary''s key';
    - comment on column "book"."name" is 'this is name of book';
    + comment on column book."name" is 'this is name of book';
```

I am sure using the platform methods I suggested will resolve this.

B4nan:
Closing in favour of #4123

PR #4126, opened 2023-03-13
fix(core): respect explicit `unsigned` option

PenguinToast:
Users don't necessarily want all numeric primary key columns to be unsigned.

Alternatively, we can change the behavior such that _only_ autoincrement pkeys are unsigned and just let numeric pkeys be signed by default, but that seems like a more major behavior change.

### DIFF
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -1152,10 +1152,7 @@ export class MetadataDiscovery {
       return;
     }
 
-    prop.unsigned = (prop.primary || prop.unsigned)
-      && prop.unsigned !== false  // Allow explicit opt-out of unsigned for primary keys
-      && this.isNumericProperty(prop)
-      && this.platform.supportsUnsigned();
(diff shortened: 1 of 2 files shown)


packages/core/src/metadata/MetadataDiscovery.ts
```diff
@@ -1152,7 +1152,10 @@ export class MetadataDiscovery {
       return;
     }
 
-    prop.unsigned = (prop.primary || prop.unsigned) && this.isNumericProperty(prop) && this.platform.supportsUnsigned();
+    prop.unsigned = (prop.primary || prop.unsigned)
+      && prop.unsigned !== false  // Allow explicit opt-out of unsigned for primary keys
+      && this.isNumericProperty(prop)
+      && this.platform.supportsUnsigned();
```
B4nan, line 1158: won't this work the same? if not, let's keep your version. 

```suggestion
    prop.unsigned ??= (prop.primary || prop.unsigned) && this.isNumericProperty(prop) && this.platform.supportsUnsigned();
```

i'd consider this as a bug, the option is public and it does not work when explicitly used

PenguinToast: ah, yeah that definitely works -- that's much cleaner, thank you!

PR #4155, opened 2023-03-21
fix(migrations-mongodb): use loggerFactory

alexandre-abrioux:
I've noticed that `@mikro-orm/migrations` does not use the logger defined through `config.loggerFactory`, but instead uses the logger defined in `config.logger`. This is an attempt to fix that.

### DIFF
--- a/packages/migrations-mongodb/src/Migrator.ts
+++ b/packages/migrations-mongodb/src/Migrator.ts
@@ -83,7 +83,7 @@ export class Migrator implements IMigrator {
       migrations,
     });
 
-    const logger = this.config.get('logger');
+    const logger = (message: string) => this.config.getLogger().log('info', message);
     this.umzug.on('migrating', event => logger(`Processing '${event.name}'`));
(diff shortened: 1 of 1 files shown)


B4nan:
This was done rather on purpose, your change would disable logging in the CLI which we definitely don't want.

If we want to run the migration logging through the ORM logger, it should have its own logger namespace and that one will be enabled automatically inside the CLI context.

alexandre-abrioux:
Got it! FYI we do not use the CLI to run migrations but use the ORM directly. Migrations are run automatically on the application bootstrap. Feel free to close this PR if you feel that it doesn't fit the standard workflow.

rubiin:
I think this can be closed

rubiin:
I think this can be closed @B4nan

alexandre-abrioux:
FYI this issue is now fixed with the release of v6.3.11: relevant commit is https://github.com/mikro-orm/mikro-orm/commit/e845c26b74a85948249e72dc75817a58e0cf7fd6 (see https://github.com/mikro-orm/mikro-orm/discussions/6039).

PR #4188, opened 2023-03-31
fix: mistake in database names

artemaniskin:


B4nan:
this is actually correct, you can use them both

PR #4195, opened 2023-04-06
v6 docs updates

SpencerKaiser:
- Adding a warning icon to the upgrade notes about typescript
- Separating `.load` into a dedicated section for clarity
- Removing stale references to `BaseEntity`

image

### DIFF
--- a/docs/docs/decorators.md
+++ b/docs/docs/decorators.md
@@ -248,16 +248,16 @@ Many instances of the current Entity refer to One instance of the referred Entit
 
 See [Defining Entities](relationships.md#manytoone) for more examples.
 
-| Parameter    | Type | Optional | Description                                                                                                                                               |
-|--------------|------|----------|-----------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `entity`     | `string` &#124; `() => EntityName` | yes | Set target entity type.                                                                                                                                   |
-| `cascade`    | `Cascade[]` | yes | Set what actions on owning entity should be cascaded to the relationship. Defaults to `[Cascade.PERSIST, Cascade.MERGE]` (see [Cascading](cascading.md)). |
-| `eager`      | `boolean` | yes | Always load the relationship.                                                                                                                             |
-| `inversedBy` | `(string & keyof T) ` &#124; ` (e: T) => any` | yes | Point to the inverse side property name.                                                                                                                  |
-| `ref`        | `boolean` | yes | Wrap the entity in [`Reference` wrapper](type-safe-relations.md).                                                                                         |
-| `primary`    | `boolean` | yes | Use this relation as primary key.                                                                                                                         |
+| Parameter    | Type | Optional | Description                                                                                                                                                |
+|--------------|------|----------|------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `entity`     | `string` &#124; `() => EntityName` | yes | Set target entity type.                                                                                                           |
+| `cascade`    | `Cascade[]` | yes | Set what actions on owning entity should be cascaded to the relationship. Defaults to `[Cascade.PERSIST, Cascade.MERGE]` (see [Cascading](cascading.md)).|
+| `eager`      | `boolean` | yes | Always load the relationship. (Discouraged for use with to-many relations.)                                                                                 |
+| `inversedBy` | `(string & keyof T) ` &#124; ` (e: T) => any` | yes | Point to the inverse side property name.                                                                               |
+| `ref`        | `boolean` | yes | Wrap the entity in [`Reference` wrapper](type-safe-relations.md).                                                                                          |
+| `primary`    | `boolean` | yes | Use this relation as primary key.                                                                                                                          |
 | `deleteRule`   | `string` | yes | [Referential integrity](cascading.md#declarative-referential-integrity).                                                                                  |
-| `updateRule` | `string` | yes | [Referential integrity](cascading.md#declarative-referential-integrity).                                                                                  |
+| `updateRule` | `string` | yes | [Referential integrity](cascading.md#declarative-referential-integrity).                                                                                    |
 
 ```ts
 @ManyToOne()
@@ -282,7 +282,7 @@ See [Defining Entities](relationships.md#onetoone) for more examples, including 
 |-----------------|-----------------------------------------------|----------|------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `entity`        | `string` &#124; `() => EntityName`            | yes      | Set target entity type.                                                                                                                                    |
 | `cascade`       | `Cascade[]`                                   | yes      | Set what actions on owning entity should be cascaded to the relationship. Defaults to `[Cascade.PERSIST, Cascade.MERGE]` (see [Cascading](cascading.md)).  |
-| `eager`         | `boolean`                                     | yes      | Always load the relationship.                                                                                                                              |
+| `eager`         | `boolean`                                     | yes      | Always load the relationship. (Discouraged for use with to-many relations.)                                                                                |
 | `owner`         | `boolean`                                     | yes      | Explicitly set as owning side (same as providing `inversedBy`).                                                                                            |
 | `inversedBy`    | `(string & keyof T) ` &#124; ` (e: T) => any` | yes      | Point to the inverse side property name.                                                                                                                   |
 | `mappedBy`      | `(string & keyof T)` &#124; `(e: T) => any`   | yes      | Point to the owning side property name.                                                                                                                    |
@@ -323,7 +323,7 @@ See [Defining Entities](relationships.md#onetomany) for more examples, including
 | `mappedBy`          | `(string & keyof T)` &#124; `(e: T) => any` | no       | Point to the owning side property name.                                                                                                                   |
 | `entity`            | `string` &#124; `() => EntityName`          | yes      | Set target entity type.                                                                                                                                   |
 | `cascade`           | `Cascade[]`                                 | yes      | Set what actions on owning entity should be cascaded to the relationship. Defaults to `[Cascade.PERSIST, Cascade.MERGE]` (see [Cascading](cascading.md)). |
-| `eager`             | `boolean`                                   | yes      | Always load the relationship.                                                                                                                             |
+| `eager`             | `boolean`                                   | yes      | Always load the relationship. (Discouraged for use with to-many relations.)                                                                               |
 | `orphanRemoval`     | `boolean`                                   | yes      | Remove the entity when it gets disconnected from the connection (see [Cascading](cascading.md#orphan-removal)).                                           |
 | `orderBy`           | `{ [field: string]: QueryOrder }`           | yes      | Set default ordering condition.                                                                                                                           |
 | `joinColumn`        | `string`                                    | yes      | Override default database column name on the owning side (see [Naming Strategy](naming-strategy.md)).                                                     |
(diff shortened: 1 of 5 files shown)


B4nan:
i just finished rebasing v6, sry about that

SpencerKaiser:
Sorry about the mess of commits; I'm not really sure what I can do on my end to clean up that history aside from making a new branch. If you'd prefer that, let me know, but I imagine you're squash merging anyways.

B4nan:
Reset the changes, stash it and create new branch from v6.

B4nan:
Or in the end, I will be squash merging, so as long as I can do that, it should be fine.

docs/docs/type-safe-relations.md
```diff
@@ -163,6 +162,22 @@ If you use different metadata provider than `TsMorphMetadataProvider` (e.g. `Ref
...
+
+After retrieving a reference, you can load the full entity by utilizing the asynchronous `Reference.load()` method.
+
+```ts
+const book1 = await em.findOne(Book, 1);
+(await book1.author.load()).name; // async safe access
+
+const book2 = await em.findOne(Book, 2);
+const author = await book2.author.load();
+author.name; // sync safe access
```
B4nan, line 175: this is not really sync-safe access - that is the `$` property from `Loaded` type. those two examples are doing exactly the same, both using the return value of the `ref.load()` method

SpencerKaiser: Yea I struggled with the verbiage there; what I was trying to highly was that you can save a copy of the newly loaded entity... if you'd prefer to have only one of the two, I'd advocate for the second as I think it's clearer. Thoughts?

SpencerKaiser:
I just deleted the local branch, created a new one from your `v6`, cherry-picked, and force pushed... all clean now 😬

SpencerKaiser:
@B4nan FYI I added a new section; I just tried to create a migrations after upgrading as a sanity check and it tried to generate a migration switching everything from `timestampz(0)` to `timestampz`. If that was intentional (I'm assuming it was), there's now a comment in the upgrading doc.

docs/docs/upgrading-v5-to-v6.md
```diff
@@ -231,6 +231,9 @@ await em.flush();
 console.log(ref.age); // real value is available after flush
 ```
 
+## Changed default `columnType` for `clock_timestamp`
+If you were previously using `clock_timestamp` as a `defaultRaw` value, you will need to specify `columnType: 'timestampz(0)'` or the next migration creation will convert columns to `timestampz`.
```
B4nan, line 235: Can you provide some details for this? First time hearing about `clock_timestamp`, but it sounds weird that a `defaultRaw` would affect column type anyhow. What changed in v6 is the runtime default (property initializer) is used to infer the database default as well as column type. But only if you don't provide your own options, and `defaultRaw` in particular shouldn't affect this at all.

> you will need to specify `columnType: 'timestampz(0)'` or the next migration creation will convert columns to `timestampz`.

This has nothing to do with the defaults you set, and indeed, this changed - postgres no longer defaults to 0 length for datetimes. #3832

SpencerKaiser: It's something my buddy introduced me to, it basically takes the time on the db server when you insert the record... which I'm just now realizing is a Postgres specific thing.

Here was what my entity used to have:
```ts
export abstract class Node<T extends AnyEntity> extends BaseEntity<T, 'id'> {
// ...
@Property({ defaultRaw: 'clock_timestamp()' })
public createdAt!: Date;
// ...
}
```

On 5.x I was up to date and running `mikro-orm migration:create` resulted in no changes. After the upgrade (and without changing anything) the same command triggered a migration which targeted all the tables based on entities that extend my `Node` class. The migration wanted to change every column from `timestampz(0)` to `timestampz`. Setting the `columnType` explicitly resulted in a no-op when I ran the migration command again.

Maybe the issue is what `Date` translates to from a `columnType` perspective? Like whatever you're using to do that changed?

B4nan: There is no issue, you dont have to explain it more :) Have you checked the PR I linked? The default datetime length in postgres changed, thats all.

SpencerKaiser: Ah ok cool, and nope I missed the PR. Just looked; want me to update the above to reflect what was changed in that PR so there's *some* mention of it?

B4nan: Yeah should be mentioned in there for sure, it just needs to be correct - it is not about `default` or `defaultRaw` options, it is about the default length in postgres, which is now aligned with how postgres behaves instead of how other drivers behave. 

I guess I will add some hooks to the discovery so this can be easily overridden globally.

SpencerKaiser: Honestly until you get feedback, I wouldn't bother... I didn't want the extra migration, but all it does it replace a couple of zeros with other numbers; the structure should be exactly the same. Honestly not even sure how/why that differentiation exists 😬 I'll re-word and feel free to tweak it after if I misused a term or something

B4nan: https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp.280.29_or_timestamptz.280.29

SpencerKaiser: Yea I saw that from the PR you mentioned; until someone asks for that explicitly, I personally wouldn't bother haha I'm actually going to revert my change in my project and migrate to the correct column type.

My changes have been pushed; take a look and let me know what you think

docs/docs/upgrading-v5-to-v6.md
```diff
@@ -231,8 +231,12 @@ await em.flush();
 console.log(ref.age); // real value is available after flush
 ```
 
-## Changed default `columnType` for `clock_timestamp`
-If you were previously using `clock_timestamp` as a `defaultRaw` value, you will need to specify `columnType: 'timestampz(0)'` or the next migration creation will convert columns to `timestampz`.
+## Changed default PostgreSQL column type for `Date`
+Using a type of `Date` would result in an inferred column type of `timestampz(0)` which is not the default for PostgreSQL and is actually [not recommended](https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp.280.29_or_timestamptz.280.29). After upgrading, be prepared for a migration which will change the column type of relevant columns to `timestampz`, which is the default. Should you chose to continue using `timestampz(0)`, simply add a `columnType` to your properties:
+
+```ts
+@Property({ columnType: 'timestamptz(0)' })
```
B4nan, line 238: maybe we should suggest this instead?
```suggestion
@Property({ length: 0 })
```

SpencerKaiser: If that does the same thing, absolutely; I just wasn't aware that had an impact. I'll try it locally and make that change if it works as expected

docs/docs/decorators.md
```diff
@@ -248,16 +248,16 @@ Many instances of the current Entity refer to One instance of the referred Entit
...
-| `primary`    | `boolean` | yes | Use this relation as primary key.                                                                                                                         |
+| Parameter    | Type | Optional | Description                                                                                                                                                |
+|--------------|------|----------|------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `entity`     | `string` &#124; `() => EntityName` | yes | Set target entity type.                                                                                                           |
+| `cascade`    | `Cascade[]` | yes | Set what actions on owning entity should be cascaded to the relationship. Defaults to `[Cascade.PERSIST, Cascade.MERGE]` (see [Cascading](cascading.md)).|
+| `eager`      | `boolean` | yes | Always load the relationship. (Discouraged for use with large collections)                                                                                 |
```
B4nan, line 255: maybe better this way? i would really discourage it for any to-many relations, simply because any collection can get large over time, turning this into a timebomb basically. 

```suggestion
| `eager`      | `boolean` | yes | Always load the relationship. (Discouraged for use with to-many relations.)                                                                                 |
```

docs/docs/upgrading-v5-to-v6.md
```diff
@@ -231,6 +231,13 @@ await em.flush();
 console.log(ref.age); // real value is available after flush
 ```
 
+## Changed default PostgreSQL column type for `Date`
+Using a type of `Date` would result in an inferred column type of `timestampz(0)` which is not the default for PostgreSQL and is actually [not recommended](https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp.280.29_or_timestamptz.280.29). After upgrading, be prepared for a migration which will change the column type of relevant columns to `timestampz`, which is the default. Should you chose to continue using `timestampz(0)`, simply add a `columnType` to your properties:
```
B4nan, line 235:
```suggestion
## Changed default PostgreSQL `Date` mapping precision

Previosly, all drivers defaulted the `Date` type mapping to a timestamp with 0 precision (so seconds). This is [discouraged in PostgreSQL](https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp.280.29_or_timestamptz.280.29), and is no longer valid - the default mapping without the `length` property being explicitly set is now `timestamptz`, which stores microsecond precision, so equivalent to `timestampz(6)`.

To revert back to the v5 behavior, you can either set the `columnType: 'timestampz(0)'`, or use `length: 0`:
```

docs/docs/upgrading-v5-to-v6.md
SpencerKaiser: @B4nan are you ready to resolve this thread?

docs/docs/decorators.md
```diff
@@ -248,16 +248,16 @@ Many instances of the current Entity refer to One instance of the referred Entit
...
-| `primary`    | `boolean` | yes | Use this relation as primary key.                                                                                                                         |
+| Parameter    | Type | Optional | Description                                                                                                                                                |
+|--------------|------|----------|------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `entity`     | `string` &#124; `() => EntityName` | yes | Set target entity type.                                                                                                           |
+| `cascade`    | `Cascade[]` | yes | Set what actions on owning entity should be cascaded to the relationship. Defaults to `[Cascade.PERSIST, Cascade.MERGE]` (see [Cascading](cascading.md)).|
+| `eager`      | `boolean` | yes | Always load the relationship. (Discouraged for use with to-many relations.)                                                                                 |
```
B4nan, line 255: this change should go to the other places too, its there 3 more times

B4nan (approved):
one last nit, and we can merge 👍

docs/docs/decorators.md
SpencerKaiser: Yep, made it locally... also caught a typo in your other suggestion and fixed the space formatting for that table after the change; waiting to push but asked a question in Slack

SpencerKaiser, line 255: @B4nan do the periods there bug you? If so, I can fix

docs/docs/upgrading-v5-to-v6.md
```diff
@@ -8,7 +8,7 @@ title: Upgrading from v5 to v6
 
 Support for older node versions was dropped. 
 
-## TypeScript 4.9+ required
+## ⚠️ TypeScript 4.9+ required
```
B4nan, line 11: one last thing, let's be progressive, i think i might need some TS 5 features like the `const` generics modifier

```suggestion
## ⚠️ TypeScript 5.0+ required
```

i am also thinking about requiring node 18 right ahead, as 16 will reach eol of half a year (afaik it should have happened much sooner, but postponed). which reminds me that i need to try out v20

PR #4202, opened 2023-04-08
fix: correct a typo in install docs

ZuBB:


B4nan:
this is not a typo

ZuBB:
are they interchangeable?

B4nan:
Yes

PR #4205, opened 2023-04-08
fix(core): prevent tuple type properties from being converted to array when serializing entities

Anytinz:
This commit fixes an issue where properties with tuple type were being converted to array
when entities were serialized. The fix ensures that tuple type properties are correctly
preserved during the serialization process.


B4nan (approved):
Thanks!

PR #4209, opened 2023-04-10
fix(core): update relationship when manually specifying PK on inverse side

Anytinz:
When creating an entity with many-to-one or one-to-one relationship on the inverse side, if the primary key is manually specified before flush(), the owning side's corresponding relationship definition will not be updated. This is because the manually specified primary key entity does not have a __identifier attribute, and the ChangeSetComputer.processToOne() method only adds a payload containing the updated relationship definition when the __identifier attribute is present. In this situation, if the owning side is only expected to update the relationship definition in this flush(), it will not trigger the onUpdate event, as the event relies on the changeSet, and an empty payload changeSet will not be added to the corresponding queue.

### DIFF
--- a/tests/features/unit-of-work/manual-pk-inverse-relation.test.ts
+++ b/tests/features/unit-of-work/manual-pk-inverse-relation.test.ts
@@ -1,13 +1,13 @@
 /* eslint-disable @typescript-eslint/naming-convention */
-import { Collection, Entity, ManyToOne, OneToMany, PrimaryKey, Rel, SimpleLogger } from '@mikro-orm/core';
-import { MikroORM, ObjectId, defineConfig } from '@mikro-orm/mongodb';
+import { Collection, Entity, ManyToOne, OneToMany, PrimaryKey, Rel } from '@mikro-orm/core';
+import { MikroORM } from '@mikro-orm/sqlite';
 import { v4 } from 'uuid';
 
 @Entity()
 class Foo {
 
   @PrimaryKey()
-  _id!: ObjectId;
+  _id!: number;
 
   @ManyToOne({ entity: () => Bar1, nullable: true })
   bar1?: Rel<Bar1>;
@@ -18,16 +18,13 @@ class Foo {
   @ManyToOne({ entity: () => Bar3, nullable: true })
   bar3?: Rel<Bar3>;
 
-  @ManyToOne({ entity: () => Bar4, nullable: true })
-  bar4?: Rel<Bar4>;
-
 }
 
 @Entity()
 class Bar1 {
 
   @PrimaryKey()
-  _id!: ObjectId;
+  _id!: number;
 
   @OneToMany({ entity: () => Foo, mappedBy: book => book.bar1 })
   foos = new Collection<Foo>(this);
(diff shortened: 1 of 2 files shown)


tests/features/unit-of-work/manual-pk-inverse-relation.test.ts
```diff
@@ -0,0 +1,138 @@
...
+    loggerFactory: options => new SimpleLogger(options),
+    debug: true,
+  }));
+});
+
+beforeEach(async () => {
+  await orm.schema.refreshDatabase();
+  foo = orm.em.create(Foo, {});
+  await orm.em.flush();
+  orm.em.clear();
```
B4nan, line 92: `foo` is detached entity, you need to reload it again after clearing, otherwise the whole test does not make much sense - such detached entity should be considered as new.

```suggestion
  foo = orm.em.create(Foo, {});
  await orm.em.flush();
//  orm.em.clear(); either this should be commented, or you need `em.findOne` call after it
```

tests/features/unit-of-work/manual-pk-inverse-relation.test.ts
```diff
@@ -0,0 +1,138 @@
...
+    entities: [
+      Foo,
+      Bar1,
+      Bar2,
+      Bar3,
+      Bar4,
+    ],
+    clientUrl: 'mongodb://db',
+    loggerFactory: options => new SimpleLogger(options),
+    debug: true,
```
B4nan, line 84:
```suggestion
```

tests/features/unit-of-work/manual-pk-inverse-relation.test.ts
```diff
@@ -0,0 +1,138 @@
...
+  @OneToMany({ entity: () => Foo, mappedBy: book => book.bar4 })
+  public foos = new Collection<Foo>(this);
+
+}
+
+let orm: MikroORM;
+let foo: Foo;
+
+beforeAll(async () => {
+  orm = await MikroORM.init(defineConfig({
```
B4nan, line 74: no need for `defineConfig` here

tests/features/unit-of-work/manual-pk-inverse-relation.test.ts
```diff
@@ -0,0 +1,138 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import { Collection, Entity, ManyToOne, OneToMany, PrimaryKey, Rel, SimpleLogger } from '@mikro-orm/core';
+import { MikroORM, ObjectId, defineConfig } from '@mikro-orm/mongodb';
+import { v4 } from 'uuid';
+
+@Entity()
+class Foo {
+
+  @PrimaryKey()
+  public _id!: ObjectId;
```
B4nan, line 10: dont use `public` modifier, its implicit

tests/features/unit-of-work/manual-pk-inverse-relation.test.ts
```diff
@@ -0,0 +1,138 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import { Collection, Entity, ManyToOne, OneToMany, PrimaryKey, Rel, SimpleLogger } from '@mikro-orm/core';
+import { MikroORM, ObjectId, defineConfig } from '@mikro-orm/mongodb';
```
B4nan, line 3: i would prefer sqlite with in memory database for tests that are not driver specific, as its the fastest

B4nan (changes requested):
Your change is breaking 123 tests... :]

Anytinz:
I have made the changes based on your suggestions.

Anytinz:
> 您的更改正在打破 123 项测试... ：]

I'm sorry..... 😢

Anytinz:
Hi @B4nan , when can we merge this request?

B4nan:
As I said, this breaks many tests, so not mergable at all.

B4nan:
So the problem was actually elsewhere, your problem was specific to using `em.create()` which uses the hydrator under the hood - and hydration did have disabled propagation of changes, or better say under hydration context the changes were considered as already saved in the database.

Fixed via 30778c35667d275bf4407a762feb1a92d91f7fe2

PR #4210, opened 2023-04-10
docs: fix wrongly named prop in polymorphic embed page

JonParton:
Looks like there was a typo after copy and pasting one of the examples from above!

Just scoping out the library and learning so sorry if I missed anything else!

### DIFF
--- a/docs/versioned_docs/version-5.6/embeddables.md
+++ b/docs/versioned_docs/version-5.6/embeddables.md
@@ -733,7 +733,7 @@ export const OwnerSchema = new EntitySchema({
   properties: {
     id: { primary: true, type: 'number' },
     name: { type: 'string' },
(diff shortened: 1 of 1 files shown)


B4nan (changes requested):
thanks!

the decorator here also shoudnt be there

https://github.com/mikro-orm/mikro-orm/pull/4210/files#diff-583adf89d040146349c9f71612131f1f86e2170abd4400ad0abc47820f3c7632R693

and please update the next docs too, in the `docs/docs` folder

JonParton:
Did think that was wrong as well but not knowing the library fully I wasn't 100% sure! 😂

I'll take a look when I get chance!

JonParton:
Sorry I never got to this @B4nan!! 🙈

Thanks for correcting! 👍

B4nan:
No worries, and thanks for reporting :)

PR #4233, opened 2023-04-17
feat(core): add customizable `LoggerContext` with labeling support

### LINKED ISSUE #4230: Add context to config `logger` property
When logging information from the ORM, I would like to be able to see additional context surrounding the query

I would like to be able to see additional information for a query when one is made; I want to see:
- The calling function (and/or file)
- The query in clean JSON format (instead of the aggressive alias format in queries which isn't super readable)

I doubt you want the extra details to get pumped into `console` so maybe an `advancedLogger` property on the config that has those extra properties and the other logger will be overridden if that property is provided.

Adding log commands after queries have executed

**Additional context**
I tried adding my own little custom log commands AFTER queries (if you try before there's no way to guarantee it'll print with the related query if others are being executed at the same time) but it's kind of messy and isn't super scalable...

```ts
/*
namespace Express {
  export interface Request {
    entityManager: EntityManager<PostgreSqlDriver>;
  }
}
*/

// `req` comes from the express handler

const author = await req.entityManager
  .fork()
  .findOne(
    Author,
    { id: 1 },
  );

if (env.dbLoggingEnabled) {
  logger.debug(`\tvia Author Middleware: ${req.originalUrl}\n`);
}
```

SpencerKaiser:
Fixes #4230

### DIFF
--- a/docs/docs/logging.md
+++ b/docs/docs/logging.md
@@ -84,6 +84,7 @@ return MikroORM.init({
 ```
 
 ### Using a custom `LoggerFactory`
+
 If you want more control over what is logged and how, use the `loggerFactory` option in your config and either extend the `DefaultLogger` class or make your `Logger` from scratch:
 
 #### Extending `DefaultLogger`
--- a/docs/docs/logging.md
+++ b/docs/docs/logging.md
@@ -107,8 +107,6 @@ return MikroORM.init({
 });
 ```
 
-
-
 #### Creating a custom logger from scratch
 We can use `loggerFactory` and use our own implementation of the `Logger` interface:
 
--- a/docs/docs/logging.md
+++ b/docs/docs/logging.md
@@ -86,11 +86,11 @@ return MikroORM.init({
 
 ### Using a custom `LoggerFactory`
 
-If you want more control over what is logged and how, use the `loggerFactory` option in your config and either extend the `DefaultLogger` class or make your `Logger` from scratch:
+If you want more control over what is logged and how, use the `loggerFactory` option in your config and extend the `SimpleLogger` class, extend the `DefaultLogger` class, or make your `Logger` from scratch:
 
-#### Extending `DefaultLogger`
+#### Extending `DefaultLogger` or `SimpleLogger`
 
-You can extend the `DefaultLogger` instead of implementing everything from scratch. It is also exported from the `@mikro-orm/core` package:
+You can extend the `DefaultLogger` or `SimpleLogger` instead of implementing everything from scratch. `DefaultLogger` and `SimpleLogger` are both exported from the `@mikro-orm/core` package with `SimpleLogger` being colorless.
 
 ```ts
 class CustomLogger extends DefaultLogger {
@@ -107,6 +107,13 @@ return MikroORM.init({
   debug: true,
   loggerFactory: (options) => new CustomLogger(options),
 });
+```
+
+To use `SimpleLogger` instead, simply replace `DefaultLogger` in the example above: 
+```ts
+class CustomLogger extends SimpleLogger {
+  // ...
+}
 ```
 
 #### Creating a custom logger from scratch
--- a/docs/docs/logging.md
+++ b/docs/docs/logging.md
@@ -32,7 +32,7 @@ It is also useful for debugging problems with entity discovery, as you will see 
 
 ## Disabling colored output
 
-To disable colored output, we can use multiple environment variables:
+To disable colored output, you can use multiple environment variables:
 
 - `NO_COLOR`
 - `MIKRO_ORM_NO_COLOR`
@@ -75,7 +75,7 @@ Several customization options exist to allow for style changes or custom logic.
 
 ### Using a custom logger
 
-We can provide our own logger function via the `logger` option:
+You can provide your own logger function via the `logger` option:
 
 ```ts
 return MikroORM.init({
@@ -89,7 +89,8 @@ return MikroORM.init({
 If you want more control over what is logged and how, use the `loggerFactory` option in your config and either extend the `DefaultLogger` class or make your `Logger` from scratch:
 
 #### Extending `DefaultLogger`
-We can extend the `DefaultLogger` instead of implementing everything from scratch. It is also exported from the `@mikro-orm/core` package:
+
+You can extend the `DefaultLogger` instead of implementing everything from scratch. It is also exported from the `@mikro-orm/core` package:
 
 ```ts
 class CustomLogger extends DefaultLogger {
@@ -109,7 +110,8 @@ return MikroORM.init({
 ```
 
 #### Creating a custom logger from scratch
-We can use `loggerFactory` and use our own implementation of the `Logger` interface:
+
+You can use `loggerFactory` and use your own implementation of the `Logger` interface:
 
 ```ts
 import { Logger, LoggerOptions, MikroORM, Configuration } from '@mikro-orm/core';
--- a/docs/docs/logging.md
+++ b/docs/docs/logging.md
@@ -122,3 +122,14 @@ MikroORM.init({
 ```
 
 For MongoDB you can use `MongoHighlighter` from `@mikro-orm/mongo-highlighter` package.
+
+## Query Labels
+
+It may often be beneficial to log the origin of a query when using [`EntityManager.find`](entity-manager.md#fetching-entities-with-entitymanager) or [`EnmtityManager.findOne`](entity-manager.md#fetching-entities-with-entitymanager) for debugging and redundancy elimination purposes. 
+
+An optional `loggerContext` option can be included within the `FindOptions` parameter of either call which will add a label to the output when debug mode is enabled.
+
+```ts
+const author = await em.findOne(Author, { id: 1 }, { loggerContext: { label: 'Author Retrieval - /authors/me' } });
+// [query] (Author Retrieval - /authors/me) select "a0".* from "Author" as "a0" where "a0"."id" = 1 limit 1 [took 21 ms]
+```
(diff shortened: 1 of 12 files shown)


SpencerKaiser:
@B4nan I’m having a really difficult time tracking how data passes through the different abstract classes and where the context actually gets passed to `logger.logQuery`.

The changes to `AbstractSqlDriver` and `QueryBuilder` are the changes **_I_** think I need to get that data to `logQuery`, but not sure. I tried to test locally by bringing those changes into a real project but `yarn link` was seemingly breaking everything due to some strange types issue… I built the project, ran the link command from inside `core`, made sure all my other dependencies were on `58`, and successfully got it linked into my other, but I get weird type errors around `Loaded` entities: 
`Types have separate declarations of a private property 'entity'`

Any idea what I’m doing wrong there? Any suggestions on the approach?

B4nan (commented):
i am usually going the old school way, compile the package and just copy paste the dist folder inside node_modules, maybe that will work? the error sounds like you have some private property with different type, maybe you need to link more packages than just core? as you did changes in knex package too

tests/Logger.test.ts
```diff
@@ -1,43 +1,111 @@
-import { Configuration, DefaultLogger } from '@mikro-orm/core';
+import { Configuration, DefaultLogger, SimpleLogger, colors } from '@mikro-orm/core';
 
-describe('Logger', () => {
+// Remove colors to allow for text matching in response
```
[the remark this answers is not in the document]
SpencerKaiser: Resolved [via Slack](https://mikroorm.slack.com/archives/C053MQN67D0/p1681829077736189); keeping it as is.

packages/core/src/logging/Logger.ts
```diff
@@ -50,3 +51,5 @@ export interface LoggerOptions {
   highlighter?: Highlighter;
   usesReplicas?: boolean;
 }
+
+export type LoggerContext = Pick<LogContext, 'label'>;
```
B4nan, line 55: maybe better this way so its universal way to pass params down to the logger?

```suggestion
export type LoggerContext = { label?: string } & Dictionary;
```

and add a note about that to the section about custom loggers

packages/knex/src/query/QueryBuilder.ts
```diff
@@ -608,7 +610,7 @@ export class QueryBuilder<T extends object = AnyEntity> {
     }
 
     const type = this.connectionType || (method === 'run' ? 'write' : 'read');
-    const res = await this.driver.getConnection(type).execute(query.sql, query.bindings as any[], method, this.context);
```
B4nan, line 611: the same needs to be supported in mongo driver

tests/Logger.test.ts
```diff
@@ -1,43 +1,111 @@
...
-    expect(mock.mock.calls.length).toBe(0);
-    logger.log('query', 'test query msg');
-    expect(mock.mock.calls.length).toBe(1);
-    logger.error('query', 'test error msg');
-    expect(mock.mock.calls.length).toBe(2);
-    logger.warn('query', 'test warning msg');
-    expect(mock.mock.calls.length).toBe(3);
+  describe('SimpleLogger', () => {
+
+        test('should print correctly without a label', () => {
```
B4nan, line 93: bad indent (not just here, maybe we need to improve the eslint config, iirc this might be already fixed in master so maybe just rebase)

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -843,9 +844,18 @@ export abstract class AbstractSqlDriver<Connection extends AbstractSqlConnection
...
+    const qb = new QueryBuilder<T>(
+      entityName,
+      this.metadata,
+      this,
+      ctx,
+      undefined,
+      connectionType,
+      undefined,
+      loggerContext,
+    );
```
B4nan, line 858: side note: we should turn (most of) this into options object parameter in v6 (not in this PR)

B4nan (changes requested):

packages/knex/src/AbstractSqlDriver.ts
SpencerKaiser: 100% Agreed... I actually started down the path of doing that but didn't want to make that decision without getting your feedback and figured it would be a bigger change than the scope of this PR should be 😬

B4nan:
Will you have time to do the mongo support? We can merge without it too, I'll be happy to do it myself. 

Planning to ship v5.7 tomorrow probably, and want to rebase v6 again afterward, so would be good to merge this before instead of battling the rebases again :]

SpencerKaiser:
@B4nan sorry for the delay, I went on a mini vacation for the weekend + today; I'll work on it here shortly and/or tomorrow. I might have a question or two about it, but if I do I'll follow up in Slack for expediency! 

Re: rebasing; feel free to rebase main, just don't pull it into this branch... when that happens, I'll see the "Update branch" button and know to reset/stash/delete/recreate 😬

tests/Logger.test.ts
SpencerKaiser: Fixed, but fwiw using `> Format Document` doesn't fix this, so yea, the eslint config might be a bit off

packages/core/src/logging/Logger.ts
SpencerKaiser: This change was also needed for `LogContext`... we might consider refactoring these at some point... one acts as a sort of public type where the other is more focused on internals unless someone uses `LoggerFactory`.

SpencerKaiser: Also I personally prefer picking the `label` just to keep as much overlap as possible. If you still want that to be unique on it's own, let me know and I can fix it

SpencerKaiser:
@B4nan just added what I _**think**_ are the only changes needed for Mongo, but I'm not able to test tonight (don't have a quick repo to test against and I'm calling it for the night).

Also I made some pretty substantial changes to the logging docs because I felt they were a little all over the place after adding more so I just reorganized. Hope you like it; if not, let me know and I can revert back to the old structure but with the new content.

packages/knex/src/query/QueryBuilder.ts
SpencerKaiser: See my comment below 😬

packages/core/src/logging/Logger.ts
```diff
@@ -50,3 +51,12 @@ export interface LoggerOptions {
   highlighter?: Highlighter;
   usesReplicas?: boolean;
 }
+
+/**
+ * Context for a logger to utilize to format output, including a label and additional properties that can be accessed by custom loggers
+ *
+ * @example
+ * await em.findOne(User, 1, { loggerContext: { label: 'user middleware' } };
+ * // [query] (user middleware) select * from user where id = 1;
```
B4nan, line 60: this should be a codeblock i guess?

docs/docs/logging.md
```diff
@@ -41,7 +83,33 @@ return MikroORM.init({
 });
 ```
 
-If we want to have more control over logging, we can use `loggerFactory` and use our own implementation of the `Logger` interface:
+### Using a custom `LoggerFactory`
```
B4nan, line 86: keep a blank line around headings (all of them)

```suggestion
### Using a custom `LoggerFactory`

```

docs/docs/logging.md
```diff
@@ -41,7 +83,33 @@ return MikroORM.init({
...
+}
+
+return MikroORM.init({
+  debug: true,
+  loggerFactory: (options) => new CustomLogger(options),
+});
+```
+
+
+
```
B4nan, line 110: should be just a single blank line here

B4nan (commented):
few nits before i have time to review this fully

thanks for the docs improvements, that's amazing!

packages/core/src/logging/Logger.ts
SpencerKaiser: I'm really not sure how to do those the proper way... the way above looks correct in VSCode: 
image

And the other way (with a code block) doesn't... 
image

Is the syntax you use a requirement for the auto-generated docs, maybe?

docs/docs/logging.md
SpencerKaiser: Whoops, good catch... I had to disable my auto-formatting because what I use for other projects is _**very**_ different from yours (I don't even like mine but at least it's consistent with format-on-save haha)

SpencerKaiser:
Also re: tests, looks like they're all failing due to this error: 
```[Error: No driver specified, please fill in the `driver` option or use `defineConfig` helper (to define your ORM config) or `MikroORM` class (to call the `init` method) exported from the driver package (e.g. `import { defineConfig } from '@mikro-orm/mysql'; export defineConfig({ ... })`).]```

But I didn't change anything related to the CLI 🤔 thoughts?

File: packages/core/package.json [hunk 9]
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mikro-orm/core",
-  "version": "5.6.16",
+  "version": "6.0.0-dev.57",
```
B4nan, line 3: this should be reverted, same for change in knex package

B4nan (commented):
the tests here are failing because you have out of sync lock file, probably because of the version change i mentioned

packages/core/src/logging/Logger.ts
SpencerKaiser: Ah ok sounds good!

SpencerKaiser:
Test are passing locally again! Should be ready for that through review once you have time 🙂 no rush though!

docs/docs/logging.md
```diff
@@ -41,7 +83,32 @@ return MikroORM.init({
...
+}
+
+return MikroORM.init({
+  debug: true,
+  loggerFactory: (options) => new CustomLogger(options),
+});
+```
+
+#### Creating a custom logger from scratch
+We can use `loggerFactory` and use our own implementation of the `Logger` interface:
```
B4nan, line 111:
```suggestion
#### Creating a custom logger from scratch

You can use `loggerFactory` and use your own implementation of the `Logger` interface:
```

B4nan (approved):

PR #4250, opened 2023-04-21
feat(migrations): add support for custom migration names

AbdlrahmanSaberAbdo:
This PR adds a new option to the `migrations:create` command that allows developers to specify custom names for migration files. Previously, the migration files were automatically named based on the current timestamp, but with this new option, developers can now give the migration file a more descriptive name that is easier to understand and manage.

This feature was inspired by the functionality in Laravel and Ruby on Rails, and I believe it will bring many benefits, including improved readability and organization of migration files.

I created a discussion before on this: https://github.com/mikro-orm/mikro-orm/discussions/3987

### DIFF
--- a/docs/docs/migrations.md
+++ b/docs/docs/migrations.md
@@ -153,6 +153,14 @@ npx mikro-orm migration:fresh    # Drop the database and migrate up to the lates
 
 > To create blank migration file, we can use `npx mikro-orm migration:create --blank`.
 
+> To specify a name for your migration file, we can use `npx mikro-orm migration:create --name=newName`
+
+You can customize the naming convention for your migration file by utilizing the `fileName` callback and overriding its default behavior.
+```sh
+    migrations: {
+      fileName: (time, name) => time + name
+    }
+```
 For `migration:up` and `migration:down` commands we can specify `--from` (`-f`), `--to` (`-t`) and `--only` (`-o`) options to run only a subset of migrations:
 
 ```sh
--- a/docs/docs/migrations.md
+++ b/docs/docs/migrations.md
@@ -262,14 +262,21 @@ await MikroORM.init({
   },
 });
 ```
-## Using custom migration names
-To specify a name for your migration file, we can use `npx mikro-orm migration:create --name=newName`
 
-> Each migration filename should contains a timestamp that allows Mikro-orm to determine the order of the migrations
+## Using custom migration names
+
+We can add a name to migration files via a CLI
+
+```
+npx mikro-orm migration:create --name=newName
+```
+
+The default behavior support `name` option, here's how it works, if we set `name` option to equal `add_email_property_to_user_table` the name of the generated file will be as follows "Migration20191013214813_add_email_property_to_user_table"
 
 You can customize the naming convention for your migration file by utilizing the `fileName` callback and overriding its default behavior.
 
-If we would like to always specfiy a name, we should throw an error in case the name is not provided.
+If we would like to always specify a name, we should throw an error in case the name is not provided.
+
 ```ts
 migrations: {
   fileName: (timestamp: string, name?: string) => {
@@ -282,7 +289,10 @@ migrations: {
 },
 ```
 
+> Each migration filename should contains a timestamp that allows MikroOrm to determine the order of the migrations
+
 If we just want to change the filename without throwing an error in case the name is not provided
+
 ```ts
 migrations: {
   fileName: (timestamp: string, name?: string) => {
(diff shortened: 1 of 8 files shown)


B4nan (changes requested):
thanks!

packages/migrations/src/MigrationGenerator.ts
```diff
@@ -12,14 +12,18 @@ export abstract class MigrationGenerator implements IMigrationGenerator {
...
     path = Utils.normalizePath(this.driver.config.get('baseDir'), path ?? defaultPath);
     await ensureDir(path);
     const timestamp = new Date().toISOString().replace(/[-T:]|\.\d{3}z$/ig, '');
     const className = this.namingStrategy.classToMigrationName(timestamp);
-    const fileName = `${this.options.fileName!(timestamp)}.${this.options.emit}`;
+    let fileName = `${this.options.fileName!(timestamp)}`;
+    if (name) {
+      fileName = `${fileName}_${name}`;
+    }
+    fileName = `${fileName}.${this.options.emit}`;
```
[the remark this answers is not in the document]
AbdlrahmanAbdo: Got it, thanks for your input.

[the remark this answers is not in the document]
AbdlrahmanSaberAbdo: Got it, thanks for your input.

packages/core/src/utils/Configuration.ts
```diff
@@ -452,7 +452,7 @@ export type MigrationsOptions = {
   snapshotName?: string;
   emit?: 'js' | 'ts' | 'cjs';
   generator?: Constructor<IMigrationGenerator>;
-  fileName?: (timestamp: string) => string;
+  fileName?: (timestamp: string, name?: string) => string;
```
B4nan, line 455: the default implementation should respect the `name` out of box

https://github.com/mikro-orm/mikro-orm/blob/4a5dd15a15bfb17d8dacbd55ffec3374b672d777/packages/core/src/utils/Configuration.ts#L100

B4nan:
and dont forget to update the docs

packages/core/src/utils/Configuration.ts
AbdlrahmanSaberAbdo: Ah, sorry, I wasn't aware that the update was necessary. Thank you for pointing out this. as for the default implementation, would it suffice to simply add the parameter, or should we consider a different approach, such as using `Migration${timestamp}${name}`? Let me know your thoughts.

B4nan: I'd go with `Migration${timestamp}${name ? '_' + name : ''}`

docs/docs/migrations.md
```diff
@@ -153,6 +153,14 @@ npx mikro-orm migration:fresh    # Drop the database and migrate up to the lates
...
 > To create blank migration file, we can use `npx mikro-orm migration:create --blank`.
 
+> To specify a name for your migration file, we can use `npx mikro-orm migration:create --name=newName`
+
+You can customize the naming convention for your migration file by utilizing the `fileName` callback and overriding its default behavior.
+```sh
+    migrations: {
+      fileName: (time, name) =>`Migration-${time}+${name}`
+    }
+```
```
B4nan, line 163: this should have its own section, its not just about "Using via CLI", should be more like "using custom migration names"

also, the parameter is not `time` but `timestamp`, we should have type hints in there for both, and the language is not `sh` but `ts`, plus bad indent. in this example, we should also throw in case the name is not provided. 

```ts
migrations: {
  fileName: (timestamp: string, name?: string) => {
    if (!name) {
      throw new Error('Specify migration name via `npx mikro-orm migration:create --name=...` option when creating the ');
    }

    return `Migration${timestamp}_${name}`;
  },
},
```

AbdlrahmanSaberAbdo: Thank you very much for your input! I have added a new section. Please let me know if I have missed anything else or if you think additional information should be included.

docs/docs/migrations.md
```diff
@@ -262,7 +262,34 @@ await MikroORM.init({
   },
 });
 ```
+## Using custom migration names
+To specify a name for your migration file, we can use `npx mikro-orm migration:create --name=newName`
 
+> Each migration filename should contains a timestamp that allows Mikro-orm to determine the order of the migrations
+
+You can customize the naming convention for your migration file by utilizing the `fileName` callback and overriding its default behavior.
```
B4nan, line 270: default behavior already supports the `name` option, we should describe how it works by default (appending the name if available), and show an example that forces the name option (that's what we already have underneath)

docs/docs/migrations.md
```diff
@@ -262,7 +262,34 @@ await MikroORM.init({
...
+
+If we would like to always specfiy a name, we should throw an error in case the name is not provided.
```
B4nan, line 272: typo in `specfiy`

B4nan (commented):
the code is ok now, lets please try to improve the docs, now its more like few disconnected notes. 

if you struggle with this, no worries, i can rewrite it myself if you want

B4nan (commented):
~improve docs~

B4nan (approved):

PR #4273, opened 2023-04-25
feat(core): allow overriding global logging options on per-query basis

### LINKED ISSUE #4223: Disable log for some query

By using sequelize I can disable log like

```ts
const user = await User.findOne({  
  logging: false,
  where: { id: 2207 }
})
```

When upsert or insert many rows, this is very helpful, I want to disable some log when using mikro orm.

Maybe add or logging to the second option ?

als control the Logger

**Additional context**

SpencerKaiser:
Closes #4223


tests/Logger.test.ts
```diff
@@ -8,7 +8,7 @@ const yellowColorFormatterSpy = jest.spyOn(colors, 'yellow').mockImplementation(
 
 const mockWriter = jest.fn();
 
-describe('Logger', () => {
+describe('Loggers', () => {
```
B4nan, line 11: some integration tests would be nice, to see the propagation from em.find works

B4nan (commented):
that was fast! i just added a comment how this could work to the issue

B4nan:
btw i am rebasing the v6 branch now, so you might wanna recreate the PR once i am done

SpencerKaiser:
@B4nan sounds good; just don't update this branch and I'll just reset everything and force pull when it's ready

Also I just realized failures kinda make this weird... should `isDisabled` prevent a failed query from printing?? Feels like it should still print otherwise you may not be aware of a potential issue ☹️ maybe instead of `isDisabled`, something like `disableWhen: boolean | 'on-success'`?

tests/Logger.test.ts
SpencerKaiser: Sounds good; can you point me in the direction of one such test? 😅

B4nan:
I don't think we want to print failed queries when global logging is disabled, but for the local override it could be a good thing.

I care more about the API in FindOptions, and there I would like the `logging` option (or something else, feel free to suggest, but not `isDisabled`), that can override in both ways (true/false).

SpencerKaiser:
I'd be game to replace `loggerOptions` inside `FindOptions` with just `logging`... is that what you're proposing? If so, I'm on board... I think it's more generic and less characters 😄 it'd also help clean up the types and make it obvious that that type is for options within find options (and later others).

Re: `enabled` - I personally think it's kind of weird to have an optional that turns something on because then it's unclear what the default is. I think your idea of the local `debug` is a good one... you're basically suggesting a mirror of the relevant arguments from the init call?

B4nan:
> Re: enabled - I personally think it's kind of weird to have an optional that turns something on because then it's unclear what the default is.

the whole point of this feature is to override the global value - so its irrelevant what the default is

B4nan:
What a painful rebase, made two super small mistakes and it has cost me an additional hour of debugging 🤦 But it should be ready now.

SpencerKaiser:
@B4nan oof, sorry man... well it should be trivial for me to fix the branch and re-push. I'll do it first thing tomorrow and read through the comments from this evening that I missed.

packages/core/src/logging/DefaultLogger.ts
```diff
@@ -60,15 +60,27 @@ export class DefaultLogger implements Logger {
...
+    if (context?.isDisabled !== undefined && context.isDisabled !== false) {
+      if (context.isDisabled === true) {
+        // Disable for ALL
+        return false;
+      } else if (context.isDisabled === 'on-success'
+        && context.level
+        && !['warning', 'error'].includes(context.level)) {
+        // Disable only on success and the level was NOT an error state
+        return false;
+      }
```
B4nan, line 73: one code style thing + no need for `else if` when you `return`

```suggestion
      }

      if (
        context.isDisabled === 'on-success' &&
        context.level &&
        !['warning', 'error'].includes(context.level)
      ) {
        // Disable only on success and the level was NOT an error state
        return false;
      }
```

btw i am not completely sold on the idea of the `on-success` value. if a query fails, it will result in error which has the query inside the error message already, i think we dont need to enforce it here when logging is disabled. try it in a real app to see, iirc it works this way.

that's also the reason why i was asking for some integration tests, to see how it will actually work for users

SpencerKaiser: Oh sorry, I didn't think you'd re-review after I resolved the merge stuff... I had some of that stuff locally but just wanted to fix the branch after the rebase. here's what I'm planning:

image

`enabled` will act as a master override; if `disabled` no logs will be output. Period.

If `debugMode` is set, it'll be used instead of what the logger was initialized with as a local override.

Sound good?

B4nan: for me this is about the find methods mainly, i'd keep it, you need to be able to set what to log (query or query + query params), right?

i guess the flush method should support this too somehow

SpencerKaiser: Just pushed the changes; let me know what you'd like to do. My recommendation is to pull the `debugMode` for now and add it back later when it's actually relevant

SpencerKaiser: Yea eventually I'd love for this to be added to `update`, `flush`, and `persistAndFlush`... that was something I hoped to work on next. If you're good with it as is, I can update the docs just not totally sure how 😬

B4nan: Before v6 branch gets merged to master, those docs are not deployed anywhere, so we can have something in progress there, not a huge deal.

I'd like to merge it within a few weeks, let's say around half of may.

SpencerKaiser:
@B4nan this is ready to go unless you want an integration test; happy to write one, just can't figure out where they live (aside from the issue tests)

B4nan:
They live everywhere :] I'd say more than 90% of the tests are integration tests. I usually use the `tests/features` folder these days, and moving the regression tests there too when they are about a specific feature.

SpencerKaiser:
Is there a feature you think most closely relates to this? Should I just use the logger.test.ts file?

Do you have a good example of a simple integration test I could use as a baseline?

B4nan:
All (or most of) the issue tests are integration test examples - they init the ORM and use the EM API.

I would create a new folder `tests/features/logging` or something like that, and move the `logger.test.ts` file there. I don't mind if the additional tests would go inside it or to a new file.

I'd like to see a test that has `debug: false` (so the default) globally and does few queries, some of them will override this via FindOptions, and the logs will be mocked and asserted. Many existing tests are already doing this, there is a `mockLogger` helper you can search for to see some examples. Then the same but with enabled logging globally and disabled locally.

SpencerKaiser:
@B4nan I made an integration test but it isn't working as expected... can you take a look and tell me why you think it isn't working? I added that console in there to help with testing; if I set `debug: true` on the `init`, it captures all the discovery stuff, but it isn't writing anything when I query ☹️

B4nan:
Why didn't you just copy the approach with the existing `mockLogger` helper? 

I think your problem is that you only adjust the `debug` option in the config, but the logger is already created and you are not changing its cached value. This is handled in the `mockLogger` helper.

Or maybe its not problem with the test, but the implementation, I can take a closer look later this week if you want.

SpencerKaiser:
I remember seeing `mockLogger` but I forgot about it; I'll take a look and see if I can utilize that instead

SpencerKaiser:
@B4nan still can't get it to work as expected; I've played around with the mock logger a lot and it seems like queries are just omitted for some reason :/  I even switched back to the current v6 with my logging test (minus the new stuff) and had the same experience: no queries get logged... any ideas?

B4nan:
You didn't clear the EM, you have the entities in context, and you query them by the PK, so it gets returned from the identity map cache right ahead. There are no queries at all, that's why there are no logs.

SpencerKaiser:
Your library is too efficient 😅 thanks for the direction; I'll go make that change

SpencerKaiser:
@B4nan tests are passing!

docs/docs/logging.md
```diff
@@ -73,6 +73,39 @@ For MongoDB you can use `MongoHighlighter` from `@mikro-orm/mongo-highlighter` p
...
+An optional `loggerContext` option can be included within the `FindOptions` parameter of either call which will add a label to the output when debug mode is enabled.
+
+```ts
+const author = await em.findOne(Author, { id: 1 }, { loggerContext: { label: 'Author Retrieval - /authors/me' } });
+// [query] (Author Retrieval - /authors/me) select "a0".* from "Author" as "a0" where "a0"."id" = 1 limit 1 [took 21 ms]
+```
+
+### Changing `debugMode` or disabling logging for specific queries
+
+If you'd like to disable queries on a per-query basis, you can leverage the `isDisabled` flag within `FindOptions`:
```
B4nan, line 89: the flag is not called `FindOptions.isDisabled`, also its disabling of logging, not disabling of queries

docs/docs/logging.md
```diff
@@ -153,25 +186,17 @@ interface LogContext extends Dictionary {
   params?: unknown[];
   took?: number;
   level?: 'info' | 'warning' | 'error';
+  enabled?: boolean;
+  debugMode?: LoggerNamespace[]
```
B4nan, line 190:
```suggestion
  debugMode?: LoggerNamespace[];
```

packages/core/src/logging/Logger.ts
```diff
@@ -53,10 +55,12 @@ export interface LoggerOptions {
...
+ * Logger options to modify format output and overrides, including a label and additional properties that can be accessed by custom loggers
+ *
+ * Differs from {@link LoggerOptions} in terms of how they are used; this type is primarily a public type meant to be used within methods like `EntityManager.Find`
  *
  * @example
  * await em.findOne(User, 1, { loggerContext: { label: 'user middleware' } };
  * // [query] (user middleware) select * from user where id = 1;
  */
-export type LoggerContext = Pick<LogContext, 'label'> & Dictionary;
+export type LoggingOptions = (Pick<LogContext, 'label' | 'enabled' | 'debugMode'> & Dictionary);
```
B4nan, line 66: i'd remove the parens here, imho they only make it worse in terms of readability

SpencerKaiser:
@B4nan great catches, sorry for the sloppiness on those 😬 should be ready again

docs/docs/logging.md
```diff
@@ -86,7 +86,7 @@ const author = await em.findOne(Author, { id: 1 }, { loggerContext: { label: 'Au
 
 ### Changing `debugMode` or disabling logging for specific queries
 
-If you'd like to disable queries on a per-query basis, you can leverage the `isDisabled` flag within `FindOptions`:
+If you'd like to disable logging on a per-query basis, you can leverage the `enabled` flag within `FindOptions`:
```
B4nan, line 89: the flag is `logging.enabled`, this still sounds a bit like it should be `FindOptions.enabled`

otherwise it looks good

SpencerKaiser: How about the new version?

docs/docs/logging.md
```diff
@@ -73,6 +73,39 @@ For MongoDB you can use `MongoHighlighter` from `@mikro-orm/mongo-highlighter` p
...
+An optional `loggerContext` option can be included within the `FindOptions` parameter of either call which will add a label to the output when debug mode is enabled.
+
+```ts
+const author = await em.findOne(Author, { id: 1 }, { loggerContext: { label: 'Author Retrieval - /authors/me' } });
+// [query] (Author Retrieval - /authors/me) select "a0".* from "Author" as "a0" where "a0"."id" = 1 limit 1 [took 21 ms]
+```
+
+### Changing `debugMode` or disabling logging for specific queries
+
+If you'd like to disable logging or change the debug mode on a per-query basis, you can leverage `FindOptions.loggingOptions` and its `enabled` or `debugMode` property:
```
B4nan, line 89: but its not `loggingOptions`, its just `logging` :D `LoggingOptions` is the interface, `logging` is the option name/key.

```suggestion
If you'd like to disable logging or change the debug mode on a per-query basis, you can leverage `FindOptions.logging` and its `enabled` or `debugMode` property:
```

PR #4294, opened 2023-04-30
fix(migrations): respect custom migration name in migration class names

srmagura:
Fixes:

- Custom migration name was not used in the migration class name for SQL drivers.
- Custom migration name was not used in the migration file name or class name for the Mongo driver.

### Side note

I am having some trouble running the Mongo tests on Ubuntu 22.04. Getting an error because libcrypto 1.1 is not found.

I manually updated the Mongo snapshot to what I think the correct value is. We'll have to see if that test passes in CI.


srmagura:
@B4nan I fixed some issues with the Mongo code, should be good now 🙂. I was able to run the tests locally by using my Mac.

B4nan (approved):
Thanks!

Issue #2388, opened 2021-11-09
Support mapping the same column across discriminator subclasses


parisholley:
**Describe the bug**
In order to make it easier for developers to understand models that rely on single-table design (in particular, managed ManyToMany tables), we "alias" some of our ManyToOne's so that you don't have to remember which logical data model maps to which generically named column. Other examples would be tables with a `to`/`from` or `parent/child` naming convention. 

```typescript
@Entity({ discriminatorValue: MemberRelationshipType.CUSTOMER })
export class BusinessCustomerMemberRelationship extends MemberRelationship<MemberRelationshipType.CUSTOMER> extends {
  @ManyToOne(() => Member, { fieldName: 'activeId', wrappedReference: true })
  customer: IdentifiedReference<Member>;

  @ManyToOne(() => Member, { fieldName: 'passiveId', wrappedReference: true })
  business: IdentifiedReference<Member>;

  // additional relationships/columns/metadata that only apply for customers
}

@Entity({ discriminatorValue: MemberRelationshipType.VENDOR })
export class BusinessVendorMemberRelationship extends MemberRelationship<MemberRelationshipType.VENDOR> extends {
  @ManyToOne(() => Member, { fieldName: 'activeId', wrappedReference: true })
  vendor: IdentifiedReference<Member>;

  @ManyToOne(() => Member, { fieldName: 'passiveId', wrappedReference: true })
  business: IdentifiedReference<Member>;

  // additional relationships/columns/metadata that only apply for vendors
}
```

When attempting to insert a discriminated subclass, the ORM will inject the activeId twice into the SQL query and it will fail. Using a getter/setter per subclass to "mimick" the alias isn't useful because it will likely cause problems when using the API (eg: filtering or populating).

**Expected behavior**
ORM should only use metadata of parent class and subclass to generate queries and access identity map.

**Versions**

```
    "@mikro-orm/cli": "^4.5.9",
    "@mikro-orm/core": "^4.5.9",
    "@mikro-orm/migrations": "^4.5.9",
    "@mikro-orm/nestjs": "^4.3.1",
    "@mikro-orm/postgresql": "^4.5.9",
    "@mikro-orm/reflection": "^4.5.9",
```

parisholley:
[...]

PR #4308, opened 2023-05-02
Reproduction for #2388

parisholley:
Ran into this again when working on my refactor to v6, it includes an issue from the original ticket as well as a new one I found when mutating a discriminated entity.

@MartinAndreev perhaps use this PR as a starting point to reproduce your problem


parisholley:
@B4nan having dug through code in this PR to find potential fixes and here is what i saw:

https://github.com/mikro-orm/mikro-orm/blob/f9530e4b24bd806e33f1997d4e1ef548e02b6b90/packages/core/src/unit-of-work/UnitOfWork.ts#L1015-L1050

this logic is of course here to help with batch inserts/updates, but perhaps we add logic to detect instances where discriminated classes have custom fields defined (instead of just inheriting 100% of them from parent class) and instead of using `rootName`, just use `name`.

https://github.com/mikro-orm/mikro-orm/blob/f9530e4b24bd806e33f1997d4e1ef548e02b6b90/packages/core/src/utils/EntityComparator.ts#L547-L549

when debugging, I see that the `comparableProps` will have each column across all sub-types, whereas `hydrateProps` will ignore fields from other discriminator classes. would it make sense to compare those two to build the comparison instead of just `comparableProps`?

PR #4311, opened 2023-05-03
failing test for multiple orm instances at the same time

Tlepel:
This test reproduces the behaviour from discussion #4304

### DIFF
--- a/tests/issues/GH4304.test.ts
+++ b/tests/issues/GH4304.test.ts
@@ -1,0 +1,132 @@
+import {
+  Entity,
+  ManyToOne,
+  PrimaryKey,
+  Property,
+  OneToOne,
+  IdentifiedReference,
+  OneToMany,
+  Collection,
+  LoadStrategy,
+} from '@mikro-orm/core';
+import { MikroORM } from '@mikro-orm/sqlite';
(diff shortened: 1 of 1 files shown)


B4nan:
Well, but that's expected behaviour, you have two ORM instances, they won't be sharing anything, by design.

B4nan:
The problem here is that the entity prototype holds a reference to the hydrator - and you use the same entity classes in both ORM instances. I tried to replace it dynamically but for some reason, it's still not working as expected.

What's the use case here, why do you want to have multiple ORM instances for the same set of entities/database?

Tlepel:
We're using multiple instances (3 actually) in our api endpoint tests:

- In the api instance
- In the tests, to edit entities and persist them, without needing to call api endpoints
- In the seeder

The seeder could probably reuse the ORM instance from the test files, however the issue is still present when we do that.

B4nan:
I still don't understand why you would need more than one, but it should indeed work. The fix might end up a bit breaking, so maybe it will go to v6 only, will let you know when I manage to find a proper way.

B4nan:
Closing as this is no longer reproducible on v6.

PR #4314, opened 2023-05-04
fix(mongo): sorting with UnderscoreNamingStrategy

### LINKED ISSUE #4313: orderBy is not working with UnderscoreNamingStrategy in Mongodb
I use MongoDB with UnderscoreNamingStrategy  
`find` and `findOne` with sort by camel case field do not work, sorting is ignored.  

The problem seems to be exactly with Mongo driver.

**To Reproduce**
```
@Entity()
export class A {
  @PrimaryKey()
  _id = randomUUID();
  
  @Property()
  complexName!: string;
}

orm = await MikroORM.init({
  entities: [A],
  clientUrl: await initMongoReplSet('mikro-orm-test'),
  driver: MongoDriver,
  namingStrategy: UnderscoreNamingStrategy,
});
await orm.schema.clearDatabase()

const a = new A();
a.complexName = 'a';
const b = new A();
b.complexName = 'b';
await orm.em.persistAndFlush(a);
await orm.em.persistAndFlush(b);
orm.em.clear();

const data2 = await orm.em.find(A, {}, {
  orderBy: { complexName: 'desc' },
});
expect(data2[0].complexName).toBe('b');
```
**Expected behavior**
Keys in `orderBy` should be transformed according to naming strategy.

**Versions**

| Dependency | Version |
| - | - |
| node | 18.14 |
| typescript | 4.9.5 |
| mikro-orm | 5 |
| your-driver | MongoDB |

funduck:
Fixes #4313 

I propose to apply `this.renameFields` to `options.orderBy` the same way it is used for `where`

### DIFF
--- a/tests/issues/GH4313.test.ts
+++ b/tests/issues/GH4313.test.ts
@@ -30,7 +30,7 @@ describe('GH issue 4313', () => {
     a.complexName = 'a';
     const b = new A();
     b.complexName = 'b';
-    await orm.em.persistAndFlush([a,b]);
+    await orm.em.persistAndFlush([a, b]);
     orm.em.clear();
   });
 
@@ -81,6 +81,5 @@ describe('GH issue 4313', () => {
       orderBy: [{ complexName: 'desc' }],
     });
     expect(data4!.complexName).toBe('b');
-
   });
 });
--- a/tests/issues/GH4313.test.ts
+++ b/tests/issues/GH4313.test.ts
@@ -1,0 +1,86 @@
+import { Entity, MikroORM, PrimaryKey, Property, UnderscoreNamingStrategy } from '@mikro-orm/core';
+import { MongoDriver } from '@mikro-orm/mongodb';
+import { randomUUID } from 'crypto';
+
+@Entity()
+export class A {
+
+  @PrimaryKey()
+  _id = randomUUID();
+
+  @Property()
+  complexName!: string;
+
+}
+
+let orm: MikroORM<MongoDriver>;
+
+describe('GH issue 4313', () => {
+
+  beforeAll(async () => {
+    orm = await MikroORM.init({
+      entities: [A],
+      clientUrl: 'mongodb://localhost:27017/mikro-orm-test',
+      driver: MongoDriver,
+      namingStrategy: UnderscoreNamingStrategy,
+    });
+    await orm.schema.clearDatabase();
+
+    const a = new A();
+    a.complexName = 'a';
+    const b = new A();
+    b.complexName = 'b';
+    await orm.em.persistAndFlush([a,b]);
+    orm.em.clear();
+  });
+
+  afterAll(async () => {
+    await orm.close();
+  });
+
+  test('Find with sort by complexName field that is camel case string', async () => {
+    const data1 = await orm.em.find(A, {}, {
+      orderBy: { complexName: 'asc' },
+    });
+    expect(data1[0].complexName).toBe('a');
+
+    const data2 = await orm.em.find(A, {}, {
+      orderBy: { complexName: 'desc' },
+    });
+    expect(data2[0].complexName).toBe('b');
+
+    const data3 = await orm.em.find(A, {}, {
+      orderBy: [{ complexName: 'asc' }],
+    });
+    expect(data3[0].complexName).toBe('a');
+
+    const data4 = await orm.em.find(A, {}, {
+      orderBy: [{ complexName: 'desc' }],
+    });
+    expect(data4[0].complexName).toBe('b');
+
+  });
+
+  test('FindOne with sort by complexName field that is camel case string', async () => {
+    const data1 = await orm.em.findOne(A, { complexName: { $ne: null } }, {
+      orderBy: { complexName: 'asc' },
+    });
+    expect(data1!.complexName).toBe('a');
+
+    const data2 = await orm.em.findOne(A, { complexName: { $ne: null } }, {
+      orderBy: { complexName: 'desc' },
+    });
+    expect(data2!.complexName).toBe('b');
+
+    const data3 = await orm.em.findOne(A, { complexName: { $ne: null } }, {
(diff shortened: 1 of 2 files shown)


tests/issues/GH4313.test.ts
```diff
@@ -0,0 +1,121 @@
...
+}
+
+describe('GH issue 4313', () => {
+
+  describe('Basic naming strategy with Mongo', () => {
+
+    it('Init orm with basic naming strategy',async () => {
+      orm = await MikroORM.init({
+        entities: [A],
+        clientUrl: await initMongoReplSet('mikro-orm-test'),
```
B4nan, line 85: no need for replica sets, that memory server implementation is rather unstable and i am using it only in places that are actually testing transactions

tests/issues/GH4313.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    const data4 = await orm.em.findOne(A, { complexName: { $ne: null } }, {
+      orderBy: [{ complexName: 'desc' }],
+    });
+    expect(data4!.complexName).toBe('b');
+
+  });
+}
+
+describe('GH issue 4313', () => {
```
B4nan, line 78: no need for nested describe blocks - actually no need for any of them, use just one `test` block on top level

tests/issues/GH4313.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      });
+      await orm.schema.clearDatabase();
+    });
+
+    testCase();
+
+    it('Close connection', async () => {
+      await orm.close();
+      await closeReplSets();
+    });
```
B4nan, line 117: this should be `afterAll` hook, not a separate test case

tests/issues/GH4313.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    it('Init orm with underscore naming strategy',async () => {
+      orm = await MikroORM.init({
+        entities: [A],
+        clientUrl: await initMongoReplSet('mikro-orm-test'),
+        driver: MongoDriver,
+        namingStrategy: UnderscoreNamingStrategy,
+      });
+      await orm.schema.clearDatabase();
+    });
```
B4nan, line 110: this should be `beforeAll` hook, not a separate test case

tests/issues/GH4313.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  });
+
+  describe('Underscore naming strategy with Mongo', () => {
+
+    it('Init orm with underscore naming strategy',async () => {
+      orm = await MikroORM.init({
+        entities: [A],
+        clientUrl: await initMongoReplSet('mikro-orm-test'),
```
B4nan, line 105: keep the defaults, just set the database name via `dbName`

tests/issues/GH4313.test.ts
```diff
@@ -0,0 +1,121 @@
...
+function testCase() {
+  test('Insert entities', async () => {
+    const a = new A();
+    a.complexName = 'a';
+    const b = new A();
+    b.complexName = 'b';
+    await orm.em.persistAndFlush(a);
+    await orm.em.persistAndFlush(b);
+    orm.em.clear();
+  });
```
B4nan, line 28: this should be also part of that (single) `beforeAll` hook

tests/issues/GH4313.test.ts
```diff
@@ -0,0 +1,121 @@
...
+let orm: MikroORM<MongoDriver>;
+
+function testCase() {
+  test('Insert entities', async () => {
+    const a = new A();
+    a.complexName = 'a';
+    const b = new A();
+    b.complexName = 'b';
+    await orm.em.persistAndFlush(a);
+    await orm.em.persistAndFlush(b);
```
B4nan, line 26: you can pass both in a single call
```suggestion
    await orm.em.persistAndFlush([a, b]);
```

tests/issues/GH4313.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    const data3 = await orm.em.findOne(A, { complexName: { $ne: null } }, {
+      orderBy: [{ complexName: 'asc' }],
+    });
+    expect(data3!.complexName).toBe('a');
+
+    const data4 = await orm.em.findOne(A, { complexName: { $ne: null } }, {
+      orderBy: [{ complexName: 'desc' }],
+    });
+    expect(data4!.complexName).toBe('b');
```
B4nan, line 73: i dont think we need to test `asc` vs `desc`, that's kinda irrelevant to the issue, one direction is enough, the problem is about the left side, not right side

tests/issues/GH4313.test.ts
```diff
@@ -0,0 +1,121 @@
...
+  _id = randomUUID();
+
+  @Property()
+  complexName!: string;
+
+}
+
+let orm: MikroORM<MongoDriver>;
+
+function testCase() {
```
B4nan, line 19: lets inline this

B4nan (changes requested):
thanks! interesting this wasnt reported before, i guess most of the mongo folks dont really use different strategies/field names

left a few comments about the test case, i'd like to see that greatly simplified, but will be happy to do that myself after merging if you'd struggle somehow with that.

tests/issues/GH4313.test.ts
funduck: one case may pass just by an accident, but both wont

tests/issues/GH4313.test.ts
funduck: Ok, I'll make it shorter

tests/issues/GH4313.test.ts
funduck: Wow, how much faster the test have become without replset! Thanks for hint!

funduck:
Fixed most of the issues you've mentioned

B4nan (approved):

PR #4321, opened 2023-05-05
feat(core): allow using dataloader for references and collections

darkbasic:
This is currently a very basic implementation so that we can start talking about where we want this headed. I left out the collection dataloader as well as options handling and kept it as simple as possible. The biggest priority right now is being able to build mikro-orm on ppc64le to actually test it: https://github.com/lerna/lerna/issues/3676
I've tried to build just `core` but I had no luck as well:
```
niko@talos2 ~/devel/mikro-orm/packages/core $ yarn build
command not found: rimraf
```


packages/core/src/EntityManager.ts
```diff
@@ -69,6 +86,16 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
   private readonly entityFactory: EntityFactory = new EntityFactory(this);
   private readonly unitOfWork: UnitOfWork = new UnitOfWork(this);
   private readonly resultCache = this.config.getResultCacheAdapter();
+  private readonly refLoader = new DataLoader<Ref<any>, any>(async refs => {
+    const groupedIds = groupPrimaryKeysByEntity(refs);
+    const promises = Array.from(groupedIds).map(
+      async ([entity, ids]) =>
+        await this.getRepository(entity).find(Array.from(ids)),
```
B4nan, line 93: don't use repositories internally
```suggestion
        await this.find(entity, Array.from(ids)),
```

also the first parameter is rather `entityName`, calling this just `entity` is pretty confusing. and i dont think we need to mark the method as async and await the inside, just pass the promise already created by `em.find`.

packages/core/src/EntityManager.ts
```diff
@@ -69,6 +86,16 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
...
+    );
+    await Promise.all(promises);
+    return await Promise.all(refs.map(async ref => await ref.load({ dataloader: false })));
```
B4nan, line 96: why? you are already loading things via the `em.find` call, or what am i missing?

packages/core/src/EntityManager.ts
```diff
@@ -50,6 +50,23 @@ import type { FlushEventArgs } from './events';
 import { EventManager, TransactionEventBroadcaster } from './events';
 import type { EntityComparator } from './utils/EntityComparator';
 import { OptimisticLockError, ValidationError } from './errors';
+import DataLoader from 'dataloader';
+
+function groupPrimaryKeysByEntity(
```
B4nan, line 55: either put this at the very end, or a private method of the class. i dont want this file to start with some random helper function.

maybe even better, put this to the `Utils` class or some other file outside. would be good to export and unit test it

packages/core/src/EntityManager.ts
```diff
@@ -50,6 +50,23 @@ import type { FlushEventArgs } from './events';
...
+  refs: readonly Reference<any>[],
+): Map<string, Set<Primary<any>>> {
+  const map = new Map<string, Set<Primary<any>>>();
+  for (const ref of refs) {
+    const className = helper(ref).__meta.className;
+    let primaryKeys = map.get(className);
```
B4nan, line 61: maybe better use the EntityMetadata objects as the key, it will be more performant

packages/core/src/EntityManager.ts
```diff
@@ -69,6 +86,16 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
   private readonly entityFactory: EntityFactory = new EntityFactory(this);
   private readonly unitOfWork: UnitOfWork = new UnitOfWork(this);
   private readonly resultCache = this.config.getResultCacheAdapter();
+  private readonly refLoader = new DataLoader<Ref<any>, any>(async refs => {
```
B4nan, line 89: since you already have a helper that does part of this, i'd like to see the whole parameter/handler created via some helper too, this part of the EM class should stay clean. maybe there should be some utils class/file for dataloader specific things

packages/core/src/entity/Reference.ts
```diff
@@ -114,6 +114,10 @@ export class Reference<T extends object> {
   async load<TT extends T, K extends keyof T = never, P extends string = never>(options?: LoadReferenceOptions<T, P> | K): Promise<Loaded<TT, P> | T[K]> {
     const opts: Dictionary = typeof options === 'object' ? options : { prop: options };
 
+    if (opts.dataloader === true) {
```
B4nan, line 117:
```suggestion
    if (opts.dataloader) {
```

packages/core/src/entity/Reference.ts
```diff
@@ -114,6 +114,10 @@ export class Reference<T extends object> {
...
+      return helper(this.entity).__em.refLoader.load(this);
```
B4nan, line 118: so just for me to understand this better, with this you can use something like `Promise.all(refs.map(ref => ref.load({ dataloader: true }))` and it will run a single em.find query under the hood?

packages/core/src/EntityManager.ts
darkbasic: This is definitely gonna move elsewhere because I will have one more method for collections dataloading and way more ones if we eventually decide to do general purpose find dataloading. I've left it there just do decide where to move it. I think we can start with `Utils` and eventually create its own file if/when it gets crowded.

darkbasic:
> Why would you need to build anything to test it? Write tests, they use ts-node, nothing needs to be build, or what am I missing? Tests will be needed either way, and you can run them in the CI so you are not blocked now.

Because I already have tons of tests in my project which already makes use of an external library version of the dataloder.
They verify the result, test performance, etc so being able to start with them would be an headstart before I implement proper tests in mikro-orm itself.

packages/core/src/EntityManager.ts
darkbasic: Instead of filtering between the find results and mapping each of them to the original reference the ref dataloader uses a shortcut and takes advantage of the already existing Mikro-ORM cache: when it calls `ref.load` it will automatically find an entry in the entity cache (because of the previous find query) and automatically sort/map them without the need for additional code. This trick won't be possible for collections dataloading where I had to implement my own mappers. I guess I should explain how it works in a comment.

packages/core/src/entity/Reference.ts
darkbasic: Yes, see https://github.com/mikro-orm/mikro-orm/pull/4321#discussion_r1185950286

B4nan:
Would turborepo be any better? We could switch to that for the task running, and keep lerna only for publishing.

darkbasic:
> Would turborepo be any better? We could switch to that for the task running, and keep lerna only for publishing.

It was, but unfortunately not anymore:
https://github.com/vercel/turbo/issues/2616
https://github.com/vercel/turbo/discussions/1891#discussioncomment-3913339

They are planning to add ppc64le support back but I fear it's not going to happen anytime soon.

darkbasic:
@B4nan if I could manage to find where that unsupported error comes from I could try to build my own ppc64 binary (or even emulate the x86 one with qemu-user) but unfortunately it's not so obvious:

```
niko@talos2 ~/devel $ git clone https://github.com/lerna/lerna.git
Cloning into 'lerna'...
remote: Enumerating objects: 43257, done.
remote: Counting objects: 100% (1411/1411), done.
remote: Compressing objects: 100% (814/814), done.
remote: Total 43257 (delta 744), reused 1131 (delta 563), pack-reused 41846
Receiving objects: 100% (43257/43257), 29.57 MiB | 4.05 MiB/s, done.
Resolving deltas: 100% (32033/32033), done.
niko@talos2 ~/devel $ cd lerna/
niko@talos2 ~/devel/lerna $ grep -ir "Unsupported architecture" .
niko@talos2 ~/devel/lerna $ grep -ir "architecture" .
./libs/commands/publish/src/lib/create-temp-licenses.ts:    // (give up on 32-bit architecture to avoid fs-extra warning)
./website/docs/features/project-graph.md:For Lerna (and Nx) to run tasks quickly and correctly, it creates a graph of the dependencies between all the projects in the repository.  Exploring this graph visually can be useful to understand why Lerna is behaving in a certain way and to get a high level view of your code architecture.
./website/src/components/about-lerna.tsx:    text: "Lerna comes with a powerful interactive workspace visualizer, helping you understand the architecture of your workspace.",
niko@talos2 ~/devel/lerna $ grep -ir "unsupported" .
./.git/hooks/fsmonitor-watchman.sample:	die "Unsupported query-fsmonitor hook version '$version'.\n" .
./.eslintrc.json:        "node/no-unsupported-features/es-syntax": "off",
./libs/e2e-utils/src/lib/fixture.ts:        throw new Error(`Unsupported package manager: ${this.packageManager}`);
./libs/e2e-utils/src/lib/fixture.ts:        throw new Error(`Unsupported package manager: ${this.packageManager}`);
./libs/e2e-utils/src/lib/fixture.ts:        throw new Error(`Unsupported package manager: ${this.packageManager}`);
```

packages/core/src/EntityManager.ts
darkbasic: No we don't need async/await, it's just [linting](https://typescript-eslint.io/rules/promise-function-async/) from my existing dataloader library which uses `standard-with-typescript`.

packages/core/src/EntityManager.ts
darkbasic: I can do so, but are you sure using an object instead of a string as the Map key would be faster? Maps performance on V8 has changed so much in the past few years that I've lost track of what is faster.

darkbasic:
Apparently the lerna error comes from nx: https://github.com/nrwl/nx/blob/master/packages/nx/src/native/index.js#L235

Do we actually use the nx stuff? I didn't even know nx was a thing in lerna...

packages/core/src/EntityManager.ts
B4nan: i think it should, but its not that important really, not without actual perf test. btw the same approach is used for the identity map.

B4nan:
> Do we actually use the nx stuff? I didn't even know nx was a thing in lerna...

It is used as the default task runner now. Maybe this could be even replaced with `yarn`? Lerna is used basically just for task running in topological order, and for versioning/publishing (which happens in CI only).

darkbasic:
> It is used as the default task runner now. Maybe this could be even replaced with `yarn`?

The problem is that using yarn as a task runner it doesn't find the root dependencies from inside the workspaces:

```
niko@talos2 ~/devel/mikro-orm $ yarn workspace @mikro-orm/core run build
command not found: rimraf
```

B4nan:
I think there was some option to allow that?

darkbasic:
> Hmm, but that's for scripts, not dependencies...

Unfortunately yes, it would simply run the top level script.

darkbasic:
https://github.com/yarnpkg/berry/issues/5429

merceyz:
That's not what the documentation says.

> Check the root workspace for scripts and/or **binaries** instead of the current one
https://yarnpkg.com/cli/run

Binaries in the documentation is what you're referring to as dependencies.

If you change https://github.com/mikro-orm/mikro-orm/blob/9ebdd87ad45fb48cdeb52ad1158a3d8d3d88ee35/packages/core/package.json#L53 to `run -T rimraf ./dist` it will use `rimraf` from the root workspace.

darkbasic:
> `yarn run -T rimraf ./dist` it will use `rimraf` from the root workspace.

@merceyz looks like it's working, but unfortunately I cannot manage to get the `plugin-workspace-tools` to run the build script in the correct dependency order.

Workspaces use a wildcard:
```
  "workspaces": [
    "packages/*"
  ],
```

`yarn workspaces foreach run build` runs them in alphabetical order I think, which leads to failure because it needs to build the packages it depends on first. Looking at the docs I've found the `-t` option (`yarn workspaces foreach -t run build`) but it lead to the correct order either. I've looked for issues in the bug tracker and I've found this: https://github.com/yarnpkg/yarn/issues/7150

Unfortunately it's not clear whether it's for berry or yarn v1 and it looks like a won't fix basically. Am I missing something?

darkbasic:
I've finally managed to build mikro-orm using yarn's own task runner and incorporate it into my project via portals. The only minor annoyance so far has been having to change `@mikro-orm/core/typings` imports to `@mikro-orm/core/dist/typings`, but portals are so convenient that it's a minor tradeoff I'm willing to pay. So far it passes all my tests, so tomorrow I'll add basic support for dataloading collections. From there we can add the remaining options and start adding some tests into mikro-orm itself.

B4nan:
> change @mikro-orm/core/typings imports

You shouldn't be using any deep imports, what types are you missing that are not exported from the root of the package?

darkbasic:
```ts
import type {
  EntityKey,
  EntityProps,
  ExpandProperty,
  ExpandScalar,
  FilterValue2,
  Loaded,
  Query,
  Scalar,
} from "@mikro-orm/core/dist/typings";
```

```ts
import type { EntityKey, IWrappedEntityInternal } from "@mikro-orm/core/dist/typings";
```

These are mostly for the find dataloader (a couple of them for the collection dataloader maybe).
I would have needed more (`FilterValue, Query, FilterObject, Compute, ObjectQuery, FilterQuery`...) but the find dataloader still doesn't cover 100% of the original types so I'm using some custom slightly smaller subtypes.

Apart from the dataloader I didn't have to use deep imports elsewhere in my app. I suggest to ignore them for the moment, I'll add a stub commit for the find dataloader and we decide if this is something which might be worth merging alongside with the ref and collection ones. If we decide to merge it we won't need these exports on the root package anymore, otherwise we can reason about exporting them.

darkbasic:
I've added a basic version of the Collections dataloader. It's not as straightforward as the Reference one, because we have to filter the result to re-assign them to the original collections, but it shouldn't be too hard to understand either. Let me know if something is not clear. I've tested it against my own project test suite and it works well.

P.S.
Unfortunately I couldn't create an incremental commit because I had to rebase it to drop commits from https://github.com/mikro-orm/mikro-orm/pull/4331
As soon as the PR to use yarn as a task runner gets merged it should be a little bit easier to follow the history.

darkbasic:
I've pushed the find dataloader as well. This one is quite a bit more complex because it basically tries to optimize any kind of possible query into the smallest possible number of queries. It doesn't always get it right so it's not something that we want to enable by default for every query, but amazingly it manages to be faster in some real world graphql scenarios (at least in my application). Once you get the grasp out of it it's not that complex (it's the second time I've rewritten it and I've focused on keeping things simple while achieving worthwhile performance), but feel free to ask me all the questions you want including a full detailed explanation if needed. I wanted to make it capable of possibly covering the whole set of operators, but I don't want to do so at the further expense of performance and I'm basically gradually adding more whenever I find the right use case in my own application. I think that even having a slightly tighter scope compared to the normal find could still be fine because that covers 90% of the use cases and you can always create your own specialized dataloader for complex queries. I use to benchmark it case by case and enable it for the queries that I know will suffer a lot from the GraphQL inherent N+1 nature. Not sure if we want to merge it, it's definitely useful but might never reach the full operators coverage (and maybe that's not even necessary).

packages/core/src/EntityManager.ts
```diff
@@ -160,7 +184,6 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
 
     const em = this.getContext();
     await em.tryFlush(entityName, options);
-    entityName = Utils.className(entityName);
```
B4nan, line 163: this seems wrong, the methods like `em.processWhere` expect a string, but `entityName` can be a class reference too.

darkbasic:
Rebased.

darkbasic:
I've added some initial tests which should give full coverage to the utils, making the code much easier to read from github without the missing coverage clutter. I've also added many more comments and renamed some variables.
I'm going to add more to cover things like composite keys.

darkbasic:
@B4nan I've added composite key tests and rebased to latest v6 and suddenly one test started to fail: did you implement automatic inverse relationships by any chance?

B4nan:
Yes exactly that commit, they have `hydrate: false` so you can distinguish.

darkbasic:
@B4nan I think this should be ready for the first round of feedback

darkbasic:
This is also the report from my project's dataloader performance test suite with latest `v6-dataloader` (plus the find dataloader, which is not part of this PR) and some real world queries:

```
 PASS  tests/graphql.test.ts (16.82 s)
  ● Console

    console.log
      W/o dataloader: 34 ms

      at Object.<anonymous> (tests/graphql.test.ts:259:13)

    console.log
      W/ dataloader: 26 ms

      at Object.<anonymous> (tests/graphql.test.ts:260:13)

    console.log
      W/o dataloader: 36 ms

      at Object.<anonymous> (tests/graphql.test.ts:308:13)

    console.log
      W/ dataloader: 25 ms

      at Object.<anonymous> (tests/graphql.test.ts:309:13)

    console.log
      W/o dataloader: 1065 ms

      at Object.<anonymous> (tests/graphql.test.ts:352:13)

    console.log
      W/ dataloader: 371 ms

      at Object.<anonymous> (tests/graphql.test.ts:353:13)

    console.log
      W/o dataloader: 96 ms

      at Object.<anonymous> (tests/graphql.test.ts:396:13)

    console.log
      W/ dataloader: 80 ms

      at Object.<anonymous> (tests/graphql.test.ts:397:13)

    console.log
      W/o dataloader: 103 ms

      at Object.<anonymous> (tests/graphql.test.ts:445:13)

    console.log
      W/ dataloader: 95 ms

      at Object.<anonymous> (tests/graphql.test.ts:446:13)

    console.log
      W/o dataloader: 161 ms

      at Object.<anonymous> (tests/graphql.test.ts:494:13)

    console.log
      W/ dataloader: 101 ms

      at Object.<anonymous> (tests/graphql.test.ts:495:13)
```

I thought about adding some performance tests but with the parallel nature of jest the results would tend to wobble quite a bit and I'm not sure if it's possible to force a worker to run lonely.

B4nan:
> It also would be nice to test that only query per Entity type gets fired by Reference.load/Collection.load, but I'm not sure if it's possible to retrieve that information from mikro-orm.

I am usually using the `SimpleLogger` and `mockLogger` helper to check the queries, take a look at how the existing tests do it.

B4nan (commented):
few initial comments.

sorry for not getting back to this earlier...

darkbasic:
@B4nan it's rebased and everything has been addressed (expect for a conversation where I have a pending question).

B4nan:
Looks like the lockfile needs to be regenerated. Also we should document this somewhere, I guess a new section about dataloaders? Could be basic for now, but I'd like to have at least something.

darkbasic:
@B4nan the lock file has been regenerated and documentation has been created. I've also added a new global option to surgically enable each dataloader.

packages/core/src/utils/Configuration.ts
```diff
@@ -525,6 +526,7 @@ export interface MikroORMOptions<D extends IDatabaseDriver = IDatabaseDriver> ex
   batchSize: number;
   hydrator: HydratorConstructor;
   loadStrategy: LoadStrategy | 'select-in' | 'joined';
+  dataloader: Dataloader;
```
B4nan, line 529: I would allow boolean values here too, `true` means `ALL`, `false` means `OFF`

B4nan (changes requested):
two final things before we merge

darkbasic:
@B4nan done, should be ready to merge.

docs/docs/dataloaders.md
```diff
@@ -0,0 +1,49 @@
...
+
+```ts
+import { Dataloader } from '@mikro-orm/core';
+
+MikroORM.init({
+  dataloader: Dataloader.ALL,
+});
+```
+
+`Dataloader.REF` enables the dataloader for References, `Dataloader.COL` enables it for Collections while `Dataloader.ALL` enables it for both.
```
B4nan, line 20: old names still here

packages/core/src/entity/Collection.ts
```diff
@@ -258,6 +258,27 @@ export class Collection<T extends object, O extends object = object> extends Arr
       return this as unknown as LoadedCollection<Loaded<TT, P>>;
     }
 
+    if (options.dataloader ?? (DataloaderUtils.getDataloaderType(this.getEntityManager().config.get('dataloader')) > Dataloader.COLLECTION)) {
```
B4nan, line 261: this is pretty unreadable, also i would rather not use enums as numbers, this could backfire

B4nan (approved):
all right, thanks for holding on with me!

Issue #3750, opened 2022-11-15
Iterating through an uninitialized collection should throw an error


parisholley:
**Is your feature request related to a problem? Please describe.**
I have run into a few instances where I've had business logic failures because of not being aware that I was iterating through a collection that had not been fetched. understandably this may be a breaking change to save for v6 (perhaps in v5 we make it a flag that is passed down to collections)

```typescript
const collection: Collection<Entity>;

// throws
collection.getItems();

// does not throw
collection[0];

// does not throw
Array.from(collection)[0];
```

B4nan:
Implemented in v6 via 1260e588ac74521ef565f0271354ca0e1efe566d

PR #4324, opened 2023-05-05
Reproduction for #4319

parisholley:


parisholley:
i should note that this reproduction also is a tie in to #3750 , if you were to replace my getItems() map with for loops, all the tests would past. i'd really like to see that added to v6 :)

B4nan:
So I guess the place that needs synchronization is `em.populate`.

> i'd really like to see that added to v6 :)

We can do that at least opt-in, I dont mind that.

B4nan:
I think the issue is kinda twofold, synchronization of `em.populate` and the fact that the `init` methods of both `Collection` and `WrappedEntity` are not using `em.populate` but a custom implementation.

When I use `em.populate` in the failing test, it actually works already, without any other changes:

```ts
await Promise.all([
    orm.em
      .findOneOrFail(
        Service,
        { id: service.id },
        { populate: ['vendor.rules'] },
      )
      .then(async service => {
        // changed here to use `em.populate()`:
        await orm.em.populate(service, ['properties.conditions']);
        // const properties = await service.properties.load({
        //   populate: ['conditions'],
        // });

        service.properties.getItems().map(f => f.conditions.getItems().map(c => c));
      }),
    orm.em.findOneOrFail(Service, service.id, {
      populate: ['properties'],
    }),
  ]);
```

parisholley:
@B4nan changed a little bit to prevent some hanging:

https://github.com/mikro-orm/mikro-orm/pull/4324/commits/e1933deb439a72b964268c62c35b285e95e2bf57

one of the "gotchas" of this implementation is there are bound to be potential deadlocks (especially around bidirectional population). in `events.mysql.test.ts`, loading happens in this order `books > publisher > books`. I was able to add a check for `populate.all = true` to cover this test, but could certainly be more.

assuming my approach is the best way to tackle this problem, we'll probably need to consider changing the populate implementation to work one level at a time, eg: `books > publishers`, then check to see if there are any "unresolved" populates, then go the next layer down.

B4nan:
You will need to recreate the PR I guess, at least that's what I am doing in such cases, it's usually simpler than battling with git. First, reset your changes, then stash them and create the branch from the new v6 head. You could probably use hard reset instead too. If you don't want to deal with this, keep targeting master instead - the v6 branch will be rebased once in a while, I need to keep that updated, otherwise, it will be a major pain to sync it later. At some point, I will merge it to master and create 5.x branch instead, so the development will go the other way around, cherry-picking fixes from v6 back to v5, wanted to do that soon, but the work on v6 got derailed with v5 fixes again.

I don't mind the changes you did in general (e.g. doing the population one level at a time), although some of them are looking a bit suspicious, like the last change with the `populate.all` seems a bit weird, e.g. why would you even store empty strings to the `__populating` set? (https://github.com/mikro-orm/mikro-orm/commit/e1933deb439a72b964268c62c35b285e95e2bf57#diff-a62cd78c400d99253a3c0fb5a9bf26e1e63e0deb511059d28ba51014b58242ddR216)

As long as things keep working the way they work now, I am open to changing the internals a bit.

B4nan:
FYI the rebasing is getting more and more painful, I will probably merge the v6 branch today/tomorrow to master.

PR #4331, opened 2023-05-08
build: use `yarn` instead of `lerna` as a task runner

darkbasic:
Please note that I had to change the top level build and clean scripts to build-all and clean-all, otherwise the workspace-tools plugin would call the top level build script as well. This might break CI, so please let me know if/what scripts I have to update.
Instead of installing the workspace-tools plugin we might as well upgrade to yarn v4 (`yarn set version canary`) where it is built-in. The advantage would be that the newer plugin is better (has progress output, parallel builds, etc..) but I guess we will have to bundle a copy of yarn v4 instead of relying on corepack to provide it.

### DIFF
--- a/docs/yarn.lock
+++ b/docs/yarn.lock
@@ -2,15 +2,15 @@
 # Manual changes might be lost - proceed with caution!
 
 __metadata:
-  version: 6
-  cacheKey: 8
+  version: 7
+  cacheKey: 9
 
 "@algolia/autocomplete-core@npm:1.7.1":
   version: 1.7.1
   resolution: "@algolia/autocomplete-core@npm:1.7.1"
   dependencies:
-    "@algolia/autocomplete-shared": 1.7.1
-  checksum: 511176e9c2a9f2e2be62552e48e72dadfcc6638cda4a2990fd3453aed3ce4e7d8ca1bd6a9ccb912430c77734b00a8b836aaad97facc1987157af4ac00f590f4a
+    "@algolia/autocomplete-shared": "npm:1.7.1"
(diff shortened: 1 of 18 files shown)


darkbasic:
Nice, I've switched to v4 then and made the build parallel. I've also changed the script name in github workflows, so let's see if CI is happy now.

B4nan:
Nice, thanks! Would be great if we could keep the `build` script name, maybe @merceyz will know some flag we could use instead of renaming the top-level scripts? :]

Not a huge deal though.

darkbasic:
That's really weird because I've just upgraded accounts.js to latest lerna 6.6.2 and apparently it works like a charm on ppc64le:
```
 >  Lerna (powered by Nx)   Successfully ran target compile for 28 projects (57s)
```

I guess this is because accounts.js uses pnpm instead of yarn and thus it hits different paths, but honestly I have no idea.
I'll add that information on the lerna/nx bug report (but they never answered so I don't have much hope) and I'll file another one on the yarn side to ask if it's possible to skip the root level scripts.

darkbasic:
Apparently there is no need file a bug upstream at yarn, because since v4 it is smart enough to exclude the calling script without any additional modification. If this can now be merged I'll rebase the dataloader PR as well.

B4nan (changes requested):
The docs will need a lockfile update as well

merceyz:
> Apparently there is no need file a bug upstream at yarn, because since v4 it is smart enough to exclude the calling script without any additional modification.

That has almost always been the case, at least since https://github.com/yarnpkg/berry/pull/216, if you're able to get it to run recursively please file a bug upstream 👍

darkbasic:
> The docs will need a lockfile update as well

Ops, done that.

> That has almost always been the case, at least since https://github.com/yarnpkg/berry/pull/216

That's weird because I was definitely using latest 3.5. Maybe it ended up fetching an outdated version of the plugin or who knows what else, but it works like a charm in v4.

PR #4349, opened 2023-05-13
chore: add notes to migrations and schema generation

micahbule:
This PR just adds notes to migration and schema generator docs regarding the usage of both

### DIFF
--- a/docs/docs/migrations.md
+++ b/docs/docs/migrations.md
@@ -1,6 +1,7 @@
 ---
 title: Migrations
 ---
+> Note: Migrations are usually used for any schema changes in an existing database. If you are deploying a service/application that uses MikroORM on a new database, you might want to run the [Schema Generator](schema-generator.md) instead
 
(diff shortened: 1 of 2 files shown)


B4nan:
this is a weird suggestion, you should be using migrations in anything that is not a toy project, and i dont see any reason to suggest schema generator over migrations for anything else than toy projects or early prototyping (before you let others in, before you push anything anywhere/). 

to me, this addition is quite confusing and not really correct.

rubiin:
> this is a weird suggestion, you should be using migrations in anything that is not a toy project, and i dont see any reason to suggest schema generator over migrations for anything else than toy projects or early prototyping (before you let others in, before you push anything anywhere/).
> 
> to me, this addition is quite confusing and not really correct.

I too agree with this. On large projects, suggesting something like from the official docs will create a backlash

B4nan:
I will close this, open for improving the docs, but this seems to be rather about a weird behavior that caused confusion (more about it [here](https://github.com/mikro-orm/mikro-orm/discussions/2243#discussioncomment-5891271)).

Issue #4371, opened 2023-05-20
Property 'fieldName' option and ORM 'namingStrategy' do not affect Embeddable for MongoDB


funduck:
**Describe the bug**
Use MongoDB driver  
Use `@Property({fieldName: 'something'})` in embedded entity or choose UnderscoreNamingStrategy for ORM
In database embedded entity will have same naming as in code.

**To Reproduce**
```
import { Embeddable, Embedded, Entity, MikroORM, PrimaryKey, Property, UnderscoreNamingStrategy } from '@mikro-orm/core';
import { MongoDriver } from '@mikro-orm/mongodb';

@Embeddable()
class E {
  @Property()
  camelCase: string = 'c';

  @Property({fieldName: 'withALIAS'})
  withAlias: string = 'w';
}

@Entity()
class A {
  @PrimaryKey()
  _id = '1';

  @Property()
  complexName = 'n';

  @Embedded({entity: () => E, object: true})
  emBedded = new E()
}

let orm: MikroORM<MongoDriver>;

describe('GH issue 0', () => {
  beforeAll(async () => {
    orm = await MikroORM.init({
      entities: [A],
      clientUrl: 'mongodb://localhost:27017/mikro-orm-test',
      driver: MongoDriver,
      namingStrategy: UnderscoreNamingStrategy,
    });
    await orm.schema.clearDatabase();

    await orm.em.persistAndFlush(new A());
    orm.em.clear();
  });

  afterAll(async () => {
    await orm.close();
  });

  test('Ensure that embedded entity has underscore naming and fieldName is applied', async () => {
    const collection = orm.em.getDriver().getConnection().getDb().collection('a');
    expect(await collection.findOne({},{projection:{_id:0}})).toEqual({
      complex_name: 'n',
      em_bedded: {
        camel_case: 'c',
        withALIAS: 'w'
      }
    });
  });
});
```
**Expected behavior**
Renaming should propagate to the depth of embedded documents.

**Versions**

| Dependency | Version |
| - | - |
| node | 18 |
| typescript | 5 |
| mikro-orm | 5.7 |
| your-driver | MongoDB |

funduck:
@B4nan I'm preparing PR but I'm not sure how `object` options should work.  
In test `tests/features/embeddables/nested-embeddables.mongo.test.ts` I see that first `@Embedded` has `object: true` and it's effect propagates to nested embedded entities despite they dont have `object: true` option.  
[...]

PR #4372, opened 2023-05-20
fix(mongo): renaming propagates to embedded entities

funduck:
#4371

Here I solve two problems:
* renameFields recursively walks all embedded entities
* renameFields carries `object` mode flag so first `object: true` will save all nested embeddables as objects ignoring their `meta`


B4nan:
I think we should leave this for v6, it feels a bit breaking.

funduck:
I see now it is more complicated, I implemented "write" part, but missed the "read".. so I'll get back to it a bit later.

funduck:
And for now I will close PR because it is not ready at all

PR #4381, opened 2023-05-23
Docs: add note about hooks calls on relations

VoytechG:
Having a One-To-Many relation and adding an entity instance to a collection doesn't trigger any event on the owning side of the relation. For first-time users this might be confusing. It is worthwhile to add a note about this to the documentation.

I have drafted a basic explanation of the topic, but it requires to be refined. I wrote what I deduced about how this logic works. Help from those who are familiar with this logic is welcome.

### DIFF
--- a/docs/versioned_docs/version-5.7/events.md
+++ b/docs/versioned_docs/version-5.7/events.md
@@ -36,6 +36,15 @@ All hooks support async methods with one exception - `@OnInit`.
 
 > `@OnInit` can be sometimes fired twice, once when the entity reference is created, and once after its populated. To distinguish between those we can use `wrap(this).isInitialized()`.
 
+## Hook calls on relation updates
+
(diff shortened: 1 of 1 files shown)


B4nan (commented):
I understand what you want to say in there, but the way you tried to describe it feels pretty confusing.

The main problem here is not that "1:m collections don't trigger hooks", it's about "1:m collections are virtual, they just propagate to the m:1 owning side - that is what fires hooks". It's the very same problem as with 1:1 inverse side, or in other words, with any inverse side.

There are also syntactic errors and typing errors, plus inconsistencies (e.g. there is no place in the docs where event names are wrapped in double quotes, also the event names are not "Update" but "beforeUpdate" and "afterUpdate", etc).

I will keep this open as a reminder, and will try to add some section myself, then we can sync on how to improve that more.

PR #4388, opened 2023-05-25
fix(cli): expose missing `migration:check` command

stefansundin:
Please let me know if you want me to make any changes. Feel free to make edits to the branch directly if you prefer that.

### DIFF
--- a/packages/cli/src/commands/MigrationCommandFactory.ts
+++ b/packages/cli/src/commands/MigrationCommandFactory.ts
@@ -80,7 +80,7 @@ export class MigrationCommandFactory {
       type: 'string',
       desc: 'Sets path to directory where to save entities',
     });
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/MigrationCommandFactory.ts
```diff
@@ -80,7 +80,7 @@ export class MigrationCommandFactory {
       type: 'string',
       desc: 'Sets path to directory where to save entities',
     });
-    args.option('name', {
+    args.option('n', {
       alias: 'name',
```
stefansundin, line 84: Here's an alternative fix if you don't want a short option.

```suggestion
    args.option('name', {
```

B4nan (approved):
thanks!

PR #4392, opened 2023-05-26
feat(cli): improve `migration:check` command to also print the schema diff

stefansundin:
I think the `migration:check` command would be better if it also reported what the diff is. I am planning to run it in CI to ensure there isn't any migration drift, and I think it is helpful to have more information when it errors.

Example output:

Screenshot 2023-05-25 at 16 46 06

This isn't super polished as I am not sure exactly how you'd prefer for it to work, so please let me know if there's any changes you'd like to see. Thanks!

### DIFF
--- a/packages/migrations-mongodb/src/Migrator.ts
+++ b/packages/migrations-mongodb/src/Migrator.ts
@@ -56,6 +56,10 @@ export class Migrator implements IMigrator {
     return true;
   }
 
+  async getSchemaDiff(blank: boolean, initial: boolean): Promise<{ up: string[]; down: string[] }> {
+    return { up: [], down: [] };
+  }
+
   /**
    * @inheritDoc
    */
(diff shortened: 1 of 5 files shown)


packages/migrations-mongodb/src/Migrator.ts
```diff
@@ -56,6 +56,10 @@ export class Migrator implements IMigrator {
     return true;
   }
 
+  async getSchemaDiff(blank: boolean, initial: boolean): Promise<{ up: string[]; down: string[] }> {
+    return { up: [], down: [] };
+  }
```
stefansundin, line 61: This is probably wrong. I don't use mongodb so I don't know what it should be.

B4nan:
Any reason you don't just use `schema:update --dump`?

stefansundin:
> Any reason you don't just use `schema:update --dump`?

I didn't know about that command. It looks like that should work for my use case. Feel free to close this PR then. 👍

rubiin:
this can be closed i think

PR #4394, opened 2023-05-28
feat(core): add option to exclude fields from upsert

### LINKED ISSUE #4325: Allow specifying which fields to exclude from update in upsert/upsertMany functions
Currently, when using the entity manager's `em.upsert` or `em.upsertMany` functions to upsert data, all fields other than the unique ones are included in the update. However, sometimes we need to keep some of the fields as they are if the row already exists, but insert a new value for them if it doesn't.

I would like to be able to specify which fields to exclude from the on conflict update clause. Maybewe can add an `excludeFieldsOnUpdate` option in the `NativeInsertUpdateOptions` which is an array of field names that should be kept as is if the row already exists, but inserted with new values if it doesn't. 

The proposed option looks like:
```ts
const author = await em.upsertMany(Author, [
  { email: 'foo@bar.com', name:"Foo Bar", age: 33 },
  { email: 'lol@lol.lol', name:"John Doe", age: 666 },
], {excludeFieldsOnUpdate:["name"]});
```

Some alternatives are to use the query builder or a raw SQL query instead of the ORM's `em.upsert` or `em.upsertMany` functions, but it would be nice to have the option through the EntityManager.

**Additional context**
This is based on discussion #4317

jsprw:
Closes #4325.

@B4nan could you have a look at the 3 failing tests for Mongo? It seems that `em.refresh` does not handle `null` values from database correctly. When it does not have a null value, it refreshes correcly, see: `tests/features/upsert/upsert.test.ts#460`. 

It works as follows:
- if the entity is found in the UoW, it is assigned all data except fields from `upsertExcludeFields` (`packages/core/src/EntityManager.ts#586`).
- if a managed entity is passed as data, it's fields will be refreshed (`packages/core/src/EntityManager.ts#570`).
- if the entity does not exist in the local cache, it will fire the query with the fields excluded (`packages/knex/src/AbstractSqlDriver.ts#428`). On platforms with that support returning statements, it will use this data to handle the excluded fields (`packages/core/src/EntityManager.ts#639`). If the platform does not support returning statements, it will refresh the ignored fields with the database (`packages/core/src/EntityManager.ts#654`).


packages/core/src/EntityManager.ts
```diff
@@ -565,6 +566,11 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
 
       if (helper(entity).__managed && helper(entity).__em === em) {
         em.entityFactory.mergeData(meta, entity, data, { initialized: true });
+
+        if (options.excludeFields?.length) {
+          await this.refresh(entity, { fields: options.excludeFields });
```
B4nan, line 571: This feels weird, can you explain why the `refresh` call? 

I think I understand what you are trying to do here, but `refresh` is not the right API, it can wipe state, and more importantly - it feels like you are mixing two features/fixes into one PR, this is not really about making parts of the upsert call configurable, right?

jsprw: When you exclude fields from an upsert, the data from the database should be assigned to the entity we're trying to persist. If you don't refresh the fields, the ignored data will be assigned to the entity.

In this case, I assumed that managed entities always exist in the database, so any changes to that field should be refreshed from the database.

You can see the test case below, when it does not exist it should insert with propName = val1, if it exists it should ignore propName and therefore it should stay null. But when you comment out that refresh call, the value of 'propName' will be 'val1'.

https://github.com/mikro-orm/mikro-orm/pull/4394/files#diff-c2487ec58d0e5b8ce63bacb5d1c08238b19d6f78a1615b54fda7baddf825f5a6R486

jsprw: Another question, what does [this line](https://github.com/mikro-orm/mikro-orm/blob/c8472be948465c7c237b21473e0aa42a43a4f11e/packages/core/src/EntityManager.ts#L565) do? It initialises the entity? It does not merge anything either as [entity and data are the same object](https://github.com/mikro-orm/mikro-orm/blob/c8472be948465c7c237b21473e0aa42a43a4f11e/packages/core/src/EntityManager.ts#L568).

B4nan: I would say you should remove the fields from `data` before calling `assign`, still not sure why would you need the refresh call. And as I said, I don't think it's the right API in the first place, it was added for testing usage mostly, refreshing entity state is not really something to do in production. The `mergeData` call is an alternative to `refresh` that is not wiping the state, so seeing those two calls next to each other is what I don't understand at all.

> Another question, what does [this line](https://github.com/mikro-orm/mikro-orm/blob/c8472be948465c7c237b21473e0aa42a43a4f11e/packages/core/src/EntityManager.ts#L565) do? It initialises the entity? It does not merge anything either as [entity and data are the same object](https://github.com/mikro-orm/mikro-orm/blob/c8472be948465c7c237b21473e0aa42a43a4f11e/packages/core/src/EntityManager.ts#L568).

Not sure what exactly are you asking here? That line just handles the different signature of upsert, as you can pass entity instances in the first param, or the entity type and use POJOs.

jsprw: When you look at line 565, you do this:

```
      entity = data as Entity;
      if (helper(entity).__managed && helper(entity).__em === em) {
        em.entityFactory.mergeData(meta, entity, data, { initialized: true });
```

I don't understand how this refreshes the fields as it merges with the same object?

B4nan: It's not refreshing anything, upsert is not about refreshing things.

B4nan: I am not entirely sure what the line is doing, but my guess is it is normalizing the entity state, adding things to the `loadedProperties` and `originalEntityData`. You could comment it out and run tests to see, maybe its not needed but I guess it is there for a reason :]

jsprw: But without refreshing, how will you know the existing value of the excluded field?

When the database record is like this:
| id (unique) | prop1 | prop2 | prop3 |
|---|---|---|---|
|  1 |  a  | b  | c  |
| 2  |  d | e  | f  |

And you upsert the following line `id = 1, prop1 = g, prop2 = h, prop3 = i` while excluding prop3 from the update. You would expect the resulting table to be:

 | id (unique) | prop1 | prop2 | prop3 |
|---|---|---|---|
|  1 |  g  | h  | c  |
| 2  |  d | e  | f  |

How will we know that prop3 has value `c` without checking the database?

I modified the method to exclude these properties from the data but could you have look at how to fix the failing tests?

jsprw: > Maybe I just didn't understand what the issue is about, I thought it was only about being able to configure what the upsert query will look like.

Yeah, that's main question. But when you change the upsert query, it will not update the entities in the UoW and then you will end up with out-of-sync entities as you can see above with the tables. Therefore this issue is more complex than just configuring the upsert query.

> Well, that branch is for managed entities, I was expecting we are talking about those that are already loaded, so all the values would be there.

The data is overwritten for this entity. When you create an entity with an id that already exists in the UoW with `em.create` -> `packages/core/src/entity/EntityFactory.ts#60`, it will overwrite the data outside of the `upsert` scope. Unless I miss something, the original data is not there anymore for these entities?

> With that said, my problem from the very beginning is that I don't think you should be using em.refresh there - not saying we can't (re)fetch some data.

Could you give a hint with what would be a better approach besides refresh? I just need the original values for all excluded fields.

B4nan: I checked your tests again, and now I think where the confusion is coming.

https://github.com/mikro-orm/mikro-orm/pull/4394/files#diff-c2487ec58d0e5b8ce63bacb5d1c08238b19d6f78a1615b54fda7baddf825f5a6R490-R493

This is not doing what you think, the `em.create` call will give you the already loaded entity from the context, as it respects the identity map and you are passing in the PK. The `factory.mergeData` will be used to modify its state. You are overriding your loaded entity, that is why the old value is no longer there. 

I'd say you should be doing this instead:

```ts
// use the POJO directly in em.upsert, dont create the entity instance explicitly
const fooBar1 = await orm.em.upsert(FooBar, { id: 1, name: 'fb1', author: 1, propName: 'val 1' });
```

jsprw: Yeah exactly. So we don't want to support this particular case where you can pass in a loaded entity?

For the POJO case the test works fine (except for Mongo): https://github.com/mikro-orm/mikro-orm/pull/4394/files#diff-c2487ec58d0e5b8ce63bacb5d1c08238b19d6f78a1615b54fda7baddf825f5a6R472-R484

B4nan: Imho that test is a misusage and you are trying to fix something in `upsert` that is not related to it - which is the `em.create` call messing up the entity state.

So yes, I don't think this is something to support. I don't think it makes sense to reload values when the user instructs something should be excluded from the update, it feels like the opposite of what the user is asking for. This is really a use case for the POJO signature, as you want to handle the entity data conditionally. 

I'd also want to first merge this to v6 branch to reduce conflicts, as the reloading of properties changed a bit there, and then we can discuss if and how to backport it to v5. I guess I will merge the v6 to master soon as the rebasing is getting painful and branch out 5.x separately.

I will give it a try myself to see if there is a better way to implement it (and maybe end up understanding that your approach is ok). I wanted to look at another (in a way pretty similar) problem in there, which is the not initialized entities in the context.

jsprw: > I'd also want to first merge this to v6 branch to reduce conflicts, as the reloading of properties changed a bit there, and then we can discuss if and how to backport it to v5. I guess I will merge the v6 to master soon as the rebasing is getting painful and branch out 5.x separately.

Haha yeah, when I rebased my branch for the full text MR I just gave up, it just kept going: just redid the whole PR. Looking forward to the changes in v6.

> So yes, I don't think this is something to support. I don't think it makes sense to reload values when the user instructs something should be excluded from the update, it feels like the opposite of what the user is asking for. This is really a use case for the POJO signature, as you want to handle the entity data conditionally.

Completely agreed. The only reason why I wanted to support this, is because it's done that way in other upsert tests as well: https://github.com/mikro-orm/mikro-orm/blob/ac5ecf36b86e746e603b54cbb582fb787413f2f0/tests/features/upsert/upsert.test.ts#LL365C1-L365C1.

> I will give it a try myself to see if there is a better way to implement it (and maybe end up understanding that your approach is ok). I wanted to look at another (in a way pretty similar) problem in there, which is the not initialized entities in the context.

Do you want me to do anything? Rebase it for v6? Or will you do it?

B4nan: > Do you want me to do anything? Rebase it for v6? Or will you do it?

Let me play with this first, I will have some time for it in a few hours.

PR #4396, opened 2023-05-29
feat(core): apply discriminator condition to relations

YanDjin:
This PR aims to apply the discriminator condition automatically when filtering on a relation that uses STI or when populating one using the joined strategy.

It works by looping on the conditions and populated fields and adding the conditions if needed.

### DIFF
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -39,7 +39,7 @@ import type {
   RequiredEntityData,
 } from './typings';
 import type { TransactionOptions } from './enums';
-import { EventType, FlushMode, LoadStrategy, LockMode, PopulateHint, ReferenceType, SCALAR_TYPES } from './enums';
+import { GroupOperator, EventType, FlushMode, LoadStrategy, LockMode, PopulateHint, ReferenceType, SCALAR_TYPES } from './enums';
 import type { MetadataStorage } from './metadata';
 import type { Transaction } from './connections';
 import type { FlushEventArgs } from './events';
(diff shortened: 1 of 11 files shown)


B4nan:
Can you please prepare an isolated reproduction first (like in `tests/issues`)? I'd rather not add yet another entity to the EM tests, which is what you did. It will slow down and clutter everything, I am rather moving away from those "fat setups" in favor of self-contained tests that define the entities inside the test file.

There is actually a fix in v6 that feels like it's doing the same: https://github.com/mikro-orm/mikro-orm/commit/45565bb164018a89aa51d5cb7b363551984fc316

So first let's create a repro, test how it works with v6, and if it does not, then please base the PR on v6 branch instead.

Also, I would call this a hot path, so we need to be careful with performance. The call to `preparePopulate` is definitely done twice this way, we don't want that. But I will be more than happy to help with that myself, just need to see the isolated repro first.

B4nan:
Amazing!

Yes, go for v6, I am pretty close to merging it to master (not releasing as stable yet, it will take some time to polish the docs). Once that happens, the website will also contain v6 changes, right now the docs changes for v6 are available only when browsing the v6 branch.

I hope v6 will be out during the summer, but I'd say it's already in better shape than v5. 

Here is a WIP of the upgrading guide, it mentioned all of the BCs:

https://github.com/mikro-orm/mikro-orm/blob/v6/docs/docs/upgrading-v5-to-v6.md

YanDjin:
Thank you very much for your help !

PR #4408, opened 2023-05-31
docs: postgres >= 13 has built-in  uuid function "get_random_uuid"

leonardbinet:
Following https://github.com/mikro-orm/mikro-orm/discussions/4403

Since Postgres 13 no additional module is necessary to use 'gen_random_uuid' function.

### DIFF
--- a/docs/docs/defining-entities.md
+++ b/docs/docs/defining-entities.md
@@ -1456,9 +1456,7 @@ export const Book = new EntitySchema<IBook>({
   </TabItem>
 </Tabs>
 
-### Using PostgreSQL [uuid-ossp](https://www.postgresql.org/docs/current/uuid-ossp.html) module function as primary key
-
-Requires enabling the module via: `create extension "uuid-ossp";`
+### Using PostgreSQL built-in [gen_random_uuid](https://www.postgresql.org/docs/current/functions-uuid.html) function as primary key
 
 <Tabs
 groupId="entity-def"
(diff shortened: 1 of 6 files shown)


B4nan (approved):
nice, didnt know about that

stefansundin:
You forgot to update these lines:

```typescript
  @PrimaryKey({ type: 'uuid', defaultRaw: 'uuid_generate_v4()' })
```

Should now be:

```typescript
  @PrimaryKey({ type: 'uuid', defaultRaw: 'get_random_uuid()' })
```

B4nan:
It actually looks like both are working nowadays? I was checking the tests that used the `uuid_generate_v4` and they are now passing even without enabling the uuid-ossp extension.

leonardbinet:
Ahh my bad!

B4nan:
> Should now be:

Damn, now I know why it wasn't working for me! You got a typo in there, it's `gen_random_uuid`, not `get_random_uuid`. And looks like both `gen_random_uuid` and `uuid_generate_v4` are supported nowadays.

leonardbinet:
I've renamed the pull-request title to `gen_random_uuid` indeed. Since both `gen_random_uuid ` and `uuid_generate_v4 ` work @B4nan what do you prefer in documentation? Maybe `gen_random_uuid` (because it is what is displayed in postgres documentation), with a disclaimer saying that `uuid_generate_v4 ` also works (for people already using it in their codebases, so they know why it works 😄 )?

B4nan:
I'd mention they both work since pg 13, and that in previous versions the extension was required.

stefansundin:
The only issue with `uuid_generate_v4()` is that you need to manually run `create extension if not exists "uuid-ossp";` in a migration, which isn't super obvious how to do if it is your first time. I think just recommending `gen_random_uuid()` in the docs is the way to go, and if someone needs more advanced uuid functions then they can hopefully find `uuid-ossp` on their own.

PR #4419, opened 2023-06-04
feat(core): support Uint8Array

### LINKED ISSUE #4418: Add support for typed arrays
Currently MikroORM works fine with Node.js `Buffer`. They are handled correctly in the `EntityDTO` type returned by the `toObject` method. Also `BlobType` can store them as expected in the database.

So far so good. But unfortunetely in browser applications we can only work with typed arrays. I'm working with a binary format (CBOR) in my API and data is transferred as `Uint8Array` since that is supported in both Node.js and browsers.

I propose to support at least Uint8Array in `BlobType` and `EntityDTO`. Since Node.js `Buffer` basically is a `Uint8Array` and it usually is the drop-in  replacement when browsers are involved.

Right now two extra conversion steps (incoming and outgoing) are necessary to make things run smoothly. The database side is not that bad actually. I created a custom type that is based on `BlobType`. The ougoign data on the other hand requires more boilerplate code if I don't want to do without type safety - which is one reason I picked MikroORM in the first place.

**Additional context**
I already made the rather trivial changes locally and can provide the necessary PR if that is helpful.

makuko:
The goal is to work properly with entities that have properties of type Uint8Array. Those properties should automatically be stored as binary data and converted back to Uint8Array without specifying a type (analogous to Buffer).

Also when converting entities to DTOs Uint8Array types should be kept unchanged.

Closes #4418

### DIFF
--- a/docs/docs/custom-types.md
+++ b/docs/docs/custom-types.md
@@ -300,7 +300,7 @@ Uint8Array type can be used to store binary data in the database.
 
 ```ts
 @Property({ type: Uint8ArrayType, nullable: true })
-blob?: Uint8ArrayType;
+blob?: Uint8Array;
 ```
 
 ### JsonType
--- a/docs/docs/custom-types.md
+++ b/docs/docs/custom-types.md
@@ -237,6 +237,7 @@ export const types = {
   datetime: DateTimeType,
   bigint: BigIntType,
   blob: BlobType,
+  uint8array: Uint8ArrayType,
   array: ArrayType,
   enumArray: EnumArrayType,
   enum: EnumType,
@@ -289,6 +290,17 @@ Blob type can be used to store binary data in the database.
 ```ts
 @Property({ type: BlobType, nullable: true })
 blob?: Buffer;
+```
+
+### Uint8ArrayType
+
+Uint8Array type can be used to store binary data in the database.
+
+> `Uint8ArrayType` will be used automatically if you specify the type hint as `Uint8Array`. This means that the following example should work even without the explicit `type: Uint8ArrayType` option (with both reflect-metadata and ts-morph providers).
+
+```ts
+@Property({ type: Uint8ArrayType, nullable: true })
+blob?: Uint8ArrayType;
 ```
 
 ### JsonType
--- a/packages/core/src/types/Uint8ArrayType.ts
+++ b/packages/core/src/types/Uint8ArrayType.ts
@@ -1,0 +1,42 @@
+import { Type } from './Type';
+import type { Platform } from '../platforms';
+import type { EntityProperty } from '../typings';
+
+export class Uint8ArrayType extends Type<Uint8Array | null> {
+
+  convertToDatabaseValue(value: Uint8Array): Buffer {
+    return Buffer.from(value);
+  }
+
+  convertToJSValue(value: Buffer): Uint8Array | null {
+    if (!value) {
+      return value;
+    }
+
+    /* istanbul ignore else */
+    if (value as unknown instanceof Buffer) {
+      return new Uint8Array(value);
+    }
+
+    /* istanbul ignore else */
+    if (value.buffer instanceof Buffer) {
+      return new Uint8Array(value.buffer);
+    }
+
+    /* istanbul ignore next */
+    return new Uint8Array(Buffer.from(value));
+  }
+
+  compareAsType(): string {
+    return 'Buffer';
+  }
+
+  ensureComparable(): boolean {
+    return false;
+  }
+
+  getColumnType(prop: EntityProperty, platform: Platform): string {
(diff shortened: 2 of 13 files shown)


rubiin (commented):
Tests seem to be missing.

packages/core/src/typings.ts
```diff
@@ -58,7 +58,7 @@ export type PrimaryProperty<T> = T extends { [PrimaryKeyProp]?: infer PK }
 export type IPrimaryKeyValue = number | string | bigint | Date | { toHexString(): string };
 export type IPrimaryKey<T extends IPrimaryKeyValue = IPrimaryKeyValue> = T;
 
-export type Scalar = boolean | number | string | bigint | symbol | Date | RegExp | Buffer | { toHexString(): string };
+export type Scalar = boolean | number | string | bigint | symbol | Date | RegExp | Uint8Array | { toHexString(): string };
```
B4nan, line 61: we want to keep `Buffer` here, or am I missing something?

packages/core/src/types/Uint8ArrayType.ts
```diff
@@ -0,0 +1,42 @@
+import { Type } from './Type';
+import type { Platform } from '../platforms';
+import type { EntityProperty } from '../typings';
+
+export class Uint8ArrayType extends Type<Uint8Array | null> {
```
B4nan, line 5: this could maybe extend the `BufferType` to reduce duplication

packages/core/src/typings.ts
makuko: `Buffer` extends `Uint8Array` so it should be fine without. I was thinking that maybe it even makes sense to use `ArrayBufferView` to make it work with all `TypedArray` and `DataView`.

packages/core/src/types/Uint8ArrayType.ts
makuko: Also, I just discovered that I can adjust the default type mapping in the configuration so this behaviour could easily be added in apps that actually need it. Maybe it's not even needed as feature for everyone. The important part is fixing the EntityDTO because that is rather a bug than a feature request.

B4nan: do you plan to finish this?

makuko: I just did. It was necessary to make the BlobType generic to achieve this. But that shouldn't be a problem.

packages/core/src/types/Uint8ArrayType.ts
```diff
@@ -1,8 +1,6 @@
-import { Type } from './Type';
-import type { Platform } from '../platforms';
-import type { EntityProperty } from '../typings';
+import { BlobType } from '.';
```
B4nan, line 1: this could cause unnecessary import cycles
```suggestion
import { BlobType } from './BlobType';
```

docs/docs/custom-types.md
```diff
@@ -291,6 +292,17 @@ Blob type can be used to store binary data in the database.
...
 
+### Uint8ArrayType
+
+Uint8Array type can be used to store binary data in the database.
+
+> `Uint8ArrayType` will be used automatically if you specify the type hint as `Uint8Array`. This means that the following example should work even without the explicit `type: Uint8ArrayType` option (with both reflect-metadata and ts-morph providers).
+
+```ts
+@Property({ type: Uint8ArrayType, nullable: true })
+blob?: Uint8ArrayType;
```
B4nan, line 303: the mapping should be `blob?: Uint8Array`, right?

docs/docs/custom-types.md
```diff
@@ -291,6 +292,17 @@ Blob type can be used to store binary data in the database.
 blob?: Buffer;
 ```
 
+### Uint8ArrayType
+
+Uint8Array type can be used to store binary data in the database.
+
+> `Uint8ArrayType` will be used automatically if you specify the type hint as `Uint8Array`. This means that the following example should work even without the explicit `type: Uint8ArrayType` option (with both reflect-metadata and ts-morph providers).
```
B4nan, line 299: I don't think we should suggest this

docs/docs/custom-types.md
makuko: Of course, you're right.  Fixed.

B4nan (approved):
thanks!

B4nan:
You will need to regenerate all the snapshots, run `yarn test -u` to do so. Or create a separate (isolated) feature test instead of modifying the existing entities.

B4nan:
Ups, misclick. And I see its quite broken, not just the snapshots...

B4nan:
[7793b82](https://github.com/mikro-orm/mikro-orm/commit/7793b82fd72c705634820d6551bf0cd374097fa3) should help

PR #4427, opened 2023-06-07
fix(core): fix returning clause for upsert with embeddables

jsprw:
…hen upserting with embeddables

Currently, when you run upsert on an entity with an embeddable, the name of the embeddable is included in the returning statement and causing `InvalidFieldNameException`.

The produced query without the change:
``` 
insert into `foo_bar_with_embeddable` (`_id`) values (1) on conflict (`_id`) do nothing returning `fooBarEmbeddable`, `foo_bar_embeddable_name`;
````

### DIFF
--- a/tests/features/upsert/upsert.test.ts
+++ b/tests/features/upsert/upsert.test.ts
@@ -474,13 +474,13 @@ describe.each(Object.keys(options))('em.upsert [%s]',  type => {
   test('em.upsert(entity) with embeddable', async () => {
     const testEntity = orm.em.create(FooBarWithEmbeddable, { fooBarEmbeddable: {} });
 
-    const insertedEntity = await orm.em.upsert(FooBarWithEmbeddable, testEntity);
+    await orm.em.upsert(testEntity);
 
-    expect(insertedEntity.id).toBeDefined();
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/features/upsert/upsert.test.ts [hunk 1]
```diff
@@ -449,4 +470,17 @@ describe.each(Object.keys(options))('em.upsert [%s]',  type => {
     await assertFooBars([fooBar1, fooBar2, fooBar3], mock);
   });
 
+
+  test('em.upsert(entity) with embeddable', async () => {
+    const testEntity = orm.em.create(FooBarWithEmbeddable, { fooBarEmbeddable: {} });
+
+    const insertedEntity = await orm.em.upsert(FooBarWithEmbeddable, testEntity);
```
B4nan, line 477: since you pass in entity instance you can just do `upsert(testEntity)`, also `testEntity === insertedEntity`, the return value makes sense only with the POJO signature

jsprw: Fixed it, basically anything can be done here. It just wanted to test that the upsert and upsertMany queries run successfully (not throw an InvalidFieldNameException).

B4nan (approved):

Issue #4423, opened 2023-06-05
V6. When populating a Many-to-Many relationship that uses STI, the ORM joins on the wrong column of the pivot table


YanDjin:
**Describe the bug**
On the version 6, when populating a Many-to-Many relation that uses STI (i did not specify a strategy), the ORM joins on the wrong column of the pivot table.
I tried the same with a relation that does not use STI and it does not seem to have the same issue.

**To Reproduce**
- use an entity that has a Many-to-Many relationship with an entity that uses STI
- use a find on the entity that owns the relation
- specify the relation to be populated, without specifying a strategy
- when loading the relation, the ORM will join on the wrong column

https://github.com/YanDjin/mikro-orm/blob/bug/find-sti-relation/tests/issues/GH4423.test.ts
this link contains the test that replicates the issue.

**Expected behavior**
To join the populate request on the correct column of the pivot table

**Versions**

| Dependency | Version |
| - | - |
| node | 18.16.0 |
| typescript | 5.0.4 |
| mikro-orm | 6.0.0-dev.82 |
| your-driver | postgres/sqlite |

YanDjin:
maybe related to #4422

B4nan:
This one is a bit more tricky, I guess I will try to do a bigger refactor to get rid of the weird `loadFromPivotTable` driver method and use `em.populate` instead.

YanDjin:
I just happened to add a hotfix PR, if you have the time to check it out :)

B4nan:
I managed to remove the weird parts from QB (together with this problematic `qb.autoJoinPivotTable` method), still have a few failing tests to deal with, but this new solution does not suffer from the problem you have in the PR.

PR #4437, opened 2023-06-10
fix(knex): joining on wrong column of pivot table when using STI

YanDjin:
#4423 

The bug is due to this._cond containing the condition not at the root but inside an $and group operator because of the discriminator condition that is added.

But, I have noticed that the discriminator condition is applied twice, once by applyDiscriminatorCondition and the other one is added on the join.


B4nan:
> But, I have noticed that the discriminator condition is applied twice, once by applyDiscriminatorCondition and the other one is added on the join.

Check the `qb.processPopulateWhere` method, IIRC that is what converts those to the `on` conditions.

Issue #4433, opened 2023-06-08
V6 - Incorrect query when using '<Ref | Collections>.*'


xmuller:
**Describe the bug**

The generated queries from a `find` with a `where.fields`  using '<relation>.*'  to not populate all the fields of the relations which seems to be wrong. 

**To Reproduce**
1. create the following two entities:

```ts
@Entity()
class A {
  @PrimaryKey()
  id: number;
  
  @Property({
    type: "text",
  })
  description: string;
  
  @ManyToOne(() => B)
  b: Ref<B>;
}

@Entity()
class B {
  @PrimaryKey()
  id: number;
  
  @Property({
    type: "text",
  })
  name: string;
}
```

2. do the following query

```ts
    const result = await this.em.findAll(A, {
      fields: ["b.*"],
      logging: { enabled: true }
    });
```

3. here is the generated query

```
select "s0"."id", "s0"."b_id" from "a_entity" as "s0"
```

**Expected behavior**

I think that the following query should be generated if i understand the api correctly:

```
select "s0".*, "h1"."id" as "b_id" from "b_entity" as "s0" left join "a_entity" as "h1" on "s0"."id" = "h1"."b_id"
```

**Additional context**

Also when i am population the relation like this and try to get all fields like the following:

```ts
    const result = await this.em.findAll(A, {
      populate: ["b"],
      fields: ["*"],
      logging: { enabled: true }
    });
```

Here is the response:
```ts
{
  id: "....",
  b: {
    id: "..."
  }
}
```

Here is what I expected:
```ts
{
  id: "....",
  description: "..."
  b: {
    id: "...",
    name: "..."
  }
}
```

This issue may have the same origin as the one presented earlier.

**Versions**

| Dependency | Version |
| - | - |
| node | 18 |
| typescript | 5.0.4 |
| mikro-orm | 6.0.0-dev.82 |
| your-driver | postgresql |

B4nan:
> I think that the following query should be generated if i understand the api correctly:

You need to enable joined strategy for that, by default it should issue two queries, one for each relation.

B4nan:
Also you are missing `ref: true` in your entity definition (unless you are using ts-morph).

xmuller:
> > I think that the following query should be generated if i understand the api correctly:
> 
> You need to enable joined strategy for that, by default it should issue two queries, one for each relation.
> 
> edit: or is that the second query only? why do you expect the join in there?

[...]

PR #4445, opened 2023-06-12
fix(#4433): serialize the whole nested field when '*' is present

xmuller:
This is my attempt to fix the serialization issue of #4433.


packages/core/src/serialization/SerializationContext.ts
```diff
@@ -78,6 +78,9 @@ export class SerializationContext<T> {
   isMarkedAsPopulated(entityName: string, prop: string): boolean {
     let populate: PopulateOptions<T>[] | undefined = this.populate;
 
+    if (populate.length === 1 && populate[0].field === '*') {
```
B4nan, line 81: maybe this would be better this way? i think there are valid use cases for `*` on different positions too

```suggestion
    if (populate.some(({ field }) === '*') {
```

packages/core/src/serialization/SerializationContext.ts
```diff
@@ -110,12 +113,16 @@ export class SerializationContext<T> {
         return true;
       }
 
+      if (fields.length === 1 && fields[0] === '*') {
```
B4nan, line 116: same here i guess

B4nan:
I was wrong, this needs to be fixed in the serialization context, but I believe I found a bit better place (61de246).

PR #4446, opened 2023-06-13
refactor(core): keep v5 `serialize` api

xmuller:
This pull request aims to fix the intellisense issue with `serialize` in V5 while keeping the v5 API.

related: https://github.com/mikro-orm/mikro-orm/blob/v6/docs/docs/upgrading-v5-to-v6.md#serialize-helper-always-returns-array

What do you think about this ? 
Also should this pull request be applied to master ?

### DIFF
--- a/packages/core/src/serialization/EntitySerializer.ts
+++ b/packages/core/src/serialization/EntitySerializer.ts
@@ -1,5 +1,6 @@
 import type { Collection } from '../entity/Collection';
 import type {
+  ArrayElement,
   AutoPath,
   Dictionary,
   EntityDTO,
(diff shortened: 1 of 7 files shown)


B4nan:
This is amazing, thank you! I spent hours trying to make this works. Yes, we should do the same for v5, and for `em.populate` too, it has exactly the same problem (in v6 it should behave the same, changing the return type would be breaking in v5).

packages/core/src/serialization/EntitySerializer.ts
```diff
@@ -238,19 +239,20 @@ export interface SerializeOptions<T, P extends string = never, E extends string
...
+ * Converts entity instance to POJO, converting the `Collection`s to arrays and unwrapping the `Reference` wrapper, while respecting the serialization options.
+ */
+export function serialize<T extends object, P extends string = never, E extends string = never>(entity: T, options?: SerializeOptions<T, P, E>): T extends object[] ? EntityDTO<Loaded<ArrayElement<T>, P>>[] : EntityDTO<Loaded<T, P>>;
 
 /**
  * Converts entity instance to POJO, converting the `Collection`s to arrays and unwrapping the `Reference` wrapper, while respecting the serialization options.
- * This method accepts either a single entity or array of entities, and always returns an array of entities. To serialize single entity, you can use array destructing,
- * or use `wrap(entity).serialize()` which handles a single entity only.
- *
- * ```ts
```
B4nan, line 247: i'd like to keep some example, it should show that the method returns different things based on the argument. and we should also mention it in the comment as it was before - so instead of removing, adjust the comment and example

xmuller: Oh sure, I will do it.

B4nan:
This does not seem to be working, I see you were concerned only with the return type, but the real problem why I moved away from this signature is the intellisense/autocomplete on the `populate` option - and that is still broken with your changes. Try using `serialize` with an array parameter and see how the `populate` parameter won't suggest anything nested. It will actually fail with an error:

image

With single entity it works fine:

image
image

xmuller:
I forgot to pass the correct type to  `SerializeOptions`, this is fix !

xmuller:
Side note, shouldn't the intelissence/autocomplete on the `populate` option only expose relation fields ?

If I understand the API correctly the following is equivalent
* `populate: [ 'coverImage.itemEntity' ]` 
*  `fields :  [ 'coverImage.itemEntity.*' ]`

But `populate: [ 'coverImage.itemEntity.name' ]` is not supported while `fields :  [ 'coverImage.itemEntity.name' ]` is, right ?

B4nan:
The reason why scalars are also allowed in the populate hint is that they can be [lazy](https://mikro-orm.io/docs/defining-entities#lazy-scalar-properties).

I am not entirely happy with this, maybe it would be better to use branded types to mark properties that can be populated and only allow those in the populate hint, want to experiment with that too.

PR #4477, opened 2023-06-22
fix(sql): schema not inherited in query with relations

### LINKED ISSUE #4472: Schema not inherited in query with relations
**Describe the bug**
When performing a query with relations and using a JOINED type strategy, the schema is not inherited by the related entity.

Remember that both entities must use the wildcard operator in schemas.
**SELECT_IN strategy works.**

**To Reproduce**
1. Create an entity with wildcard schema
2. Create another entity with wildcard schema
3. Perform a query operation using the JOINED strategy in a schema other than the default one.

**Versions**

| Dependency | Version |
| - | - |
| mikro-orm | 5.7.12 |

mlusca:
When both tables have wildcard schema, join not use the correct schema.

Closes #4472

### DIFF
--- a/tests/issues/GH4472.test.ts
+++ b/tests/issues/GH4472.test.ts
@@ -1,19 +1,9 @@
-import {
-  Collection,
-  Entity,
-  LoadStrategy,
-  ManyToOne,
-  MikroORM,
-  OneToMany,
-  PrimaryKey,
-  Property,
-  wrap,
-} from '@mikro-orm/core';
+import { Collection, Entity, LoadStrategy, ManyToOne, MikroORM, OneToMany, PrimaryKey, Property } from '@mikro-orm/core';
 import { PostgreSqlDriver } from '@mikro-orm/postgresql';
 import { mockLogger } from '../helpers';
 
 @Entity({ schema: '*' })
-export class Topic {
+class Topic {
 
   @PrimaryKey()
   id!: number;
@@ -27,7 +17,7 @@ export class Topic {
 }
 
 @Entity({ schema: '*' })
-export class Category {
+class Category {
 
   @PrimaryKey()
   id!: number;
@@ -53,7 +43,7 @@ describe('multiple connected schemas in postgres', () => {
     }
 
     // `*` schema will be ignored
-    await orm.schema.updateSchema(); // `*` schema will be ignored
+    await orm.schema.updateSchema();
 
     // we need to pass schema for book
     await orm.schema.updateSchema({ schema: 'n2' });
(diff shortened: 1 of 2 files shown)


tests/issues/GH4472.test.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    expect(mock.mock.calls[0][0]).toMatch(
+      'select "t0"."id", "c1"."id" as "c1__id" from "n5"."topic" as "t0" left join "n5"."category" as "c1" on "t0"."id" = "c1"."topic_id" where "t0"."id" = 1',
+    );
+  });
+
+  test('should default schema on not define schema', async () => {
+    const topic = new Topic();
+    topic.name = 't1';
+    orm.em.persist(topic);
```
B4nan, line 112: you dont flush, so the database stays empty. in other words, those tree lines are not really needed, they dont do anything right now, and since you assert the actual query instead of the return value, its fine to just remove them

tests/issues/GH4472.test.ts
```diff
@@ -0,0 +1,130 @@
...
+  test('should same schema', async () => {
+    const topic = new Topic();
+    topic.name = 't1';
+    wrap(topic).setSchema('n5');
+    orm.em.persist(topic);
+
+    const category = new Category();
+    category.topic = topic;
+    topic.category.add(category);
+    await orm.em.flush();
```
B4nan, line 87: kinda the same here, while you flush, you dont assert the return value, so this part is a bit irrelevant. 

will keep it up to you, either remove it or add assertions for the return value too

tests/issues/GH4472.test.ts
```diff
@@ -0,0 +1,130 @@
...
+      driver: PostgreSqlDriver,
+    });
+    await orm.schema.ensureDatabase();
+
+    for (const ns of ['n2', 'n5']) {
+      await orm.schema.execute(`drop schema if exists ${ns} cascade`);
+    }
+
+    // `*` schema will be ignored
+    await orm.schema.updateSchema(); // `*` schema will be ignored
```
B4nan, line 56:
```suggestion
    await orm.schema.updateSchema();
```

tests/issues/GH4472.test.ts
```diff
@@ -0,0 +1,130 @@
...
+  OneToMany,
+  PrimaryKey,
+  Property,
+  wrap,
+} from '@mikro-orm/core';
+import { PostgreSqlDriver } from '@mikro-orm/postgresql';
+import { mockLogger } from '../helpers';
+
+@Entity({ schema: '*' })
+export class Topic {
```
B4nan, line 16:
```suggestion
class Topic {
```

tests/issues/GH4472.test.ts
```diff
@@ -0,0 +1,130 @@
...
+  @Property()
+  name!: string;
+
+  @OneToMany(() => Category, e => e.topic)
+  category = new Collection<Category>(this);
+
+}
+
+@Entity({ schema: '*' })
+export class Category {
```
B4nan, line 30:
```suggestion
class Category {
```

B4nan (changes requested):
thanks! just few nits

mlusca:
> thanks! just few nits

Thanks

B4nan (approved):

PR #4479, opened 2023-06-22
Update defining-entities.md on default values

VoytechG:
Hello, I would like to suggest creating this link to another part of documentation. I had difficulties web searching for this specific information and got here via an issue on Github. So maybe this link will be helpful for others.

Also, do you agree with the notion of the property being optional only when the entity of this property is created or updated?
I also would find such mention helpful, to indicate it has a different effect than simply using `?` before someone even click the link. Maybe even adding this notion in the actual Type validation document would also be helpful.

### DIFF
--- a/docs/versioned_docs/version-5.7/defining-entities.md
+++ b/docs/versioned_docs/version-5.7/defining-entities.md
@@ -437,6 +437,9 @@ properties: {
   </TabItem>
 </Tabs>
 
+### Type validation of properties with default values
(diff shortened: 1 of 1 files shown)


docs/versioned_docs/version-5.7/defining-entities.md
```diff
@@ -437,6 +437,9 @@ properties: {
   </TabItem>
 </Tabs>
 
+### Type validation of properties with default values
```
B4nan, line 440: I think this should rather go to the section about [default values](https://mikro-orm.io/docs/defining-entities#default-values), and the existing examples should be using the `OptionalProps` symbol right ahead.

I will be doing a docs checkup before v6 gets out, so planning to polish things like this. Feel free to modify the PR, but if you'd like to merge it, the changes also need to be reflected in the next docs version, you modified only the v5.7 snapshot.

PR #4541, opened 2023-07-18
fix(seeder): allow run method to return without promise

noftaly:
I noticed the `Seeder#run` method only allows returning a `Promise<void>`, although it is completely possible for it to return only `void`, without being async.

This isn't anything major, but I have ESLint rules that forbid me to have an `async` function if it isn't really async, and i'd like to stick by it. It'd be great if the signature of the function didn't force us to have it `async`.

Thank you!


B4nan (changes requested):
you will need to do the same for the `Seeder` interface defined in the core package

B4nan (approved):

PR #4553, opened 2023-07-24
[wip] feat: Pagination

arafatkn:
Added offset based pagination. I have added two pagination function.
- `paginate` function will return paginated data and a meta object containing total matched rows in the table. 
- `simplePaginate` function will return paginated data and a meta object without total matched rows or page count. But it will return a property to determine whether more page is available.

### Fetching Paginated Results
To paginate the results, we can use `em.paginate()` or `em.simplePaginate` method:

```ts
const { data, meta } = await em.paginate(Author, { ... }, { page: 2, perPage: 20 });

console.log(data); // Author[]
console.log(meta); // { totalItems: number; perPage: number; totalPages: number; currentPage: number; }
```
`em.paginate()` will return total count of entities before applying limit and offset, so it is a bit slower than `em.simplePaginate()`, which will return only the paginated results:

```ts
const { data, meta } = await em.simplePaginate(Author, { ... }, { page: 2, perPage: 20 });

console.log(data); // Author[]
console.log(meta); // { perPage: number; currentPage: number; hasNextPage: boolean; }
```

Going add tests for these soon.

### Next PR
Cursor based pagination.


B4nan:
Thanks, will take a closer look later, for now some food for thought:

I am not sure if we really need two methods for this, it could be just another option to `em.paginate`. Also calling things "simple" could be a bit misleading, maybe `includeTotal: false` or something like that would be better?

> Cursor based pagination.

That is already implemented in v6 (https://github.com/mikro-orm/mikro-orm/pull/3975). Please target v6 branch here as well, v5 is in maintenance mode, I don't plan any more feature releases for it (mainly, but not only, because the rebasing of v6 branch is getting more and more painful).

B4nan:
Just FYI I will wait with a review before you base this on v6.

B4nan:
You will need to rebase this. Which could be a bit tedious, so it might be easier to just start from fresh v6 branch and port your changes manually.

PR #4571, opened 2023-07-31
fix(core): rework `Collection` initialization to use `em.populate()`

### LINKED ISSUE #4464: Incomplete Entities after using FindOption 'fields'
After loading an entity with the 'fields' FindOption Subsequent load operations will return incomplete entities.
See included testcase

```
import 'reflect-metadata';
import { Collection, Entity, ManyToMany, ManyToOne, MikroORM, PrimaryKey, Property, wrap, Embeddable, Embedded, EntityCaseNamingStrategy, OneToMany } from '@mikro-orm/core';
import { PostgreSqlDriver } from '@mikro-orm/postgresql';

@Entity()
export class A {
  @PrimaryKey()
    id!: number

  @Property()
    test1!: string

  @ManyToMany({ entity: () => B, mappedBy: 'a' })
    b = new Collection<B>(this)
}

@Entity()
export class B {
  @PrimaryKey()
    id!: number

  @Property()
    test2!: string

  @Property()
    test3!: string

  @ManyToMany({ entity: () => A, inversedBy: 'b' })
    a = new Collection<A>(this)
}


describe('loadItems with fields', () => {
    let orm: MikroORM<SqliteDriver>;

    beforeAll(async () => {
        orm = await MikroORM.init({
            entities: [A, B],
            dbName: ':memory:',
            type: 'sqlite',
        })
    await orm.schema.refreshDatabase()
    })

    test('loadItems with fields', async () => {
      const b = orm.em.create(B, {test2: 'qwer', test3: 'asdf'})
      const a = orm.em.create(A, {test1: 'yxcv', b: [b]})

      await orm.em.persistAndFlush([a, b])
      orm.em.clear()

      await orm.em.findOne(B, 1, {fields: ['test2']}) // fail
      // await orm.em.findOne(B, 1) // ok

      const a2 = await orm.em.findOne(A, 1)
      const test3 = (await a2.b.loadItems())[0].test3
      
      // no 'find' or 'populate' specified, but still undefined if 'fields' was previously used
      expect(test3).toMatch('asdf') 
    })
})
```

B4nan:
This unifies the population mechanism for to-many relations by using the `em.populate()` internally.

Previously, the `Collection.init` (and methods using it, e.g. `loadItems`) were using custom implementation, which in some cases resulting in different results as opposed to `em.populate()`. With this PR, the `init` method is also using `em.populate`, yielding the same results as if you would populate the relation.

Closes #4464

### DIFF
--- a/packages/core/src/entity/ArrayCollection.ts
+++ b/packages/core/src/entity/ArrayCollection.ts
@@ -247,7 +247,7 @@ export class ArrayCollection<T extends object, O extends object> {
   }
 
   protected propagateToInverseSide(item: T, method: 'add' | 'remove' | 'takeSnapshot'): void {
-    const collection = item[this.property.inversedBy as keyof T] as unknown as ArrayCollection<O, T>;
+    const collection = item[this.property.inversedBy as keyof T] as ArrayCollection<O, T>;
 
     if (this.shouldPropagateToCollection(collection, method)) {
       collection[method as 'add'](this.owner);
@@ -303,7 +303,7 @@ export class ArrayCollection<T extends object, O extends object> {
   }
 
   protected incrementCount(value: number) {
-    if (typeof this._count === 'number') {
+    if (typeof this._count === 'number' && this.initialized) {
       this._count += value;
     }
   }
--- a/packages/core/src/entity/Collection.ts
+++ b/packages/core/src/entity/Collection.ts
@@ -23,6 +23,7 @@ import { Reference } from './Reference';
 import type { Transaction } from '../connections/Connection';
 import type { FindOptions } from '../drivers/IDatabaseDriver';
 import { helper } from './wrap';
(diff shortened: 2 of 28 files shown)


parisholley:
@B4nan by chance, does this refactor fix a potential issue where calling `collection.loadItems()` on an already initialized collection (or perhaps uninitialized but one of the resulting items is already in UOW) will cause any mutations (eg: changing a string property) on an item in that collection to be reverted (I assume a hard "refresh" was happening to the UOW).  I worked around it for now by instead using `entity.populate()` (on dev v89). If this change is already in v89 I can issue a pull otherwise, i'll wait to test it again in the latest dev release.

B4nan:
Yes, this PR is part of 6.0.0-dev.89.

Keep in mind that `em.populate()` has `refresh: false` by default, while the one this PR adds into the `coll.init()` will have `refresh: true`.

PR #4575, opened 2023-08-01
feat: Pagination in EntityManager and EntityRepository

arafatkn:
Added offset based pagination. I have added two pagination function.
- `paginate` function will return paginated data and a meta object containing total matched rows in the table. 
- `simplePaginate` function will return paginated data and a meta object without total matched rows or page count. But it will return a property to determine whether more page is available.

### Fetching Paginated Results
To paginate the results, we can use `em.paginate()` or `em.simplePaginate` method:

```ts
const { data, meta } = await em.paginate(Author, { ... }, { page: 2, perPage: 20 });

console.log(data); // Author[]
console.log(meta); // { totalItems: number; perPage: number; totalPages: number; currentPage: number; }
```
`em.paginate()` will return total count of entities before applying limit and offset, so it is a bit slower than `em.simplePaginate()`, which will return only the paginated results:

```ts
const { data, meta } = await em.simplePaginate(Author, { ... }, { page: 2, perPage: 20 });

console.log(data); // Author[]
console.log(meta); // { perPage: number; currentPage: number; hasNextPage: boolean; }
```

Going to add tests for these soon.


jrobeson:
should there be configurable defaults for offset and limit in the mikro-orm config? Although i'm not sure where they'd go.

B4nan (changes requested):
thanks, few initial thouths

and dont forget to add some tests, you can check those for inspiration https://github.com/mikro-orm/mikro-orm/tree/v6/tests/features/cursor-based-pagination

B4nan:
This got closed automatically because the v6 branch is now removed. Do you plan to finish the PR or should I take over?

If so, please target the master branch now (which is v6).

darkbasic:
@B4nan I don't remember if cursor based pagination is in this PR or has already been merged because I would like to test it before v6 gets released.

B4nan:
That was introduced earlier and is already part of v6:

https://mikro-orm.io/docs/next/entity-manager#cursor-based-pagination

arafatkn:
> This got closed automatically because the v6 branch is now removed. Do you plan to finish the PR or should I take over?
> 
> If so, please target the master branch now (which is v6).

I will finish the PR this weekend.

PR #4590, opened 2023-08-04
docs: Add instructions for using sqlite extensions

### LINKED ISSUE #4588: docs: Provide example for using SQLite extensions like sqlean
SQLite extensions like [sqlean](https://github.com/nalgeon/sqlean) can help bring it much closer to feature parity with other database servers like Postgres (e.g. regexp, statistics). Unfortunately, they're difficult to install in a MikroORM connection with the current API, and there's no example explaining how to do it.

After a lot of research and experimenting, I finally found that the `pool.afterCreate` property of the SQLite initialization options can be used for this purpose. For example, after downloading and extracting the sqlean plugins, you can load them like so:

```typescript
const orm = await MikroORM.init<BetterSqliteDriver>({
  // ...
  pool: {
    afterCreate: (conn: any, done: any) => {
      conn.loadExtension(
        '/Users/stabai/Downloads/sqlean-macos-arm64/sqlean'
      );
      done(null, conn);
    },
  },
});
```

It would be even better if there were a simpler API to do this, but this at least seems worth documenting for others who need these additional features in SQLite.

stabai:
fixes #4588

### DIFF
--- a/docs/docs/usage-with-sql.md
+++ b/docs/docs/usage-with-sql.md
@@ -191,3 +191,21 @@ qb.select('*').where({ id: { $in: [...] } });
 const res = await em.getDriver().execute(qb);
 console.log(res); // unprocessed result of underlying database driver
 ```
+
+## Using SQLite extensions
+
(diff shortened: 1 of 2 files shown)


B4nan:
Thanks!

Issue #4592, opened 2023-08-05
Add more methods to Collections


bakura10:
Hi :),

I am coming from Doctrine world, where collections are having a lot of convenient methods (https://www.doctrine-project.org/projects/doctrine-collections/en/stable/index.html).

The workaround in Mikro-ORM seems to converting first to an array to have access to all JS array related methods. Arrays are more flexible in JS than they are in JS, but maybe exposing those methods to Mikro-ORM collections would reduce the need for this conversion. Here are the one I often use in Doctrine:

* Reduce
* Find
* Find first
* Exists
* Filter

Except from this I really love how Mikro-ORM is thought. It _feels_ like Doctrine, but with so many pain points being lifted. That's really a breath of fresh air :)

Thanks!

B4nan:
FYI the collection items are stored in a `Set` internally, so you need to convert that to an array to be able to use the native array methods. I wanted to add some of those for v6, not sure if we need `reduce`, and not sure what should be the difference in `find` and `findFirst`? Also we already have `contains` method.

rubiin:
`exists`, `slice`, `removeElement` and `isEmpty` seems useful as well.

B4nan:
Another good candidate would be `map()`.

PR #4599, opened 2023-08-09
feat(core): add `Collection.isEmpty()` method

rubiin:
Adds the isExists which returns either true false based on the collection count same as doctrine

### DIFF
--- a/docs/docs/collections.md
+++ b/docs/docs/collections.md
@@ -41,6 +41,7 @@ console.log(author.books.contains(book)); // false
 author.books.add(book);
 console.log(author.books.count()); // 1
 author.books.removeAll();
+console.log(author.books.isEmpty()); // true
 console.log(author.books.contains(book)); // false
 console.log(author.books.count()); // 0
 console.log(author.books.getItems()); // Book[]
--- a/packages/core/src/entity/ArrayCollection.ts
+++ b/packages/core/src/entity/ArrayCollection.ts
@@ -64,6 +64,7 @@ export class ArrayCollection<T extends object, O extends object> {
     }) as unknown as U[];
   }
 
+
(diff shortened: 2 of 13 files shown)


rubiin:
> let's at least mention the method in here
> 
> [mikro-orm.io/docs/collections](https://mikro-orm.io/docs/collections)
> 
> should be enough to include it in the code block, I hope I will find some time to rework this section a bit and explain the methods too

Must have missed it. I will update the pr

rubiin:
https://github.com/mikro-orm/mikro-orm/issues/4592

rubiin:
I have also implemented slice but don't know where I should add the tests. Is there a centralized file for collections test.

B4nan (approved):
thanks

rubiin:
Thanks to you for such a great ORM :)

B4nan:
Regarding the slice method, I'd like that to be a bit smarter than just converting all the items from `Set` to an array just to call `slice` on it - ideally just iterate on the `Set` and create a new array only from the items that should be there (and finish iterating once you reached the end of the range, not the whole `Set`).

I don't think we have any test suite for collections, feel free to create one in the `tests/features`, we could move other collection-related tests in there (out of the huge EM test suites, those are a bit legacy tests).

rubiin:
Sure thing

rubiin:
Onething we could do is , keep a counter and iterate over the set something like this but one gotcha will be we cant jump 
off to a particular index and always have to start from 0

```
let start = 0;

for (const element of set1) {
    start += 1;

    if (start === end) {
        break;
        '
    }

}


```

rubiin:
on it. Seems like the same thing I have in mind

PR #4600, opened 2023-08-09
fix(platform): allow enum column type to be set

RWOverdijk:


B4nan:
Closing in favor of #4601

PR #4617, opened 2023-08-11
test: add `features/collection` for collection related tests

rubiin:
Adds a separate test file for collection


rubiin:
@B4nan need your input on this

tests/features/collection/collection.test.ts
```diff
@@ -0,0 +1,103 @@
...
+const { BookTag3 } = require('../../entities-js').BookTag3;
+
+describe('EntityManagerSqlite', () => {
+
+  let orm: MikroORM<SqliteDriver>;
+
+  beforeAll(async () => orm = await initORMSqlite());
+  beforeEach(async () => orm.schema.clearDatabase());
+
+  test('isConnected()', async () => {
```
B4nan, line 16: this is unrelated copypaste

tests/features/collection/collection.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    await orm.close(true);
+    expect(await orm.isConnected()).toBe(false);
+    await orm.connect();
+    expect(await orm.isConnected()).toBe(true);
+
+    // as the db lives only in memory, we need to re-create the schema after reconnection
+    await orm.schema.createSchema();
+  });
+
+
```
B4nan, line 27: you keep adding 2 blank lines on random places, please don't, there should be just one

tests/features/collection/collection.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    const tag2 = new BookTag3('funny');
+    const tag3 = new BookTag3('sick');
+    const tag4 = new BookTag3('strange');
+    const tag5 = new BookTag3('sexy');
+    book1.tags.add(tag1, tag3);
+    book2.tags.add(tag1, tag2, tag5);
+    book3.tags.add(tag2, tag4, tag5);
+
+    await orm.em.persist(book1);
+    await orm.em.persist(book2);
```
B4nan, line 43: persist is sync method

tests/features/collection/collection.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    await orm.em.persist(book3).flush();
+
+
+    const tagRepository = orm.em.getRepository<typeof BookTag3>(BookTag3);
```
B4nan, line 47: if you wouldn't copy this from the JS entity test, it would be cleaner. those JS entities require such weird casting when used inside TS test

tests/features/collection/collection.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    // count
+    expect(book.tags.count()).toBe(0);
+
+
+    // isEmpty
+    expect(book.tags.isEmpty()).toBe(true);
+  });
+
+
+  afterAll(async () => {
```
B4nan, line 99: this should go to the top ideally, where the other hooks are

B4nan (changes requested):

tests/features/collection/collection.test.ts
rubiin: Yeah I copied most from the Sqlite test file . Will do that from another file

tests/features/collection/collection.test.ts
rubiin: Thats my IDE . Dont know why it happens. Perhaps we can add this rule https://eslint.org/docs/latest/rules/no-multiple-empty-lines

B4nan:
If this should add any value, you should extract the collection tests from the EM tests - move them, so we don't just duplicate what is already tested somewhere else.

rubiin:
> If this should add any value, you should extract the collection tests from the EM tests - move them, so we don't just duplicate what is already tested somewhere else.
> 
> And I guess I would rather do that in the v6 branch.

Yeah . I saw them on many files , one of the reason that i didnt move them out

rubiin:
Will be closing this for now and creating a new one

PR #4626, opened 2023-08-14
refactor: add `includeCount` option to `em.findByCursor`

bakura10:
Hi !

This is my first contribution to a TypeScript project, let me know if there is anything wrong.

This PR adds the ability to opt-out for count in cursor pagination.

In the tests I wanted to add an assertion to make sure that "find" is called and "findAndCount" is not called, but I could not find any equivalent to $mock->expects($this->never()->method('findAndCount'). Is there a way to do that easily with mocks in JS?

PS: I don't really know why the lock files are part of the PR, I did not update any dependencies:/


docs/docs/entity-manager.md
```diff
@@ -288,7 +288,11 @@ console.log(count); // total count, e.g. 1327
 
 As an alternative to the offset based pagination with `limit` and `offset`, we can paginate based on a cursor. A cursor is an opaque string that defines specific place in ordered entity graph. You can use `em.findByCursor()` to access those options. Under the hood, it will call `em.find()` and `em.count()` just like the `em.findAndCount()` method, but will use the cursor options instead.
 
-Supports `before`, `after`, `first` and `last` options while disallowing `limit` and `offset`. Explicit `orderBy` option is required.
+Supports `before`, `after`, `first` and `last` options while disallowing `limit` and `offset`. Explicit `orderBy` 
+option is required. It also supports the `includeCount` (default to true) option. When explicitly set to false, 
+entity manager will perform a `find` instead of `findAndCount`. The cursor `totalCount` will be set to null instead. 
+This can be used as a performance optimization to avoid an expensive SQL count query, when knowing the exact number 
+of pages is not important.
```
B4nan, line 296: remove all the line breaks

docs/docs/entity-manager.md
```diff
@@ -330,7 +341,7 @@ Cursor<User> {
     User { ... },
     ...
   ],
-  totalCount: 50,
+  totalCount: 50, // always null when `includeCount` is set to false
```
B4nan, line 344: this should be `undefined`, not `null`
```suggestion
  totalCount: 50, // not defined when `includeCount` is set to false
```

docs/docs/entity-manager.md
```diff
@@ -318,6 +322,13 @@ const nextCursor2 = await em.findByCursor(User, {}, {
   after: { id: lastSeenId }, // entity-like POJO
   orderBy: { id: 'desc' },
 });
+
+const currentCursorWithoutCount = await em.findByCursor(User, {}, {
+  first: 10,
+  after: previousCursor, // cursor instance
+  orderBy: { id: 'desc' },
+  includeCount: false
```
B4nan, line 330:
```suggestion
  includeCount: false,
```

packages/core/src/EntityManager.ts
```diff
@@ -451,7 +455,7 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
    *     User { ... },
    *     User { ... },
    *   ],
-   *   totalCount: 50,
+   *   totalCount: 50, // not included if count is excluded
```
B4nan, line 458:
```suggestion
   *   totalCount: 50, // not included if `includeCount: false`
```

packages/core/src/utils/Cursor.ts
```diff
@@ -57,21 +57,23 @@ export class Cursor<Entity extends object, Hint extends string = never, Fields e
 
   readonly hasPrevPage: boolean;
   readonly hasNextPage: boolean;
+  readonly totalCount: number | null;
```
B4nan, line 60: better to keep this optional instead of allowing `null` values
```suggestion
  readonly totalCount?: number;
```

and ideally this will be conditionally mapped so the property is not defined only when `includeCount: false`, I can take a look at that myself if you have no idea how to do that with TS

packages/core/src/utils/Cursor.ts
```diff
@@ -57,21 +57,23 @@ export class Cursor<Entity extends object, Hint extends string = never, Fields e
...
     meta: EntityMetadata<Entity>,
+    totalCount?: number | undefined,
   ) {
     const { first, last, before, after, orderBy, overfetch } = options;
     const limit = first || last;
     const isLast = !first && !!last;
     const hasMorePages = !!overfetch && limit != null && items.length > limit;
     this.hasPrevPage = before || after ? true : (isLast && hasMorePages);
     this.hasNextPage = !(isLast && !before && !after) && hasMorePages;
+    this.totalCount = totalCount ?? null;
```
B4nan, line 76:
```suggestion
    this.totalCount = totalCount;
```

packages/core/src/utils/Cursor.ts
```diff
@@ -57,21 +57,23 @@ export class Cursor<Entity extends object, Hint extends string = never, Fields e
...
+  readonly totalCount: number | null;
 
   private readonly definition: (readonly [EntityKey<Entity>, QueryOrder])[];
 
   constructor(
     readonly items: Loaded<Entity, Hint, Fields>[],
-    readonly totalCount: number,
     options: FindByCursorOptions<Entity, Hint, Fields>,
     meta: EntityMetadata<Entity>,
+    totalCount?: number | undefined,
```
B4nan, line 68: this is the same

```suggestion
    totalCount?: number,
```

tests/features/cursor-based-pagination/simple-cursor.test.ts
```diff
@@ -652,6 +652,26 @@ describe.each(['sqlite', 'better-sqlite', 'mysql', 'postgresql', 'mongo'] as con
...
+      orderBy: { id: 'asc' },
+      includeCount: false,
+    });
+    expect(cursor1).toBeInstanceOf(Cursor);
+    expect(cursor1.items).toMatchObject([
+      { id: 1, name: 'User 1' },
+      { id: 2, name: 'User 2' },
+      { id: 3, name: 'User 3' },
+    ]);
+    expect(cursor1.totalCount).toBe(null);
```
B4nan, line 668:
```suggestion
    expect(cursor1.totalCount).toBeUndefined();
```

tests/features/cursor-based-pagination/simple-cursor.test.ts
```diff
@@ -652,6 +652,26 @@ describe.each(['sqlite', 'better-sqlite', 'mysql', 'postgresql', 'mongo'] as con
     ]);
   });
 
+  test('exclude count when `includeCount` is set to false', async () => {
+    const mock = mockLogger(orm, ['query', 'query-params']);
```
B4nan, line 656: this is not used apparently - check other tests how they assert the logged queries, you can use that to ensure only the select query is fired and no count query is there

B4nan (changes requested):
the lock file changes will need to go. i can see you used npm instead of yarn and you committed their lockfile too. and then I'd guess you used wrong yarn version (do you have corepack enabled?) which probably ended up with the yarn lockfile changes too.

just remove them from the git diff

packages/core/src/utils/Cursor.ts
bakura10: I will let you do that, I am not sure how to tackle this :).

bakura10:
This should be good!

tests/features/cursor-based-pagination/simple-cursor.test.ts
```diff
@@ -653,7 +653,6 @@ describe.each(['sqlite', 'better-sqlite', 'mysql', 'postgresql', 'mongo'] as con
   });
 
   test('exclude count when `includeCount` is set to false', async () => {
-    const mock = mockLogger(orm, ['query', 'query-params']);
```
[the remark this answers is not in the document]
bakura10: I have added the "toMatchSnapshot" (I've figured out this was adding extra queries in the snapshot file, so I think this is what I needed here no?

B4nan (approved):
I guess this is good to go, let's just wait for the green CI.

Thanks!

bakura10:
I don't really understand the tests failure. When I try locally I indeed have some errors with mysql and mongo (most likely because I don't have them installed on my machine?), but SQLite works properly. However it seems that this may due to the tests not running locally, that the snapshot are not properly generated.

What is the best way to solve such issues?

B4nan:
Having the tests pass locally is the very first step. Have you followed the contributing guide? It should be just about `docker compose up` nowadays.

bakura10:
After doing the compose up, all the tests are now passing locally! Thanks, this PR should be good to go :)

rubiin (commented):
LGTM

Nexfader:
@B4nan Any reason it hasn't been merged into the `main` branch with `v6`? I could help port it to the upcoming `v7`

B4nan:
no idea, i guess it could get lost during a bad rebase. feel free to PR it to the master branch, i don't see why this couldn't be released in v6.

B4nan:
hmm maybe the change in the cursor class params is a bit breaking, I'll leave it up to you, either come up with a non breaking way for v6 or target v7

Issue #4602, opened 2023-08-10
Control the column on upsert


bakura10:
Hi !

I have an entity where I generate the UUID on application side (can't use database generation), where the email is the unique field. I try to do the upsert like this:

```
const customer = entityManager.create(Customer, request.body);
const managedCustomer = await entityManager.upsert(customer);
```

However MikroORM generate this SQL:

```
insert into customers (email, first_name, id) values ('test@test.com', 'foo', '71b804dc-7c92-4287-aff8-6e2c536832be') on conflict (id) do update set email = excluded.email, first_name = excluded.first_name returning last_name;
```

However, I would like MikroORM to check the conflict on the email, not the ID, so generate this query:

```
insert into customers (email, first_name, id) values ('test@test.com', 'foo', '71b804dc-7c92-4287-aff8-6e2c536832be') on conflict (email) do update set email = excluded.email, first_name = excluded.first_name returning last_name;
```

Is there a way to override the column that MikroORM is using? Maybe somehing like that:

```
await entityManager.upsert(customer, {conflictColumn: 'email'});
```

Thanks!

B4nan:
What you can do now is to pass in DTO without the id and let the database generate it.

```ts
const customer = await em.upsert(Customer, { email: '...' });
```

Note that in your code, `customer === managedCustomer`, you won't get a new entity instance from that call.

bakura10:
I would like to generate uuidv7, which is not yet supported natively (I generate it server side).

B4nan:
I will add 4 new options:
- `onConflictFields?: (keyof T)[];` (to control the conflict clause)
- `onConflictAction?: 'ignore' | 'merge';` (used `ignore` and `merge` as that is how the QB methods are called)
- `onConflictMergeFields?: (keyof T)[];` (to control the merge clause)
- `onConflictExcludeFields?: (keyof T)[];` (to omit fields from the merge clause)

Which should make the upsert queries completely configurable. Already have a WIP that handles the queries correctly, the tricky part will be handling the entity mapping (both to existing entity instances and to the POJO signatures) and reloading the missing data (for mysql that does not have a returning clause).

This should cover also #4325, which is about being able to omit a field from the merge data.

bakura10:
Thanks a lot, this looks like a perfect solution to this problem! Great job!

PR #4627, opened 2023-08-15
feat: allow to select fields for upsert

bakura10:
This PR attempts to solve #4602 by allowing passing specific columns on which the upsert condition is done. This allows to use entities and generate an ID server-side.

I'm not 100% sure about the approach here, but the change I did is that if you select a different column to act as the on conflict column _and_ that the PK is also present, then it won't update any field on conflict.

### DIFF
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -666,10 +666,10 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
       }
     }
 
-    const unique = meta.props.filter(p => p.unique).map(p => p.name);
+    const unique = options.onConflictFields || meta.props.filter(p => p.unique).map(p => p.name);
     const propIndex = unique.findIndex(p => data![p] != null);
 
-    if (where == null) {
+    if (where == null || options.onConflictFields) {
       if (propIndex >= 0) {
         where = { [unique[propIndex]]: data[unique[propIndex]] } as FilterQuery<Entity>;
       } else if (meta.uniques.length > 0) {
--- a/packages/core/src/drivers/IDatabaseDriver.ts
+++ b/packages/core/src/drivers/IDatabaseDriver.ts
@@ -151,6 +151,7 @@ export interface NativeInsertUpdateOptions<T> {
   schema?: string;
   /** `nativeUpdate()` only option */
   upsert?: boolean;
+  onConflictFields?: string[];
 }
 
 export interface NativeInsertUpdateManyOptions<T> extends NativeInsertUpdateOptions<T> {
--- a/packages/knex/src/AbstractSqlDriver.ts
+++ b/packages/knex/src/AbstractSqlDriver.ts
@@ -488,10 +488,16 @@ export abstract class AbstractSqlDriver<Connection extends AbstractSqlConnection
       if (options.upsert) {
         /* istanbul ignore next */
         const uniqueFields = Utils.isPlainObject(where) ? Utils.keys(where) as EntityKey<T>[] : meta!.primaryKeys;
+        const uniqueFieldsContainPK = uniqueFields.filter(value => meta!.primaryKeys.includes(value)).length > 0;
+        const pk = pks.map(pk => Utils.extractPK<T>(data[pk] || where, meta)!) as Primary<T>[];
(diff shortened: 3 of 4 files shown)


packages/core/src/drivers/IDatabaseDriver.ts
```diff
@@ -151,6 +151,7 @@ export interface NativeInsertUpdateOptions<T> {
   schema?: string;
   /** `nativeUpdate()` only option */
   upsert?: boolean;
+  onConflictFields?: string[];
```
B4nan, line 154: this should go only to the `em.upsert` options ideally, having it on this level seems confusing as it won't do anything with `em.insert()` and similar methods.

B4nan (commented):
I would actually like to see this targetting the master branch, I want to do a feature release next and I wanted to take a look exactly at those upsert improvements like this one

packages/core/src/EntityManager.ts
```diff
@@ -666,10 +666,10 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
       }
     }
 
-    const unique = meta.props.filter(p => p.unique).map(p => p.name);
+    const unique = options.onConflictFields || meta.props.filter(p => p.unique).map(p => p.name);
```
B4nan, line 669: this could be an empty array too, would be good to have a test case for that (it might be already properly handled by existing validations)

tests/features/upsert/GH4602.test.ts
```diff
@@ -0,0 +1,48 @@
...
+});
+
+afterAll(async () => {
+  await orm.close(true);
+});
+
+test('GH issue 4602', async () => {
+  const mock = mockLogger(orm);
+  const user1 = new User(123, 'test@test.com');
+  await orm.em.upsert(User, user1, { onConflictFields: ['email'] });
```
B4nan, line 39: test this with `upsertMany` too, those two methods need to work the same when it comes to the options

tests/features/upsert/GH4602.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    loggerFactory: options => new SimpleLogger(options),
+  });
+  await orm.schema.createSchema();
+});
+
+afterAll(async () => {
+  await orm.close(true);
+});
+
+test('GH issue 4602', async () => {
```
B4nan, line 36: this file should have a better name, only bug reproductions can have a name based on issue number, not features (same for the test description)

maybe it would be best to move this into the existing suite for upsert methods, where we already test on all the drivers automatically.

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -488,10 +488,16 @@ export abstract class AbstractSqlDriver<Connection extends AbstractSqlConnection
       if (options.upsert) {
         /* istanbul ignore next */
         const uniqueFields = Utils.isPlainObject(where) ? Utils.keys(where) as EntityKey<T>[] : meta!.primaryKeys;
+        const uniqueFieldsContainPK = uniqueFields.filter(value => meta!.primaryKeys.includes(value)).length > 0;
```
B4nan, line 491: I see you haven't changed anything in the mongo driver, not sure if this can have some support in mongo or we should throw in case the option is used.

packages/core/src/drivers/IDatabaseDriver.ts
B4nan: I guess we can keep this the way you have it for v5 and polish it in v6 together with the `upsert` option that has a similar problem. There should be different a interface for upsert options.

Btw this should be `keyof T` to have some intellisense, right?

B4nan (changes requested):
I can see there are some TS and linter errors...

B4nan:
Do you plan to work on this in the near future? I'd like to ship new feature release (from master, so this would need to be targetting that instead of v6), and wanted to have this in there as well.

Ideally, I'd ship it during this week, so will be happy to take over if you don't have time. I don't mind the proposal for `onConflictFields` and `onConflictAction`.

bakura10:
Hi @B4nan ,

Sorry, I am on holiday for two weeks and won't be able to work on this for this week :(.

B4nan:
Closing in favor of #4669

PR #4635, opened 2023-08-17
docs: add note about using regexp literals in queries

kamiyo:
Document regexp query filter to inform users about passing the regexp object directly as the field value instead of using $re.

### DIFF
--- a/docs/docs/query-conditions.md
+++ b/docs/docs/query-conditions.md
@@ -79,8 +79,8 @@ const res = await orm.em.find(Author, [1, 2, 7]);
 The `$re` operator takes a string as input value, and by default uses the case-sensitive operator. If you would like to use a `RegExp` object, i.e. to be able to set flags, then search directly on the field name without using the operator:
 
 ```ts
-const res = await orm.em.find(Painter, {
-  lastName: /m[oa]net/i        // or new RegExp('m[oa]net', 'i')
+const res = await em.find(Painter, {
+  lastName: /m[oa]net/i, // or `new RegExp('m[oa]net', 'i')`
 });
 ```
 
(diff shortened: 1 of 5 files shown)


docs/docs/query-conditions.md
```diff
@@ -74,6 +74,16 @@ const res = await orm.em.find(Author, [1, 2, 7]);
...
+## Regular Expressions
+
+The `$re` operator takes a string as input value, and by default uses the case-sensitive operator. If you would like to use a `RegExp` object, i.e. to be able to set flags, then search directly on the field name without using the operator:
+
+```ts
+const res = await orm.em.find(Painter, {
+  lastName: /m[oa]net/i        // or new RegExp('m[oa]net', 'i')
+});
+```
+
```
rubiin, line 86: remove the extra line

B4nan: this one is correct, there should be blank lines around the headings

docs/docs/query-conditions.md
```diff
@@ -74,6 +74,16 @@ const res = await orm.em.find(Author, [1, 2, 7]);
 | `$not`   | Inverts the effect of a query expression and returns documents that do not match the query expression.  |
 | `$or`    | Joins query clauses with a logical OR returns all documents that match the conditions of either clause. |
 
+## Regular Expressions
+
+The `$re` operator takes a string as input value, and by default uses the case-sensitive operator. If you would like to use a `RegExp` object, i.e. to be able to set flags, then search directly on the field name without using the operator:
+
+```ts
+const res = await orm.em.find(Painter, {
+  lastName: /m[oa]net/i        // or new RegExp('m[oa]net', 'i')
```
B4nan, line 83:
```suggestion
const res = await em.find(Painter, {
  lastName: /m[oa]net/i, // or `new RegExp('m[oa]net', 'i')`
```

B4nan (approved):
thanks!

PR #4680, opened 2023-09-07
fix(core): default baseUrl value to '.' when registering ts-node

### LINKED ISSUE #4679: Unable to register `ts-node` when the `baseUrl` compiler option is not set
**Describe the bug**

[Since TypeScript 4.1, the `paths` compiler option can be used without the `baseUrl` compiler option.](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#paths-without-baseurl) However, the `ConfigurationLoader` assumes that the `baseUrl` option is set when the `paths` option is set when MikroORM's `useTsNode` flag in `package.json` is set to `true`:

https://github.com/mikro-orm/mikro-orm/blob/b0ecd21a7143bf6d42623f14f17a819e6ad739a5/packages/core/src/utils/ConfigurationLoader.ts#L135-L148

In the above code snippet, the `register` function from `tsconfig-paths` expects `baseUrl` to be a `string`. When the `baseUrl` compiler option is not set, the error below occurs.

**Stack trace**
```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:399:5)
    at validateString (node:internal/validators:163:11)
    at Object.isAbsolute (node:path:1157:5)
    at configLoader (/Users/ianyong/<redacted>/node_modules/.pnpm/tsconfig-paths@4.1.0/node_modules/tsconfig-paths/src/config-loader.ts:50:34)
    at Object.register (/Users/ianyong/<redacted>/node_modules/.pnpm/tsconfig-paths@4.1.0/node_modules/tsconfig-paths/src/register.ts:79:42)
    at Function.registerTsNode (/Users/ianyong/<redacted>/node_modules/.pnpm/@mikro-orm+core@5.7.12_@mikro-orm+migrations@5.7.12_@mikro-orm+postgresql@5.7.12_@mikro-orm+seeder@5.7.12/node_modules/@mikro-orm/core/utils/ConfigurationLoader.js:118:71)
    at Function.configure (/Users/ianyong/<redacted>/node_modules/.pnpm/@mikro-orm+cli@5.7.12_@mikro-orm+migrations@5.7.12_@mikro-orm+postgresql@5.7.12_@mikro-orm+seeder@5.7.12_pg@8.11.0/node_modules/@mikro-orm/cli/CLIConfigurator.js:29:61)
    at async /Users/ianyong/<redacted>/node_modules/.pnpm/@mikro-orm+cli@5.7.12_@mikro-orm+migrations@5.7.12_@mikro-orm+postgresql@5.7.12_@mikro-orm+seeder@5.7.12_pg@8.11.0/node_modules/@mikro-orm/cli/cli.js:13:18 {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

**To Reproduce**
1. Enable the `useTsNode` flag in `package.json`.
1. Set the `paths` compiler option but not the `baseUrl` compiler option in `tsconfig.json`.
1. Run any command that makes use of `ts-node`, such as `mikro-orm migration:fresh`.

**Expected behavior**

When `baseUrl` is not specified, it should default to `'.'` instead of throwing an error.

**Additional context**

The expected behaviour described above is effectively the same approach taken by the `tsconfig-paths` library itself when reading directly from a `tsconfig.json` file which does not have the `baseUrl` compiler option specified (https://github.com/dividab/tsconfig-paths/pull/208). The reason why I believe this issue should be fixed within MikroORM and not upstream is because in this case, MikroORM is not respecting the interface of the `register` method by passing in an argument of type `string | undefined` when `string` is expected.

**Versions**

| Dependency | Version |
| - | - |
| node | 18.16.0 |
| typescript | 5.0.3 |
| mikro-orm | 5.7.12 |

ianyong:
Fixes #4679. See the linked issue for the full context of this PR.

### DIFF
--- a/packages/core/src/utils/ConfigurationLoader.ts
+++ b/packages/core/src/utils/ConfigurationLoader.ts
@@ -142,7 +142,7 @@ export class ConfigurationLoader {
 
     if (Object.entries(options?.paths ?? {}).length > 0) {
       Utils.requireFrom('tsconfig-paths', tsConfigPath).register({
-        baseUrl: options.baseUrl,
+        baseUrl: options.baseUrl ?? '.',
         paths: options.paths,
       });
     }
--- a/tests/features/cli/CLIHelper.test.ts
+++ b/tests/features/cli/CLIHelper.test.ts
@@ -16,6 +16,9 @@ jest.mock(process.cwd() + '/tsconfig.extended-abs.json', () => tscExtendedAbs, {
 const tscExtended = { extends: './tsconfig.extended-abs.json', compilerOptions: { module: 'commonjs' } } as any;
 jest.mock(process.cwd() + '/tsconfig.extended.json', () => tscExtended, { virtual: true });
 
+const tscWithoutBaseUrl = { compilerOptions: { paths: { '@some-path/some': './libs/paths' } } };
+jest.mock(process.cwd() + '/tsconfig.without-baseurl.json', () => tscWithoutBaseUrl, { virtual: true });
+
 const tsc = { compilerOptions: { } } as any;
 jest.mock(process.cwd() + '/tsconfig.json', () => tsc, { virtual: true });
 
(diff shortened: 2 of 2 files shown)


B4nan (approved):
thanks!

PR #4702, opened 2023-09-16
feat(core): Allow to use `.ts` files as configuration without `ts-node`

### LINKED ISSUE #4701: Allow to pass `.ts` files as config with Bun when `ts-node` is disabled

Hi, I'm experimenting with MikroORM and Bun, and I find out that MikroORM's configuration loader ignores `.ts` files when `ts-node` is not enabled. And while this constraint is suitable for node, it is not necessary for Bun since it support TS files [out of the box](https://bun.sh/docs#design-goals).

I think it would be nice to add at least an option for configurator loader to pass `.ts` files when this option enabled. Something like `bypassTsFiles` flag in `mikro-orm` section of `package.json`. When enabled, [this](https://github.com/mikro-orm/mikro-orm/blob/3f30c16a1614ab059be71f3dbf51ccdd40dd9033/packages/core/src/utils/ConfigurationLoader.ts#L110) filter will be ignored same was as if `ts-node` was enabled.

Enabling `ts-node` works with Bun 1.0.1, but this running code from `ts-node` with Bun is not necessary.

**Additional context**

No additional context. But I think I can send a PR upon agreement on implementation details.

octet-stream:
### Details

This PR introduces two new options to allow `.ts` files as configs when `ts-node` is not enabled:

- `alwaysAllowTs` flag in `mikro-orm` section of `package.json`
- `MIKRO_ORM_CLI_ALWAYS_ALLOW_TS` environment variable

### Changes

- [x] Add optional `alwaysAllowTs` field on `Settings` interface in `ConfigurationLoader`;
- [x] Return `.ts` configs files from `ConfigurationLoader.getConfigPaths()` when `alwaysAllowTs` option is enabled;
- [x] Add tests;
- [x] Add documentation for these new options

Fixes #4701

### DIFF
--- a/docs/docs/installation.md
+++ b/docs/docs/installation.md
@@ -186,13 +186,14 @@ For CLI to be able to access our database, we will need to create `mikro-orm.con
 
 > ORM configuration file can export the Promise, like: `export default Promise.resolve({...});`.
 
-TypeScript is also supported, just enable `useTsNode` flag in our `package.json` file. There we can also set up array of possible paths to `mikro-orm.config` file, as well as use different file name. The `package.json` file can be located in the current working directory, or in one of its parent folders.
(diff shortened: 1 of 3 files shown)


B4nan (approved):
Looking good, thanks!

PR #4713, opened 2023-09-20
fix(core): fix query processing when PK is falsy

Robert-Schirmer:
Fix case of `processWhere` returning empty object when the where clause is an entity with an int pk of 0

### DIFF
--- a/packages/core/src/utils/QueryHelper.ts
+++ b/packages/core/src/utils/QueryHelper.ts
@@ -95,11 +95,7 @@ export class QueryHelper {
       QueryHelper.inlinePrimaryKeyObjects(where as Dictionary, meta, metadata);
     }
 
-    where = QueryHelper.processParams(where);
-
-    if (where == null) {
-      where = {};
-    }
+    where = QueryHelper.processParams(where) ?? {};
 
     /* istanbul ignore next */
     if (!root && Utils.isPrimaryKey<T>(where)) {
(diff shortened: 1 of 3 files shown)


packages/core/src/utils/QueryHelper.ts
```diff
@@ -95,7 +95,11 @@ export class QueryHelper {
       QueryHelper.inlinePrimaryKeyObjects(where as Dictionary, meta, metadata);
     }
 
-    where = QueryHelper.processParams(where) || {};
+    where = QueryHelper.processParams(where);
+
+    if (!where && typeof where !== 'number') {
```
B4nan, line 100: this should do the same, right?
```suggestion
    if (where == null) {
```

Robert-Schirmer: I think empty string as pk should still return `{}`, so not same for that case

B4nan: rest of the codebase is using the `null` checks, so this part should do it the same way. i disagree that an empty string should be treated differently than `0`

Robert-Schirmer: Ok will need to update other test that fails with that change

Robert-Schirmer: Could do `QueryHelper.processParams(where) ?? {}` then

B4nan: yeah correct, lets do it that way

B4nan:
thanks!

PR #4717, opened 2023-09-21
feat(core): allow setting default `schema` on `EntityManager`

EmmEm:
Hi,

I want to start with thanking you all for this great ORM that is so actively maintained and offer so many great features. One thing that really catches my eye is the per request (shell)EntityManager that I think will make this ORM a winner when it comes to multi tenancy. Almost all of the other big ORM libraries are based on a global concept. 

My aim with this PR is to make a shoot at implementing schema based multi tenancy support with help of the per request EntityManager. With this PR it's possible to set a schema when forking the EntityManager. That schema will be applied as a default schema if no other schema is present. It makes use of the schema options introduced in: https://github.com/mikro-orm/mikro-orm/pull/2296 

If this seems like an approach to pursue there is some more work to do:
- Add a hook in the other packages like Nestjs to allow for setting schema / request
- Documentation
- More extensive tests

Some questions
- Should the number of `this.context()` calls be reduced?
- When `.fork()` an existing fork with schema, should schema be inherited and should it be possible to clear schema?

### DIFF
--- a/docs/docs/multiple-schemas.md
+++ b/docs/docs/multiple-schemas.md
@@ -32,6 +32,41 @@ const qb = em.createQueryBuilder(User);
 await qb.insert({ email: 'foo@bar.com' }).withSchema('client-123');
 ```
 
+## Default schema on `EntityManager`
+
+Instead of defining schema per entity or operation it's possible to `.fork()` EntityManger and define a default schema that will be used with wildcard schemas.
+
+```ts
+const fork = em.fork({ schema: 'client-123' });
+await fork.findOne(User, { ... });
+
+// Will yield the same result as 
+const user = await em.findOne(User, { ... }, { schema: 'client-123' });
+```
+
+When creating an entity the fork will set default schema
+
+```ts
+const fork = em.fork({ schema: 'client-123' });
+const user = new User();
+user.email = 'foo@bar.com';
+await fork.persistAndFlush(user);
+
+// Will yield the same result as
+const qb = em.createQueryBuilder(User);
+await qb.insert({ email: 'foo@bar.com' }).withSchema('client-123');
+```
+
+You can also set or clear schema
+
+```ts
+em.schema = 'client-123';
+const fork = em.fork({ schema: 'client-1234' });
+fork.schema = null;
+```
+
+`EntityManager.schema` Respects the context, so global EM will give you the contextual schema if executed inside [request context handler](https://mikro-orm.io/docs/identity-map#-requestcontext-helper)
+
 ## Wildcard Schema
 
 Since v5, MikroORM also supports defining entities that can exist in multiple schemas. To do that, we just specify wildcard schema:
@@ -57,7 +92,7 @@ export class Book {
 
 Entities like this will be by default ignored when using `SchemaGenerator`, as we need to specify which schema to use. For that we need to use the `schema` option of the `createSchema/updateSchema/dropSchema` methods or the `--schema` CLI parameter.
 
-On runtime, the wildcard schema will be replaced with either `FindOptions.schema`, or with the `schema` option from the ORM config.
+On runtime, the wildcard schema will be replaced with either `FindOptions.schema`, `EntityManager.schema` or with the `schema` option from the ORM config.
 
 ### Note about migrations
 
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -311,7 +311,6 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
     Entity extends object,
     Hint extends string = never,
   >(entityName: string, where: FilterQuery<Entity>, options: FindOptions<Entity, Hint> | FindOneOptions<Entity, Hint>, type: 'read' | 'update' | 'delete'): Promise<FilterQuery<Entity>> {
-    options.schema ??= this._schema;
     where = QueryHelper.processWhere({
       where: where as FilterQuery<Entity>,
       entityName,
@@ -349,7 +348,6 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
   }
 
   protected async applyJoinedFilters<Entity extends object>(meta: EntityMetadata<Entity>, cond: ObjectQuery<Entity>, options: FindOptions<Entity, any> | FindOneOptions<Entity, any>): Promise<ObjectQuery<Entity>> {
-    options.schema ??= this._schema;
     const ret = {} as ObjectQuery<Entity>;
     const populateWhere = options.populateWhere ?? this.config.get('populateWhere');
 
@@ -392,7 +390,6 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
    * @internal
    */
   async applyFilters<Entity extends object>(entityName: string, where: FilterQuery<Entity>, options: Dictionary<boolean | Dictionary> | string[] | boolean, type: 'read' | 'update' | 'delete', findOptions: FindOptions<any, any> | FindOneOptions<any, any> = {}): Promise<FilterQuery<Entity>> {
-    findOptions.schema ??= this._schema;
     const meta = this.metadata.find<Entity>(entityName);
     const filters: FilterDef[] = [];
     const ret: Dictionary[] = [];
@@ -1234,7 +1231,6 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
   merge<Entity extends object>(entityName: EntityName<Entity> | Entity, data?: EntityData<Entity> | EntityDTO<Entity> | MergeOptions, options: MergeOptions = {}): Entity {
     const em = this.getContext();
 
-
     if (Utils.isEntity(entityName)) {
       return em.merge((entityName as Dictionary).constructor.name, entityName as unknown as EntityData<Entity>, data as MergeOptions);
     }
@@ -1287,7 +1283,6 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
    * Shortcut for `wrap(entity).assign(data, { em })`
    */
   assign<Entity extends object>(entity: Entity, data: EntityData<Entity> | Partial<EntityDTO<Entity>>, options: AssignOptions = {}): Entity {
-    options.schema ??= this._schema;
     return EntityAssigner.assign(entity, data, { em: this.getContext(), ...options });
   }
 
@@ -1708,7 +1703,6 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
   }
 
   private async lockAndPopulate<T extends object, P extends string = never>(entityName: string, entity: T, where: FilterQuery<T>, options: FindOneOptions<T, P>): Promise<Loaded<T, P>> {
-    options.schema ??= this._schema;
     if (options.lockMode === LockMode.OPTIMISTIC) {
       await this.lock(entity, options.lockMode, {
         lockVersion: options.lockVersion,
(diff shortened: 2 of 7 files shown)


packages/core/src/EntityManager.ts
```diff
@@ -163,7 +164,8 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
   async find<
     Entity extends object,
     Hint extends string = never,
-  >(entityName: EntityName<Entity>, where: FilterQuery<Entity>, options: FindOptions<Entity, Hint> = {}): Promise<Loaded<Entity, Hint>[]> {
+  >(entityName: EntityName<Entity>, where: FilterQuery<Entity>, _options: FindOptions<Entity, Hint> = {}): Promise<Loaded<Entity, Hint>[]> {
```
B4nan, line 167: please dont rename the parameters, there is no need for this, you just alter the options interface but retain the type, so you can just reassign to the same `options` variable
```suggestion
  >(entityName: EntityName<Entity>, where: FilterQuery<Entity>, options: FindOptions<Entity, Hint> = {}): Promise<Loaded<Entity, Hint>[]> {
```

packages/core/src/EntityManager.ts
```diff
@@ -1539,6 +1574,7 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
 
     fork.filters = { ...em.filters };
     fork.filterParams = Utils.copy(em.filterParams);
+    fork._schema = options.schema || em.schema;
```
B4nan, line 1577:
```suggestion
    fork._schema = options.schema ?? em.schema;
```

packages/core/src/EntityManager.ts
```diff
@@ -163,7 +164,8 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
   async find<
     Entity extends object,
     Hint extends string = never,
-  >(entityName: EntityName<Entity>, where: FilterQuery<Entity>, options: FindOptions<Entity, Hint> = {}): Promise<Loaded<Entity, Hint>[]> {
+  >(entityName: EntityName<Entity>, where: FilterQuery<Entity>, _options: FindOptions<Entity, Hint> = {}): Promise<Loaded<Entity, Hint>[]> {
+    let options = this.mergeOptions<FindOptions<Entity, Hint>>(_options);
```
B4nan, line 168: i also dont think we need the `mergeOptions` method, its already a one-liner (and if we do, it cant be a public method). i would say it should be just this instead:

```suggestion
    options.schema ??= this._schema;
```

B4nan (changes requested):
i like the idea, but the implementation can simplified

packages/core/src/EntityManager.ts
EmmEm: I agree that it doesn't need to be public, but I added the method to allow for other similar options to be applied easily. But that can be done when/if it's needed. I will implement your feedback.

packages/core/src/EntityManager.ts
EmmEm: Resolved

packages/core/src/EntityManager.ts
```diff
@@ -346,6 +349,7 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
   }
 
   protected async applyJoinedFilters<Entity extends object>(meta: EntityMetadata<Entity>, cond: ObjectQuery<Entity>, options: FindOptions<Entity, any> | FindOneOptions<Entity, any>): Promise<ObjectQuery<Entity>> {
+    options.schema ??= this._schema;
```
B4nan, line 352: i'd expect this should be only in the `find` methods, not here (same for the `applyFilters`, `lockAndPopulate`, `assign`, and few others)? maybe also `create` methods

packages/core/src/EntityManager.ts
```diff
@@ -1216,10 +1234,12 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
   merge<Entity extends object>(entityName: EntityName<Entity> | Entity, data?: EntityData<Entity> | EntityDTO<Entity> | MergeOptions, options: MergeOptions = {}): Entity {
     const em = this.getContext();
 
+
```
B4nan, line 1237:
```suggestion
```

packages/core/src/EntityManager.ts
```diff
@@ -1320,7 +1343,11 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
     Entity extends object,
     Hint extends string = never,
   >(entityName: EntityName<Entity>, where: FilterQuery<Entity> = {} as FilterQuery<Entity>, options: CountOptions<Entity, Hint> = {}): Promise<number> {
-    options = { ...options };
+    // Shallow copy options since the object will be modified when deleting orderBy
+    options = {
+      schema: this._schema,
+      ...options,
+    };
```
EmmEm, line 1350: I got in an racing condition error, for example in https://github.com/mikro-orm/mikro-orm/blob/master/tests/issues/GH228.test.ts with `findAndCount()`. The problem was that `count()` was able to delete `options.orderBy` before the `find()` sql was created.

B4nan (changes requested):

packages/core/src/EntityManager.ts
EmmEm: That seems fair, the only place where it should be needed is the entry points. Then it should cascade from there. I will go trough it and check.

packages/core/src/unit-of-work/UnitOfWork.ts
```diff
@@ -255,6 +255,8 @@ export class UnitOfWork {
 
   computeChangeSet<T extends object>(entity: T, type?: ChangeSetType): void {
     const wrapped = helper(entity);
+    // Set entityManager default schema
+    wrapped.__schema ??= this.em.schema;
```
B4nan, line 259: i dont think those changes in UoW are needed, managed entities will have the schema already set in their state object (wrapped entity)

B4nan: actually, this method shouldn't mutate the entity state anyhow. 

do you have tests that show those changes are needed? if they are needed, we should find a better place. i'd say this should work the same as the global schema option, and since its only about inserts, it should be enough to do it here: 

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/unit-of-work/UnitOfWork.ts#L927

EmmEm: It seems to be enough to have it in `findNewEntities`. But if that is the optimal placement is up to you to decide. You know more about the repo then anyone. :)

B4nan: Ok, I guess we can keep it there, it will have a similar effect to the `uow.persist`.

B4nan:
So I guess the last missing thing is to document this somewhere? We should also come up with a better commit message as this is a bit too generic for what this PR is actually doing.

EmmEm:
I've tested some more with request context and found an issue. I think we need to change `this._schema` for `this.schema` otherwise it wont work when using request context. `this._schema` will refer to the global `EntityManager` instance which will be wrong.

B4nan:
I don't follow, change it where?

packages/core/src/EntityManager.ts
```diff
@@ -488,6 +492,8 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
     Entity extends object,
     Hint extends string = never,
   >(entityName: EntityName<Entity>, where: FilterQuery<Entity>, options: FindOneOptions<Entity, Hint> = {}): Promise<Loaded<Entity, Hint> | null> {
+    options.schema ??= this._schema;
+
```
EmmEm, line 496:
```suggestion
    Entity extends object,
    Hint extends string = never,
  >(entityName: EntityName<Entity>, where: FilterQuery<Entity>, options: FindOneOptions<Entity, Hint> = {}): Promise<Loaded<Entity, Hint> | null> {
    options.schema ??= this.schema;

```
If you use NestJs package and get a repo for example
```
constructor(@InjectRepository(Comment) private readonly commentRepo: EntityRepository<Comment>) {}

findComment() {
        const repoManager = this.commentRepo.getEntityManager();
        // Will set schema on the contextual EntityManager
        repoManager.schema = 'client_4';
        
       // client_4
       // Will output the contextual schema
        console.log(repoManager.schema)

       // undefined
       // Will output the schema on global entity manager
        console.log(repoManager._schema)

        // Will query against the orm schema since we use this._schema in findOne
        const comment = await this.commentRepo.findOne({
            id: 2,
        });
}
```

So I guess we need to change to use the contextual schema (`this.schema`)

B4nan: Right, what you should do is to use the EM from context, all the methods first get that and work with it instead of using `this`.

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/EntityManager.ts#L500

No need to duplicate the `getContext` calls for this.

EmmEm: Great, that should solve the problem

EmmEm: I will try and add a test for context schema as well

packages/core/src/unit-of-work/UnitOfWork.ts
```diff
@@ -559,6 +559,8 @@ export class UnitOfWork {
       return;
     }
 
+    // Set entityManager default schema
+    wrapped.__schema ??= this.em.schema;
```
EmmEm, line 563: Should we avoid spamming getContext here as well? Should we send em here or move this?

packages/core/src/EntityManager.ts
```diff
@@ -165,9 +165,9 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
     Entity extends object,
     Hint extends string = never,
   >(entityName: EntityName<Entity>, where: FilterQuery<Entity>, options: FindOptions<Entity, Hint> = {}): Promise<Loaded<Entity, Hint>[]> {
-    options.schema ??= this._schema;
     if (options.disableIdentityMap ?? this.config.get('disableIdentityMap')) {
       const em = this.getContext(false);
+      options.schema ??= em._schema;
```
B4nan, line 170: no need to do it here

packages/core/src/EntityManager.ts
```diff
@@ -492,10 +491,9 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
     Entity extends object,
     Hint extends string = never,
   >(entityName: EntityName<Entity>, where: FilterQuery<Entity>, options: FindOneOptions<Entity, Hint> = {}): Promise<Loaded<Entity, Hint> | null> {
-    options.schema ??= this._schema;
-
     if (options.disableIdentityMap ?? this.config.get('disableIdentityMap')) {
       const em = this.getContext(false);
+      options.schema ??= em._schema;
```
B4nan, line 496: neither here

packages/core/src/EntityManager.ts
EmmEm: true, it will be sorted in the next find

packages/core/src/unit-of-work/UnitOfWork.ts
B4nan: `this.em` is already the contextual EM, so it should be safe, as it will have `useContext = false` on it and `getContext` will just return `this` right ahead.

B4nan (approved):
Unless you plan to make more changes, we can merge.

docs/docs/multiple-schemas.md
```diff
@@ -32,6 +32,41 @@ const qb = em.createQueryBuilder(User);
 await qb.insert({ email: 'foo@bar.com' }).withSchema('client-123');
 ```
 
+## Default schema on `EntityManager`
```
B4nan, line 35: one more thing, i dont plan to ship any more feature releases for some time, so this will go out in a patch update - and it wont have its own docs snapshot, so please add this part to the versioned docs for v5.8 too

EmmEm: ok, I will fix that tomorrow. Otherwise I'm happy with the changes 👍

B4nan:
Thanks!

PR #4744, opened 2023-09-25
feat(core): deprecate `UseRequestContext` decorator

ruscon:


packages/core/src/decorators/CreateRequestContext.ts
```diff
@@ -0,0 +1,22 @@
+import { MikroORM } from '../MikroORM';
+import { RequestContext } from '../utils/RequestContext';
+
+export function CreateRequestContext<T>(getContext?: MikroORM | ((type?: T) => MikroORM)): MethodDecorator {
```
B4nan, line 4: i would just reexport the current implementation under different name, no need to duplicate the code

ruscon: will be the same Error message with  @CreateRequestContext() text

PR #4745, opened 2023-09-25
feat(core): add `EntityRepository.getEntityName()` method

ruscon:


packages/core/src/entity/EntityRepository.ts
```diff
@@ -286,6 +286,10 @@ export class EntityRepository<T extends object> {
     return this._em;
   }
 
+  getEntityName(): EntityName<T> {
+    return this.entityName;
```
B4nan, line 290: maybe we should always return string here? the type could stay the same for better inference support

ruscon: It would be better to return `Constructor<T>`, but I understand this is not possible now :)

ruscon: In any case, I don’t know what cases other people will have.

Those who need a string can use `Utils.className(this.getRepository().getEntityName())`

B4nan: It's supposed to be _entity name_, so I think a string is actually what most people would expect. While we accept both string name and class reference, the getters should be consistent in what they return.

If you need the class reference, you can always get that from the metadata, e.g. `em.getMetadata('Foo').class`.

ruscon: Ok, done

packages/core/src/entity/EntityRepository.ts
```diff
@@ -286,8 +286,8 @@ export class EntityRepository<T extends object> {
     return this._em;
   }
 
-  getEntityName(): EntityName<T> {
-    return this.entityName;
+  getEntityName(): string {
+    return Utils.className(this.entityName)
```
B4nan:
```suggestion
    return Utils.className(this.entityName);
```

ruscon:
Please make a new npm version when you have time. I want to test it. Thanks.

B4nan:
every commit to master is published to npm as dev version

PR #4752, opened 2023-09-27
fix(core): make `em.create` strictly typed for relations too

### LINKED ISSUE #4748: Embeddable Properties Incorrectly Marked as Optional
**Describe the bug**
There's an issue with the type definitions in MikroOrm. When you have an entity that includes an "Embeddable", TypeScript marks all properties of that Embeddable as optional. This leads to entities being created without requesting parameters that are actually required.

**Example**
```typescript
@Embeddable()
export class UserEmail {
  @Property({
    hidden: true,
  })
  code!: number;

  @Property()
  codeSentAt?: Date;

  @Enum({
    items: () => UserEmailStatuses,
  })
  status!: UserEmailStatuses;

  @Property({
    unique: true,
  })
  value!: string;
}

@Embeddable()
export class UserPhoneMobile {
  @Property({
    hidden: true,
  })
  code!: number;

  @Property()
  codeSentAt?: Date;

  @ManyToOne()
  country!: Country;

  @Property()
  number!: string;

  @Enum({
    items: () => UserPhoneMobileStatuses,
  })
  status!: UserPhoneMobileStatuses;
}

@Embeddable()
export class UserPhone {
  @Embedded({
    object: true,
  })
  mobile!: UserPhoneMobile;
}

@Entity({
  collection: 'users',
})
export class User extends Base {
  @Embedded({
    object: true,
  })
  email?: UserEmail;

  @Embedded({
    object: true,
  })
  phone!: UserPhone;

  @Property({
    hidden: true,
  })
  password?: string;

  @Property({
    unique: true,
  })
  username?: string;
}
```
If you try to create an entity named "User", you'll notice that all the properties of the "Embeddable UserEmail" appear as optional. Yet, it has required properties such as "code", "status", and "value". The same situation arises with the "Embeddable UserPhone": the "mobile" property is required, but when you create the "User" entity, "mobile" is displayed as optional, as are all properties contained within "mobile".

**Expected behavior**
Optional properties should remain optional, and required ones should stay required.

**Additional context**
I'm using the metadata provider "TsMorphMetadataProvider".

**Versions**
| Dependency | Version |
| - | - |
| node | 20.5.1 |
| typescript | 5.2.2 |
| mikro-orm | 5.8.3 |
| mikro-orm/mongodb | 5.8.3 |
| nestjs | 10.2.6 |

B4nan:
When using `em.create()`, we can create the entity graph in one step, including relations. This commit makes the typing strict for the relations too, requiring their mandatory props to be present

There is one common issue with this - the `em.create` automatically propagates things to inverse sides, in other words, doing this is completely fine, as the author will be propagated to the new books:

```ts
em.create(Author, {
  name: '...',
  books: [{ title: 'foo' }], // Book.author is required, but will be propagated from the upper scope
});
```

To get around it, the child properties matching their parent type are automatically considered as optional - in other words, since we are processing contents of `Author.books`, we will make all properties of type `Author` in the `Author.books` as optional. It's not perfect, as it will mark every `Author` property as optional, not just the one that is an inverse of `Author.books`, but I feel like it's the tradeoff we need to make.

Closes #4748

### DIFF
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -1204,7 +1204,7 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
 
     data = QueryHelper.processObjectParams(data);
     em.validator.validateParams(data, 'insert data');
-    const res = await em.driver.nativeInsert<Entity>(entityName, data, { ctx: em.transactionContext, ...options });
+    const res = await em.driver.nativeInsert<Entity>(entityName, data as EntityData<Entity>, { ctx: em.transactionContext, ...options });
 
     return res.insertId!;
   }
@@ -1248,7 +1248,7 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
 
     data = data.map(row => QueryHelper.processObjectParams(row));
     data.forEach(row => em.validator.validateParams(row, 'insert data'));
-    const res = await em.driver.nativeInsertMany<Entity>(entityName, data, { ctx: em.transactionContext, ...options });
+    const res = await em.driver.nativeInsertMany<Entity>(entityName, data as EntityData<Entity>[], { ctx: em.transactionContext, ...options });
 
     return res.insertedIds!;
   }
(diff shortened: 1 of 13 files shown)


devcaeg:
Wow, such a quick solution! I've tried it and it works correctly.

Issue #4711, opened 2023-09-18
Incorrect select when selecting an embedded property from a joined entity if prefix is set to false


dirslashls:
**Describe the bug**
The following code generates incorrect SQL

```ts
		const query = em.qb(TestB,'b')
				.leftJoin('b.a','a')
				.select(['b.*','a.name'])
				.getQuery();
		console.debug(query);
```

```sql
select `b`.*, `b`.`name` from `test_b` as `b` left join `test_a` as `a` on `b`.`a_id` = `a`.`id`
```

Instead of `a`.`name` , the SQL is generated with `b`.`name`. Here a.name is from an embedded entity with prefix set to false.

**To Reproduce**

Below are the entity definitions.

```ts
@Embeddable()
class TestNamedEntity {
	@Property({ length: 100 })
	name!: string;

	@Property({ length: 250 })
	description!: string;
}

@Entity()
class TestA {
	@PrimaryKey()
	id: number;

	@Embedded({ prefix: false })
	identity: TestNamedEntity;
}

@Entity()
class TestB {
	@PrimaryKey()
	id: string;

	@ManyToOne('TestA',{ cascade: [] })
	a?: Ref<TestA>;
}
```

**Expected behavior**
The SQL should be 

```sql
select `b`.*, `a`.`name` from `test_b` as `b` left join `test_a` as `a` on `b`.`a_id` = `a`.`id`
```

**Additional context**
This happens with prefix = false. 

**Versions**

| Dependency | Version |
| - | - |
| node |  v16.15.0  |
| typescript | 5.0.2 |
| mikro-orm | 5.8.1 |
| your-driver | "@mikro-orm/mysql": "^5.8.1" |

PR #4824, opened 2023-10-15
fix(query-builder): add alias to embedded property only if embedded is not prefixed

mlsm-trl:
When querying an entity that has prefixed Embedded, `TableNotFoundException` is thrown due to wrong generated query syntax.

Here is the reproduction and contextualization of the bug:
[mlsm-trl/mikro-orm-bug-virtual-properties-required-when-seeding](https://github.com/mlsm-trl/mikro-orm-bug-alias-embedded)

The idea would be to check if property not comes from a prefixed Embedded, and then add the alias. 

I don't know if we have a better/reliable way to do this than the string comparison I made.

### DIFF
--- a/packages/knex/src/query/QueryBuilder.ts
+++ b/packages/knex/src/query/QueryBuilder.ts
@@ -903,7 +903,8 @@ export class QueryBuilder<T extends object = AnyEntity> {
       }
 
       if (prop?.embedded) {
-        const name = this._aliases[a] ? `${a}.${prop.fieldNames[0]}` : prop.fieldNames[0];
+        const hasPrefix = prop.name.startsWith(`${prop.embedded[0]}_`);
+        const name = this._aliases[a] && !hasPrefix ? `${a}.${prop.fieldNames[0]}` : prop.fieldNames[0];
         const fieldName = this.helper.mapper(name, this.type) as string;
         ret.push(fieldName);
         return;


B4nan (changes requested):
Why should a prefixed embedded property be treated somehow differently? Sounds like you are too concerned with the fact that it's prefixed, but the problem will be probably something else.

Let's start by turning that repro into a test case (check the `tests/issues` folder), so it's easier for me to try to find a better solution, as this one seems wrong - we can't make things behave differently based on a property name...

B4nan:
Looking at your repro, is this related to the custom types (which you use there)? As you haven't mentioned them anywhere, but I can surely see some issues with prefixing such properties - but that has nothing to do with the property name.

mlsm-trl:
> Looking at your repro, is this related to the custom types (which you use there)? As you haven't mentioned them anywhere, but I can surely see some issues with prefixing such properties

Yes, `geolocation` prop inside embedded is a custom type, which then sets the prop with `hasConvertToDatabaseValueSQL` and `hasConvertToJSValueSQL`, as I mentioned.

> Let's start by turning that repro into a test case (check the `tests/issues` folder), so it's easier for me to try to find a better solution, as this one seems wrong - we can't make things behave differently based on a property name...

Yeah, it is very unreliable as I said, I don't have deep knowledge of codebase to came with a better solution... But the idea I think remains the same.

> ... I can surely see some issues with prefixing such properties - but that has nothing to do with the property name.
> ...
> Why should a prefixed embedded property be treated somehow differently? Sounds like you are too concerned with the fact that it's prefixed, but the problem will be probably something else.

In v5.7.14 this same repro works. I don't know what could have changed from v5.7.14 to v5.8.0, but whatever it is affected the emitted query. I suspect that has some to do with  issue [#4711](https://github.com/mikro-orm/mikro-orm/issues/4711) as aliasing changed  there, making the `mapper` not find the property in [QueryBuilderHelper.ts#L76...L79](https://github.com/mikro-orm/mikro-orm/blob/66278ea466df5434dbee36a8b45225998b888a94/packages/knex/src/query/QueryBuilderHelper.ts#L76...L79) (I did test with some console.log's)...

```JS
let ret = field;
...
console.log(field)
// u0.addresses.geolocation
const [a, f] = this.splitField(field);
console.log(a, f)
// u0 addresses.geolocation
const prop = this.getProperty(f, a);
console.log(prop)
// undefined
```

but I don't know if the change in alias really is the culprit...

In v5.7.14, this is the produced query:

```SQL
select "u0".*, "u0"."addresses" from "user" as "u0" where "u0"."id" = 1 limit 1
````

and then in v5.8.0...v5.8.8:

```SQL
select "u0".*, "u0"."addresses"."geolocation" from "user" as "u0" where "u0"."id"
```

which is incorrect

mlsm-trl:
> Looking at your repro, is this related to the custom types (which you use there)? As you haven't mentioned them anywhere, but I can surely see some issues with prefixing such properties

Yes, `geolocation` prop inside embedded is a custom type, which then sets the prop with `hasConvertToDatabaseValueSQL` and `hasConvertToJSValueSQL`, as I mentioned.

> Let's start by turning that repro into a test case (check the `tests/issues` folder), so it's easier for me to try to find a better solution, as this one seems wrong - we can't make things behave differently based on a property name...

Yeah, it is very unreliable as I said, I don't have deep knowledge of codebase to came with a better solution... But the idea I think remains the same.

> ... I can surely see some issues with prefixing such properties - but that has nothing to do with the property name.
> ...
> Why should a prefixed embedded property be treated somehow differently? Sounds like you are too concerned with the fact that it's prefixed, but the problem will be probably something else.

In v5.7.14 this same repro works. I don't know what could have changed from v5.7.14 to v5.8.0, but whatever it is affected the emitted query. I suspect that has some to do with  issue [#4711](https://github.com/mikro-orm/mikro-orm/issues/4711) as aliasing changed  there, making the `mapper` not find the property in [QueryBuilderHelper.ts#L76...L79](https://github.com/mikro-orm/mikro-orm/blob/66278ea466df5434dbee36a8b45225998b888a94/packages/knex/src/query/QueryBuilderHelper.ts#L76...L79) (I did test with some console.log's)...

```JS
let ret = field;
...
console.log(field)
// u0.addresses.geolocation
const [a, f] = this.splitField(field);
console.log(a, f)
// u0 addresses.geolocation
const prop = this.getProperty(f, a);
console.log(prop)
// undefined
```

and in v5.7.14 the field is found, because the "u0" is not there:

```JS
let ret = field;
...
console.log(field)
// addresses.geolocation
const [a, f] = this.splitField(field);
console.log(a, f)
// addresses geolocation
const prop = this.getProperty(f, a);
console.log(prop)
/*
{
  name: 'addresses',
  type: 'Address',
  reference: 'embedded'
  ...
}
*/
```

but I don't know if the change in alias really is the culprit...

In v5.7.14, this is the produced query:

```SQL
select "u0".*, "u0"."addresses" from "user" as "u0" where "u0"."id" = 1 limit 1
````

and then in v5.8.0...v5.8.8:

```SQL
select "u0".*, "u0"."addresses"."geolocation" from "user" as "u0" where "u0"."id"
```

which is incorrect

mlsm-trl:
> Looking at your repro, is this related to the custom types (which you use there)? As you haven't mentioned them anywhere, but I can surely see some issues with prefixing such properties

Yes, `geolocation` prop inside embedded is a custom type, which then sets the prop with `hasConvertToDatabaseValueSQL` and `hasConvertToJSValueSQL`, as I mentioned.

> Let's start by turning that repro into a test case (check the `tests/issues` folder), so it's easier for me to try to find a better solution, as this one seems wrong - we can't make things behave differently based on a property name...

Yeah, it is very unreliable as I said, I don't have deep knowledge of codebase to came with a better solution... But the idea I think remains the same.

> ... I can surely see some issues with prefixing such properties - but that has nothing to do with the property name.
> ...
> Why should a prefixed embedded property be treated somehow differently? Sounds like you are too concerned with the fact that it's prefixed, but the problem will be probably something else.

In v5.7.14 this same repro works. I don't know what could have changed from v5.7.14 to v5.8.0, but whatever it is affected the emitted query. I suspect that has some to do with  issue [#4711](https://github.com/mikro-orm/mikro-orm/issues/4711) as aliasing changed  there, making the `mapper` not find the property in [QueryBuilderHelper.ts#L76...L79](https://github.com/mikro-orm/mikro-orm/blob/66278ea466df5434dbee36a8b45225998b888a94/packages/knex/src/query/QueryBuilderHelper.ts#L76...L79) (I did test with some console.log's)...

```JS
let ret = field;
...
console.log(field)
// u0.addresses.geolocation
const [a, f] = this.splitField(field);
console.log(a, f)
// u0 addresses.geolocation
const prop = this.getProperty(f, a);
console.log(prop)
// undefined
```

and in v5.7.14 the field is found, because the `u0` alias is not there:

```JS
let ret = field;
...
console.log(field)
// addresses.geolocation
const [a, f] = this.splitField(field);
console.log(a, f)
// addresses geolocation
const prop = this.getProperty(f, a);
console.log(prop)
/*
{
  name: 'addresses',
  type: 'Address',
  reference: 'embedded'
  ...
}
*/
```

but I don't know if the change in alias really is the culprit...

In v5.7.14, this is the produced query:

```SQL
select "u0".*, "u0"."addresses" from "user" as "u0" where "u0"."id" = 1 limit 1
````

and then in v5.8.0...v5.8.8:

```SQL
select "u0".*, "u0"."addresses"."geolocation" from "user" as "u0" where "u0"."id"
```

which is incorrect

B4nan:
> Yes, geolocation prop inside embedded is a custom type, which then sets the prop with hasConvertToDatabaseValueSQL and hasConvertToJSValueSQL, as I mentioned.

Your repro is using those on an embedded array property, in other words, on a JSON property - this is not supported, the SQL conversion methods will work only with a real column. So it actually did not work before either, it just didn't fail on the query level like this, but the SQL convertors weren't there - and that is the only reason why this was added to the select clause.

The correct fix is to completely omit this bit from the select clause.

PR #4844, opened 2023-10-18
fix(reflection): ensure complete stripping of relative paths with multiple leading slashes

AbdlrahmanSaberAbdo:
This PR addresses an issue where relative paths with leading slashes and dots were not being stripped correctly. The problem arises when the 'path' contains multiple leading slashes like `../../`, as the current logic strips it to `/../`, which is incorrect and prevents the entity from being discovered.

This code modification ensures that any dots followed by slashes are stripped to provide the full path without any dots and slashes at the beginning.

**Please note:** that the current behavior is such that if you pass `./my/path` the result is `/my/path` However, with the updated code, the result will be `my/path` This change is made because, in the end, we use the resulting path to check if the `source` contains the path using `endsWith`, and the leading slash is unnecessary for this purpose.

## Update
I added a slash again after @B4nan clarified a case that we need it in this [thread](https://mikroorm.slack.com/archives/CFTFEK39D/p1697634946389959?thread_ts=1697550457.817269&cid=CFTFEK39D) 

so the result now for `./my/path` will be `/my/path`

### DIFF
--- a/packages/reflection/src/TsMorphMetadataProvider.ts
+++ b/packages/reflection/src/TsMorphMetadataProvider.ts
@@ -136,9 +136,7 @@ export class TsMorphMetadataProvider extends MetadataProvider {
     if (!this.sources) {
       await this.initSourceFiles();
     }
-
-    const source = this.sources.find(s => s.getFilePath().endsWith(tsPath.replace(/^\.+/, '')));
-
+    const source = this.sources.find(s => s.getFilePath().endsWith(Utils.stripRelativePath(tsPath)));
     if (!source && validate) {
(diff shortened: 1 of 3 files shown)


packages/reflection/src/TsMorphMetadataProvider.ts
```diff
@@ -136,9 +136,7 @@ export class TsMorphMetadataProvider extends MetadataProvider {
     if (!this.sources) {
       await this.initSourceFiles();
     }
-
-    const source = this.sources.find(s => s.getFilePath().endsWith(tsPath.replace(/^\.+/, '')));
-
+    const source = this.sources.find(s => s.getFilePath().endsWith(Utils.stripRelativePath(tsPath)));
```
B4nan, line 139: lets keep the blank lines

B4nan (approved):

B4nan:
Thanks!

PR #4857, opened 2023-10-20
chore(core): export schema types

### LINKED ISSUE #4856: Expose Schema Types
As my team is more and more using Schemas, we're creating some utility methods to build them more easily. However, we would like to have access to some of the types defined in the `EntitySchema` file, like:

* [Metadata](https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/metadata/EntitySchema.ts#L23)
* [EntityMetadata](https://github.com/mikro-orm/mikro-orm/blob/9316c377d69818ee28f0a9dcec7d44d511335db7/packages/core/src/metadata/EntitySchema.ts#L1C17-L1C17)
* [Property](https://github.com/mikro-orm/mikro-orm/blob/9316c377d69818ee28f0a9dcec7d44d511335db7/packages/core/src/metadata/EntitySchema.ts#L15C6-L15C14)

Just export these types so we can build features around them.

We've considered duplicating these types, but that's a lot of effort and we would have to keep it up to date on each MikroORM change

**Additional context**

thiagomini:
Closes #4856

### DIFF
--- a/packages/core/src/metadata/EntitySchema.ts
+++ b/packages/core/src/metadata/EntitySchema.ts
@@ -12,7 +12,7 @@ import { EnumArrayType } from '../types/EnumArrayType';
 
 type TypeType = string | NumberConstructor | StringConstructor | BooleanConstructor | DateConstructor | ArrayConstructor | Constructor<Type<any>>;
 type TypeDef<T> = { type: TypeType } | { customType: Type<any> } | { entity: string | (() => string | EntityName<T>) };
-type Property<T, O> =
+export type Property<T, O> =
   | ({ reference: ReferenceType.MANY_TO_ONE | 'm:1' } & TypeDef<T> & ManyToOneOptions<T, O>)
   | ({ reference: ReferenceType.ONE_TO_ONE | '1:1' } & TypeDef<T> & OneToOneOptions<T, O>)
   | ({ reference: ReferenceType.ONE_TO_MANY | '1:m' } & TypeDef<T> & OneToManyOptions<T, O>)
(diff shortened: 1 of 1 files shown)


thiagomini:
> The names are a bit too generic, I surely dont mind exporting the types, but we should pick a bit more specific names ideally.

Sure, Martin! Do you have any suggestions for them? Maybe, `SchemaPropertyDefinition`?

packages/core/src/metadata/EntitySchema.ts
```diff
@@ -25,6 +25,9 @@ type Metadata<T, U> =
   & ({ name: string } | { class: Constructor<T>; name?: string })
   & { properties?: { [K in keyof Omit<T, keyof U> as ExcludeFunctions<Omit<T, keyof U>, K>]-?: Property<ExpandProperty<NonNullable<T[K]>>, T> } };
 
+export type SchemaProperty<T, O> = Property<T, O>;
+export type SchemaMetadata<T, U> = Metadata<T, U>;
```
B4nan, line 29: i would just rename the types instead of aliasing them. and maybe lets use the full `EntitySchema` prefix so its clearer

i will need to look into those types in v6, as i recently discovered its no longer type-safe as it was (no excess property checks, which would be pretty bad for v6 migration, where `reference` option changed to `kind` - and this wont fail the compilation currently. and i also wanted to try to implement some kind of type inference, as with zod, so if you would define the schema without providing any class or interface, you could still infer the interface from it.

thiagomini: Done, Martin! Thanks for the suggestion :)

B4nan: looks like you forgot to export them

thiagomini: You are right, lol, fixed again!

thiagomini: Hey again, @B4nan , it seems we had a coverage issue by exporting these types 🤔 is this relevant?

B4nan (approved):

Issue #4859, opened 2023-10-20
Custom serializer does not take into account getters


ivanpondal:
**Describe the bug**
When trying to use custom property serialization on class methods it gets ignored.

**To Reproduce**
1. Define an entity method property with custom serialize function.
2. Serialize and compare result.

**Expected behavior**
Custom serialization should run.

**Versions**

| Dependency | Version |
| - | - |
| node | ? |
| typescript | ? |
| mikro-orm | ? |
| your-driver | ? |

ivanpondal:
I'll be working on this, created the issue only to have a tracker id 👍

PR #4860, opened 2023-10-20
fix(serialization): run custom serializer on getters

ivanpondal:
Fixes https://github.com/mikro-orm/mikro-orm/issues/4859

### DIFF
--- a/packages/core/src/serialization/EntitySerializer.ts
+++ b/packages/core/src/serialization/EntitySerializer.ts
@@ -96,13 +96,13 @@ export class EntitySerializer {
 
     // decorated getters
     meta.props
-      .filter(prop => prop.getter && typeof entity[prop.name] !== 'undefined' && isVisible(meta, prop.name, options))
+      .filter(prop => prop.getter && prop.getterName === undefined && typeof entity[prop.name] !== 'undefined' && isVisible(meta, prop.name, options))
       .forEach(prop => ret[this.propertyName(meta, prop.name, wrapped.__platform)] = this.processProperty(prop.name, entity, options));
 
     // decorated get methods
     meta.props
       .filter(prop => prop.getterName && entity[prop.getterName] as unknown instanceof Function && isVisible(meta, prop.name, options))
-      .forEach(prop => ret[this.propertyName(meta, prop.name, wrapped.__platform)] = (entity[prop.getterName!] as unknown as () => T[keyof T & string])());
+      .forEach(prop => ret[this.propertyName(meta, prop.name, wrapped.__platform)] = this.processProperty(prop.getterName as keyof T & string, entity, options));
 
     return ret;
   }
@@ -129,6 +129,9 @@ export class EntitySerializer {
 
     /* istanbul ignore next */
     if (!options.ignoreSerializers && serializer) {
+      if (entity[prop] as unknown instanceof Function) {
+        return serializer((entity[prop] as unknown as () => T[keyof T & string])());
+      }
       return serializer(entity[prop]);
     }
 
--- a/packages/core/src/serialization/EntitySerializer.ts
+++ b/packages/core/src/serialization/EntitySerializer.ts
@@ -97,7 +97,7 @@ export class EntitySerializer {
     // decorated getters
     meta.props
       .filter(prop => prop.getter && typeof entity[prop.name] !== 'undefined' && isVisible(meta, prop.name, options))
-      .forEach(prop => ret[this.propertyName(meta, prop.name, wrapped.__platform)] = entity[prop.name]);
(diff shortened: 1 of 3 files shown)


packages/core/src/serialization/EntitySerializer.ts
```diff
@@ -97,7 +97,7 @@ export class EntitySerializer {
     // decorated getters
     meta.props
       .filter(prop => prop.getter && typeof entity[prop.name] !== 'undefined' && isVisible(meta, prop.name, options))
-      .forEach(prop => ret[this.propertyName(meta, prop.name, wrapped.__platform)] = entity[prop.name]);
+      .forEach(prop => ret[this.propertyName(meta, prop.name, wrapped.__platform)] = this.processProperty(prop.name, entity, options));
 
     // decorated get methods
     meta.props
```
B4nan, line 103: we should do the same here as well, right?

tests/features/serialization/GH4859.test.ts
```diff
@@ -0,0 +1,55 @@
+import { Entity,  PrimaryKey, Property, serialize } from '@mikro-orm/core';
+import { MikroORM } from '@mikro-orm/sqlite';
+
+@Entity()
+export class TimeSeriesEntityTest {
```
B4nan, line 5: no need to export things from the tests
```suggestion
class TimeSeriesEntityTest {
```

tests/features/serialization/GH4859.test.ts
```diff
@@ -0,0 +1,55 @@
+import { Entity,  PrimaryKey, Property, serialize } from '@mikro-orm/core';
```
B4nan, line 1:
```suggestion
import { Entity, PrimaryKey, Property, serialize } from '@mikro-orm/core';
```

B4nan (changes requested):
two more nits

packages/core/src/serialization/EntitySerializer.ts
```diff
@@ -97,7 +97,7 @@ export class EntitySerializer {
     // decorated getters
     meta.props
       .filter(prop => prop.getter && typeof entity[prop.name] !== 'undefined' && isVisible(meta, prop.name, options))
-      .forEach(prop => ret[this.propertyName(meta, prop.name, wrapped.__platform)] = entity[prop.name]);
```
B4nan, line 100: actually, this also needs to be addressed in the `EntityTransformer` class, which is used for implicit serialization. which reminds me i wanted to look into deduplicating this in v6, now the two classes are almost the same

packages/core/src/serialization/EntitySerializer.ts
ivanpondal: Yep, had to do some small changes. First off, added a `getterName === undefined` check to the accessor serializer. This was to avoid it being called twice (for method getters too). Then I modified the `processProperty` function so that if the property is a function (the method getter case) then it gets the corresponding value by calling the function. Let me know if I'm missing something or if you can think of a more elegant way of doing it 👍

packages/core/src/serialization/EntitySerializer.ts
ivanpondal: Could you point me to some examples where `EntityTransformer` is being used as to test it accordingly? Thanks!

packages/core/src/serialization/EntitySerializer.ts
```diff
@@ -129,6 +129,9 @@ export class EntitySerializer {
 
     /* istanbul ignore next */
     if (!options.ignoreSerializers && serializer) {
+      if (entity[prop] as unknown instanceof Function) {
+        return serializer((entity[prop] as unknown as () => T[keyof T & string])());
```
ivanpondal, line 133: This is the same cast that was being used to [obtain the getter method value](https://github.com/mikro-orm/mikro-orm/pull/4860/files#diff-8fd27dc733983b948f557254ee3f4f004f49b7ab4188eab091b9052d90624451L105).

packages/core/src/serialization/EntitySerializer.ts
B4nan: `wrap(entity).toObject()` is using it, same for `JSON.stringify(entity)`

ivanpondal: added a test case for `wrap(entity).toObject()` 👍

ivanpondal:
Will squash commits once the pr is ready to conform with commit format.

ivanpondal:
I'm looking into the test failure. Things broke with the `EntityTransformer` changes.

B4nan (approved):
thanks!

Issue #4863, opened 2023-10-22
Unnecessary inserting entity after deleting it


vitaliy-art:
**Describe the bug**
I described three entities, one of them is owned by others via `@ManyToOne` attribute. Then i create first entity, create second entity and create third entity, owned by first two. Then i delete first entity, update second entity and delete second entity. I expects that all entities is deleted from database, but before update second entity entity manager creates first entity again. And after all this steps i have in database first entity which was deleted.

**Stack trace**
There is no errors in stack trace.

**To Reproduce**
```typescript
import { MikroORM, defineConfig } from '@mikro-orm/sqlite';
import { Entity, Property, PrimaryKey, ManyToOne, Cascade } from '@mikro-orm/core';

@Entity()
class Department {
    @PrimaryKey()
    public id!: number;

    @Property()
    public name!: string;
}

@Entity()
class Person {
    @PrimaryKey()
    public id!: number;

    @Property()
    public name!: string;
}

@Entity()
class Employee {
    @PrimaryKey()
    public id!: number;

    @ManyToOne({ onDelete: 'cascade' })
    public department!: Department;

    @ManyToOne({ onDelete: 'cascade' })
    public person!: Person;
}

(async () => {
    const config = defineConfig({
        entities: [Department, Person, Employee],
        dbName: 'db.bd',
        type: 'sqlite',
        allowGlobalContext: true,
        driverOptions: {
            'foreign_keys': 'ON',
        },
        debug: true,
    });

    const orm = await MikroORM.init(config);
    await orm.schema.updateSchema();
    const em = orm.em;

    const dep = new Department();
    dep.name = 'Department';
    await em.persistAndFlush(dep);

    const per = new Person();
    per.name = 'Person';
    await em.persistAndFlush(per);

    const emp = new Employee();
    emp.department = dep;
    emp.person = per;
    await em.persistAndFlush(emp);

    await em.removeAndFlush(dep);

    per.name = 'New Person';
    await em.persistAndFlush(per);
    await em.removeAndFlush(per);
    await orm.close();
})();
```

**Starting log**
```log
Debugger attached.
[info] MikroORM version: 5.8.10
[discovery] ORM entity discovery started, using ReflectMetadataProvider
[discovery] - processing entity Department
[discovery] - processing entity Person
[discovery] - processing entity Employee
[discovery] - entity discovery finished, found 3 entities, took 19 ms
[info] MikroORM successfully connected to database db.bd
[...]

PR #4864, opened 2023-10-22
fix(core): recursively remove MANY_TO_ONE entities

vitaliy-art:
Added test `tests\issues\GH4863.test.ts` reproduces behavior from issue.

https://github.com/mikro-orm/mikro-orm/issues/4863

### DIFF
--- a/packages/core/src/unit-of-work/UnitOfWork.ts
+++ b/packages/core/src/unit-of-work/UnitOfWork.ts
@@ -306,12 +306,21 @@ export class UnitOfWork {
   }
 
   remove<T extends object>(entity: T, visited?: Set<AnyEntity>, options: { cascade?: boolean } = {}): void {
+    const entityHelper = helper(entity);
     this.removeStack.add(entity);
-    this.queuedActions.add(helper(entity).__meta.className);
+    this.queuedActions.add(entityHelper.__meta.className);
     this.persistStack.delete(entity);
 
+    // remove entities referencing to entity via MANY_TO_ONE with onDelete = cascade
+    this.identityMap.values().filter(v => {
+      const refHelper = helper(v);
+      const relations = refHelper.__meta.relations.filter(r => r.type === entityHelper.__meta.className && r.onDelete === 'cascade' && r.reference === ReferenceType.MANY_TO_ONE);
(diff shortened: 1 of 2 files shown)


vitaliy-art:
Added possibly fix.

packages/core/src/unit-of-work/UnitOfWork.ts
```diff
@@ -306,12 +306,21 @@ export class UnitOfWork {
   }
 
   remove<T extends object>(entity: T, visited?: Set<AnyEntity>, options: { cascade?: boolean } = {}): void {
+    const entityHelper = helper(entity);
     this.removeStack.add(entity);
-    this.queuedActions.add(helper(entity).__meta.className);
+    this.queuedActions.add(entityHelper.__meta.className);
     this.persistStack.delete(entity);
 
+    // remove entities referencing to entity via MANY_TO_ONE with onDelete = cascade
```
B4nan, line 314: this feels wrong, the cascading is already implemented elsewhere (happens on commit), and we surely don't need to traverse the whole identity map to do this. and I'd argue its wrong to even cascade the `remove` call on those, the problem is about removing the link, not removing the referenced entities - basically what the code under those changed lines is doing. 

also you are comparing `referencedColumnNames` (so column names) with the PK (a property name), this will fail if case of property/fieldName mismatch. also the same problem will happen with 1:1 relations probably.

vitaliy-art: Thanks for clearing it up! Where i can find code removing the link?

B4nan: Right after your changes:

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/unit-of-work/UnitOfWork.ts#L314

I believe that code is there to deal with this exact issue (but with a different setup of relations). The fix will be rather about doing some `delete` than cascading the `remove` operation (probably).

I will try to review this myself soon, those issues are pretty hard to fix correctly without knowing all the internals. Thanks for the test case, that is surely helpful!

vitaliy-art: Many thanks!

Issue #3484, opened 2022-09-11
Schema first approach with generated client


B4nan:
Original idea coming from this tweet: https://twitter.com/MikroOrm/status/1568534840370237441

TODO: 

- [ ] generate simple repositories next to the entities (in separate files, linking them via `EntityRepositoryType` symbol)
- [ ] allow to generate JS/D.TS, again in separate files - for code gen we would have to go with this approach probably
- [ ] in the generated repository, we should extend the driver package, for that we will need a way to get package name and driver class (there is no way currently)

Last step would be that "generated client". I have this simple script I used for the PoC, it generates entities, repositories and a client file which initializes the ORM and exports a map of services:

```ts
import { MikroORM } from '@mikro-orm/core';
import { dirname } from 'node:path';
import fs from 'fs-extra';
import { fileURLToPath } from 'node:url';

const orm = await MikroORM.init();
const dirName = dirname(fileURLToPath(import.meta.url));
const driver = { package: '@mikro-orm/better-sqlite', className: 'BetterSqliteDriver' };

// clean up everything in the folder except dot files
await fs.emptyDir(dirName + '/entities');
await fs.emptyDir(dirName + '/repositories');

const entityGenerator = orm.config.get('entityGenerator');
entityGenerator.identifiedReferences = true;
entityGenerator.bidirectionalRelations = true;
// orm.config.set('entityGenerator', entityGenerator);
const ret = await orm.entityGenerator.generate({
  baseDir: dirName + '/entities',
  save: true,
});

const metadata = orm.getMetadata().getAll();
const entities = Object.values(metadata).filter(meta => !meta.pivotTable && !meta.embeddable && !meta.virtual);

for (const meta of entities) {
  const code = [
    `import { EntityRepository } from '${driver.package}';`,
    `import { ${meta.className} } from '../entities/${meta.className}.js';`,
    '',
    `export class ${meta.className}Repository extends EntityRepository<${meta.className}> { }`,
  ];
  await fs.writeFile(`${dirName}/repositories/${meta.className}Repository.ts`, code.join('\n'));
}

const client: string[] = [];
const coreImports: string[] = [];

client.push(`import { ${driver.className} } from '${driver.package}';`);

for (const meta of entities) {
  client.push(`import { ${meta.className} } from './entities/${meta.className}.js';`);
  client.push(`import { ${meta.className}Repository } from './repositories/${meta.className}Repository.js';`);
}

client.push('');
coreImports.push('MikroORM');
[...]

PR #4892, opened 2023-11-03
feat(entity-generator): allow generating scalar properties for FKs

boenrobot:
When a pivot table's primary key is referenced by a foreign key constraint, render an entity for the pivot table.

NOTE: I am using Windows, and the test suite seems to not pass on it. I'm also not sure how to add a new test for this. But I did test it with my code base, where it does end up generating the correct entities and references to them.


B4nan (changes requested):
> NOTE: I am using Windows, and the test suite seems to not pass on it. 

You should use WSL, I believe that should do it.

> I'm also not sure how to add a new test for this. 

I don't want to accept any code changes I don't understand without seeing the why, so a test case is needed. I looked at the change you did and no idea why you would need to make things conditional based on that. Sounds like you have a wrong entity definition rather than its something to fix in the ORM, but I might be wrong - hard to guess without the failing test case first.

What do you not understand about creating one? Check this folder, it's full of examples for bug reproductions:

https://github.com/mikro-orm/mikro-orm/tree/master/tests/issues

If you struggle with the tests, it can be as well a separate repository, all I need is to understand what you do, I can help with making a test case out of it. 

---

Also, you should drop all the unrelated code formatting changes (e.g. I can see you changed the order of imports for no reason).

B4nan:
Thanks for the details, now I see what you are trying to fix, that's a good first step. The generated entity is indeed wrong (even if the pivot entity would be generated, there should not be the `senderId` column at all, only the relation one is supposed to be there), but the fix you made still does not feel correct (or better say it's still very confusing). I will have a look later today and try to come up with something cleaner and more general, as you are trying to fix only the edge case you face, while the issue feels rather a bit wider.

> I am using a schema first approach

MikroORM is a code-first framework, entity generator is a way to bootstrap your project from existing schema, not a schema-first approach per se. Over time it's getting better and better, so in a way you can use it this way too, but it never was the supposed usage. I've created #3484 some time ago with some ideas of what should be done to make this a proper schema-first approach, but it has low priority, again because MikroORM is a code-first framework.

B4nan:
Isn't the right fix here to generate the pivot entity as a real file? It feels like that's a more sane approach, in the end, your schema references it from other places.

(fwiw, you can have a pivot entity defined as a regular entity nowadays, see https://mikro-orm.io/docs/collections#custom-pivot-table-entity)

B4nan:
I would prefer to keep things as they were for existing tests (so not printing the field names if they patch the naming strategy, etc).

tests/issues/__snapshots__/GH4892.test.ts.snap
```diff
@@ -0,0 +1,81 @@
...
+@Index({ name: 'fk_email_messages_log_sender_emails1_idx', properties: ['senderId', 'sender'] })
+export class EmailMessagesLog {
+
+  [PrimaryKeyProp]?: 'logId';
+
+  @PrimaryKey()
+  logId!: number;
+
+  @ManyToOne({ entity: () => SenderEmails, fieldNames: ['sender_id', 'sender_email_id'], updateRule: 'cascade' })
+  senderId!: SenderEmails;
```
B4nan, line 19: why is this called `senderId` and not `sender`?

tests/issues/__snapshots__/GH4892.test.ts.snap
```diff
@@ -0,0 +1,81 @@
...
+@Entity()
+export class SenderEmails {
+
+  [PrimaryKeyProp]?: ['senderId', 'emailId'];
+
+  @ManyToOne({ entity: () => Sender, fieldName: 'sender_id', updateRule: 'cascade', deleteRule: 'cascade', primary: true })
+  senderId!: Sender;
+
+  @ManyToOne({ entity: () => Emails, fieldName: 'email_id', updateRule: 'cascade', primary: true, index: 'fk_sender_emails_emails1_idx' })
+  emailId!: Emails;
```
B4nan, line 76: same here, this should adhere to naming strategy instead of using explciit `fieldName`

B4nan (changes requested):
overall good job, left some comments, mostly about code style and complexity. looking at the snapshot, there are few things that i'd like to see polished, mainly the `senderId` vs `sender` property name, relation properties shouldnt have the `_id` suffix (defined in naming strategy) - this is handled propertly for the regular relations so i am sure you can fine in the code how its done and replicate it to your changes

tests/issues/__snapshots__/GH4892.test.ts.snap
```diff
@@ -0,0 +1,84 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`4892: mysql-entity-dump 1`] = `
+[
+  "import { Entity, Index, ManyToOne, PrimaryKey, PrimaryKeyProp } from '@mikro-orm/core';
+import { Emails } from './Emails';
+import { SenderEmails } from './SenderEmails';
+
+@Entity()
+@Index({ name: 'fk_email_messages_log_sender_emails1_idx', properties: ['senderId', 'sender'] })
```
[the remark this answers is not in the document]
boenrobot: Isn't this whole property idea somewhat flawed?

Like, if I have two composite FKs that are both covering one column, and I have an index on that one column in addition to the index I'd normally use for the sake of the FK.

e.g.
- columns a,b,c (all nullable, or all non-nullable; the problem exists either way)
- fk_x over a and b (with an according index)... renders as prop a
- fk_y over b and c (with an according index)... renders as prop c
- index b_idx over b

For the indexes used by the FKs, sure, the one prop for the FK is the one to be listed, but what about the index for the column itself? Prop b would not be rendered, because it is swallowed by the two FKs. So this leaves b_idx to either under specify props, i.e. nothing... or over specify props - specify props a and c, because b is part of them both.

Would the correct behavior of the generator here be to include all FKs with this column if there is no field for the column?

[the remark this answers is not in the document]
B4nan: I still don't understand why are you talking about nullability (nor why the code does things conditionally based on it), or about indexes. This is only about entity properties and database columns, I don't see how indexes and nullability have any meaning in this problem.

> Like, if I have two composite FKs that are both covering one column, and I have an index on that one column in addition to the index I'd normally use for the sake of the FK.

One thing to say out loud, this use case was never really supported, IMO this is a very specific setup most people simply don't need. Also worth saying this is a code-first framework, the entity generator is not a schema-first approach, it's a way to bootstrap projects and it is supposed to be used only once (and the output is subject to manual changes).

> Would the correct behavior of the generator here be to include all FK props with this column if there is no field for the column?

No idea, provide a code example and we can talk, this is too abstract for me to say anything :]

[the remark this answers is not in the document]
boenrobot: > I still don't understand why are you talking about nullability (nor why the code does things conditionally based on it),

At least in MySQL and MariaDB (idk about Postgresql, but I imagine there too), if you have a composite FK, where some or all of the columns are nullable, the FK is not enforced at all if one or more of the columns are set to NULL. As soon as all columns are not null, the entire FK is enforced.

Therefore, it is ok for the ORM to treat a FK prop's value as undefined if one of the FKs' columns is NULL. Any non-nullable columns however must be set somehow for the row to still be valid.

Hm... now that I'm spelling this out, I'm thinking

```typescript
      if (specificColumnNames.length === currentFk.columnNames.length) {
        // All columns involved with this FK are only covered by this one FK.
        // It is safe to attach this FK to the first nullable column,
        // or if the FK has all non-nullable columns, attach it to the first column.
        const columnName = nullableColumnsInFk.at(0) ?? currentFk.columnNames[0];
        fksOnColumnProps.set(columnName, currentFk);
        continue;
      }
```

may not actually be safe in this form... If you have multiple nullable columns in a FK, even if this is the only FK over those columns, it still means you can set only some of those columns to something other than NULL, and keep the others NULL. That's a valid DB row, without the FK being enforced, and due to the above fragment, that mixed state is not representable. I'll have to adjust this too I guess. In the case of multiple nullable columns, I guess the FK prop would have to be named after the referenced table if possible, and all columns that are part of the FK would need to be rendered as props without the FK attached.

So the above index problem is really only a problem with all non-nullable columns...

- columns a,b,c (~~all nullable, or~~ all non-nullable; ~~the problem exists either way~~)
- fk_x over a and b (with an according index)... renders as prop a
- fk_y over b and c (with an according index)... renders as prop c
- index b_idx over b

> One thing to say out loud, this use case was never really supported

I understand... that is what I'm trying to help with. Help to support this case :)

> IMO this is a very specific setup most people simply don't need. 

In my experience, this is a chicken-and-the-egg problem. People do need this... BUT they never do it in the DB layer because their ORM of choice doesn't support it, and their ORM of choice doesn't support it because their users don't use it, because their DB designs are already made with that or competing ORM's limited capabilities in mind. The same kind of constraints that I'm trying to enforce on a DB level are more often enforced on the application level due to these limitations, but they are frequently needed, especially as the requirements of the application crystalize.

> Also worth saying this is a code-first framework, the entity generator is not a schema-first approach, it's a way to bootstrap projects and it is supposed to be used only once (and the output is subject to manual changes).

Still... if it's a valid SQL schema, the generator should be able to produce usable code out of it... even if that code is later manually refined or requires special care to be used correctly.

I will admit I do intend to have a schema-first approach, where I'll regenerate the entities with every migration (or every major MikroORM upgrade), but I don't expect the generator to cater to this. I expect that I'll always have to post-process the output from the entity generator... one ts-morph, eslint and prettier pass at minimum, in this order. And it's OK for that to be out of scope for the generator. I expect my app to potentially fail to compile with every entity regeneration, even if I didn't change the schema, but especially if I changed the schema... and that too is OK.

> No idea, provide a code example and we can talk, this is too abstract for me to say anything :]

The "overlap_fk_example" in the new tests in one such example. "nullable_fk_example" is another. I can come up with a few more similar examples.

I guess the correct answer for the index depends on what MikroORM uses the index meta data for... I can't seem to find any uses beyond just "keeping a record of it" and validating if said records are correct upon meta data discovery... If the intent is to f.e. in the future add "force index" when all involved props are requested, then I guess it would be ok to feature all properties in which "b" is part of. If the intent is for MikroORM to be able to generate the SQL that targets the correct columns during a migration generation, changing "properties" to be "columns" may be more appropriate... Or maybe add both in the "index" declaration, and use either "properties" or "columns" based on context? As far as the generator is concerned, it would generate both the list of columns, and the list of properties in which any one of those columns is involved.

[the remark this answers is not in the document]
B4nan: > At least in MySQL and MariaDB (idk about Postgresql, but I imagine there too), if you have a composite FK, where some or all of the columns are nullable, the FK is not enforced at all if one or more of the columns are set to NULL. As soon as all columns are not null, the entire FK is enforced.

What I don't understand is how is all of that relevant to entity/schema generators. The FK is in the schema and as such should be generated into a relation property in the entity generator, its not relevant that some of the targeting columns are nullable or not. What you say (I think) is that it has a runtime effect, what I say is that I don't see how it's connected to what the entity generator generates.

> I understand... that is what I'm trying to help with. Help to support this case :)

Yeah sure, and I am more than happy to accept PRs for that, thanks!

> Still... if it's a valid SQL schema, the generator should be able to produce usable code out of it... even if that code is later manually refined or requires special care to be used correctly.

Sure, again, I am fine with any improvements if it was previously failing completely with a type error or similar.

> I guess the correct answer for the index depends on what MikroORM uses the index meta data for...

Only for the schema generator, there is no runtime effect on the ORM level. Unique constraints do have some effect (namely for `em.upsert()`, I don't recall any other place), but not a regular index.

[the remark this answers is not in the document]
boenrobot: > What I don't understand is how is all of that relevant to entity/schema generators. The FK is in the schema and as such should be generated into a relation property in the entity generator, its not relevant that some of the targeting columns are nullable or not. What you say (I think) is that it has a runtime effect, what I say is that I don't see how it's connected to what the entity generator generates.

For each table, the generator needs to generate a class that can represent any possible state of any row in that table.

It would be easy to just generate props for every column named after the column name, then also generate separate props for every FK, named after the FK constraint... But this is not what the generator is currently doing, so it won't be backwards compatible... It's also not exactly ergonomic, so trying to be smarter about the names of props and combining FK props into column props where it's safe to do so is indeed a welcome addition to DX.

> One more thing to mention, when it comes to complex indexes, the indexes should be generated as index expressions, no need to deal with the column<->property mapping at all for those - recent connected issue about that is https://github.com/mikro-orm/mikro-orm/issues/4911.

OK... that is a valid solution... I can detect if all columns in the index are covered by a single prop each (rather than a prop with an attached composite FK), and if it's not safe to just list all props, fallback to an expression. Sounds perfectly reasonable to me. If things change in the future, the generator could potentially be made smarter to further support those index features.

> And maybe one last note - I am fine merging this as long as it does not break the existing tests (which it does not right now), it's overall an improvement even if it produces some things wrongly (better than type errors).

Let me fix some of your other nit picks first (I kind of already did fix most, just haven't pushed yet...). I'll see if I can squeeze in this index fix too.

tests/features/entity-generator/__snapshots__/EntityGenerator.test.ts.snap
```diff
@@ -1294,6 +1298,12 @@ export class User2 {
   @Property({ nullable: true })
   foo?: number;
 
+  @Property({ length: 100, nullable: true })
+  favouriteCarName?: string;
+
+  @Property({ nullable: true })
+  favouriteCarYear?: number;
```
B4nan, line 1305: FYI having the relation property side by side with scalar columns for the same is (or at least historically was) not really supported.

over time, i made some changes to make this possible, i believe the values should be hydrated correctly, but its not tested at all. and i do expect some issues with schema generator. with this said, i wanted to support this use case in v6 anyway, but if you'd like this to be backported to v5, we should probably not generate those. the main use case for entity generator is bootstraping entities in existing project, but after that the normal user would start using migrations or schema generator instead of going schema first and regenerating entities - and that use case need to keep working as it is now.

i will test this a bit to see where we actually are, maybe things are already working fine'ish

boenrobot: > but if you'd like this to be backported to v5, we should probably not generate those

Honestly, I don't personally care to backport this. I am only now beginning to use MikroORM in earnest for a real project. I've previously only toyed with it on the side... So I don't need any BC.

B4nan: i gave this another thought, and even if it will work fine, we should not generate those. this should be **opt-in**, and **consistent**, now it's generated just for some entities (apparently those with composite FK as PK).

tests/issues/__snapshots__/GH4892.test.ts.snap
B4nan: this is still not addressed

tests/features/entity-generator/__snapshots__/EntityGenerator.test.ts.snap
boenrobot: > now it's generated just for some entities (apparently those with composite FK as PK).

Column properties in addition to the FK are generated for any composite FK with more than one nullable column, regardless of whether the FK is at the PK or not, and regardless of whether the referenced column is in PK or not.

Non-nullable column properties are also generated if the FK only has one nullable column.

This is consistent IMO... Ultimately, it allows CRUD of rows where some columns are filled, but not others, at which point the row does not really have a reference, but still exists in the DB.

> this should be opt-in

Fair enough. I understand not everyone needs to deal with rows containing partially filled references. I'll add that.

tests/issues/__snapshots__/GH4892.test.ts.snap
boenrobot: > And with regards to the `senderId` vs `sender` thing... I think my test cases are hitting cases where the code doesn't treat that part as removable... Specifically, I have `email_messages_log.sender_id` reference `email_senders.sender_id`... My naming convention - to just use the same column name everywhere, unless the same table is referenced multiple times from the same table, in which case, use a meaningful prefix - is not covered by the regex. The regex would have worked if I had `email_messages_log.email_sender_id` reference `email_senders.sender_id`, as it would then replace the `_sender_id` in `email_sender_id` with an empty string, resulting in the prop being called email.

If this is really bothering you too much, I can edit the example schema to use "id" in the origin table, and "sender_id" in the table that references the origin table, thus triggering the removal.

But the naming algorithm works as designed for its earlier use cases. There's no way for the generator to figure out that "_id" is the only removable part in the above combination.

tests/features/entity-generator/__snapshots__/EntityGenerator.test.ts.snap
B4nan: Well, to me this is not consistent, I'd like to generate all of them or none of them, now it's only some of them. I get that the rule you defined here probably makes sense, but it is not consistent - the flag should control whether you want to generate scalar props for FKs in general, nothing related to composite FKs, nor their nullability.

tests/issues/__snapshots__/GH4892.test.ts.snap
B4nan: > But the naming algorithm works as designed for its earlier use cases. There's no way for the generator to figure out that "_id" is the only removable part in the above combination.

I don't follow, it's a relation, the naming strategy is where we append `_id` suffix to relations. It worked fine till now (and it still works for all the other tests), so please don't tell me things are not designed for it, this is the very same case as for other entities, it is not about properties being removable or not - as mentioned in the other thread, I'd like to have this configurable and consistent, the flag I'd like to see would force generate the scalars here is well, and you are defining the relations with names of what should be scalars. 

> If this is really bothering you too much, I can edit the example schema to use "id" in the origin table, and "sender_id" in the table that references the origin table, thus triggering the removal.

I don't want you to adjust the tests so they pass, I want to address this problem, as it wasn't there before.

B4nan:
Just want to say this again and explicitly, so you don't feel discouraged anyhow by my comments - thank you for the PR, and for all the thoughts about the complex use cases. I am a bit cautious here, as I don't want to merge something which will in turn result in other bug reports (as it happened so many times throughout the years).

B4nan:
> codecov/project — 99.11% (-0.12%) compared to d5d54d6

This is also something to deal with before merging, 0,12% drop is pretty huge (however small it may sound like).

packages/core/src/utils/Configuration.ts
```diff
@@ -135,7 +136,7 @@ export class Configuration<D extends IDatabaseDriver = IDatabaseDriver> {
     },
     preferReadReplicas: true,
     dynamicImportProvider: /* istanbul ignore next */ (id: string) => import(id),
-  };
+  } satisfies MikroORMOptions;
```
B4nan, line 139: is this somehow better or why this change? feels unrelated

(i know what `satisfies` is for, and i dont think this is the place where we would benefit from it anyhow)

packages/entity-generator/src/EntityGenerator.ts
```diff
@@ -47,6 +48,9 @@ export class EntityGenerator {
     const baseDir = Utils.normalizePath(options.baseDir ?? (this.config.get('baseDir') + '/generated-entities'));
     const schema = await DatabaseSchema.create(this.connection, this.platform, this.config);
 
+    const scalarFieldsForCompositeRelations = this.config.get('entityGenerator').scalarFieldsForCompositeRelations
+      ?? Configuration.DEFAULTS.entityGenerator.scalarFieldsForCompositeRelations;
```
B4nan, line 52: the value from `this.config.get` will already respect the defaults, you dont need to deal with that anywhere

packages/knex/src/schema/DatabaseTable.ts
```diff
@@ -175,45 +191,276 @@ export class DatabaseTable {
     return this.platform.getIndexName(this.name, columnNames, type);
   }
 
-  getEntityDeclaration(namingStrategy: NamingStrategy, schemaHelper: SchemaHelper): EntityMetadata {
+  getEntityDeclaration(
+    namingStrategy: NamingStrategy,
+    schemaHelper: SchemaHelper,
+    scalarFieldsForCompositeRelations: NonNullable<MikroORMOptions['entityGenerator']['scalarFieldsForCompositeRelations']> = Configuration.DEFAULTS.entityGenerator.scalarFieldsForCompositeRelations,
```
B4nan, line 197: this is really weird line. either this comes from the config class (which handles the defauls), or it comes explicitly from user (who provides the boolean directly). you should not use the `DEFAULTS` anywhere explicitly, and this method is rather internal, no need to expect users will be calling themselves, no need to make this configurable here or have some default parameters for it.

B4nan (changes requested):
lets not rebase anymore, it makes impossible to tell what you are changing

(i will be squash merging the PR, if you want to have multiple commits/changelog records, you need to create multiple PRs)

packages/core/src/utils/Configuration.ts
boenrobot: I had a type error which was easily solvable with this... And it seems like a good thing in this case overall, which didn't affect anything, so I just went ahead and did it.

packages/knex/src/schema/DatabaseTable.ts
B4nan: you can adjust the configuration object, which should be the single source of truth, the `DEFAULTS` is used only when creating this config instance to fill the missing values, its not used anywhere else in the codebase and I'd like to keep it that way.

```ts
const entityGeneratorOptions = orm.config.get('entityGenerator');
entityGeneratorOptions.whatEver = 123;
```

docs/docs/entity-generator.md
```diff
@@ -53,6 +53,10 @@ By default, the `EntityGenerator` generates only owning sides of relations (e.g.
 - `esmImport` to use esm style import for imported entities e.x. when `esmImport=true`, generated entities include `import Author from './Author.js'`
 - `skipTables` to ignore some database tables (accepts array of table names)
 - `skipColumns` to ignore some database tables columns (accepts an object, keys are table names with schema prefix if available, values are arrays of column names)
+- `scalarFieldsForCompositeRelations` to control how scalar column fields are generated for composite foreign key relations. Possible values:
```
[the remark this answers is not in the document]
B4nan: hmm looks like one of my comments got lost - just wanted to say i would much rather have a simple bool to control this and remove all the magic that considers whether a FK is a composite one, and whether some parts of it are nullable or not, or the whole indexes thing. that is all irrelevant for runtime, i would rather let users to switch this on/off if they want/need the scalar properties instead of having all that complex code to maintain.

packages/core/src/utils/Configuration.ts
boenrobot: > The previous solution would report wrongly named keys as well, but that's why I am asking

Satisfies also does this... I checked... I tried to type in a default config that does not actually satisfy the config type, including invalid types of existing configs or adding non-existent properties or not adding required properties... fails to build in any of those cases.

The difference is that with satisfies, you can reference the defaults with their true values, which in turn is what enabled me to avoid the "!" earlier.

At this point, I don't need it anymore, since you asked that I rather modify the test case that uses the internal method instead. But it's worth keeping. In general, sure, I'm doing my best to minimize unrelated changes.

tests/features/entity-generator/EntityGenerator.test.ts
```diff
@@ -34,6 +34,32 @@ describe('EntityGenerator', () => {
...
+    await orm.schema.dropDatabase();
+    await orm.close(true);
+  });
+
+  test('generate entities with columns for some composite foreign key properties [mysql]', async () => {
+    const orm = await initORMMySql('mysql', { entityGenerator: { scalarFieldsForCompositeRelations: 'smart' } }, true);
+    const dump = await orm.entityGenerator.generate({ save: true, baseDir: './temp/entities' });
+
+    // The test schema only has one instance where column props for composite FK are generated,
+    // and it's the case covered by smart field creation. So the output should be the same as in the previous snapshot.
```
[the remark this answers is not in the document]
boenrobot: fair enough... moved into a separate test file, and with a separate snapshot for each test case.

tests/features/entity-generator/EntityGenerator.test.ts
```diff
@@ -34,6 +34,32 @@ describe('EntityGenerator', () => {
     await orm.close(true);
   });
 
+  test('generate entities with columns for all composite foreign key properties [mysql]', async () => {
```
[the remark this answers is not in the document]
boenrobot: Done

B4nan:
Btw maybe you could first do some tiny PR (e.g. some small docs change) we would merge first so you don't have to wait for me to trigger the CI? It will work automatically once you are a contributor.

I will be happy to add you to the contributors list too if you plan to PR more things, so you can just create a branch instead of forking.

docs/docs/entity-generator.md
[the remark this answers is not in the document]
B4nan: fair enough, as long as your changes will be covered with tests, i dont mind. i still need to review (or improve) how it works when those fields are generated, as right now i think its needed to flag one of those with `persist: false` so its not part of the schema, but thats exactly the bit i'd like to fix.

also lets keep the string options, that's fine, better than mixing strings with boolean

packages/core/src/metadata/EntitySchema.ts
```diff
@@ -225,11 +225,11 @@ export class EntitySchema<Entity = any, Base = never> {
     this.addProperty(name, type, prop);
   }
 
-  addIndex<T>(options: Required<Omit<IndexOptions<T>, 'name' | 'type' | 'options' | 'expression'>> & { name?: string; expression?: string; options?: Dictionary }): void {
+  addIndex<T>(options: IndexOptions<Dictionary<EntityKey>>): void {
     this._meta.indexes.push(options as any);
   }
 
-  addUnique<T>(options: Required<Omit<UniqueOptions<T>, 'name' | 'options' | 'expression'>> & { name?: string; options?: Dictionary }): void {
+  addUnique<T>(options: UniqueOptions<Dictionary<EntityKey>> & Pick<IndexOptions<Dictionary<EntityKey>>, 'expression'>): void {
```
B4nan, line 232: the `T` is currently unused, if it is on purpose, we should remove it completely

docs/docs/entity-generator.md
```diff
@@ -53,6 +53,10 @@ By default, the `EntityGenerator` generates only owning sides of relations (e.g.
 - `esmImport` to use esm style import for imported entities e.x. when `esmImport=true`, generated entities include `import Author from './Author.js'`
 - `skipTables` to ignore some database tables (accepts array of table names)
 - `skipColumns` to ignore some database tables columns (accepts an object, keys are table names with schema prefix if available, values are arrays of column names)
+- `scalarFieldsForRelations` to control how scalar column fields are generated for foreign key relations. Possible values:
```
B4nan, line 56: i dont like this name very much, there is not a single place in the codebase where we would refer to something as "scalar fields". 
- entity has properties
- db tables have columns
- `fields` option is used for partial loading

`scalarPropertiesForRelations` sounds more correct to me, not perfect, but it feels better

B4nan (commented):
will review this later this evening, left just two quick comments now. 

also, what issues is this addressing, is it just #4898 or is #4911 also covered?

packages/core/src/metadata/EntitySchema.ts
boenrobot: For completeness' sake, I used it as the arg in EntityKey... In practice, right now, no point in the code uses it, but in theory, you could force it down for a particular entity later on. I imagine this may be useful for users who are tinkering with the internals, even though they probably shouldn't.

boenrobot: Nevermind... the EntitySchema has an Entity generic, and yet setting it introduces a type error... so... rather than trying to come up with a complex type jujitsu, I'll just remove the T.

B4nan: I can try to help with that. I don't really like to see `Dictionary<EntityKey>` either, it only adds CPU load while it will (probably) work the same as just `Dictionary`. 

The change you did feels like it makes things less type safe but on a second look, it feels like it was never type safe (given the `T` type arg instead being used instead of the one from the `EntitySchema`) :D

boenrobot: Ah... `Dictionary<EntityKey>`... That gave me a good hint... I just replaced that with `Entity`, and seems to be fine now. There's still the "as any" cast inside, which is kind of unfortunate, but that requires bigger type changes that are not needed now.

docs/docs/entity-generator.md
```diff
@@ -53,6 +53,10 @@ By default, the `EntityGenerator` generates only owning sides of relations (e.g.
 - `esmImport` to use esm style import for imported entities e.x. when `esmImport=true`, generated entities include `import Author from './Author.js'`
 - `skipTables` to ignore some database tables (accepts array of table names)
 - `skipColumns` to ignore some database tables columns (accepts an object, keys are table names with schema prefix if available, values are arrays of column names)
+- `scalarPropertiesForRelations` to control how scalar column properties are generated for foreign key relations. Possible values:
```
B4nan, line 56: so i took a look at how this works (having the scalar properties side by side with relation ones), and we need to add `persist: false` to the scalar props, otherwise they will trigger validation if you dont provide value for both when creating an entity (there might be more problems, this is the first one i faced). with `persist: false`, they wont be considered for change tracking, validation, or schema diffing. they will be still hydrated as usual.

boenrobot: Fixed... I think... I haven't added tests to see if the generated entities work ok, but I changed it to add `persist: false` if a scalar column is rendered without a FK attached and yet a FK featuring it exists.

tests/features/entity-generator/__snapshots__/ScalarPropsForFks.test.ts.snap
```diff
@@ -0,0 +1,1120 @@
...
+  "import { Entity, OneToOne, PrimaryKey, PrimaryKeyProp, Property } from '@mikro-orm/core';
+import { Author2 } from './Author2';
+
+@Entity()
+export class Address2 {
+
+  [PrimaryKeyProp]?: ['authorId', 'author'];
+
+  @PrimaryKey({ persist: false })
+  authorId!: number;
```
B4nan, line 14: i would prefer to move the declarations after the actual relation properties (ideally separated, now they are all printed as a group of non persistent props as visible on the Author2 entity)

tests/features/entity-generator/__snapshots__/ScalarPropsForFks.test.ts.snap
```diff
@@ -0,0 +1,1120 @@
...
+  @Property({ columnType: 'date', nullable: true })
+  born?: string;
+
+  @Index({ name: 'born_time_idx' })
+  @Property({ columnType: 'time', nullable: true })
+  bornTime?: string;
+
+  @Index({ name: 'author2_favourite_book_uuid_pk_index' })
+  @Property({ length: 36, nullable: true, persist: false })
+  favouriteBookUuidPk?: string;
```
B4nan, line 75: for clarity, id like to see this order of properties:
1. favouriteAuthor
2. favouriteAuthorId
3. favouriteBook
4. favouriteBookUuidPk

B4nan (changes requested):
left few final comments

packages/knex/src/schema/DatabaseTable.ts
```diff
@@ -13,7 +13,7 @@ import {
   type EntityProperty,
   type MikroORMOptions,
   type NamingStrategy,
-  type UniqueOptions,
+  type UniqueOptions, PropertyOptions,
```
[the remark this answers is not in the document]
boenrobot: Oops. Sorry, that shouldn't even be there... I thought I'd make the property declaration returned be type safe, but it turns out this internal output doesn't quite match up to PropertyOptions, so I abandoned that idea... but forgot to also remove the import.

B4nan (approved):
all right, good job, lets move this forward!

B4nan:
The tests are pretty crazy (the one for #4898), would be good to simplify the schema or think of a way to make them faster. Ideally 2-3 tables, with as few properties as possible, that way it could be reasonably fast on its own. Or split them into test per schema, so they can run in parallel (we would ideally do both).

(I've spend a lot of time making the tests "fast" again - it used to be 5x slower - and this adds ~20% to the run time :])

B4nan:
A few tables are ok, let's say up to 5, but that test is using 10 tables (in 6 different schemas). And agreed, also wanted to move the test to the right folder :]

Btw be sure to pull latest changes, I found few small bugs in your changes and fixed them as part of 64a39f82c7d391d28e28c639512a810c516f08a9.

PR #4931, opened 2023-11-16
test: all tests can now run on Windows

boenrobot:
In order to do so without breaking Unix compatibility, and still respect Jest's mock rules (namely, no imports for mocked module names), additional mocks are added where the path's slashes are replaced.

Utils.normalizePath() or path.resolve() is used in other places, based on the expected value.

Also fixed building on Windows, by making copy.mjs be responsible for copying esm.cmd.

### DIFF
--- a/scripts/copy.mjs
+++ b/scripts/copy.mjs
@@ -143,6 +143,9 @@ if (options['pin-versions']) {
 copy('README.md', root, target);
 copy('LICENSE',  root, target);
 copy('package.json', process.cwd(), target);
+if (resolve(process.cwd()) === resolve(root, 'packages/cli')) {
+  copy('esm.cmd', resolve(process.cwd(), 'src'), target);
+}
 rewrite(resolve(target, 'package.json'), pkg => {
   return pkg.replace(/dist\//g, '').replace(/src\/(.*)\.ts/g, '$1.js');
 });
--- a/scripts/copy.mjs
+++ b/scripts/copy.mjs
@@ -143,9 +143,11 @@ if (options['pin-versions']) {
 copy('README.md', root, target);
 copy('LICENSE',  root, target);
 copy('package.json', process.cwd(), target);
+
 if (resolve(process.cwd()) === resolve(root, 'packages/cli')) {
   copy('esm.cmd', resolve(process.cwd(), 'src'), target);
 }
+
 rewrite(resolve(target, 'package.json'), pkg => {
   return pkg.replace(/dist\//g, '').replace(/src\/(.*)\.ts/g, '$1.js');
 });
--- a/tests/features/cli/CLIHelper.test.ts
+++ b/tests/features/cli/CLIHelper.test.ts
@@ -8,28 +8,37 @@ import { resolve } from 'node:path';
 import type * as pathModule from 'node:path';
 import type * as mikroOrmCoreModule from '@mikro-orm/core';
 
-jest.mock(jest.requireActual<typeof mikroOrmCoreModule>('@mikro-orm/core').Utils.normalizePath(process.cwd()) + '/mikro-orm.config.js', () => ({ driver: MongoDriver, dbName: 'foo_bar', entities: ['tests/foo'] }), { virtual: true });
-jest.mock(jest.requireActual<typeof mikroOrmCoreModule>('@mikro-orm/core').Utils.normalizePath(process.cwd()) + '/mikro-orm.config.ts', () => ({ driver: MongoDriver, dbName: 'foo_bar', entities: ['tests/foo'] }), { virtual: true });
-jest.mock(jest.requireActual<typeof mikroOrmCoreModule>('@mikro-orm/core').Utils.normalizePath(process.cwd()) + '/mikro-orm-async.config.js', () => (Promise.resolve({ driver: MongoDriver, dbName: 'foo_bar', entities: ['tests/foo'] })), { virtual: true });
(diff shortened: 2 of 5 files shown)


scripts/copy.mjs
```diff
@@ -143,6 +143,9 @@ if (options['pin-versions']) {
 copy('README.md', root, target);
 copy('LICENSE',  root, target);
 copy('package.json', process.cwd(), target);
+if (resolve(process.cwd()) === resolve(root, 'packages/cli')) {
+  copy('esm.cmd', resolve(process.cwd(), 'src'), target);
+}
```
B4nan, line 148:
```suggestion

if (resolve(process.cwd()) === resolve(root, 'packages/cli')) {
  copy('esm.cmd', resolve(process.cwd(), 'src'), target);
}

```

tests/features/cli/CLIHelper.test.ts
```diff
@@ -4,28 +4,40 @@ import { CLIConfigurator, CLIHelper } from '@mikro-orm/cli';
...
 jest.mock(process.cwd() + '/mikro-orm.config.ts', () => ({ driver: MongoDriver, dbName: 'foo_bar', entities: ['tests/foo'] }), { virtual: true });
+jest.mock(process.cwd().replaceAll('\\', '/') + '/mikro-orm.config.ts', () => ({ driver: MongoDriver, dbName: 'foo_bar', entities: ['tests/foo'] }), { virtual: true });
 jest.mock(process.cwd() + '/mikro-orm-async.config.js', () => (Promise.resolve({ driver: MongoDriver, dbName: 'foo_bar', entities: ['tests/foo'] })), { virtual: true });
+jest.mock(process.cwd().replaceAll('\\', '/') + '/mikro-orm-async.config.js', () => (Promise.resolve({ driver: MongoDriver, dbName: 'foo_bar', entities: ['tests/foo'] })), { virtual: true });
 jest.mock(process.cwd() + '/mikro-orm-async-catch.config.js', () => (Promise.reject('FooError')), { virtual: true });
+jest.mock(process.cwd().replaceAll('\\', '/') + '/mikro-orm-async-catch.config.js', () => (Promise.reject('FooError')), { virtual: true });
+
 const pkg = { 'mikro-orm': {} } as any;
 jest.mock(process.cwd() + '/package.json', () => pkg, { virtual: true });
+jest.mock(process.cwd() + '\\package.json', () => pkg, { virtual: true });
```
B4nan, line 20: those double mocks are looking pretty weird, why cant we use the same approach here with `path.resolve`?

boenrobot: Because Jest complains about resolve being unaccesible. Any imports are unaccessible in fact. "process" is a global that is not imported, which is why calls to it work.

B4nan: I've never seen any project that would be doing this, do you know any? Nor a blog post or a mention in jest docs on this topic, so it feels rather wrong/workaround'ish than a proper solution.

(with that said, I hate `jest.mock` in general, would be great to come up with something else)

boenrobot: `jeck.mock` is matching against the literal argument of `require()`, with no processing on it.

I suppose a "proper" workaround would be to ensure all places in the code always use paths that are both normalized and relative before importing, so that you don't need `process.cwd()` at all, and the rest of the path is sure to have "/".

boenrobot: ok, I found an alternative approach that works cross platform... by using jest.requireActual directly into the module name.

Using it outside still produces an error due to the constant itself not being accessible, but using it inline works.

tests/features/cli/CLIHelper.test.ts
```diff
@@ -4,28 +4,32 @@ import { CLIConfigurator, CLIHelper } from '@mikro-orm/cli';
 import { SchemaCommandFactory } from '../../../packages/cli/src/commands/SchemaCommandFactory';
 import { MongoDriver } from '@mikro-orm/mongodb';
 import { SqliteDriver } from '@mikro-orm/sqlite';
+import { resolve } from 'node:path';
+import type * as pathModule from 'node:path';
+import type * as mikroOrmCoreModule from '@mikro-orm/core';
+
+jest.mock(jest.requireActual<typeof mikroOrmCoreModule>('@mikro-orm/core').Utils.normalizePath(process.cwd()) + '/mikro-orm.config.js', () => ({ driver: MongoDriver, dbName: 'foo_bar', entities: ['tests/foo'] }), { virtual: true });
```
B4nan, line 11: i guess we can extract the `jest.requireActual<typeof mikroOrmCoreModule>('@mikro-orm/core')` and similar to variables to reduce the duplication?

boenrobot: I wish... but no... even local const and let declarations don't work (as in, they give a reference error), and var... passes, but is treated as undefined, which is also an error.

B4nan: maybe we are trying to solve this on a wrong level, you mentioned the problem is with the input paths, if its about the ORM config, that is being imported here:

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/utils/ConfigurationLoader.ts#L28

but the path is already normalized...

boenrobot: I've added a monkey patch to the global module as a separate commit (so that I can easily rollback if you prefer the duplicate version). ESLint prevents declarations, so I had to change that rule also. Or I guess I could ignore the specific line if you prefer.

B4nan (approved):
thanks, lets not waste more time on this, its ugly but it was ugly before as well :]

PR #4935, opened 2023-11-17
feat(entity-generator): added ability to output type option in decorator

boenrobot:
I tried to run the tests with `@swc/jest` (yes, I did saw #3200, but I thought I take a fresh crack at it, since that branch is over a year old...), and an interesting error that popped up was the inability to discover the entity type.

Featuring the type is one way to avoid this... another way is to not use SWC or to use `EntitySchema` (which is the solution I intend to use for my projects personally...).

So... for users who intend to use SWC, and would like to use decorators, I think it would be very beneficial to be able to generate the "type" option. For everyone else, it would just provide a minor boost in startup performance of a project using generated entities with decorators.

Being a very limited option in scope, implementation was easy. Honestly, whether this should be an option and whether it should default to `true` or `false` is debatable... but I thought I just set it as an option defaulting to `false` for the sake of BC to avoid this debate... I mean, I think it's OK for it to not be an option and just always default to `true`... but that requires modification in existing tests if nothing else.

Side note: Because of the way SWC works, I fear the only way to possibly switch to it would be to support __BOTH__ `@swc/jest` and `ts-jest` in all tests but those that require reflection... and in CI, run those that require reflection separately with `ts-jest` while most others would run in `@swc/jest`.

### DIFF
--- a/tests/features/entity-generator/TypesForScalarDecorators.test.ts
+++ b/tests/features/entity-generator/TypesForScalarDecorators.test.ts
@@ -1,0 +1,38 @@
+import { EntityGenerator } from '@mikro-orm/entity-generator';
+import { MikroORM } from '@mikro-orm/mysql';
+
+let orm: MikroORM;
+beforeAll(async () => {
+  orm = await MikroORM.init({
+    dbName: 'types_for_scalar_decorators',
+    port: 3308,
+    discovery: { warnWhenNoEntities: false },
+    extensions: [EntityGenerator],
+    multipleStatements: true,
+  });
+  await orm.schema.ensureDatabase();
+});
+
+afterAll(async () => {
+  await orm.schema.dropDatabase();
+  await orm.close(true);
+});
+
+describe('TypesForScalarDecorators', () => {
+  test('generate entities from schema [mysql]', async () => {
+    await orm.schema.execute(`
+CREATE TABLE IF NOT EXISTS \`users\`
+(
+    \`user_id\` INT UNSIGNED NOT NULL AUTO_INCREMENT,
+    \`username\` VARCHAR(255) NOT NULL,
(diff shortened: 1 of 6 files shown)


B4nan:
> I tried to run the tests with @swc/jest (yes, I did saw https://github.com/mikro-orm/mikro-orm/pull/3200, but I thought I take a fresh crack at it, since that branch is over a year old...), and an interesting error that popped up was the inability to discover the entity type.

I still don't get it, I actually rebased that branch a month ago, and it's still the very same story, it does not bring any speed improvements over ts-jest with isolated modules (so no type checking). So not interested in switching away from what works fine without any added benefits.

tests/features/entity-generator/TypesForScalarDecorators.test.ts
```diff
@@ -0,0 +1,37 @@
+import { pathExists, remove } from 'fs-extra';
+import { initORMMySql, initORMPostgreSql, initORMSqlite } from '../../bootstrap';
+
+describe('EntityGenerator', () => {
+
+  test('generate entities from schema [mysql]', async () => {
+    const orm = await initORMMySql('mysql', {}, true);
```
B4nan, line 7: lets not use those fat ORM instances, its enough to define one entity with few different property types for this (better say one schema as we test the entity generator)

B4nan (changes requested):
looking good, lets just simplify the tests so they can be faster, i also dont think we need to test multiple drivers here as the changes are not driver specific

tests/features/entity-generator/TypesForScalarDecorators.test.ts
boenrobot: ok... simplified down to one table with a few columns.

B4nan (approved):

PR #4941, opened 2023-11-20
docs: property-validation.md

roderik:
The code provided was not copy pastable, T was not defined

### DIFF
--- a/docs/docs/property-validation.md
+++ b/docs/docs/property-validation.md
@@ -64,7 +64,7 @@ When you want to define some optional properties in your own base entity class, 
 ```ts
 @Entity()
 // highlight-next-line
-class MyBaseEntity<T extends object, Optional extends keyof T = never> {
+class MyBaseEntity<Entity extends object, Optional extends keyof Entity = never> {
 
   // highlight-next-line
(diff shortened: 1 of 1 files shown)


docs/docs/property-validation.md
```diff
@@ -64,7 +64,7 @@ When you want to define some optional properties in your own base entity class,
 ```ts
 @Entity()
 // highlight-next-line
-class MyBaseEntity<Optional extends keyof T = never> {
+class MyBaseEntity<T extends object, Optional extends keyof T = never> {
```
B4nan, line 67: not sure what should we do here, i've removed the argument as it is no longer needed for the ORM `BaseEntity`, here its technically not needed either, only for the intellisense of the optional keys

this would work too, but without intellisense

```ts
class MyBaseEntity<Optional extends string = never> {
```

i guess its better to keep what we had before. but lets use `Entity` instead of `T`

```suggestion
class MyBaseEntity<Entity extends object, Optional extends keyof Entity = never> {
```

---

too bad we cannot do this instead (it works with `<Opt extends keyof typeof this = never>` but it only checks for the base properties)

```ts
class MyBaseEntity<Optional extends keyof this = never> {
```

PR #4943, opened 2023-11-21
docs(entity-generator): install note and option reorganization

boenrobot:
Reorganized the config options into a table,
and documented the generate() options in a separate table.

Also added a note about the entity-generator needing to be installed, be aligned with the database driver, and the lack of support for MongoDB.

### DIFF
--- a/docs/docs/entity-generator.md
+++ b/docs/docs/entity-generator.md
@@ -4,6 +4,20 @@ title: Entity Generator
 
 To generate entities from existing database schema, you can use `EntityGenerator` helper.
 
+Install it with
+
+```sh
+yarn add @mikro-orm/entity-generator
+```
+
+or
+
+```sh
+npm i -s @mikro-orm/entity-generator
+```
+
+> The version needs to be aligned with the `@mikro-orm/core` package and the database driver package.
+
 You can use it via CLI:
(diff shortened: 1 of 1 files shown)


docs/docs/entity-generator.md
```diff
@@ -45,27 +59,38 @@ $ ts-node generate-entities
...
+| `save`        | `boolean`                          | `false`                                              | Whether to save the generated entities as files.                                                                                                                                                                                                                                        |
+| `baseDir`     | `string`                           | `this.config.get('baseDir') + '/generated-entities'` | Folder to save the generated entities in, if saving is enabled. Defaults to a folder called `generated-entities` inside the `baseDir` of the main config.                                                                                                                               |
+
+Example configuration and run:
 
 ```ts
+orm.config.set('entityGenerator', {
```
B4nan, line 86: i'd actually like to see another PR first, which will allow overriding the config options in the `generate` method params, its pretty confusing now that you have some options there and some not.

also i am not sure if this is the right material for table layout, i'd prefer to have separate sections (so headings instead of bullit points) with a simple example (where applicable) so its clear what it does. but will first check how it looks locally, maybe its not that bad idea

Issue #4945, opened 2023-11-21
upsertMany error within nullable and composite key reference.


z4o4z:
### Describe the bug

ORM generates incorrect `Ref` for nullable and composite key references. 

Here's an example:
```ts
class EntityA {
  @Primary()
  id!: string;
  
  @Primary()
  envID!: string;
}

class EntityB {
  @Primary()
  id!: string;
  
  @Primary()
  envID!: string;

  @Property({ type: 'text' })
  name!: string;

  @ManyToOne(() => EntityA, {
    name: 'entity_a_id',
    default: null,
    onDelete: 'set default',
    nullable: true,
    fieldNames: ['entity_a_id', 'env_id'],
  })
  entityA: Ref<EntityA> | null = null;


}
```

when I try to upsert entities like below, I get an error `ValidationError: Value for EntityB.envID is required, 'undefined' found`.

```ts
await this.em.upsertMany(EntityA, [{ id: 'entity-a-1', envID: 'env-1' }, { id: 'entity-a-2', envID: 'env-1' }]);

await this.em.upsertMany(EntityB, [{ id: 'entity-b-1', envID: 'env-1', name: 'entity-b-1', entityA: ref({ id: 'entity-a-1', envID: 'env-1' }) }, { id: 'entity-b-2', envID: 'env-1', name: 'entity-b-2', entityA: ref({ id: 'entity-a-2', envID: 'env-1' }) }]);
```

After some debugging, I've found that [finalize method in the `QueryBuilderHelper`](https://github.com/mikro-orm/mikro-orm/blob/master/packages/knex/src/query/QueryBuilderHelper.ts#L663-L673) adds `entity_a_id` as a returning field, cause it has default value, I guess that makes sense. But when the [returned value is mapped](https://github.com/mikro-orm/mikro-orm/blob/a737b207744a85b2034a5e4a2b362e4aaefea570/packages/core/src/EntityManager.ts#L998C8-L998C8) the original `entityA` ref (that has both `id` and `envID`) replaced with incorrect one (that has only id, no env id), it happens cause in the `mapReturnedValues` `entityA` is set to `entity_a_id` that is returned from the DB, but it should be a composite value (array of `entity_a_id` and `envID`). 

Not sure what's the best solution here, probably ignore nullable + composite key references during `mapReturnedValues`?

### Reproduction

see above

### What driver are you using?

@mikro-orm/postgresql

### MikroORM version

5.9.4

### Node.js version

v16.19.0

[...]

PR #4949, opened 2023-11-22
fix(core): ignore nullable composite references in the upsert (#4945)

z4o4z:
The PR implements the `cleanupUpsertRow` helper to ignore composite foreign keys. More info on this issue: https://github.com/mikro-orm/mikro-orm/issues/4945.

Btw, I've also found out that `MySQL` and `MariaDB` drivers (could be for others as well) generate incorrect create table SQLs for the entities from the test. In the test `entity_a_id` is a composite foreign key, it's nullable, and `null` is a default value. The generated SQL for `MySQL` and `MariaDB` looks like this:

```sql
create table `entity_b` (`id` varchar(255) not null, `env_id` varchar(255) not null default null, `name` text not null, `entity_a_id` varchar(255) null default null, primary key (`id`, `env_id`)) default character set utf8mb4 engine = InnoDB
```

As you can see, `default null` is used for the `env_id` field as well, even if it's a required field. Not sure if it's an issue with drivers/knex scheme generation...

### DIFF
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -591,6 +591,28 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
     return entity;
   }
 
+  private cleanupUpsertRow(meta: EntityMetadata<any>, row?: Dictionary) {
+    if (!row) {
+      return row;
+    }
+
+    const copy: Dictionary = {};
+
+    for (const [key, value] of Object.entries(row)) {
+      const prop = meta.hydrateProps.find(p => p.fieldNames[0] === key);
+
+      // ignore composite foreign keys with default value that are not null
+      if (prop && prop.reference !== ReferenceType.SCALAR && prop.fieldNames.length > 1 && prop.defaultRaw && value != null) {
+        continue;
+      }
+
+
+      copy[key] = value;
+    }
+
+    return copy;
+  }
+
   /**
    * Creates or updates the entity, based on whether it is already present in the database.
    * This method performs an `insert on conflict merge` query ensuring the database is in sync, returning a managed
@@ -699,7 +721,7 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
       convertCustomTypes: true,
     });
 
-    em.unitOfWork.getChangeSetPersister().mapReturnedValues(entity, data, ret.row, meta, true);
+    em.unitOfWork.getChangeSetPersister().mapReturnedValues(entity, data, this.cleanupUpsertRow(meta, ret.row), meta, true);
     const uniqueFields = options.onConflictFields ?? (Utils.isPlainObject(where) ? Object.keys(where) : meta!.primaryKeys) as (keyof Entity)[];
     const returning = getOnConflictReturningFields(meta, data, uniqueFields, options) as string[];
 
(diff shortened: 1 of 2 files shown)


packages/core/src/EntityManager.ts
```diff
@@ -591,6 +591,28 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
...
+    if (!row) {
+      return row;
+    }
+
+    const copy: Dictionary = {};
+
+    for (const [key, value] of Object.entries(row)) {
+      const prop = meta.props.find(p => p.fieldNames && p.fieldNames[0] === key);
+
+      // ignore composite foreign keys with default value that are not null
```
B4nan, line 604: I will need to review this myself, as I don't really understand why we should ignore anything here.

tests/features/upsert/GH4945.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  entityA: Ref<EntityA> | null = null;
+
+  [PrimaryKeyType]?: [string, string];
+
+}
+
+const options = {
+  'sqlite': { dbName: ':memory:' },
+  'better-sqlite': { dbName: ':memory:' },
+  'postgresql': { dbName: 'mikro_orm_upsert' },
```
B4nan, line 48: we need mysql here too, as that driver does not support retuning statements and therefore uses different code paths (but should have the same results)

tests/features/upsert/GH4945.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  @Property({ type: 'text' })
+  name!: string;
+
+  @ManyToOne(() => EntityA, {
+    name: 'entity_a_id',
+    default: null,
+    onDelete: 'set default',
+    nullable: true,
+    fieldNames: ['entity_a_id', 'env_id'],
```
B4nan, line 37: so sharing one column (`env_id`) for multiple properties (`envID` and `entityA`), right?

B4nan (commented):
thanks, will try to look into this soon

packages/core/src/EntityManager.ts
z4o4z: we ignore only composite foreign keys with the default value, otherwise, we override composite reference with a normal reference (that has only id in it), then `.flush` triggers updating entity second time, cause ref changed, and it's incorrect one (doesn't have envID), so mikro fails cause it detects that ref should be a composite one but it's not...

tests/features/upsert/GH4945.test.ts
z4o4z: as I mentioned in the PR comment `MySQL` and `MariaDB` drivers are broken, cause they generate an incorrect create table sql for this use case:
```sql
create table `entity_b` (`id` varchar(255) not null, `env_id` varchar(255) not null default null, `name` text not null, `entity_a_id` varchar(255) null default null, primary key (`id`, `env_id`)) default character set utf8mb4 engine = InnoDB
```

tests/features/upsert/GH4945.test.ts
z4o4z: i just wanna say to DB that the composite foreign key should be created, like this one:
```
alter table "designer"."response_variant" add constraint "response_variant_condition_id_environment_id_foreign" foreign key ("condition_id", "environment_id") references "designer"."condition" ("id", "environment_id") on update cascade on delete set null;
```

packages/core/src/EntityManager.ts
B4nan: FYI generally speaking, the ORM requires you to provide values for composite keys, having a database default is not supported (it might work, but its not tested at all, and there are places in the codebase that expect the values to be there if its a composite key).

tests/features/upsert/GH4945.test.ts
B4nan: well, then this needs to be addressed too, otherwise we dont know if your fix works for those drivers. will have a look at that too.

z4o4z: Yeah, but it's completely unrelated to main issue I've found and it should be done in another PR

B4nan:
FYI the test is passing in v6 without any changes.

B4nan:
It was fixed as part of a bigger and unrelated refactoring #4391. I'll close this, as the correct fix is to modify the returning results mapping, not skip some columns from it.

With that said, there seems to be a different bug (in both v5 and v6) related to this, the reloading fallback is not working - sometimes the returning clause won't work (namely when there is nothing to update) and there should be a select query fired to get the missing fields, and that seems to be broken with your test.

PR #4968, opened 2023-11-26
feat(core): validate duplicate field names

### LINKED ISSUE #4359: Duplicate fieldName in entity causes hard-to-understand error
**Describe the bug**
When a `fieldName` propertyis accidentally duplicated in an `entity` this cause a `Column count doesn't match value count` error during record insertion.  It took me time and luck to find the cause of the problem.

**Stack trace**
DriverException: insert into `glims_codes` (`exportId`, `glimsCode`, `hprimCode`, `label`, `sequencer`, `origin`) values (0, 'Info', 'Info', 'Informatie', 0, '0010', 'nodes') - (conn=3999, no: 1136, SQLState: 21S01) Column count doesn't match value count at row 1
sql: insert into `glims_codes` (`exportId`, `glimsCode`, `hprimCode`, `label`, `sequencer`, `origin`) values (0, 'Info', 'Info', 'Informatie', 0, '0010', 'nodes') - parameters:[]

**To Reproduce**
```typescript
import { Entity, Enum, PrimaryKey, Property } from '@mikro-orm/core'

export enum CodeOrigin {
  NODES = 'nodes',
  CODESET_CODES = 'codeset-codes',
  CODESET_NODES = 'codeset-nodes',
}

export const TEST_TITLE_LEVEL = -1

@Entity({ tableName: 'glims_codes' })
export class GlimsCode {
  @PrimaryKey({ columnType: 'int', default: 0, fieldName: 'exportId', nullable: false })
  exportId = 0

  @PrimaryKey({ columnType: 'varchar', length: 100, default: '', nullable: false, fieldName: 'glimsCode' })
  glimsCode = ''

  @PrimaryKey({ columnType: 'varchar', length: 100, nullable: true, default: null, fieldName: 'hprimCode' })
  hprimCode: string | null = null

  @Property({ columnType: 'int', nullable: false, default: TEST_TITLE_LEVEL, fieldName: 'hprimCode' /* duplicate ! */})
  titleLevel: number = TEST_TITLE_LEVEL

  @PrimaryKey({ columnType: 'varchar', length: 250, nullable: true, default: null, fieldName: 'label' })
  label: string | null = null

  @Property({ columnType: 'varchar', length: 250, nullable: true, default: null, fieldName: 'sequencer' })
  sequencer: string | null = null

  @Enum({ items: () => CodeOrigin, default: 'nodes', nullable: false, fieldName: 'origin' })
  origin: CodeOrigin = CodeOrigin.NODES
}
```

**Expected behavior**
Better flag this error during entity discovery.

**Versions**
```json
    "@mikro-orm/cli": "^5.6.15",
    "@mikro-orm/core": "^5.6.15",
    "@mikro-orm/mariadb": "^5.6.15",
    "@mikro-orm/migrations": "^5.6.15",
    "@mikro-orm/nestjs": "^5.1.8",
```
| Dependency | Version |
| - | - |
| node | v18.13.0|
| typescript |4.9 |
| mikro-orm | 5.6 |

B4nan:
When you use the same `fieldName` for two properties in one entity, error will be thrown:

```ts
@Entity()
class User {

  @PrimaryKey()
  id!: number;

  @Property({ name: 'custom_name' })
  name!: number;

  @Property({ name: 'custom_name' })
  age!: number;

}
```

This does not apply to virtual properties:

```ts
@Entity()
class User {

  @PrimaryKey()
  id!: number;

  @ManyToOne(() => User, { name: 'parent_id' })
  parent!: User;

  @Property({ name: 'parent_id',  })
  parentId!: number;

}
```

> This validation can be disabled via `discovery.checkDuplicateFieldNames` ORM config option.

Closes #4359

### DIFF
--- a/docs/docs/upgrading-v5-to-v6.md
+++ b/docs/docs/upgrading-v5-to-v6.md
@@ -456,3 +456,43 @@ const users = await em.find(User, {}, {
 ```
 
 `populate: false` is still allowed and serves as a way to disable eager loaded properties.
+
+## Duplicate field names are now validated
+
+When you use the same `fieldName` for two properties in one entity, error will be thrown:
+
+```ts
+@Entity()
+class User {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property({ name: 'custom_name' })
+  name!: number;
+
+  @Property({ name: 'custom_name' })
+  age!: number;
+
+}
+```
+
+This does not apply to virtual properties:
+
+```ts
+@Entity()
+class User {
+
+  @PrimaryKey()
+  id!: number;
+
+  @ManyToOne(() => User, { name: 'parent_id' })
+  parent!: User;
+
+  @Property({ name: 'parent_id',  })
+  parentId!: number;
+
+}
+```
+
+> This validation can be disabled via `discovery.checkDuplicateFieldNames` ORM config option.
--- a/packages/core/src/MikroORM.ts
+++ b/packages/core/src/MikroORM.ts
@@ -229,7 +229,7 @@ export class MikroORM<D extends IDatabaseDriver = IDatabaseDriver> {
     entities = Utils.asArray(entities);
     const tmp = this.discovery.discoverReferences(entities);
     const options = this.config.get('discovery');
-    new MetadataValidator().validateDiscovered([...Object.values(this.metadata.getAll()), ...tmp], options.warnWhenNoEntities, options.checkDuplicateTableNames);
+    new MetadataValidator().validateDiscovered([...Object.values(this.metadata.getAll()), ...tmp], options);
     const metadata = this.discovery.processDiscoveredEntities(tmp);
     metadata.forEach(meta => this.metadata.set(meta.className, meta));
     this.metadata.decorate(this.em);
--- a/packages/core/src/errors.ts
+++ b/packages/core/src/errors.ts
@@ -224,6 +224,10 @@ export class MetadataError<T extends AnyEntity = AnyEntity> extends ValidationEr
     return new MetadataError(`Duplicate ${subject} are not allowed: ${paths.join(', ')}`);
   }
 
+  static duplicateFieldName(className: string, names: [string, string][]): MetadataError {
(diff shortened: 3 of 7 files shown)


ladal1:
Seems the virtual entity exception does not work, our entity fails the validation (which it didnt before) with this definition:
```ts
  @Property({ nullable: true })
  accountId?: number;

  @ManyToOne({ entity: () => Account, joinColumn: 'account_id', nullable: true })
  account?: Account;

```

With Error:
```
return new MetadataError(`Duplicate fieldNames are not allowed: ${names.map(n => `${className}.${n[0]} (fieldName: '${n[0]}')`).join(', ')}`);
               ^
MetadataError: Duplicate fieldNames are not allowed: Entity.accountId (fieldName: 'accountId'), Entity.account (fieldName: 'account')
```

I've tried multiple ways of rewriting the definition but it doesn't pass validation in any way I write it

B4nan:
Uh, I forgot to add `persist: false` to the docs, it works correctly, what you do is wrong and exactly the thing that is now validated, you should have `persist: false` on one of the properties. Relations are not "virtual entities", and this has nothing to do with those, the note in the docs is about "virtual properties" = those marked with `persist: false`.

https://mikro-orm.io/docs/next/defining-entities#virtual-properties

ladal1:
Okay, that makes more sense, so canonically the only correct approach is keeping only the relation and if we need assignment by id use `ref(...)`?

B4nan:
Yes, it was always like that, having two properties side-by-side mapping to the same column is possible nowadays (they will be mapped correctly), but it will bring issues down the path, e.g. property validation is triggered for both those properties, so for new entities you need to provide both values (which is of course irrelevant for nullable props), and I would also expect problems in schema generator.

B4nan:
FWIW, I wanted to support this in v6 somehow, not yet sure how. Technically you can already do something like this:

```ts
@ManyToOne({ entity: () => Account, joinColumn: 'account_id', nullable: true })
account?: Account;

@Property({ nullable: true }) // getters are automatically marked with `persist: false`
get accountId(): number | undefined {
  return this.account?.id;
}

set accountId(val:  number): void {
  this.account = rel(Account, val);
}
```

PR #4974, opened 2023-11-28
feat(entity-generator): detect more ManyToMany relations

boenrobot:
Auto increment columns in pivot entities are now detected.

In addition, pivot tables are allowed to contain additional props, including relations, but only if those props would not hinder the ORM's ability to freely insert and remove records from the collection.

In other words, those additional props need to be optional, and have defaults that are either null or non-unique.

----

I wonder if maybe the additional props thing should be behind an option (let's say called "allowAdditionalPropertiesInPivotTables")? Surely detecting the auto increment for the sake of fixed ordering should be unconditional though.

P.S. I'll add some additional tests for coverage's sake after getting some feedback on the above point...


B4nan:
I'd actually say this whole thing should be behind an option (could be opt-out, but it should be configurable). 

I'd keep it disabled by default and when allowed, I'd include all columns regardless of their type. Ignoring some columns will most probably break the schema diffing.

B4nan:
> For MikroORM to successfully manage a collection of a M:N relation, it needs to be able to freely insert and remove rows in the pivot entity table

You can define a pivot entity in any way you want, if the values don't have defaults, you need to create the records explicitly (through the pivot entity), this itself is not a problem really. I still don't think we need to distinguish that.

tests/features/entity-generator/__snapshots__/EntityGenerator.mysql.test.ts.snap
```diff
@@ -746,7 +758,7 @@ export class Account {
 
   [OptionalProps]?: 'active' | 'receiveEmailNotifications';
 
-  @PrimaryKey()
+  @PrimaryKey({ autoincrement: true })
```
B4nan, line 761: This is not necessary, numeric PKs are always considered as autoincrement.

(I guess it won't hurt anything to have it there, but I'd rather stay with the rule of not printing things that don't have any effect, as this could also mislead users to think it's required to mark such PKs as autoincrement.)

B4nan (commented):
Sorry for the silence, I've been off for a week. Will try to get back to this over the weekend.

I'd appreciate it if you don't try to do too many things in a single PR as it will be harder to review (and therefore to merge). Let's do one thing at a time, it will be also easier to understand from the changelog what has changed. So let's focus on the fixed order collections first, and we can allow more things later on.

> If the ORM is this permissive... What about tables where instead of a composite PK, you have a non-nullable unique composite key? I've added a test case in which the table is called "completed_orders"... We have a PK that is an auto increment, and a single composite unique key, made up from exactly two relations.

You cannot have a relation like that, relations can be built only based on FKs/PKs, not some random unique constraint. We would have to first support that in the ORM, adding support to the entity generator is the very last step to be done. I already have some ideas on how to support this but will need proper testing to see if it's feasible. The population mechanism depends on the fact that entities are always referred by a PK. 

I surely won't have time for that before v6 is out in stable, that is the priority now. But I will be more than happy to improve the entity generator over time, so I appreciate the suggestions.

tests/features/entity-generator/__snapshots__/EntityGenerator.mysql.test.ts.snap
B4nan: > In a different PR, I'm thinking of adding a few tests that are meant to "prove" a 1:1 match between the entity generator and schema generator...

Yeah, we should indeed do that, probably not for all the entity generator tests, as I don't want to increase the test duration too much, but we should know things work fine without me carefully checking the snapshots :] 

> e.g. DESC column indexes, fulltext and spatial indexes, index options.

This should all ideally generate an index expression to be safe. But as I mentioned many times, the purpose of entity generator is mainly bootstrapping, it does not have to be 100% correct. But it needs to work, the entities need to be valid and pass the discovery phase just fine, that's what I am after. Like with your first PR, the generated entities were not syntactically correct and failed to be required (because of unmatched paren or something like that).

> This thing with AI on bigint is an example feature which would have been detected in such a test.

Not sure I follow here, the change you did in the entity schema shouldn't have any effect, and the generated entities were fine as well, they just had some unneeded options (matching the default behavior).

tests/features/entity-generator/__snapshots__/EntityGenerator.mysql.test.ts.snap
```diff
@@ -457,13 +458,22 @@ export const FooBar2Schema = new EntitySchema({
     blob2: { type: 'Buffer', length: 65535, nullable: true },
     array: { type: 'string', columnType: 'text', nullable: true },
     objectProperty: { type: 'any', columnType: 'json', nullable: true },
+    fooParam2: {
```
B4nan, line 461: why is this generated? [FooParam2](https://github.com/mikro-orm/mikro-orm/blob/master/tests/entities-sql/FooParam2.ts) is entity with composite PK (formed from two FKs), it's not a pivot between foobar and foobaz. you are apparently not taking compound FKs into account properly

tests/features/entity-generator/ManyToManyRelations.mysql.test.ts
```diff
@@ -0,0 +1,242 @@
+import { MikroORM, MikroORMOptions } from '@mikro-orm/mysql';
```
B4nan, line 1: i would prefer sqlite since this does not contain anything mysql specific, and mysql is the slowest driver out of all when it comes to creating the db schema (as well as querying it, but the main difference is in the info schema queries)

or how long does it take to execute this one? maybe its fine, since the discovery should be instant (we dont discover anything), but i still expect the info schema queries to be slower than necessary

tests/features/entity-generator/ManyToManyRelations.mysql.test.ts
```diff
@@ -0,0 +1,242 @@
...
+  CONSTRAINT \`fk_user_email_flags_flags1\`
+    FOREIGN KEY (\`flag_id\`)
+      REFERENCES \`flags\` (\`flag_id\`)
+      ON DELETE NO ACTION
+      ON UPDATE NO ACTION
+)
+  ENGINE = InnoDB;
+  `;
+
+beforeAll(async () => {
```
B4nan, line 179: why so many init calls, one should be enough? the state we store is the entity discovery, but since there are no discovered entities, i dont think it matters?

tests/features/entity-generator/ManyToManyRelations.mysql.test.ts
```diff
@@ -0,0 +1,242 @@
...
+    port: 3308,
+    discovery: { warnWhenNoEntities: false },
+    extensions: [EntityGenerator],
+    multipleStatements: true,
+  });
+});
+
+afterEach(async () => {
+  await orm.close(true);
+});
```
B4nan, line 204: i would drop this completely

tests/features/entity-generator/ManyToManyRelations.mysql.test.ts
```diff
@@ -0,0 +1,242 @@
...
+
+afterAll(async () => {
+  orm = await MikroORM.init({
+    dbName: schemaName,
+    port: 3308,
+    discovery: { warnWhenNoEntities: false },
+    extensions: [EntityGenerator],
+    multipleStatements: true,
+  });
+  await orm.schema.dropDatabase();
```
B4nan, line 214: and only keep `close` call here, no need to drop the database when the name is not random

B4nan (changes requested):
few comments on the tests for now, haven't checked the implementation much 

i'd also think about better names for the config options, i dont like that very much (but its a minor issue so keeping that for later)

tests/features/entity-generator/__snapshots__/AmbiguousFks.mysql.test.ts.snap
```diff
@@ -54,7 +54,7 @@ export class ProductColors {
   product!: Products;
 
   @Index({ name: 'fk_product_colors_products1_idx' })
-  @Property({ persist: false })
+  @Property({ autoincrement: true, persist: false })
```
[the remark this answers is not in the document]
boenrobot: autoincrement option can be applied on any numeric column in place of a default value.

An autoincrement column doesn't have to be PK, and it may be a FK.

In this case, the column is all three - it's part of a composite PK within the `product_colors` table, the `product_id` column has a foreign key to the `products` table, and it has autoincrement.

See

```sql
CREATE TABLE IF NOT EXISTS `product_colors` (
  `color_id` INT UNSIGNED NOT NULL,
  `product_id` INT UNSIGNED NOT NULL AUTO_INCREMENT,
  PRIMARY KEY (`color_id`, `product_id`),
  INDEX `fk_product_colors_products1_idx` (`product_id` ASC) VISIBLE,
  CONSTRAINT `fk_product_colors_colors1`
    FOREIGN KEY (`color_id`)
    REFERENCES `colors` (`color_id`)
    ON DELETE NO ACTION
    ON UPDATE NO ACTION,
  CONSTRAINT `fk_product_colors_products1`
    FOREIGN KEY (`product_id`)
    REFERENCES `products` (`product_id`)
    ON DELETE NO ACTION
    ON UPDATE NO ACTION)
ENGINE = InnoDB;
```

And I imagine someone may intentionally do such a design if they want to f.e. pre-make the referenced column rows, and then ensure only the first N ones get populated in the other table... not a typical scenario, I know, but valid, and thus something that should be handled.

Having said all that... The above autoincrement was added by mistake 😅 
If anything is wrong here, it's the SQL, not the generator.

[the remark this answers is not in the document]
B4nan: > autoincrement option can be applied on any numeric column in place of a default value.

in fact this only works in postgres, but yes

> An autoincrement column doesn't have to be PK, and it may be a FK.

nope, FK is not autoincrement, FK is poiting to a PK which is autoincrement. the flag shouldn't be there.

tests/features/entity-generator/ManyToManyRelations.mysql.test.ts
boenrobot: MySQL has Workbench, which lets me very quickly and visually design schemas with complex relations or other exotic options to tables and columns, that I can then forward engineer into SQL create statements.

There's no equivalent to that for SQLite and Postgresql. There are table managers where you can define those things, sure, but you can't _draw_ these things out first and forward engineer them once you're ready.

I've seen maybe one or two Postgresql editors that allegedly support this functionality, but are paid (and I'm not paying for that...), while Workbench is free, and I'm not aware of any equivalent (even a paid one) for SQLite.

If the schema, in its current form, can be switched to sqlite with no edits, fine, I can do that... but I refuse to design SQLite schemas from scratch, as that's too much tortune.

tests/features/entity-generator/ManyToManyRelations.mysql.test.ts
boenrobot: At one point I found out that the entity generator also adds the generated entities' metadata to the ORM's state... Not sure if that's a bug, but it appears to be the current behaviour.

So if you generate with one set of options, and then regenerate with a different set of options that results in a different meta data (say, `bidirectionalRelations: true` vs `bidirectionalRelations: false`), you'll actually get... ehh... I'm not sure if a merge, or just the first or just last set... but certainly not the same result as generating the first time.

tests/features/entity-generator/ManyToManyRelations.mysql.test.ts
B4nan: we are talking about library tests, you are talking about designing schema for complex apps through UI... 

> but I refuse to design SQLite schemas from scratch, as that's too much torture.

well, and i refuse to merge slow tests (and as said, mysql is the slowest driver we support, especially when it comes to working with information schema) :] 

i'd prefer if you change this mindset. working on libraries requires some sacrifices, your (or mine) own comfort (as a contributor) is not the most important part here - it's the produced code (and its quality, performance, and whatnot). imagine i say i dont like mysql because its slow (or any other reason), so i decide not to test anything with it... 

the difference here is can be really huge, easily 10x, as creating schema with mysql on its own takes ages (lets say hundreds of milliseconds), while with sqlite you only use memory and its close to instant (as you dont write anything to the disk).

anyway, back to my question, how long does the test take? as that's the important bit here, if its a couple hundreds of milliseconds, we can probably keep things with mysql just fine, but i dont want to merge things that take seconds to test (we already have several schema related tests that do, guess what, all of them are in mysql :])

tests/features/entity-generator/ManyToManyRelations.mysql.test.ts
B4nan: not sure if it was already there, but you did some changes that alter the metadata yourself, e.g. here:

https://github.com/mikro-orm/mikro-orm/pull/4892/files?w=1#diff-41bc1ce63697a8a667a96cd32696603b53981ac0b10b598d91ddd22ef5ff663aR151

and the same PR also makes the schema inference dependent on the entity generator ORM config, so that's another problem. i guess lest keep it this way.

tests/features/entity-generator/ManyToManyRelations.mysql.test.ts
B4nan: btw tried to run the test locally and its around 250ms with my m1 pro, so that's good enough

tests/features/entity-generator/__snapshots__/AmbiguousFks.mysql.test.ts.snap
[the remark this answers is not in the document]
boenrobot: > in fact this only works in postgres, but yes

idk about the entity manager or knex level, but on a raw SQL level, the following works fine in MySQL 8.0.35, and probably some earlier MySQL versions too (though I haven't checked such):

```sql
INSERT INTO colors (`name`) VALUES ('red'), ('white'), ('blue');

INSERT products (`name`) VALUES ('Shirt'), ('Pants'), ('Phone');

INSERT INTO product_colors (color_id) VALUES (3), (1), (2);
```

The AI is part of the PK, and that column is also a FK. It's giving exactly what you'd expect:

```
color_id	product_id
3		1
1		2
2		3
```

In the `many_to_many_variants` schema, this also works fine on a raw SQL level on MySQL 8.0.35:

```sql
INSERT INTO users (`name`) VALUES ('a'), ('b'), ('c');

INSERT INTO orders (`name`) VALUES ('shirt'), ('pants'), ('phone');

INSERT INTO user_orders (user_id, order_id) VALUES (2,3), (1, 1), (1, 2), (3, 1);
```

giving, as you'd expect

```
user_id	product_id	priority
2	3		1
1	1		2
1	2		3
3	1		4
```
and the "priority" column there is not part of the PK, and is not a FK.

tests/features/entity-generator/__snapshots__/EntityGenerator.mysql.test.ts.snap
boenrobot: Yeah, it shouldn't be there... Though not because it has a PK that is a composite key of two FKs... I mean, that part is required for pivot tables... But because it has a required property that is not part of the FK. I've simplified the check to just check whether the additional props are optional.

tests/features/entity-generator/ManyToManyRelations.mysql.test.ts
boenrobot: FWIW, you do have a point that the names of the schemas are not random, so... I've now changed the test files with many cases to not drop the schemas in afterAll(), and check for existence on beforeAll().

While this won't make much difference in CI (since containers are recreated anyway), it will make a difference in local development - as long as the container is still running, you'll get a boost during those tests, as they'll skip the schema creation.

B4nan (approved):
I guess I kinda like the option names in the end, let's keep them.

Let's merge, found just a few formatting issues.

B4nan:
Thanks!

tests/features/entity-generator/AmbiguousFks.mysql.test.ts
```diff
@@ -169,8 +169,11 @@ beforeAll(async () => {
     extensions: [EntityGenerator],
     multipleStatements: true,
   });
-  await orm.schema.ensureDatabase();
-  await orm.schema.execute(schema);
+  const driver = orm.config.getDriver();
```
B4nan, line 172: FYI `ensureDatabase` returns true if the database was created, so this can be simplified:

```ts
if (await orm.schema.ensureDatabase()) {
  await orm.schema.execute(schema);
}
```

I guess it would make sense to add `createSchema: boolean` option there too.

PR #4978, opened 2023-11-29
Fixes Reference.load(), Collection.loadItems(), dataloader enabling logic. Dataloaders honor options and filters.

darkbasic:
It's mostly done and tests are passing but I'm marking this as draft until I add more tests for filters and options with the Collection dataloader.

### DIFF
--- a/packages/core/src/entity/Collection.ts
+++ b/packages/core/src/entity/Collection.ts
@@ -338,7 +338,7 @@ export class Collection<T extends object, O extends object = object> extends Arr
     }
 
     const populate = Array.isArray(options.populate)
-      ? options.populate.map(f => `${this.property.name}.${f}`)
+      ? options.populate.map(f => f === '*' ? f : `${this.property.name}.${f}`)
       : [`${this.property.name}${options.ref ? ':ref' : ''}`];
     const schema = this.property.targetMeta!.schema === '*'
       ? helper(this.owner).__schema
--- a/packages/core/src/entity/Collection.ts
+++ b/packages/core/src/entity/Collection.ts
@@ -314,7 +314,7 @@ export class Collection<T extends object, O extends object = object> extends Arr
 
     const em = this.getEntityManager();
 
-    if (options.dataloader ?? (DataloaderUtils.getDataloaderType(em.config.get('dataloader')) > Dataloader.COLLECTION)) {
+    if (options.dataloader ?? [Dataloader.ALL, Dataloader.COLLECTION].includes(DataloaderUtils.getDataloaderType(em.config.get('dataloader')))) {
       const order = [...this.items]; // copy order of references
       const customOrder = !!options.orderBy;
       const items: TT[] = await em.colLoader.load(this);
--- a/packages/core/src/entity/Collection.ts
+++ b/packages/core/src/entity/Collection.ts
@@ -317,7 +317,7 @@ export class Collection<T extends object, O extends object = object> extends Arr
     if (options.dataloader ?? [Dataloader.ALL, Dataloader.COLLECTION].includes(DataloaderUtils.getDataloaderType(em.config.get('dataloader')))) {
       const order = [...this.items]; // copy order of references
       const customOrder = !!options.orderBy;
-      const items: TT[] = await em.colLoader.load(this);
+      const items: TT[] = await em.colLoader.load([this, options]);
 
       if (!customOrder) {
         this.reorderItems(items, order);
--- a/packages/core/src/entity/Reference.ts
+++ b/packages/core/src/entity/Reference.ts
@@ -115,7 +115,7 @@ export class Reference<T> {
   async load<TT extends T, K extends keyof T = never, P extends string = never>(options?: LoadReferenceOptions<T, P> | K): Promise<Loaded<TT, P> | T[K]> {
     const opts: Dictionary = typeof options === 'object' ? options : { prop: options } as LoadReferenceOptions<T, P>;
 
-    if (opts.dataloader ?? (DataloaderUtils.getDataloaderType(helper(this.entity).__em.config.get('dataloader')) > Dataloader.REFERENCE)) {
+    if (opts.dataloader ?? [Dataloader.ALL, Dataloader.REFERENCE].includes(DataloaderUtils.getDataloaderType(helper(this.entity).__em.config.get('dataloader')))) {
       return helper(this.entity).__em.refLoader.load(this);
     }
 
--- a/packages/core/src/entity/Reference.ts
+++ b/packages/core/src/entity/Reference.ts
@@ -117,11 +117,7 @@ export class Reference<T> {
 
     if (!this.isInitialized() || opts.refresh) {
       if (opts.dataloader ?? dataloader ?? [Dataloader.ALL, Dataloader.REFERENCE].includes(DataloaderUtils.getDataloaderType(helper(this.entity).__em.config.get('dataloader')))) {
-        await helper(this.entity).__em.refLoader.load([this, {
-          ...(opts?.populate && { populate: opts?.populate }),
-          ...(opts?.lockMode && { lockMode: opts?.lockMode }),
-          ...(opts?.connectionType && { populate: opts?.connectionType }),
-        }]);
+        await helper(this.entity).__em.refLoader.load([this, opts]);
       } else {
         await helper(this.entity).init(undefined, opts?.populate, opts?.lockMode, opts?.connectionType);
       }
--- a/packages/core/src/entity/Reference.ts
+++ b/packages/core/src/entity/Reference.ts
@@ -106,21 +106,25 @@ export class Reference<T> {
    * Ensures the underlying entity is loaded first (without reloading it if it already is loaded).
    * Returns the requested property instead of the whole entity.
    */
-  async load<K extends keyof T>(prop: K): Promise<T[K]>;
+  async load<K extends keyof T>(prop: K, dataloader?: boolean): Promise<T[K]>;
 
   /**
    * Ensures the underlying entity is loaded first (without reloading it if it already is loaded).
    * Returns either the whole entity, or the requested property.
    */
-  async load<TT extends T, K extends keyof T = never, P extends string = never>(options?: LoadReferenceOptions<T, P> | K): Promise<Loaded<TT, P> | T[K]> {
+  async load<TT extends T, K extends keyof T = never, P extends string = never>(options?: LoadReferenceOptions<T, P> | K, dataloader?: boolean): Promise<Loaded<TT, P> | T[K]> {
     const opts: Dictionary = typeof options === 'object' ? options : { prop: options } as LoadReferenceOptions<T, P>;
 
-    if (opts.dataloader ?? [Dataloader.ALL, Dataloader.REFERENCE].includes(DataloaderUtils.getDataloaderType(helper(this.entity).__em.config.get('dataloader')))) {
-      return helper(this.entity).__em.refLoader.load(this);
-    }
-
     if (!this.isInitialized() || opts.refresh) {
-      await helper(this.entity).init(undefined, opts?.populate, opts?.lockMode, opts?.connectionType);
+      if (opts.dataloader ?? dataloader ?? [Dataloader.ALL, Dataloader.REFERENCE].includes(DataloaderUtils.getDataloaderType(helper(this.entity).__em.config.get('dataloader')))) {
+        await helper(this.entity).__em.refLoader.load([this, {
+          ...(opts?.populate && { populate: opts?.populate }),
+          ...(opts?.lockMode && { lockMode: opts?.lockMode }),
+          ...(opts?.connectionType && { populate: opts?.connectionType }),
+        }]);
+      } else {
+        await helper(this.entity).init(undefined, opts?.populate, opts?.lockMode, opts?.connectionType);
+      }
     }
 
     if (opts.prop) {
--- a/packages/core/src/utils/DataloaderUtils.ts
+++ b/packages/core/src/utils/DataloaderUtils.ts
@@ -3,7 +3,7 @@ import type {
   Primary,
   Ref,
 } from '../typings';
-import { Collection } from '../entity/Collection';
+import { Collection, type InitOptions } from '../entity/Collection';
 import { helper } from '../entity/wrap';
 import { type EntityManager } from '../EntityManager';
 import type DataLoader from 'dataloader';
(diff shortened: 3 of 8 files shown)


darkbasic:
It's ready for review. Please don't squash the commits when you merge it because I'd like to keep track of all the changes if something breaks. I've tested it against my project as well and the suite passes.

packages/core/src/entity/Collection.ts
```diff
@@ -335,7 +342,7 @@ export class Collection<T extends object, O extends object = object> extends Arr
     }
 
     const populate = Array.isArray(options.populate)
-      ? options.populate.map(f => `${this.property.name}.${f}`)
+      ? [...(options.populate.includes('*') ? Object.keys(this.property.targetMeta!.properties) : options.populate)].map(f => `${this.property.name}.${f}`)
```
B4nan, line 345: This looks wrong, the star needs to work recursively and you are only expanding one level here.

The idea was that we pass down the star and it's gonna be handled in the EntityLoader. So it should be rather this:

```suggestion
      ? options.populate.map(f => f === '*' ? f : `${this.property.name}.${f}`)
```

B4nan (commented):
thanks, left few initial comments

packages/core/src/entity/Collection.ts
```diff
@@ -311,10 +311,17 @@ export class Collection<T extends object, O extends object = object> extends Arr
 
     const em = this.getEntityManager();
 
-    if (options.dataloader ?? (DataloaderUtils.getDataloaderType(em.config.get('dataloader')) > Dataloader.COLLECTION)) {
+    if (options.dataloader ?? [Dataloader.ALL, Dataloader.COLLECTION].includes(DataloaderUtils.getDataloaderType(em.config.get('dataloader')))) {
       const order = [...this.items]; // copy order of references
       const customOrder = !!options.orderBy;
-      const items: TT[] = await this.getEntityManager().colLoader.load(this);
+      const items: TT[] = await this.getEntityManager().colLoader.load([this, {
```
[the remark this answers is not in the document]
B4nan: Got it, I thought the `load` method is our own implementation.

packages/core/src/entity/Collection.ts
```diff
@@ -311,10 +311,17 @@ export class Collection<T extends object, O extends object = object> extends Arr
...
+        ...(options?.populate && { populate: options?.populate }),
+        ...(options?.orderBy && { orderBy: options?.orderBy }),
+        ...(options?.where && { where: options?.where }),
+        ...(options?.lockMode && { lockMode: options?.lockMode }),
+        ...(options?.connectionType && { connectionType: options?.connectionType }),
+        ...(options?.logging && { logging: options?.logging }),
```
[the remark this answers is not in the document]
darkbasic: Just making sure we don't pass unneeded properties, for example dataloader.
Wouldn't hurt anyway because we don't support dataloader in find queries yet, but mostly I hate the fact that Typescript's structural subtyping allows you to pass additional properties like `dataloader` despite option's type is specifically `Omit<InitOptions<any, any>, 'dataloader'>`. We can safely pass options as is I guess.

packages/core/src/entity/Collection.ts
darkbasic: I've tried your suggestion but `await Promise.all(colsA.map(col => col.loadItems({ populate: [ '*' ] })));` times out.

B4nan: Yes, because there is some bug in there, I saw this in another test just a few days ago. But it's the correct way of doing this.

packages/core/src/entity/WrappedEntity.ts
```diff
@@ -138,7 +138,7 @@ export class WrappedEntity<Entity extends object> {
       throw ValidationError.entityNotManaged(this.entity);
     }
 
-    await this.__em.findOne(this.entity.constructor.name, this.entity, { refresh: true, lockMode, populate, connectionType, schema: this.__schema });
+    await this.__em.findOne(this.entity.constructor.name, this.entity, { filters: false, refresh: true, lockMode, populate, connectionType, schema: this.__schema });
```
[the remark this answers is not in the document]
B4nan: >  If you don't ignore filters you won't have loaded the reference:

Yes, and that is correct behavior, as it is not valid with current default filters. Your change would break one of the main use cases for filters - soft deletes. 

> I expect to be able to load references anyway,

You can, by disabling filters - explicitly.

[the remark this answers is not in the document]
darkbasic: > Your change would break one of the main use cases for filters - soft deletes.

If that's a desired feature it should return null/undefined rather than keep returning the id, like if the property was nullable. Otherwise if find it rather confusing.
There should also be a way to disable the filters and to customize them, like `Collection.loadItems` does (but currently there is no `where` option for `Reference.load`.
Also would be nice to have an option to globally disable default filters on References and/or Collections.
Niche bonus: the collection/reference should have a property storing the combination of filters which have been used to load it.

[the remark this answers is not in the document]
darkbasic: I've removed the commit which disables default filters for References. I'll update the bug report later today.

packages/core/src/entity/Collection.ts
darkbasic: Rebased against https://github.com/mikro-orm/mikro-orm/pull/5001

packages/core/src/entity/Reference.ts
```diff
@@ -106,21 +106,25 @@ export class Reference<T> {
...
    */
-  async load<K extends keyof T>(prop: K): Promise<T[K]>;
+  async load<K extends keyof T>(prop: K, dataloader?: boolean): Promise<T[K]>;
 
   /**
    * Ensures the underlying entity is loaded first (without reloading it if it already is loaded).
    * Returns either the whole entity, or the requested property.
    */
-  async load<TT extends T, K extends keyof T = never, P extends string = never>(options?: LoadReferenceOptions<T, P> | K): Promise<Loaded<TT, P> | T[K]> {
+  async load<TT extends T, K extends keyof T = never, P extends string = never>(options?: LoadReferenceOptions<T, P> | K, dataloader?: boolean): Promise<Loaded<TT, P> | T[K]> {
```
B4nan, line 115: what's the reason for the additional parameter? we already have such option, if we need some kind of override (which i still dont really think, how would it be different than overriding the already existing option), i'd rather see it as yet another option instead of a separate parameter, the signature this way feels weird

packages/core/src/entity/Reference.ts
```diff
@@ -106,21 +106,25 @@ export class Reference<T> {
...
-      return helper(this.entity).__em.refLoader.load(this);
-    }
-
     if (!this.isInitialized() || opts.refresh) {
-      await helper(this.entity).init(undefined, opts?.populate, opts?.lockMode, opts?.connectionType);
+      if (opts.dataloader ?? dataloader ?? [Dataloader.ALL, Dataloader.REFERENCE].includes(DataloaderUtils.getDataloaderType(helper(this.entity).__em.config.get('dataloader')))) {
+        await helper(this.entity).__em.refLoader.load([this, {
+          ...(opts?.populate && { populate: opts?.populate }),
+          ...(opts?.lockMode && { lockMode: opts?.lockMode }),
+          ...(opts?.connectionType && { populate: opts?.connectionType }),
```
B4nan, line 123: this is not needed, right? lets pass whole options object

packages/core/src/utils/DataloaderUtils.ts
```diff
@@ -3,29 +3,42 @@ import type {
...
+    refsWithOpts: readonly [Ref<any>, Omit<LoadReferenceOptions<any, any>, 'dataloader'>?][],
+  ): Map<string, Set<Primary<any>>> {
+    const map = new Map<string, Set<Primary<any>>>();
+    for (const [ref, opts] of refsWithOpts) {
+      /* The key is a combination of the className and a stringified version if the load options because we want
+         to map each combination of entities/options into separate find queries in order to return accurate results.
+         This could be further optimized finding the "lowest common denominator" among the different options
+         for each Entity and firing a single query for each Entity instead of Entity+options combination.
+         The former is the approach taken by the out-of-tree "find" dataloader: https://github.com/darkbasic/mikro-orm-dataloaders
+         In real world scenarios (GraphQL) msot of the times you will end up batching the same sets of options anyway,
```
B4nan, line 29:
```suggestion
         In real-world scenarios (GraphQL) most of the time you will end up batching the same sets of options anyway,
```

packages/core/src/utils/DataloaderUtils.ts
```diff
@@ -91,31 +107,41 @@ export class DataloaderUtils {
...
-            populate: Array.from(filterMap.keys()).filter(
-              // We need to do so only if the inverse side is a collection, because we can already retrieve the PK from a reference without having to load it
-              prop => em.getMetadata().get(className).properties[prop]?.ref !== true,
-            ) as any,
-          },
-        ),
-    );
+        {
+          // We need to populate the inverse side of the relationship in order to be able to later retrieve the PK(s) from its item(s)
+          // FIXME: loadItems "ref" option must be taken into consideration as well
```
B4nan, line 134: can we resolve this? sounds rather easy

packages/core/src/utils/DataloaderUtils.ts
```diff
@@ -91,31 +107,41 @@ export class DataloaderUtils {
...
+          populate: [...(opts.populate === false ? [] : opts.populate ?? []), ...Array.from(filterMap.keys()).filter(
+            // We need to do so only if the inverse side is a collection, because we can already retrieve the PK from a reference without having to load it
+            prop => em.getMetadata().get(className).properties[prop]?.ref !== true)] as any,
+          ...(opts?.orderBy && { orderBy: opts?.orderBy }),
+          ...(opts?.lockMode && { lockMode: opts?.lockMode }),
+          ...(opts?.connectionType && { connectionType: opts?.connectionType }),
+          ...(opts?.logging && { logging: opts?.logging }),
```
B4nan, line 141: same here, iirc we removed it from one place already and it just stayed here and there, or is it needed? does it have effect on the map keys if we keep everything? if thats the reason, it sounds better to introduce a helper to compute a key out of all options, which will only respect those we can about

packages/core/src/utils/DataloaderUtils.ts
```diff
@@ -91,31 +107,41 @@ export class DataloaderUtils {
...
-            ) as any,
-          },
-        ),
-    );
+        {
+          // We need to populate the inverse side of the relationship in order to be able to later retrieve the PK(s) from its item(s)
+          // FIXME: loadItems "ref" option must be taken into consideration as well
+          populate: [...(opts.populate === false ? [] : opts.populate ?? []), ...Array.from(filterMap.keys()).filter(
+            // We need to do so only if the inverse side is a collection, because we can already retrieve the PK from a reference without having to load it
+            prop => em.getMetadata().get(className).properties[prop]?.ref !== true)] as any,
```
B4nan, line 137:
```suggestion
            prop => em.getMetadata(className).properties[prop]?.ref !== true)] as any,
```

B4nan (changes requested):
can we also add the `filters` option to both `InitOptions` and `LoadReferenceOptions`? that way users can disable default filters or enable the opt-in ones (and provide arguments to them), that seems to be the last missing thing

packages/core/src/entity/Reference.ts
darkbasic: I's because of this overload: `async load<K extends keyof T>(prop: K, dataloader?: boolean): Promise<T[K]>;`
Previously there was no way to pass options when you picked a specific `prop`, so I had to add an additional second parameter to allow enabling/disabling the dataloader.

packages/core/src/utils/DataloaderUtils.ts
darkbasic: Should be a matter of adding `:ref` to the populate array right? I didn't do so because there were issues with populate so I wanted the dust to settle before doing so.

packages/core/src/entity/Reference.ts
darkbasic: No it's not needed: I removed it from Collection but forgot to do so from Reference as well.

packages/core/src/utils/DataloaderUtils.ts
darkbasic: Yeah we can pass opts directly, we just need to make sure to do so **before** specifying the custom populate.

packages/core/src/entity/Reference.ts
B4nan: I was actually thinking about removing this signature to simplify how it works, wdyt about that? The `Collection` class also has `load` method which only takes options, I wanted to align this in v6.

We could introduce a new method for this instead, e.g. `loadProperty` (as we already have `getProperty` which returns without checking the state).

Alternatively, we could add `options` parameter to the property getter overload, as it should be there as well, that feels like an omission.

packages/core/src/utils/DataloaderUtils.ts
B4nan: is there actually something to fix? i mean, the `:ref` should be already added by user in `options.populate`, why would it require some additional handling?

tests/issues/GH4977.test.ts
```diff
@@ -0,0 +1,136 @@
...
+  PrimaryKey,
+  Property,
+  SimpleLogger,
+  ManyToOne,
+  OneToMany,
+  Collection,
+  Ref,
+  ref,
+} from '@mikro-orm/sqlite';
+import { mockLogger } from '../helpers';
```
B4nan, line 13: this is also not used :]

darkbasic: I'm pretty sure there is an issue with your eslint/prettier config at this point because I'm certain that lint-staged did run after amending the commit yet it didn't remove it. I'll remove the unused import manually.

B4nan: There is no issue, those rules are simply disabled in the tests as more often than not, it's fine to have some statement that does something and never use its value (e.g. because you assert the queries made by the ORM).

darkbasic: I agree for statements that don't use its value, but what's the utility of unused imports?

B4nan: Looks more like we don't have a lint rule for this at all.

packages/core/src/utils/DataloaderUtils.ts
darkbasic: @B4nan `find` doesn't have a `ref` property in its options while `loadItems` does, so it won't work without additional handling. I guess that conditionally adding `:ref` to the populate array every time `loadItems` has the `ref` property set should do the trick.

packages/core/src/entity/Reference.ts
darkbasic: @B4nan yes we should simplify the signature if possible: either `loadProperty` or an additional `options` param or both.
I suggest making a separate PR for this together with the `InitOptions`/`LoadReferenceOptions` `filters` option so that we can merge this, release a new alpha and start testing it in my app.

packages/core/src/utils/DataloaderUtils.ts
darkbasic: image
Are you suggesting that we should get rid of the `ref` prop in `InitOptions`?

B4nan: no, thats for collections, i am saying that the option (nor `:ref` populate hints) make sense for M:1 and 1:1 relations, that's what i am referring to as owning sides. collections are technically always inverse sides (m:n have a concept of owning side but technically it is not, the owning sides are the m:1 relations in pivot entity)

B4nan: i guess i just confused myself, i thought you are talking about the reference wrapper (and therefore about m:1/1:1 relations)

darkbasic: Good, so if the proposed fix is fine I think this is ready to be merged and I will open a new PR filters and loadProperty.

B4nan:
lets improve in another PR, this is indeed too huge already. 

(and of course, thanks a lot!)

PR #4998, opened 2023-12-08
feat(core): respect `schema` parameter in `clientUrl`

### LINKED ISSUE #4997: Use "schema" query parameter in "clientUrl" in case it exists
It might not be a problem for everybody. The schema query parameter is part of a database url. When setting the clientUrl with a url that contains the schema, the `getConnectionOptions` method does not parse it. In this scenario, you have to manually pass the "schema" property inside the "defineConfig" method as a separate argument.

By simply providing a clientUrl that has the schema, we should be able to connect to that schema, like it happens for the dbName, username and password. This can be achieved by providing the ConnectionConfig interface with an optional "schema" string property, and also by setting the schema inside the options of the Connection class like so:

image

I think the solution above is the most straightforward way to do this. It seems to be the most plausible one as well.

kaionesyan:
Closes #4997 

This allows for developers to select the database schema they want to use inside the clientUrl instead of having to manually pass down the "schema" property.

So basically this:
`
defineConfig({
  clientUrl: "postgres://localhost:5432/mydb?schema=otherschema",
}),
`

instead of this:
`
defineConfig({
  clientUrl: "postgres://localhost:5432/mydb?schema=thisisirrelevant",
  schema: "otherschema",
}),
`

### DIFF
--- a/packages/core/src/connections/Connection.ts
+++ b/packages/core/src/connections/Connection.ts
@@ -105,7 +105,6 @@ export abstract class Connection {
       this.options.user = ret.user = this.options.user ?? this.config.get('user', decodeURIComponent(url.username));
       this.options.password = ret.password = this.options.password ?? this.config.get('password', decodeURIComponent(url.password));
       this.options.dbName = ret.database = this.options.dbName ?? this.config.get('dbName', decodeURIComponent(url.pathname).replace(/^\//, ''));
-      this.options.schema = ret.schema = this.options.schema ?? this.config.get('schema', decodeURIComponent(url.searchParams.get('schema') ?? ''));
     }
 
     return ret;
@@ -114,8 +113,12 @@ export abstract class Connection {
   getClientUrl(): string {
     const options = this.getConnectionOptions();
     const url = new URL(this.config.getClientUrl(true));
+    const password = options.password ? ':*****' : '';
+    const schema = options.schema && options.schema !== this.platform.getDefaultSchemaName()
+      ? `?schema=${options.schema}`
+      : '';
 
-    return `${url.protocol}//${options.user}${options.password ? ':*****' : ''}@${options.host}:${options.port}${options.schema !== 'public' ? `?schema=${options.schema}` : ''}`;
+    return `${url.protocol}//${options.user}${password}@${options.host}:${options.port}${schema}`;
   }
 
   setMetadata(metadata: MetadataStorage): void {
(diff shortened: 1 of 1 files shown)


packages/core/src/connections/Connection.ts
```diff
@@ -97,6 +97,7 @@ export abstract class Connection {
       this.options.user = ret.user = this.options.user ?? decodeURIComponent(url.username);
       this.options.password = ret.password = this.options.password ?? decodeURIComponent(url.password);
       this.options.dbName = ret.database = this.options.dbName ?? decodeURIComponent(url.pathname).replace(/^\//, '');
+      this.options.schema = ret.schema = this.options.schema ?? decodeURIComponent(url.searchParams.get('schema') ?? '');
```
B4nan, line 100: i removed the other change, as that shouldnt be needed, the other branch uses default connection urls, those never have a schema (or if they do, its the default schema), and tried to fix the tests, lets see

B4nan:
this needs to be resolved too, the `schema` is now apparently always set, which is wrong

```
  ● Console
    console.error
      Ignoring invalid configuration option passed to Connection: schema. This is currently a warning, but in future versions of MySQL2, an error will be thrown if you pass an invalid configuration option to a Connection
      at new ConnectionConfig (node_modules/mysql2/lib/connection_config.js:88:17)
      at Object.exports.createConnection (node_modules/mysql2/index.js:10:35)
      at node_modules/knex/lib/dialects/mysql/index.js:67:38
      at Client_MySQL2.acquireRawConnection (node_modules/knex/lib/dialects/mysql/index.js:66:12)
```

packages/core/src/connections/Connection.ts
B4nan, line 100:
```suggestion

      if (this.options.schema || url.searchParams.has('schema')) {
        this.options.schema = ret.schema = this.options.schema ?? decodeURIComponent(url.searchParams.get('schema'));
      }
```

packages/core/src/connections/Connection.ts
```diff
@@ -97,6 +97,10 @@ export abstract class Connection {
       this.options.user = ret.user = this.options.user ?? decodeURIComponent(url.username);
       this.options.password = ret.password = this.options.password ?? decodeURIComponent(url.password);
       this.options.dbName = ret.database = this.options.dbName ?? decodeURIComponent(url.pathname).replace(/^\//, '');
+
+      if (this.options.schema || url.searchParams.has('schema')) {
+        this.options.schema = ret.schema = this.options.schema ?? decodeURIComponent(url.searchParams.get('schema'));
```
B4nan, line 102:
```suggestion
        this.options.schema = ret.schema = this.options.schema ?? decodeURIComponent(url.searchParams.get('schema')!);
```

B4nan (changes requested):
ok, so tests are passing, the last bit is adding some tests to cover the missing lines. 

i'd just add some unit test to `Connection.test.ts`

PR #5021, opened 2023-12-14
feat(mysql): support `order by nulls first/last`

### LINKED ISSUE #5004: MySQL Order By `*_NULLS_(LAST|FIRST)` incorrect syntax
### Describe the bug

When using `orderBy` in a find operation, using `ASC_NULLS_LAST` or equivalent will insert the syntax `order by ... asc nulls last`. This is incorrect syntax.

If we have a table

```sql
CREATE TABLE foo
(
  id            INT UNSIGNED AUTO_INCREMENT
    PRIMARY KEY,
  text          VARCHAR(255) NOT NULL,
  optional_text VARCHAR(255) NULL,
);
```

Then to sort nulls on `optional_text`, we need to do the following:

| Operation        | Syntax                                                   |
|------------------|----------------------------------------------------------|
| ASC NULLS FIRST  | `order by optional_text is not null, optional_text`      |
| ASC NULLS LAST   | `order by optional_text is null, optional_text`          |
| DESC NULLS FIRST | `order by optional_text is not null, optional_text desc` |
| DESC NULLS LAST  | `order by optional_text is null, optional_text desc`     |

### Reproduction

https://github.com/kpervin/MikroORM-Issue-Example-Repo/blob/c446d1fd0b313da3c1ccb97915fc6a2fd09003ae/test/mysql-order-by-issue.spec.ts#L49-L73

@mikro-orm/mysql

### MikroORM version

5.9.4

### Node.js version

20.9.0

### Operating system

Ubuntu 20.04 LTS

### Validations

kpervin:
Fixes #5004

- Added `getOrderByExpression` to AbstractSqlPlatform.ts as an overridable method to format `orderBy` queries. This uses syntax that is recognized by both SQLite and PostGreSQL
- Added override of `getOrderByExpression` to MariaDbPlatform.ts and MySqlPlatform.ts that adds functionality to handle `(ASC|DESC) NULLS (FIRST|LAST)` syntax

### DIFF
--- a/packages/mariadb/src/MariaDbPlatform.ts
+++ b/packages/mariadb/src/MariaDbPlatform.ts
@@ -81,13 +81,17 @@ export class MariaDbPlatform extends AbstractSqlPlatform {
     [QueryOrder.desc_nulls_last]: 'is null',
   } as const;
 
+  /* istanbul ignore next */
   override getOrderByExpression(column: string, direction: string): string[] {
     const ret: string[] = [];
-    const dir = direction.toLowerCase();
+    const dir = direction.toLowerCase() as keyof typeof this.ORDER_BY_NULLS_TRANSLATE;
+
     if (dir in this.ORDER_BY_NULLS_TRANSLATE) {
-      ret.push(`${column} ${this.ORDER_BY_NULLS_TRANSLATE[dir as keyof typeof this.ORDER_BY_NULLS_TRANSLATE]}`);
+      ret.push(`${column} ${this.ORDER_BY_NULLS_TRANSLATE[dir]}`);
     }
+
     ret.push(`${column} ${dir.replace(/(\s|nulls|first|last)*/gi, '')}`);
+
     return ret;
   }
 
--- a/packages/mariadb/src/MariaDbPlatform.ts
+++ b/packages/mariadb/src/MariaDbPlatform.ts
@@ -74,7 +74,7 @@ export class MariaDbPlatform extends AbstractSqlPlatform {
     return `alter table ${quotedTableName} add fulltext index ${quotedIndexName}(${quotedColumnNames.join(',')})`;
   }
 
-  private readonly NULLS_TRANSLATE = {
+  private readonly ORDER_BY_NULLS_TRANSLATE = {
     [QueryOrder.asc_nulls_first]: 'is not null',
     [QueryOrder.asc_nulls_last]: 'is null',
     [QueryOrder.desc_nulls_first]: 'is not null',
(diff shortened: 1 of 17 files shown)


packages/mariadb/src/MariaDbPlatform.ts
```diff
@@ -74,4 +74,21 @@ export class MariaDbPlatform extends AbstractSqlPlatform {
...
   }
 
+  private readonly NULLS_TRANSLATE = {
+    [QueryOrder.asc_nulls_first]: 'is not null',
+    [QueryOrder.asc_nulls_last]: 'is null',
+    [QueryOrder.desc_nulls_first]: 'is not null',
+    [QueryOrder.desc_nulls_last]: 'is null',
+  } as const;
+
+  override getOrderByExpression(column: string, direction: string): string[] {
```
B4nan, line 84: i guess we dont need to test the mariadb explicitly, lets just add `/* istanbul ignore next */` here
```suggestion
  /* istanbul ignore next */
  override getOrderByExpression(column: string, direction: string): string[] {
```

B4nan (commented):
can you check how sqlite behaves? we might need the same there

> This uses syntax that is recognized by both SQLite and PostGreSQL

edit: i should read the whole thing first :]

kpervin:
> can you check how sqlite behaves? we might need the same there
> 
> > This uses syntax that is recognized by both SQLite and PostGreSQL
> 
> edit: i should read the whole thing first :]

It probably would make sense to create test suites for each language. Currently the QueryBuilder.test.ts file is only testing against MySQL from what I could see.

packages/mariadb/src/MariaDbPlatform.ts
```diff
@@ -74,4 +74,22 @@ export class MariaDbPlatform extends AbstractSqlPlatform {
     return `alter table ${quotedTableName} add fulltext index ${quotedIndexName}(${quotedColumnNames.join(',')})`;
   }
 
+  private readonly NULLS_TRANSLATE = {
```
B4nan, line 77: maybe lets remove those to `ORDER_BY_NULLS_TRANSLATE`?

packages/mariadb/src/MariaDbPlatform.ts
```diff
@@ -74,4 +74,22 @@ export class MariaDbPlatform extends AbstractSqlPlatform {
...
+
+  /* istanbul ignore next */
+  override getOrderByExpression(column: string, direction: string): string[] {
+    const ret: string[] = [];
+    const dir = direction.toLowerCase();
+    if (dir in this.NULLS_TRANSLATE) {
+      ret.push(`${column} ${this.NULLS_TRANSLATE[dir as keyof typeof this.NULLS_TRANSLATE]}`);
+    }
+    ret.push(`${column} ${dir.replace(/(\s|nulls|first|last)*/gi, '')}`);
+    return ret;
```
B4nan, line 92:
```suggestion
    const dir = direction.toLowerCase() as keyof typeof this.NULLS_TRANSLATE;

    if (dir in this.NULLS_TRANSLATE) {
      ret.push(`${column} ${this.NULLS_TRANSLATE[dir]}`);
    }

    ret.push(`${column} ${dir.replace(/(\s|nulls|first|last)*/gi, '')}`);

    return ret;
```

B4nan (commented):
just two nits, please apply those suggestions to both drivers

otherwise ready to merge, thanks!

> It probably would make sense to create test suites for each language. Currently the QueryBuilder.test.ts file is only testing against MySQL from what I could see.

part of that file is also about postgres, but agreed this would deserve a split and other drivers, probably not a complete copy of what we have there but at least for the differences

packages/mariadb/src/MariaDbPlatform.ts
kpervin: Sorry, did you mean rename (not remove)?

B4nan: yes

B4nan (approved):

PR #5023, opened 2023-12-15
feat(core): allow setting logger context on EM level

### LINKED ISSUE #5022: Feature: Forked Logger
I'm using Mikro with Express with Winston as a logger (but this doesn't really matter) - when a request is received, we add context to the logger so that all subsequent logs have metadata added to it to easier tracing/filtering in Datadog.  We'd like to be able to add this same context to all Mikro logs, but since the `logger` config property is global, there's no way of adding that per request context to the logs with accuracy.  

As part of calling `RequestContext.create(orm.em, next, { ... });`, you could add `logger` as part of the `CreateContextOptions`, which would then be passed to the forked EntityManager.  There are some things that need to be global no matter what, but I was hoping that anything that could be request context specific be logged as such.

I've tried working with the global logger, logger factory, etc, but because we have many requests per seconds, they keep "competing" with each other over who gets to set the context and most of the data coming out of it is not related to the correct request.  Only way to do this would be to initialize Mikro for every request, which seems ridiculous as it would add the overhead of connecting to the DB for every request and make our API slower.

Thanks for all the work that's been done on this library, it's been very useful for us.  Cheers.

B4nan:
The logger context can now be also set on `EntityManager` level, e.g. via `em.fork()`:

```ts
const fork = em.fork({
  loggerContext: { meaningOfLife: 42 },
});
const res = await fork.findAll(Author);
```

Closes #5022

### DIFF
--- a/docs/docs/logging.md
+++ b/docs/docs/logging.md
@@ -197,10 +197,10 @@ interface LogContext extends Dictionary {
 
 ### Providing additional context to a custom logger
 
-If you have implemented your own `LoggerFactory` and need to access additional contextual values inside your customer logger implementation, utilize the `LoggerContext` property of `FindOptions`. Adding additional key/value pairs to that object will make them available inside your custom logger:
+If you have implemented your own `LoggerFactory` and need to access additional contextual values inside your customer logger implementation, utilize the `loggerContext` property of `FindOptions`. Adding additional key/value pairs to that object will make them available inside your custom logger:
 
 ```ts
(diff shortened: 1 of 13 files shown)


mboudreau:
:star_struck:

mboudreau:
When's this expected to be release?

B4nan:
Its part of v6, every commit is published to npm as a dev version, so you can try this out already. Will ship new rc version during today and stable maybe within few days.

mboudreau:
Amazing, thank you

PR #5058, opened 2023-12-28
refactor: use `flush` option with `writeFile`

boenrobot:
In all places of the code, fs-extra's writeFile is used. This package however is actually a re-export of the callback version of node:fs' writeFile.
Meanwhile, all packages seem to assume this is the version from node:fs/promises.

This fix in turn fixes tests that can sometimes be flaky, due to assuming file operations are done, when that isn't necessarily true.

Also reordered imports in the affected files alphabetically.

---

I noticed this while running tests locally. Random tests sometimes fail, and all they had in common was the fact that they were checking for a file's existence. I thought it's maybe an fs-extra specific issue with Windows, but then discovered it's not re-exporting the promise, hence this PR.

### DIFF
--- a/scripts/copy.mjs
+++ b/scripts/copy.mjs
@@ -26,7 +26,7 @@ function rewrite(path, replacer) {
   try {
     const file = readFileSync(path).toString();
     const replaced = replacer(file);
-    writeFileSync(path, replaced);
+    writeFileSync(path, replaced, { flush: true });
   } catch {
     // not found
   }
@@ -121,7 +121,7 @@ if (options.canary) {
   // eslint-disable-next-line no-console
   console.info(`canary: setting version to ${nextVersion}`);
 
-  writeFileSync(pkgPath, `${JSON.stringify(pkgJson, null, 2)}\n`);
+  writeFileSync(pkgPath, `${JSON.stringify(pkgJson, null, 2)}\n`, { flush: true });
 }
 
 if (options['pin-versions']) {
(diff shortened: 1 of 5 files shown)


B4nan:
> This package however is actually a re-export of the callback version of node:fs' writeFile.

Are you sure about that? The library itself is using it by awaiting internally:

https://github.com/jprichardson/node-fs-extra/blob/master/lib/ensure/file.js#L24

Also the types say it's the promisified version.

If it's true, it sounds like something worth reporting and fixing upstream.

FWIW, I don't think I saw a flaky test that would point to this problem, so could be surely windows specific.

B4nan:
I don't think this is true, when I console.log the method without awaiting, I get a Promise instance back.

```ts
import { writeFile } from 'fs-extra';

console.log(writeFile('foo', 'bar'));

Promise { <pending> }
```

scripts/copy.mjs
```diff
@@ -137,7 +137,7 @@ if (options['pin-versions']) {
   // eslint-disable-next-line no-console
   console.info(`pin-versions: version ${version}`, pkgJson.dependencies);
 
-  writeFileSync(pkgPath, `${JSON.stringify(pkgJson, null, 2)}\n`);
+  writeFileSync(pkgPath, `${JSON.stringify(pkgJson, null, 2)}\n`,{ flush: true });
```
B4nan, line 140:
```suggestion
  writeFileSync(pkgPath, `${JSON.stringify(pkgJson, null, 2)}\n`, { flush: true });
```

B4nan (approved):

PR #5071, opened 2024-01-04
fix(postgres): parse timestamp dates less than year 1000

Robert-Schirmer:
### Bug
Parsing a timestamp column in postgresql with a date less than year 1000 will create an invalid Date when using the `forceUtcTimezone` set to `true`.
### Reproduction
Created test for reproduction, run without the fix to `PostgresSqlConnection.ts`.
### Fix
Use the date parsing function from `postgres-date` that is used by default in `pg` for parsing dates but add the `Z` to force the utc timezone. `new Date('0022-01-01 09:17:58.000Z')` when parsing the timestamp column str creates an invalid date in js.

### DIFF
--- a/packages/postgresql/package.json
+++ b/packages/postgresql/package.json
@@ -59,7 +59,8 @@
   },
   "dependencies": {
     "@mikro-orm/knex": "5.9.7",
-    "pg": "8.11.3"
+    "pg": "8.11.3",
+    "postgres-date": "2.1.0"
   },
   "devDependencies": {
     "@mikro-orm/core": "^5.9.7"
--- a/packages/postgresql/src/PostgreSqlConnection.ts
+++ b/packages/postgresql/src/PostgreSqlConnection.ts
@@ -1,4 +1,5 @@
 import TypeOverrides from 'pg/lib/type-overrides';
+import parseDate from 'postgres-date';
 import type { Dictionary } from '@mikro-orm/core';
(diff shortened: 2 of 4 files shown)


B4nan:
please target master branch, then once we merge, we can cherry pick to 5.x

Robert-Schirmer:
> please target master branch, then once we merge, we can cherry pick to 5.x

Thats what I originally wanted to do but there seems to be a refractor of the `PostgreSqlConnection` in `master`, it will require two different fixes. How do you want to do that?

B4nan (changes requested):
few notes about the test

B4nan:
> Thats what I originally wanted to do but there seems to be a refractor of the PostgreSqlConnection in master, it will require two different fixes. How do you want to do that?

Then I'd like to first see a PR for master/v6, keep this one targeting v5. Once we have it fixed in v6, we can merge this one, but I don't want to fix things only here.

B4nan:
Btw, just curious, what kind of dates are you storing that you've hit this? :]

B4nan:
gave this a closer look, and its not just about postgres, neither it is about year 1000, just about some particular years close to 0. looks like its enough to add `T` into the datetime string instead of the space to fix this. fixing this in v6 is actually much easier than v5, as date parsing happens on a single place for all drivers. since nobody else asked about this, i am fine with merging this in for v5, and keeping the proper fix to all drivers in v6 only

PR #5111, opened 2024-01-10
Include options with nativeBinding in getKnexOptions

pg-seagic:
Added the options field for both sqlite3 and better-sqlite3. 
This give the possibility to forward the knex options which include the nativeBinding setting for knex. This is helpfull when using other bindings than the standard sqlite3 or better-sqlite3 bindings.

### DIFF
--- a/packages/sqlite/src/SqliteConnection.ts
+++ b/packages/sqlite/src/SqliteConnection.ts
@@ -35,6 +35,7 @@ export class SqliteConnection extends AbstractSqlConnection {
       client: type,
       connection: {
         filename: this.config.get('dbName'),
+        options: this.config.get('options'),
       },
       pool: this.config.get('pool'),
       useNullAsDefault: true,


packages/better-sqlite/src/BetterSqliteConnection.ts
```diff
@@ -36,6 +36,7 @@ export class BetterSqliteConnection extends AbstractSqlConnection {
       client: type,
       connection: {
         filename: this.config.get('dbName'),
+        options: this.config.get('options'),
```
B4nan, line 39: what's this supposed to do? there is no `options` key in the ORM configuration

if you want to add a new option, it needs to be first added to the `Configuration` class (and `Options` interface). and it surely cant be named `options`, thats way too generic.

do you know about `driverOptions`? it sounds like thats what you want

https://mikro-orm.io/docs/configuration#driver

B4nan: where exactly? it is in the quick start as well as in the list in the first page from getting started (as well as in the API docs)

B4nan: thanks, will add it there, let me know if you find more places where its missing

PR #5113, opened 2024-01-10
feat(entity-generator): allow post processing the metadata

### LINKED ISSUE #5010: Entity generator post processing

I understand MikroORM is a "code first" library, but the entity generator is not too far from enabling a fully "schema first" approach already.

That said, there are some things that are impossible to infer from the schema alone, and right now require further code alterations. The ones I've managed to find so far include
- Lazy loaded scalar properties
- Eager loaded relations
- Hidden properties
- STI value map

For the first one, I imagine a flag to make all scalar properties lazy could be added, but that would hinder ergonomics of using the generated code in most cases, and it's an approach that doesn't work for the other things anyway.

I'm willing to implement in a PR something that will let users guide the entity generator, so that it will in turn generate those things only where it needs to.

However, I'm not sure of the best approach, so I'm making this issue to gather what @B4nan would be willing to accept, and what others might like.

My initial thought was to NOT implement anything specific to any of the above options, and instead implement a generic "metadata post processor" mechanism.

That is, in the config, one can specify an array of callbacks that receive the `EntityMetadata[]` and can alter it however they like, before that meta is ultimately serialized into files. The SourceFile.ts would of course be amended to support any known options, including ones needed for all of the above features.

This has the obvious downside that it exposes internals to users, without stability guarantees, which may in turn give some bad user experience across versions. Depending on the how bullet proof the serialization is, it may also end up allowing users to generate invalid code, which is also kind of a bummer, even if it is technically the user's fault. For complex features, these downsides may be OK (it's the price one pays for wanting complicated things...), but for simple features that may be frustrating.

The next best solution to me seems a set of specialized callbacks that will be triggered for each qualifying item.

so f.e.

```
isHiddenCallback: (entityName: string, propertyName: string): boolean | Promise<boolean>,
isLazyColumn: (tableName: string, columnName: string): boolean | Promise<boolean>,
isEagerRelation: (foreignKey: Readonly<ForeignKey>): boolean | Promise<boolean>,
tableInheritanceValueMap: (baseTable: string, childTables: [string, ...string[]]): {discriminatorColumn: string, valueMap: Record<string/*tableName*/, string | number | bigint /*value in base table that points to this table*/
```

The upside of this is that if there is some kind of general policy that the project follows, they can define that in the logic of the callback. The downside is that this callback is needlessly triggered on all entities/properties, which will decrease performance.

But as I was writing this issue now, I also realized there's a 3rd option too...

A prop in the config that includes those amendments... e,g,

```
amendments: Record<string/*entity name*/, {entity: EntityOptions, properties: Record<string/*prop name*/, {}/*arbitrary props that are merged into the definition, and rendered if supported by the ORM*/
```

so f.e. you'd declare in the config

```
amendments: {
    Person: {
        entity: {
            discriminatorColumn: 'discr',
            discriminatorMap: { person: 'Person', employee: 'Employee' },
        },
        properties: {
          password: {hidden: true},
          bio: {lazy: true},
          favoriteBook: {eager: true}
        }
    }
}
```

The upside is that there's little extra cost to the generator (comparable to a custom post processor for the entire meta data), and that simple one off cases are easy to add. The downside is that to apply some sort of a consistent policy (e.g. all columns that contain the word "password" must be hidden), projects need to do a custom inspection of their tables/columns and generate the amendments config.

I suppose those options are not mutually exclusive, so we could have f.e. both amendments and meta data processors, where users can use whichever option fits better for their needs.

**Additional context**

boenrobot:
Closes #5010

### DIFF
--- a/docs/docs/entity-generator.md
+++ b/docs/docs/entity-generator.md
@@ -84,6 +84,8 @@ Available options:
 | `scalarPropertiesForRelations: 'never' \| 'always' \| 'smart'` | <ul><li> `'never'` (default) - Do not generate any scalar properties for columns covered by foreign key relations. This effectively forces the application to always provide the entire relation, or (if all columns in the relation are nullable) omit the entire relation.</li><li> `'always'` - Generate all scalar properties for all columns covered by foreign key relations. This enables the application to deal with code that disables foreign key checks.</li><li> `'smart'` - Only generate scalar properties for foreign key relations, where doing so is necessary to enable the management of rows where a composite foreign key is not enforced due to some columns being set to NULL. This enables the application to deal with all rows that could possibly be part of a table, even when foreign key checks are always enabled.</li></ul> |
 | `onlyPurePivotTables: boolean`                                 | By default, M:N relations are allowed to use pivot tables containing additional columns. If set to `true`, M:N relations will not be generated for such pivot tables.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        |
 | `readOnlyPivotTables: boolean`                                 | By default, M:N relations are only generated if the collection would be writable, i.e. any additional columns need to be optional and have non-unique default values. If set to `true`, also generate M:N relations even if the collection would be read only (meaning the only way to write to it is by using the pivot entity directly). Such collections will include the `persist: false` option. This setting is effectively meaningless if `onlyPurePivotTables` is set to `true`.                                                                                                                                                                                                                                                                                                                                                                     |
+| `onInitialMetadata: MetadataProcessor`                         | A function can be provided to do processing of the initially collected metadata collection. Triggered after checking for tables and columns affected by skipping, but before ManyToMany and bi-directional relations detections. This hook is the appropriate place to add virtual and embeddable entities, and any owning sides of OneToOne or ManyToOne references. The generator can then figure out the inverse side and/or if any ManyToMany relations are affected. The function may return a promise, in which case, the promise will be awaited before moving on.                                                                                                                                                                                                                                                                                    |
+| `onProcessedMetadata: MetadataProcessor`                       | A function can be provided to do processing of the final metadata collection. Triggered after ManyToMany and bi-directional relation checking, just before outputting the metadata into files. This is the appropriate place for changes that need to be universally applied, including to ManyToMany and inverse side relation properties. The function may return a promise, in which case, the promise will be awaited before moving on.                                                                                                                                                                                                                                                                                                                                                                                                                  |
 
 Example configuration:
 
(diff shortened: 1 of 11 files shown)


packages/entity-generator/src/EntityGenerator.ts
```diff
@@ -97,6 +97,11 @@ export class EntityGenerator {
 
     metadata = metadata.filter(table => !options.skipTables || !options.skipTables.includes(table.tableName));
 
+    const onInitialMetadata = options.onInitialMetadata;
+    if (typeof onInitialMetadata === 'function') {
+      await onInitialMetadata(metadata, this.platform);
+    }
```
B4nan, line 103: could be just
```suggestion
    await options.onInitialMetadata?.(metadata, this.platform);
```

B4nan (commented):
gave it just a quick look now, but its looking pretty good, thanks!

docs/docs/entity-generator.md
```diff
@@ -68,22 +68,24 @@ The behaviour of the entity generator can be adjusted either via `entityGenerato
...
+| `onInitialMetadata: (metadata: EntityMetadata[], platform: Platform) => MaybePromise<void>`   | A function can be provided to do processing of the initially collected metadata collection. Triggered after checking for tables and columns affected by skipping, but before ManyToMany and bi-directional relations detections. This hook is the appropriate place to add virtual and embeddable entities, and any owning sides of OneToOne or ManyToOne references. The generator can then figure out the inverse side and/or if any ManyToMany relations are affected. The function may return a promise, in which case, the promise will be awaited before moving on.                                                                                                                                                                                                                                                                                    |
```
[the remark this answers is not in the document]
boenrobot: I've introduced a separate type, and referenced that in both the code and the docs. It being exported allows people who use this feature to easily type check their function for compatibility.

But more to your point, it also shortens the first column's length in the docs.

docs/docs/entity-generator.md
```diff
@@ -84,6 +84,8 @@ Available options:
...
+| `onInitialMetadata: EntityGeneratorMetadataProcessor`          | A function can be provided to do processing of the initially collected metadata collection. Triggered after checking for tables and columns affected by skipping, but before ManyToMany and bi-directional relations detections. This hook is the appropriate place to add virtual and embeddable entities, and any owning sides of OneToOne or ManyToOne references. The generator can then figure out the inverse side and/or if any ManyToMany relations are affected. The function may return a promise, in which case, the promise will be awaited before moving on.                                                                                                                                                                                                                                                                                    |
```
B4nan, line 87: two more general notes (sorry, dont have time to review the code yet)

- will ship this in patch release, so those changes need to be in the 6.0 docs snapshot too
- this would probably deserve a complete example, maybe a recipe

boenrobot: Copied over the docs in the 6.0 versioned docs.

I'm not entirely sure what steps are needed to properly have a new recipe listed (in addition to writing the new md file that is)... Is it ok if I just add a section in the entity generator pages?

B4nan: Yeah, I'd just add a section to the page now, later on we can add a schema first approach guide somewhere, but we are not there yet I guess.

(otherwise its just about adding new MD file and adding it to the sidebars.js - which for the recipe folder works automatically)

B4nan:
> And on a related note, is there any metadata piece that can hold the entity config? I couldn't find one, but I'm thinking the new Config symbol and the forceObject option may be a useful thing to manually add on in those hooks. Right now, you can not.

There is no point in attaching that to the entity metadata, this is a global configuration option.

B4nan:
Btw there is also `EagerProps`, that one does make sense only for using `Ref` wrapper on `eager: true` relation. I tried to provide `Eager` type that would work the same as `Hidden` and `Opt` but that unfortunately didn't work (it broke the inference on various places). I guess we don't need to deal with that right now.

B4nan:
> But... then again, if that's the case, that kind of defeats the purpose of the Config symbol in a entity, even if it's a custom base one.

How come? There is no other way than this (correct me if I'm wrong), the serialization only has the entity type, so this information needs to be present on it on type level to it can be respected.

B4nan (changes requested):
few nits

B4nan:
Yes, result cache tests are flaky on slower machines (especially CI), and `Promise.all` can result in the same. There is not much we can do about them I think, not without making them "slow by default" (which I don't want to).

packages/core/src/typings.ts
```diff
@@ -1151,3 +1153,5 @@ export interface Seeder<T extends Dictionary = Dictionary> {
 }
 
 export type ConnectionType = 'read' | 'write';
+
+export type EntityGeneratorMetadataProcessor = (metadata: EntityMetadata[], platform: Platform) => MaybePromise<void>;
```
B4nan, line 1157: i think we could call this just `MetadataProcessor` or `MetadataHook` (that would be probably more in line with the rest)

boenrobot: But these callbacks have different signatures from `onMetadata` (they accept the array, rather than an individual meta data object, in order to enable metadata additions/removals), and a generic name like those ones would speak to them being used outside of the generator, like in `onMetadata`.

Or are you thinking of maybe later changing `onMetadata` to also take in an array?

B4nan: > But these callbacks have different signatures from onMetadata 

I am talking about the type name, how is this relevant? 

> would speak to them being used outside of the generator

The type as is can be used elsewhere too, there is nothing in the signature that would point to entity generator.

B4nan:
> so yes, "[query]" prefix again... but also some second row being queued, which shouldn't be, according to the expected output.

It's not about the prefix, we usually use `toMatch()` for query assertion, and the expected part is just the query itself, without prefix (the `[query]` part and suffix (time took).

This is exactly the flakyness I am well aware and can't do much, queries are sometimes not batched when using `Promise.all` - those tests are kinda destined to be flaky. Maybe we could remove the query assertions from those. It has nothing to do with other tests doing something, it's about the load.

packages/core/src/typings.ts
boenrobot: > > But these callbacks have different signatures from onMetadata
> 
> I am talking about the type name, how is this relevant?
> 
> > would speak to them being used outside of the generator
> 
> The type as is can be used elsewhere too, there is nothing in the signature that would point to entity generator.

I'm talking about the social / human / layman reading this out of context aspect.

Technically, yes, this is just a type name, yes, it can be used elsewhere. There's no technical reason to use one or the other.

But without digging into souce, if you're not already familiar with the prototypes, if you enter a code base that uses those and they have imported this name... In that context, the name matters for the human. And as a human in that position, I would think the generic name is used also in onMetadata, and I would be surprised of the build failure until I check the source.

But anyhow, ok, I'll fix this in a few hours (when I'm at a computer...).

boenrobot: Done. Changed to `MetadataProcessor`.

B4nan:
I have only some code style issues I can deal with myself after merging, overall looks great, thanks!

PR #5115, opened 2024-01-10
docs: update remaining `uuid_generate_v4()` calls to `gen_random_uuid()`

stefansundin:
Since https://github.com/mikro-orm/mikro-orm/pull/4408 the docs have been in a weird inconsistent state regarding the uuid functions, recommending one in the section header but then using another one in the example code.

`uuid_generate_v4()` is still used in a few tests but I am not updating that here.


B4nan (approved):
thanks!

PR #5132, opened 2024-01-15
test: try to fix flaky `Promise.all()` test

boenrobot:
Fix by explicitly checking on every tick if the entity with the previous ID is loaded into the entity manager.
Thus, while the promises are technically all started at the same time, they are resolved in a forced order, making the actual output of the query log consistent.

---

A potential solution to the situation described [here](https://github.com/mikro-orm/mikro-orm/pull/5113#issuecomment-1890953639).


B4nan (approved):
interesting that this actually works, as by calling `em.getReference()` you add the entity to the identity map... 

let's see, worst case we can revert it. in the end its not that important to assert those queries, its important that it works.

PR #5140, opened 2024-01-16
test(timers): replaced all timeouts with fake timers, or remove them as appropriate

boenrobot:
Also tweaked the test entities-js/Author3.js and entities-schema/BaseEntity5.ts to derive the update out of the create or (in the case of the entity schema) vice-versa. This refactor, combined with fake timers in the affected tests, means that no tolerance is required for those columns - they start off as equal, always.

### DIFF
--- a/tests/entities-schema/BaseEntity5.ts
+++ b/tests/entities-schema/BaseEntity5.ts
@@ -13,7 +13,7 @@ export const BaseEntity5 = new EntitySchema<IBaseEntity5>({
   abstract: true,
   properties: {
     id: { type: 'number', primary: true },
-    createdAt: { type: 'Date', onCreate: () => new Date(), nullable: true },
-    updatedAt: { type: 'Date', onCreate: () => new Date(), onUpdate: () => new Date(), nullable: true },
+    createdAt: { type: 'Date', onCreate: owner => (owner.updatedAt ? new Date(owner.updatedAt) : new Date()), nullable: true },
+    updatedAt: { type: 'Date', onCreate: owner => (owner.createdAt ? new Date(owner.createdAt) : new Date()), onUpdate: () => new Date(), nullable: true },
   },
 });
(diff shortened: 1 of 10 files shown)


B4nan:
Amazing, thank you!

> Also tweaked the test entities-js/Author3.js and entities-schema/BaseEntity5.ts to derive the update out of the create or (in the case of the entity schema) vice-versa. This refactor, combined with fake timers in the affected tests, means that no tolerance is required for those columns - they start off as equal, always.

FYI the problem with that case was not that the initial value would be different, it was about `onUpdate` being triggered and the value recomputed during flush. Not sure if the problem won't be there anymore with the fake timers, I'd guess it's still valid.

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/unit-of-work/ChangeSetComputer.ts#L90

tests/entities-schema/BaseEntity5.ts
```diff
@@ -13,7 +13,7 @@ export const BaseEntity5 = new EntitySchema<IBaseEntity5>({
   abstract: true,
   properties: {
     id: { type: 'number', primary: true },
-    createdAt: { type: 'Date', onCreate: () => new Date(), nullable: true },
-    updatedAt: { type: 'Date', onCreate: () => new Date(), onUpdate: () => new Date(), nullable: true },
+    createdAt: { type: 'Date', onCreate: owner => (owner.updatedAt ? new Date(owner.updatedAt) : new Date()), nullable: true },
+    updatedAt: { type: 'Date', onCreate: owner => (owner.createdAt ? new Date(owner.createdAt) : new Date()), onUpdate: () => new Date(), nullable: true },
```
B4nan, line 17: FYI I will revert this, it actually makes some tests flaky (namely [this one](https://github.com/mikro-orm/mikro-orm/blob/master/tests/EntityManager.sqlite2.test.ts#L1024))

as discussed in the comments, this wasn't the problem causing flakiness in the first place

B4nan: hmm it made things better, but i saw the same happening after i reverted, so its not just about that. but before it was reproducing very often, after the revert i saw it only once.

PR #5157, opened 2024-01-19
docs(fake-timers): added a page for using MikroORM with Jest, with fake timers info

### LINKED ISSUE #4994: docs: convey that `jest.useFakeTimers()` can break MikroORM without specifying `doNotFake: ["nextTick"]`
In using MikroORM in tests, sometimes we need to fake timers. However, when using `jest.useFakeTimers()` without `doNotFake: ["nextTick"]`, MikroORM hangs indefinitely and times out the test. 

[Example](https://github.com/kpervin/MikroORM-Issue-Example-Repo/blob/bug/jest-fake-timers/test/fake-timers.spec.ts)
```ts
import { MikroORM } from "@mikro-orm/core";
import { MikroOrmModule } from "@mikro-orm/nestjs";
import { Test, TestingModule } from "@nestjs/testing";
import { Bar } from "../src/database/foo/entities/bar.entity";
import config from "../src/mikro-orm.config";

describe("Fake Timers issue", () => {
  let orm: MikroORM;

  beforeEach(async () => {
    const module: TestingModule = await Test.createTestingModule({
      imports: [
        MikroOrmModule.forRoot({
          ...config,
          allowGlobalContext: true,
        }),
      ],
    }).compile();

    orm = await module.resolve(MikroORM);
  });

  afterEach(async () => {
    await orm.close();
    jest.useRealTimers();
  });

  async function insertBar(): Promise<boolean> {
    const bar = orm.em.create(Bar, {});
    await orm.em.persistAndFlush(bar);
    return true;
  }

  it("will timeout", async () => {
    jest.useFakeTimers();
    /**
     * This will exceed timeout because it never resolves
     */
    await expect(insertBar()).resolves.toEqual(true);
  });

  it("should insert", async () => {
    jest.useFakeTimers({ doNotFake: ["nextTick"] });
    await expect(insertBar()).resolves.toEqual(true);
  });
});
```

Please add to the documentation that using `jest.useFakeTimers()` can break functionality.

boenrobot:
Closes #4994.

---

I know what I said before about `glob` on init, but... I tried to locate the exact source, and couldn't find it now... I'm guessing maybe they changed in some recent versions, and maybe I was looking at an outdated one at the time. Anyhow, I haven't included it in the docs due to that.

The code shown for a workaround is a more polished version of what I showed in #4994, and after some research, I found a workaround for Mongo that I also included.

---

EDIT: Workaround provided is probably a bit too hacky... I'm researching something more elegant and less memory leak-y.

---

EDIT2: OK, I think I have a better workarounds for each driver now, targeting specifically the things that absolutely need a real `process.nextTick()` and nothing else.

### DIFF
--- a/docs/docs/usage-with-jest.md
+++ b/docs/docs/usage-with-jest.md
@@ -1,0 +1,197 @@
+---
+title: Usage with Jest
+---
+
+When testing your own code that uses [Jest](https://jestjs.io/) for the tests, and MikroORM for the database, there are some things to consider.
+
+## Using fake timers
+
+Jest allows using [timer mocks](https://jestjs.io/docs/timer-mocks), which can be a very useful feature when testing your application's logic for time sensitive tasks. Making sure the data in your database is accurate based on time is also part of this testing.
+
+One caveat to be aware of though is that one of the functions faked by Jest is [`process.nextTick()`](https://nodejs.org/docs/latest/api/process.html#processnexttickcallback-args), and this function is in turn used by some of MikroORM's dependencies.
+
+If you know your code is not sensitive to the passage of the event loop's ticks (and is instead maybe only sensitive to the system's clock), you can safely set Jest to not fake this function by using:
+
+```typescript
+jest.useFakeTimers({ doNotFake: ['nextTick'] });
+```
+
+This way, you can still use the rest of Jest's timer mock API to control the system's clock, and any timers in use by your code or in MikroORM (most notably, result caches).
+
+If you need better control in your test code in relation to the micro-task queue, keep reading the rest of this section.
+
+### Known usages of `process.nextTick()` by dependencies
+
+- All DB clients that use connection pool do so by an algorithm that lets them obtain the connection they requested at the next tick, or be queued up if no connection is free at the time of the request. Similarly, on release, if there's a request in the queue, it gets scheduled to get the released connection at the next tick, or else the connection is put back in the pool.
+- The MongoDB client has no option of not using connection pool. Even if you set a pool of one connection, the client will still reach for that pool, thus needing `process.nextTick()`.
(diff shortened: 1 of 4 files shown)


B4nan:
> Not necessarily...

Even if it's not the case, I still want one testing page in docs, not multiple recipes for different frameworks, as each has its own quirks and I want to have this documented at one place (later on, when we have more content, we can have more pages in a dedicated section).

> I'm not sure what to include in such a section that wouldn't already be in the getting started guide of the test framework or the initial setup of MikroORM already.

Vitest requires [additional setup for folder-based discovery to work](https://mikro-orm.io/docs/guide/project-setup#testing-the-endpoint), that's one thing, and later on even more (e.g. mocking). Getting started guide is not supposed to be the single source of truth, it already duplicates most of the information that is in there. 

And there are more things we should add there going forward, like the required metadata discovery for working with entity classes (so propagation can work). People often ask for some suggestions around testing, we need to have a place to put them into.

> This has a better chance to be kept up to date if someone more keen would take it on, 

Well, by merging it into master, you are making it my responsibility. I get what you mean, but realistically, if you as its author won't be the one who maintains it, it will likely get outdated in the long run anyway. But let's not discuss that further, I am fine merging it in, it's just weird to have that in the docs as it's pretty complex code on its own.

> and better likelyhood to be discovered and actually used if it's part of this monorepo or at least under this organization.

That's not really an argument, the package would be linked and used in the examples in the docs here, making it discoverable the same way. I'd argue it's the opposite, if you turn it into a package, its no longer "some piece of code in MikroORM docs" but something universal people can use with other ORMs too - as the problem is not directly connected to the ORM but to its dependencies.

B4nan:
I'll move things around once we have the dedicated testing section, let's move this forward as is. 

Thanks again!

PR #5180, opened 2024-01-29
fix(core): ensure `em.insertMany` returns array of PKs for a single item too

### LINKED ISSUE #5179: v6.0.5 insertMany
### Describe the bug

when i insert only one record, because i expecting 1 or more items i use insertMany, but when i insert only one, then returned ids insertIds are undefined and it should return insertedId from response in array
```{ affectedRows: 1, insertId: 1, row: { id: 1 }, rows: [ { id: 1 } ] }```

### Reproduction

// Works as a workaround
await em.insertMany(
          Test,
          [{
            test: "test"
          }],
        );

@mikro-orm/postgresql

### MikroORM version

6.0.5

### Node.js version

v18.18.2

### Operating system

mac

### Validations

### PR DESCRIPTION
**Pull request by tkudlicka**
Closes #5179

### DIFF
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -1410,7 +1410,10 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
     data.forEach(row => em.validator.validateParams(row, 'insert data'));
     const res = await em.driver.nativeInsertMany<Entity>(entityName, data as EntityData<Entity>[], { ctx: em.transactionContext, ...options });
 
-    return res.insertedIds!;
+    if(!res.insertedIds) {
+      return [res.insertId];
+    }
+    return res.insertedIds;
   }
 
   /**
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/core/src/EntityManager.ts [hunk 1]
```diff
@@ -1410,7 +1410,10 @@ export class EntityManager<D extends IDatabaseDriver = IDatabaseDriver> {
     data.forEach(row => em.validator.validateParams(row, 'insert data'));
     const res = await em.driver.nativeInsertMany<Entity>(entityName, data as EntityData<Entity>[], { ctx: em.transactionContext, ...options });
 
-    return res.insertedIds!;
+    if(!res.insertedIds) {
+      return [res.insertId];
+    }
+    return res.insertedIds;
```
B4nan, line 1416:
```suggestion
    if (res.insertedIds) {
      return res.insertedIds;
    }

    return [res.insertId];
```

PR #5212, opened 2024-02-06
refactor(entity-generator): emit `Opt` and `Hidden` prop types instead of the symbols

boenrobot:
As mentioned we should do in #5113 , but left it for a separate PR due to the increase in scope.

Pretty much all entity generator snapshots are affected, since most schemas have some form of an optional property involved.

I also adjusted the meta data post processing test to have one property that is both hidden and optional, to ensure the combination of the two is outputted fine too.

### DIFF
--- a/tests/features/entity-generator/__snapshots__/MetadataHooks.mysql.test.ts.snap
+++ b/tests/features/entity-generator/__snapshots__/MetadataHooks.mysql.test.ts.snap
@@ -18,7 +18,7 @@ export class Address2 {
 
 }
 ",
-  "import { Cascade, Collection, EagerProps, Embedded, Entity, HiddenProps, Index, ManyToMany, ManyToOne, OneToMany, OptionalProps, PrimaryKey, Property, Unique } from '@mikro-orm/core';
+  "import { Cascade, Collection, EagerProps, Embedded, Entity, Hidden, Index, ManyToMany, ManyToOne, OneToMany, Opt, PrimaryKey, Property, Unique } from '@mikro-orm/core';
 import { Book2 } from './Book2';
 import { IdentitiesContainer } from './IdentitiesContainer';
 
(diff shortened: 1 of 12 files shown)


B4nan (approved):
thanks!

tests/features/entity-generator/__snapshots__/MetadataHooks.mysql.test.ts.snap
```diff
@@ -82,7 +78,7 @@ export class Author2 {
   identity?: IdentitiesContainer[];
 
   @ManyToMany({ entity: () => Author2, pivotTable: 'author_to_friend', joinColumn: 'author2_1_id', inverseJoinColumn: 'author2_2_id', hidden: true })
-  authorToFriend = new Collection<Author2>(this);
+  authorToFriendCollection<Author2> & Hidden = new Collection<Author2>(this);
```
B4nan, line 81: this is missing a space

boenrobot: Oops. Fixed.

B4nan:
I think its more about you removing code that was covered, so the total project coverage decreased tiny bit.

B4nan:
Used the token [here](https://github.com/mikro-orm/mikro-orm/commit/f1c0058b8e2f40fbde0def77fdcbc4a052560c04?w=1) (we had it in the secrets for last 3 years, I hope its still valid)

B4nan:
> But now that I think about it, I can see why it would be more convenient for you too to not have me in a fork... because it makes edits in my PRs easier on you. So, ok, I'll use this repo instead of the fork in the future.

Yes, that's it, it makes quite a huge difference for me, as I can switch to your branch and commit there directly. Doing that with a fork is 100x more cumbersome :]

PR #5217, opened 2024-02-07
feat(entity-generator): support functions in extension hooks

boenrobot:
If a function is provided, its toString() representation (its source) will be put in the generated file.
Support spots for functions include
- onCreate
- onUpdate
- serializer
- expression in virtual entities

Also added support for the `@Formula` decorator (used whenever a "formula" option is included).

Simplified meta data edits to not use quotes in places where only string types are expected, such as comment, expression (as string) and STI values.

----

btw, I tried to push this branch to this repo instead of my fork, but I got 403. Seems I have no permissions to push new branches (or at all?) 😕


B4nan:
You need to accept the invitation first.

PR #5218, opened 2024-02-07
feat(entity-generator): support functions in extension hooks

boenrobot:
If a function is provided, its toString() representation (its source) will be put in the generated file.
Support spots for functions include
- onCreate
- onUpdate
- serializer
- expression in virtual entities

Also added support for the `@Formula` decorator (used whenever a "formula" option is included).

Simplified meta data edits to not use quotes in places where only string types are expected, such as comment, expression (as string) and STI values.

### DIFF
--- a/packages/entity-generator/src/SourceFile.ts
+++ b/packages/entity-generator/src/SourceFile.ts
@@ -35,12 +35,14 @@ export class SourceFile {
 
   generate(): string {
     let ret = '';
-    if (this.meta.embeddable) {
-      this.coreImports.add('Embeddable');
-      ret += `@Embeddable(${this.getEmbeddableDeclOptions()})\n`;
-    } else {
-      this.coreImports.add('Entity');
-      ret += `@Entity(${this.getEntityDeclOptions()})\n`;
+    if (this.meta.embeddable || this.meta.collection) {
+      if (this.meta.embeddable) {
+        this.coreImports.add('Embeddable');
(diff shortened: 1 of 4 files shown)


packages/entity-generator/src/SourceFile.ts
```diff
@@ -35,12 +35,14 @@ export class SourceFile {
 
   generate(): string {
     let ret = '';
-    if (this.meta.embeddable) {
-      this.coreImports.add('Embeddable');
-      ret += `@Embeddable(${this.getEmbeddableDeclOptions()})\n`;
-    } else {
-      this.coreImports.add('Entity');
-      ret += `@Entity(${this.getEntityDeclOptions()})\n`;
+    if (this.meta.embeddable || this.meta.collection) {
```
B4nan, line 38: why is that, base entities? they can technically have the decorator, they just need to set `abstract: true` in its options

boenrobot: Yes, base entities... I thought I lay the ground for the base entity that includes the `[Config]` symbol.

It seems to me like such classes don't have to be abstract, nor do they have to be annotated... just as long as entities inherit from them. And I wanted to enable extensions to potentially define classes in that way, even if by default, the generator would never end up doing that.

B4nan: > It seems to me like such classes don't have to be abstract

If the decorator is not there, it literally means they are abstract (as in `abstract: true`, this is not about the TS keyword).

PR #5220, opened 2024-02-07
refactor(docs): upgraded doc related dependencies

boenrobot:
Also fixed non TypeDoc related broken links that are reported.

----

I noticed builds are taking forever due to the docs taking ages to build, so this is my attempt at improving the situation there.


docs/docs/configuration.md
```diff
@@ -12,7 +12,7 @@ MikroORM.init({
 });
 ```
 
-We can also use folder based discovery by providing list of paths to the entities we want to discover (globs are supported as well). This way we also need to specify `entitiesTs`, where we point the paths to the TS source files instead of the JS compiled files (see more at [Metadata Providers](metadata-providers.md)).
+We can also use folder based discovery by providing list of paths to the entities we want to discover (globs are supported as well). This way we also need to specify `entitiesTs`, where we point the paths to the TS source files instead of the JS compiled files (see more at [Metadata Providers](metadata-providers)).
```
B4nan, line 15: those were broken links? i hope not, as i've got autocomplete on such links in my IDE

B4nan: if this change is not fixing anything (i dont think it is), i'd rather revert it

docs/docusaurus.config.js
```diff
@@ -33,6 +33,10 @@ module.exports = {
   projectName: 'mikro-orm',
   scripts: ['/js/custom.js'],
   trailingSlash: false,
+  onBrokenAnchors: 'ignore',
+  onBrokenLinks: 'ignore',
+  onBrokenMarkdownLinks: 'ignore',
+  onDuplicateRoutes: 'ignore',
```
B4nan, line 39: i really dont want to ignore broken links, if this is the reason why things sped up, i can see why and i am not interested in doing that

docs/versioned_docs/version-2.7/api-typedoc.json
```diff
@@ -33233,5 +33233,6 @@
 			"line": 1,
 			"character": 0
 		}
-	]
-}
\ No newline at end of file
+	],
+	"symbolIdMap": {}
```
B4nan, line 33237: i would rather not touch those old docs snapshots at all

docs/docs/configuration.md
boenrobot: They are reported as broken by docusaurus at least, though they are rendered correctly at the end, so I am guessing they are doing this analysis too early in the flow.

Yes, I did do search and replace for ".md#" and ".md)". The report doesn't include any new broken links after that replace, and they do still function fine in the end. Also in my IDE, except the links in the old version's index.md.

All the old broken links are gone, except one which is hard coded as HTML "a" rather than markdown, which is probably why it doesn't get picked up.

Also typedoc related links are reported as broken, despite working and having the expected code be there in the output. Again, I suspect this is due to this link analysis being done too early by Docusaurus.

docs/versioned_docs/version-2.7/api-typedoc.json
boenrobot: I didn't want to either, but the projects fail to build without this key being present, due to the new version of TypeDoc.

docs/docusaurus.config.js
boenrobot: Fair. I was just testing the diff on CI. I'll remove this shortly.

Locally, skipping this only improved my time by a few seconds... not much in the grand scheme of things. Upgrading all the deps made the biggest difference.

B4nan: it's weird that they were outdated in the first place. i can understand the typedoc plugin (and therefor typedoc), but docusaurus should have been updated automatically. or was it just about typedoc?

docs/versioned_docs/version-2.7/api-typedoc.json
B4nan: i guess its time to (re)move those anyway, i think it makes sense to keep v4+, maybe v5+ would be enough too. ideally they would be rendered separately. i already have this https://github.com/mikro-orm/docs for old docs snapshots, so i can try to move the old ones there

docs/docs/configuration.md
B4nan: i think the links can surely be file paths, and i would rather keep using that. if this was reported as wrong link, correct fix is not removing the extension blindly, but adding `./` in front of the link to make it properly detected as a file path (which it always was)

docs/docusaurus.config.js
boenrobot: I think due to the patch of `docusaurus-plugin-typedoc-api`, the update bots can't pick that one up. Its new version carries a new version of TypeDoc, which has all the related improvements.

And on that note... The patch I thought was required, as builds initially broke, but after doing the rest of the fixes, it seems the patch is not really needed. I've now removed it... update bots should be able to pick up new versions now.

B4nan: I think it's no longer needed because you regenerated the typedoc definition files. Which is something pretty dangerous, as the code that was used to generate old version was using old TS, and its not guaranteed to work correctly with the current TS.

So let's remove it if the build works now.

docs/docs/configuration.md
boenrobot: ok, adding "./" in front of every .md link worked to not make Docusaurus complain about those links. Now changed to that.

docs/docusaurus.config.js
boenrobot: I didn't regenerate the typedoc definition files... I manually altered them to add the "symbolIdMap" key.

And after some more investigation, it seem the patch was also needed due to another typedoc update... getSignature is a single object, not an array in newer versions...

I've manually edited the old typedoc version files to use objects instead of arrays, and thanks to that, the patch is no longer needed.

B4nan (approved):

PR #5231, opened 2024-02-10
test(docs): made the tests workflow build a more restricted set of versions

boenrobot:
Only build the earliest and latest versions, plus the current docs, and any version mentioned in the footer (currently 5.9).

This shaves off around 2 minutes from any build,
while still ensuring that no breaking changes are accidentally introduced.

Full deploys of the docs (the docs workflow) still builds all versions, as would any local build without the newly introduced "MIKRO_ORM_DOCS_TESTING" environment variable.

---

I wasn't sure whether to include this in my last PR, so here it is separately. IMO, the trade off of not building all versions is worth it.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -78,6 +78,7 @@ jobs:
         run: |
           cd docs
           yarn
+          export MIKRO_ORM_DOCS_TESTING=1
           yarn build
 
   lint:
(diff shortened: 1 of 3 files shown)


B4nan (approved):
Nice, too bas the tests still take so long, locally its like 90s on m1 pro, but the ci is super slow due to low concurrency.

PR #5232, opened 2024-02-10
feat(entity-generator): add the ability to use custom and/or core base entity

boenrobot:
If the option for a custom base entity is set, it adds the Config symbol to it.
The two new options can be combined to make all entities inherit from the custom base entity,
while the custom base entity will inherit from the core.

Users who with to not have the Config symbol can achieve that effect
by using the metadata processing hooks to manually add a custom base entity
and extend all entities with it.

Also fixed ESM imports when using entity schema.


B4nan (approved):
looking good, thanks!

PR #5241, opened 2024-02-12
feat(entity-generator): support `mapToPk` option

boenrobot:
As a nice side effect, relations in metadata hooks now feature information about the types of the columns involved in FKs.


B4nan (approved):
thanks!

PR #5248, opened 2024-02-16
docs: updated dependencies

boenrobot:
Also, now that no broken links are reported, modified the config to throw if any are introduced.

Note that the version of @mdx-js/react is fixed to 3.0.0, as 3.0.1 seems to break the anchor collection in docusaurus' mdx-loader. Not sure if docusaurus is interacting with it erroniously or if @mdx-js/react unintentionally broke what used to work, or if maybe one of @mdx-js/react's dependencies broke something, so I haven't reported this to either repo yet.

----

Not sure why renovate still didn't pick up the new plugin version. My best guess is because it's not part of the root's workspace. If successfully enabled, it might try to update @mdx-js/react, and right now, it will probably fail due to the new version introducing broken anchors.

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -98,6 +98,10 @@ module.exports = {
   projectName: 'mikro-orm',
   scripts: ['/js/custom.js'],
   trailingSlash: false,
+  onBrokenLinks: 'throw',
+  onBrokenAnchors: 'throw',
+  onBrokenMarkdownLinks: 'throw',
+  onDuplicateRoutes: 'throw',
   themeConfig: {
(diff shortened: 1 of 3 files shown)


B4nan:
Thanks!

> Not sure why renovate still didn't pick up the new plugin version

Weird, renovate used to be working in the docs folder too, looking at the lock file history, it stopped [~10 months ago](https://github.com/mikro-orm/mikro-orm/commit/348e5d24ff3d9bb90dafecc7da7563c99773616b?w=1)

B4nan:
ok, looks like this unblocked it somehow

#5249

PR #5252, opened 2024-02-16
fix(entity-generator): use `Ref` wrapper on all lazy properties

boenrobot:
For convenience in the common case, when the identifiedReferences is set to true, lazy relations should also be wrapped by Ref,
thus simplifying what onInitialMetadata needs to do.

There should be no change when identifiedReferences is set to false.

Metadata hooks can selectively opt out of some Ref wrappers by setting Ref to false in onProcessedMetadata,
or ensure they never get Ref wrappers regardless of identifiedReferences by only adding lazy flags in onProcessedMetadata.

Since the docs are also touched anyway to clarify this behavior, listings about the recently added features are also added.


packages/entity-generator/src/SourceFile.ts
```diff
@@ -204,7 +204,9 @@ export class SourceFile {
 
     if (prop.ref) {
       this.coreImports.add('Ref');
-      this.entityImports.add(propType);
+      if (typeof prop.kind === 'string' && prop.kind !== ReferenceKind.SCALAR) {
```
B4nan, line 207: whats the point of `typeof prop.kind === 'string'`? it should be always true

boenrobot: Not necessarily if you add a custom entity meta in the hook. At least on a type level, it's not enforced.

Granted hooks need to know better, so we _could_ not do this check and just crash, but when trying to write tests, crashes for things like this did trip me up, so I do believe it will be convenient for users to not have to explicitly set this.

B4nan (approved):
otherwise lgtm

packages/entity-generator/src/SourceFile.ts
B4nan: Can you provide an example where this wouldn't be set? We normalize those things internally through `EntitySchema` class during discovery. If not for this code path, we could/should.

boenrobot: Pretty much most places in the tests where properties are added, such as [https://github.com/mikro-orm/mikro-orm/blob/e49dacfa6a4e965cb73d0cfcf28af1f30e13408a/tests/features/entity-generator/MetadataHooks.mysql.test.ts#L32) the line about adding a property with a `formula`.

PR #5254, opened 2024-02-18
fix(entity-generator): fixed generation of unsigned columns

boenrobot:


packages/entity-generator/src/SourceFile.ts
```diff
@@ -498,6 +498,9 @@ export class SourceFile {
       assign('precision');
       assign('scale');
     }
+    if (prop.unsigned && this.platform.supportsUnsigned()) {
+      assign('unsigned');
+    }
```
B4nan, line 503: this is a general rule, all control statements like if/for/while should have blank lines around for better readability

```suggestion

    if (prop.unsigned && this.platform.supportsUnsigned()) {
      assign('unsigned');
    }

```

it's not super important, but since you are contributing more and more (and the rest of the codebase is written this way), please try to adhere to this

tests/features/entity-generator/__snapshots__/AmbiguousFks.mysql.test.ts.snap
```diff
@@ -11,7 +11,7 @@ export class Colors {
 
   [PrimaryKeyProp]?: 'colorId';
 
-  @PrimaryKey()
+  @PrimaryKey({ unsigned: true })
```
B4nan, line 14: this shouldnt be here, PKs are marked as unsigned automatically, same for FKs (as that works based on the target PK), and it clearly makes no sense for `persist: false` properties

over all, i dont see many cases in the changed snapshots that would be required (havent found any but havent looked too deep)

similarly the `autoincrement: true` on `persist: false` makes no sense either

tests/features/entity-generator/__snapshots__/AmbiguousFks.mysql.test.ts.snap
```diff
@@ -75,7 +75,7 @@ export class ProductCountries {
   product!: ProductSizes;
 
   @Index({ name: 'fk_product_countries_product_sizes1_idx' })
-  @Property({ persist: false })
+  @Property({ unsigned: true, persist: false })
```
B4nan, line 78: this is actually also wrong, the index decorator should be above the actual PK (which is the FK), not the virtual property

B4nan (changes requested):

tests/features/entity-generator/__snapshots__/AmbiguousFks.mysql.test.ts.snap
boenrobot: > this shouldnt be here, [PKs are marked as unsigned automatically](https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/metadata/MetadataDiscovery.ts#L1433),

Oh. I never realized that's the default for numeric PKs. It's not documented anywhere, and it's not what MySQL does by default, so that's surprising.

But OK, I've now changed the generator to reflect that.

Also, now that you pointed me to the specific code, I realize the way the generator detects numeric columns is not consistent with that discovery, so I modified it to use the same call... which showed me that TinyInt is not detected as a numeric type, so I've added it in the list of numeric types. All non- entity generator tests seem to pass without modifications even with that change, so this tells me that type should've been there... right?

> similarly the autoincrement: true on persist: false makes no sense either

I'm guessing by the same token, all other DB related things are not useful either, correct? Only the nullability maybe... I've made the scalar options return early for persist: false scalars, to avoid all excess, and on the way there, I also realized the hydrate and trackChanges options are not serialized (hooks can add them...) so I now added those too.

tests/features/entity-generator/__snapshots__/AmbiguousFks.mysql.test.ts.snap
boenrobot: Hm, may we leave fixing this for a separate PR if it doesn't break anything for it to be under the wrong property?

The scope of this PR already went a bit bigger than originally expected right now, and I'm guessing that placing the index decorator into the correct spot will involve some less trivial code changes, in addition to more changes in the snapshots.

B4nan: Yeah sure, the index issue can be resolved separately as it's not introduced by this PR.

packages/entity-generator/src/SourceFile.ts
```diff
@@ -461,25 +461,34 @@ export class SourceFile {
...
+    const columnTypeFromMappedColumnType = mappedTypeFromColumnType.getColumnType(
+      { ...prop, autoincrement: false },
+      this.platform,
+    );
 
-    if (columnType1 !== columnType2 || [mappedType1, mappedType2].some(t => t instanceof UnknownType)) {
-      options.columnType = this.quote(columnType2);
+    if (columnTypeFromMappedPropType !== columnTypeFromMappedColumnType ||
+      [mappedTypeFromPropType, mappedTypeFromColumnType].some(t => t instanceof UnknownType)
+    ) {
```
B4nan, line 490:
```suggestion
    if (
      columnTypeFromMappedPropType !== columnTypeFromMappedColumnType
      || [mappedTypeFromPropType, mappedTypeFromColumnType].some(t => t instanceof UnknownType)
    ) {
```

B4nan (approved):
thanks!

PR #5264, opened 2024-02-20
fix(entity-generator): optional and hidden properties get type option + string defaults

### LINKED ISSUE #5260: property: Date & Opt; translates to varchar in migration
### Describe the bug

Hello.

I am using `@mikro-orm/entity-generator` with `@mikro-orm/migrations` to create a small program to generate the migrations at runtime and update the database according to the output of the entity-generator schemas. 

The databases are not all equal and this is a way that I found to update them without having to do it manually since we have a lot of changes.

When I run the entity-generator it creates the date properties as expected but it also adds a `lenght: 0` option and the Opt type. Check both nullable and not nullable property expressions bellow:
`@Property({ length: 0, defaultRaw: 'CURRENT_TIMESTAMP' })
dateCreate!: Date & Opt;`

`@Property({ length: 0, nullable: true, defaultRaw: 'CURRENT_TIMESTAMP' })
  dateCreate?: Date;`

With the expressions above, the schema generated converts the `Date` type properties to `varchar` when I generate the migrations as follows bellow:
`modify 'date_create' varchar(0) not null default CURRENT_TIMESTAMP`

As it says in the [documentation](https://mikro-orm.io/docs/defining-entities#default-values) _...we use the Opt type to intersect with the property type to tell the ORM (on type level)..._ so it should not change the database type at all.

After some investigations, I am convinced that there are 3 issues:
1. The `@mikro-orm/entity-generator` is adding the `length: 0` to the Date types 
2. It's also making non nullable fields in the database with a default value as Optional in the generated schema (not 100% sure yet)
3. When the `@mikro-orm/migrations` generates the query, translates `Date & Opt` to varchar

I tested the migration generation with both MySql and Postrgresql drivers, and both had the same issue.

**Notes:**
- The type `number & Opt` will not be outputed as varchar in the migration
- I cannot remove the `& Opt` manually since I run the entity-generator everytime I need to make another update
- I tried to mess with the entity-generator options to remove the Opt type from the Date type fields with no success

Thanks in advance!

### Reproduction

https://github.com/bernardojdias/mikro-orm-reproduction

**Steps to reproduce:**
- run `yarn` command
- make a copy of `.env.example`, rename it to `.env` and enter your database credentials
- run `yarn test` command

**If all the tests run with success, it means that you've reproduced the issue**

**Note:** Due to SQLite column data types not including datetime, I had to use MySql driver in the reproduction example

@mikro-orm/mysql

### MikroORM version

6.1.4

### Node.js version

Node.js: 18.17.1 | Typescript: 5.2.2

### Operating system

Windows

### Validations

boenrobot:
The "default" option is no longer emitted into the decorator, in favor of it being specified in the property's declaration at the entity class. The "defaultRaw" option is still used in cases where the SQL expression does not unambiguously map to the "default" option's value.

The MySQL schema helper translates TINYINT(1) with a value of 0 to "false". This ensures behavior consistent with other drivers. And if this value accidentally finds its way back in a query, it is also OK, as MySQL treats this literal as "0".

The "optional" option is now added early, in the DatabaseTable stage. This has also unlocked some ManyToMany relations that were previously not detected.

Closes #5260

### DIFF
--- a/packages/entity-generator/src/SourceFile.ts
+++ b/packages/entity-generator/src/SourceFile.ts
@@ -197,9 +197,7 @@ export class SourceFile {
         })()
       : prop.type;
 
-    // string defaults are usually things like SQL functions, but can be also enums, for that `useDefault` should be true
-    const isEnumOrNonStringDefault = prop.enum || typeof prop.default !== 'string';
-    const useDefault = prop.default != null && isEnumOrNonStringDefault;
+    const useDefault = prop.default != null;
     const optional = prop.nullable ? '?' : (useDefault ? '' : '!');
 
     if (prop.ref) {
@@ -217,7 +215,7 @@ export class SourceFile {
     }
     ret += hiddenType;
 
-    if (useDefault || (optional !== '?' && (typeof prop.default !== 'undefined' || prop.generated))) {
+    if (useDefault || (prop.optional && !prop.nullable)) {
       this.coreImports.add('Opt');
       ret += ' & Opt';
     }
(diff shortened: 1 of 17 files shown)


B4nan (approved):
strictly speaking, we dont need the explicit `type` when we have the property initializer, but it wont hurt to have it there, lets keep it

Issue #5255, opened 2024-02-18
Enum requires 'type' field with reflect-metadata


NeelDigonto:
### Describe the bug

https://mikro-orm.io/docs/defining-entities#enums

I tried to use enums like above, and do a initial migration with `orm.schema.getUpdateSchemaSQL();` but won't work.
See error below.

I have `import 'reflect-metadata';` at the top of mikro-orm init file, so that isn't the problem and, other stuff like `OneToMany` and `ManyToOne` works fine.

### Reproduction

```ts
@Entity()
export class Log {
  @PrimaryKey({ type: 'int' })
  id!: number;

  @Enum({
    type: () => LogLevel,
  })
  level: LogLevel;
}
```
Will work

```ts
@Entity()
export class Log {
  @PrimaryKey({ type: 'int' })
  id!: number;

  @Enum(() => LogLevel)
  level: LogLevel;
}
```
But this won't, and errors out with

```sh
Error: Please provide either 'type' or 'entity' attribute in Log.level. If you are using decorators, ensure you have 'emitDecoratorMetadata' enabled in your tsconfig.json.
    at ReflectMetadataProvider.initPropertyType (/home/fathomless/repo/redacted/gateway/node_modules/@mikro-orm/core/metadata/ReflectMetadataProvider.js:30:19)
    at ReflectMetadataProvider.initProperties (/home/fathomless/repo/redacted/gateway/node_modules/@mikro-orm/core/metadata/ReflectMetadataProvider.js:23:22)
    at ReflectMetadataProvider.loadEntityMetadata (/home/fathomless/repo/redacted/gateway/node_modules/@mikro-orm/core/metadata/ReflectMetadataProvider.js:10:14)
    at MetadataDiscovery.discoverEntity (/home/fathomless/repo/redacted/gateway/node_modules/@mikro-orm/core/metadata/MetadataDiscovery.js:284:31)
    at MetadataDiscovery.discoverReferences (/home/fathomless/repo/redacted/gateway/node_modules/@mikro-orm/core/metadata/MetadataDiscovery.js:220:18)
    at MetadataDiscovery.findEntities (/home/fathomless/repo/redacted/gateway/node_modules/@mikro-orm/core/metadata/MetadataDiscovery.js:139:14)
    at MetadataDiscovery.discover (/home/fathomless/repo/redacted/gateway/node_modules/@mikro-orm/core/metadata/MetadataDiscovery.js:43:20)
[...]

PR #5267, opened 2024-02-22
docs: added information about SWC to docs

boenrobot:
Renamed the page to "usage-with-transpilers" to reflect the broader meaning, and added a redirect for the sake of any external link shared that may exist.

The defining-entities.md page also mentions "swc" next to babel.


B4nan:
nice, thanks. do you think we need to modify the old versions too? i'd be fine with just the latest 6.x snapshot + next, but since you already did it, we can keep the changes in older snapshots too

docs/docs/usage-with-transpilers.md
```diff
@@ -0,0 +1,57 @@
...
+
+```json
+{
+  "jsc": {
+    "parser": {
+      "syntax": "typescript",
+      "decorators": true
+    },
+    "transform": {
+      "decoratorMetadata": true
```
B4nan, line 42: dont we also need `legacyDecorator` here (unless the project is not ESM)?

boenrobot: Well... I haven't personally tried every possible combination, but in #5255, @NeelDigonto reported that `legacyDecorator` was not needed, and works with or without... And from the provided .swcrc there, it seems he's using commonjs, not ESM, so... I guess MikroORM in particular does not require it apparently.

B4nan: I am rather sure it is needed, without it you use ESM spec, and that (most probably) means there won't be any decorator metadata. Note that in #5255 he specifies the property types explicitly except the enum which doesn't need it, so his repro would work without reflect-metadata just fine.

Let's add the `legacyDecorator` option here to be safer.

boenrobot: Ah. I see. OK, added to all.

docs/docusaurus.config.js
```diff
@@ -255,6 +255,10 @@ module.exports = {
             from: '/docs/entity-references',
             to: '/docs/type-safe-relations',
           },
+          {
+            from: '/docs/usage-with-babel',
+            to: '/docs/usage-with-transpilers',
+          },
```
B4nan, line 261: this only creates a redirect for the current stable version, if we want to keep the changes in older snapshots, we also need redirects for all of them here (similarly we should have a redirect for the `next` version i guess, but that's a minor issue and i am fine with hot having it here)

boenrobot: I refactored the whole redirects section to do renames for all versions on everything, not just this page.

If the target exists at the current version, it gets the new name. Exception is made for the "installation" page, which would be a duplicated route in all versions prior to 6.0. I've added an option for minimal version where the redirect happens to accommodate it.

B4nan (approved):

NeelDigonto:
Thanks for making these changes to help the swc users.

PR #5273, opened 2024-02-23
fix(entity-generator): fix generation of entities with identifiers that overlap with core

boenrobot:
Accomplished by adding ts-morph as a dependency, and using it to post process the generated sources. On a possible conflict, user identifiers now include a marker ("U$"), to disambiguate them in the input to ts-morph.

On conflict, the prefix "Mikro" is added to MikroORM's imports as an alias, and if even that is not enough, the same prefix will keep being added until the conflict is resolved. Previously, there was a special case for the BaseEntity class. This is now removed in favor of this more general purpose conflict resolution, allowing an entity being named like "Entity" or "Property".

Fixed erroneously included skipped tables.

Ensured that tables that start with a number can also be referenced properly, not merely defined, by moving the "E" prefix addition into the naming strategy.

### DIFF
--- a/packages/entity-generator/package.json
+++ b/packages/entity-generator/package.json
@@ -59,7 +59,7 @@
   },
   "dependencies": {
     "@mikro-orm/knex": "6.2.9",
-    "fs-extra": "11.2.0"
+    "ts-morph": "22.0.0"
   },
   "devDependencies": {
     "@mikro-orm/core": "^6.2.9"
--- a/packages/entity-generator/src/CoreImportsHelper.ts
+++ b/packages/entity-generator/src/CoreImportsHelper.ts
@@ -1,3 +1,47 @@
+import {
+  BaseEntity,
+  Collection,
+  Config,
+  EagerProps,
+  Embeddable,
+  Embedded,
+  Entity,
+  EntitySchema,
+  Enum,
+  Formula,
+  Index,
+  ManyToMany,
+  ManyToOne,
+  OneToMany,
+  OneToOne,
+  PrimaryKey,
+  PrimaryKeyProp,
+  Property,
+  Unique,
+} from '@mikro-orm/core';
+
+export const POSSIBLE_IMPORTS = [
+  BaseEntity.name,
+  Collection.name,
+  Config.description!,
+  EagerProps.description!,
+  Embeddable.name,
+  Embedded.name,
+  Entity.name,
+  EntitySchema.name,
+  Enum.name,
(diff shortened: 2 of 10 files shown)


packages/entity-generator/package.json
```diff
@@ -59,7 +59,7 @@
   },
   "dependencies": {
     "@mikro-orm/knex": "6.1.5",
-    "fs-extra": "11.2.0"
+    "ts-morph": "21.0.1"
```
B4nan, line 62: I am not sure I like to introduce this dependency, ts-morph means dependency on the whole TS compiler. Not to mention this has a significant perf hit on its own.

boenrobot: Yeah, I was having some doubts myself due to the performance concerns, but the more I tried to avoid it, the more I realized I am effectively duplicating ts-morph, except in a more error prone and less flexible way.

I had an early version where I tried to have an array of string tokens and custom "import reference" tokens in SourceFile, instead of composing a single string... with the plan being to then deal with the conflicts, and turn the "import reference" token objects into the resolved identifiers, joining all the strings together.

The serializeObject() then proved to be a big obstacle to that plan. Trying to work it into the whole array of tokens idea while still keeping it generic enough would likely incur asimilar performance penalty of its own... Not to mention it would make the SourceFile code infinitely more complex to manage.

So after reading a bit about how to efficiently use ts-morph, I settled on this - doing analysis first (because reading after write causes reparsing...), do only renaming in ts-morph (minimal modifications in general...), and doing only one save in the end to flush all changes (batching IO ops...).

I am not sure how much "dependency on the TS compiler" is an issue, given that it is used during generation, not application runtime. Users should be advised to use the entity generator as a dev dependency anyway, where they likely depend on tsc anyway, for type checking, even if doing builds with babel or swc.

B4nan: > except in a more error prone and less flexible way.

You forgot to mention _much more faster_ way, that's why it's done this way. The tests now take +90s because of this change.

B4nan:
FYI I am still not sold on the idea of introducing ts-morph here, I don't really plan to merge this unless you find a way to get rid of that dependency. The problem itself is valid but feels small enough for this solution, people can already adjust the metadata dynamically and change the names of the entities where needed, right? We could as well have a simple boolean flag to enforce prefixing all the symbols from the ORM, I can imagine namespace import like `import ORM from @mikro-orm/core` and using something like `@ORM.Entity()`.

PR #5280, opened 2024-02-26
fix(mysql): apply current context when fetching `auto_increment_increment`

### LINKED ISSUE #5279: Cannot Create Entity in MySQL When Max Pool Set to 1
### Describe the bug

An application hangs when using MySQL to create a new instance of an entity and the max pool size is set to 1.

This is because the SQL statement:

`show variables like 'auto_increment_increment'`

Is not using the current context and this request creates a new connection to the database.

The PR fixes this issue by passing the context to the `getAutoIncrementIncrement` function.

### Reproduction

Test case and PR to follow.

@mikro-orm/mysql

### MikroORM version

6.1.5

### Node.js version

20.4.0

### Operating system

OSX

### Validations

steve-keep:
Closes #5279

### DIFF
--- a/packages/mysql/src/MySqlDriver.ts
+++ b/packages/mysql/src/MySqlDriver.ts
@@ -1,4 +1,4 @@
-import type { Configuration, EntityDictionary, NativeInsertUpdateManyOptions, QueryResult } from '@mikro-orm/core';
+import type { Configuration, EntityDictionary, NativeInsertUpdateManyOptions, QueryResult, Transaction } from '@mikro-orm/core';
 import { AbstractSqlDriver } from '@mikro-orm/knex';
 import { MySqlConnection } from './MySqlConnection';
 import { MySqlPlatform } from './MySqlPlatform';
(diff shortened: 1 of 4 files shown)


packages/mysql/src/MySqlDriver.ts
```diff
@@ -31,8 +31,9 @@ export class MySqlDriver extends AbstractSqlDriver<MySqlConnection, MySqlPlatfor
   override async nativeInsertMany<T extends object>(entityName: string, data: EntityDictionary<T>[], options: NativeInsertUpdateManyOptions<T> = {}): Promise<QueryResult<T>> {
     options.processCollections ??= true;
     const res = await super.nativeInsertMany(entityName, data, options);
-    const pks = this.getPrimaryKeyFields(entityName);
-    const autoIncrementIncrement = await this.getAutoIncrementIncrement();
+		const pks = this.getPrimaryKeyFields(entityName);
+		const ctx = options.ctx;
```
B4nan, line 35:
```suggestion
    const pks = this.getPrimaryKeyFields(entityName);
    const ctx = options.ctx;
```

B4nan (changes requested):
nice, thanks! same fix needs to be applied to the mariadb driver

B4nan (approved):

PR #5311, opened 2024-03-06
fix(entity-generator): emit missing imports in `EntitySchema` generated files

boenrobot:
Also fixed extra imports when the mapToPk option is set.

Also removed the usage of getEntityName() in SourceFile, thus ensuring that adjustments made by the de-duplication are honored.


B4nan (approved):
thanks!

B4nan:
Btw I am still unsure about the other PR - I would rather not depend on ts-morph just to support conflicts in naming, and it all sounds like a relatively simple thing to implement without ts-morph too, but maybe I am missing something (haven't checked the current implementation much).

PR #5317, opened 2024-03-07
fix(entity-generator): output type import statements for type only core imports

boenrobot:
Helps transpilers and bundlers to more efficiently strip out types.


B4nan:
merging as a `fix` so it gets to the changelog, thanks again!

PR #5320, opened 2024-03-08
fix(core): fix value of `hasPrev[/Next]Page` when paginating backwards

linkthai:
When using cursor pagination, the `hasPrevPage` and `hasNextPage` has inconsistency behavior that display the wrong value when query to the last page of last/before (which is the first page of the first/after).

### Changes

```ts
const { first, last, before, after, orderBy, overfetch } = options;
const limit = first || last;
const isLast = !first && !!last;
const hasMorePages = !!overfetch && limit != null && items.length > limit;
// this.hasPrevPage = before || after ? true : (isLast && hasMorePages);
// this.hasNextPage = !(isLast && !before && !after) && hasMorePages;
this.hasPrevPage = isLast ? hasMorePages : !!after; 
this.hasNextPage = isLast ? !!before : hasMorePages;
```
The old `hasPrevPage` is true if either `before` or `after` exists, which is wrong because having `before` doesn't mean it wouldn't have more page beyond the result, `hasMorePages` decides that if a backward pagination has a previous page. 

Similar to `hasNextPage`, it fails when `isLast` is true and `before` exists because it relies on hasMorePages, in a backward pagination `hasMorePages` can not determine if it has the next page which is the page beyond the `before`.

### Fixing the test cases

You can see the inconsistency in the test cases here, I have fixed to be aligned with the changes.

```ts 
// These ones are correct with the following hasPrevPage/hasNextPage order: 
// true/false => true/true => false/true

// 1. page
const cursor1 = await orm.em.findByCursor(User, {}, {
  first: 3,
  orderBy: { id: 'asc' },
});
expect(cursor1.hasNextPage).toBe(true);
expect(cursor1.hasPrevPage).toBe(false);

// 2. page
const cursor2 = await orm.em.findByCursor(User, {}, {
  first: 3,
  after: cursor1,
  orderBy: { id: 'asc' },
});
expect(cursor2.hasNextPage).toBe(true);
expect(cursor2.hasPrevPage).toBe(true);

....

// 5. page (last)
const cursor5 = await orm.em.findByCursor(User, {}, {
  first: 40,
  after: cursor4,
  orderBy: { id: 'asc' },
});
expect(cursor5.hasNextPage).toBe(false);
expect(cursor5.hasPrevPage).toBe(true);
```

This is where I made the changes
```ts
// These ones are incorrect with the following hasPrevPage/hasNextPage order: 
// false/true => true/true => false/true
// I changed the last one into true/false

// 1. page
const cursor1 = await orm.em.findByCursor(User, {}, {
  last: 3,
  orderBy: { id: 'asc' },
});
expect(cursor1.hasNextPage).toBe(false);
expect(cursor1.hasPrevPage).toBe(true);

// 2. page
const cursor2 = await orm.em.findByCursor(User, {}, {
  last: 3,
  before: cursor1,
  orderBy: { id: 'asc' },
});
expect(cursor2.hasNextPage).toBe(true);
expect(cursor2.hasPrevPage).toBe(true);

// 5. page (last)
const cursor5 = await orm.em.findByCursor(User, {}, {
  last: 40,
  before: cursor4,
  orderBy: { id: 'asc' },
});
// expect(cursor5.hasNextPage).toBe(false);
// expect(cursor5.hasPrevPage).toBe(true);
expect(cursor5.hasNextPage).toBe(true); // new change
expect(cursor5.hasPrevPage).toBe(false); // new change
```

### Additional tests

I added a new `bidirectional-cursor.test.ts`, in which it will test the cursor moving forwards using `first/after` from the first page to the last page and then backward using `last/before` from the last cursor to the first page, and vice versa.

Also, fixing the word simple to complex in the `complex-cursor.test.ts`

### DIFF
--- a/tests/features/cursor-based-pagination/simple-cursor.test.ts
+++ b/tests/features/cursor-based-pagination/simple-cursor.test.ts
@@ -408,8 +408,8 @@ describe.each(['sqlite', 'better-sqlite', 'mysql', 'postgresql', 'mongo'] as con
     expect(cursor5.totalCount).toBe(50);
     expect(cursor5.startCursor).toBe('WzFd');
     expect(cursor5.endCursor).toBe('WzFd');
-    expect(cursor5.hasNextPage).toBe(false);
-    expect(cursor5.hasPrevPage).toBe(true);
+    expect(cursor5.hasNextPage).toBe(true);
+    expect(cursor5.hasPrevPage).toBe(false);
     queries = mock.mock.calls.map(call => call[0]).sort();
     expect(queries).toMatchSnapshot();
     orm.em.clear();
@@ -426,8 +426,8 @@ describe.each(['sqlite', 'better-sqlite', 'mysql', 'postgresql', 'mongo'] as con
     expect(cursor6.totalCount).toBe(50);
     expect(cursor6.startCursor).toBeNull();
     expect(cursor6.endCursor).toBeNull();
-    expect(cursor6.hasNextPage).toBe(false);
-    expect(cursor6.hasPrevPage).toBe(true);
+    expect(cursor6.hasNextPage).toBe(true);
+    expect(cursor6.hasPrevPage).toBe(false);
     queries = mock.mock.calls.map(call => call[0]).sort();
     expect(queries).toMatchSnapshot();
     orm.em.clear();
@@ -533,8 +533,8 @@ describe.each(['sqlite', 'better-sqlite', 'mysql', 'postgresql', 'mongo'] as con
     expect(cursor5.totalCount).toBe(50);
     expect(cursor5.startCursor).toBe('WzUwXQ');
     expect(cursor5.endCursor).toBe('WzUwXQ');
-    expect(cursor5.hasNextPage).toBe(false);
-    expect(cursor5.hasPrevPage).toBe(true);
+    expect(cursor5.hasNextPage).toBe(true);
+    expect(cursor5.hasPrevPage).toBe(false);
     queries = mock.mock.calls.map(call => call[0]).sort();
     expect(queries).toMatchSnapshot();
     orm.em.clear();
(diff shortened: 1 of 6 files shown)


tests/features/cursor-based-pagination/simple-cursor.test.ts
```diff
@@ -408,8 +408,8 @@ describe.each(['sqlite', 'better-sqlite', 'mysql', 'postgresql', 'mongo'] as con
     expect(cursor5.totalCount).toBe(50);
     expect(cursor5.startCursor).toBe('WzFd');
     expect(cursor5.endCursor).toBe('WzFd');
-    expect(cursor5.hasNextPage).toBe(false);
-    expect(cursor5.hasPrevPage).toBe(true);
+    expect(cursor5.hasNextPage).toBe(true);
+    expect(cursor5.hasPrevPage).toBe(false);
```
B4nan, line 412: this looks wrong to me, cursor5 is the last, why should it have `hasNextPage` set to true?

linkthai: The last page of a backward cursor (last/before) is the first page of a forward cursor (first/after), that means it should be the same as cursor1 of the first test, it should have no previous page (due to it reaches the start of the query and has no more pages) and has the next page (due to it has a `before` value).

I look at the code closely and hasNextPage/hasPrevPage means the same thing either forward or backward, they are not reversed if you use `first` or `last`, `before` or `after`. Contrary to `hasMorePages`, which means the opposite depends on if you use `first` or `last`.

Those two test cases are the literal opposite of each other going they also start and end in the opposite of each other. So the first one is correct, `true/false => true/true => false/true` but the second one is currently `false/true => true/true => false/true`. This just makes no sense right, it should be `false/true => true/true => true/false`.

B4nan: I guess you are right, will wait for @darkbasic approval as we were discussing this exact problem some time ago too.

darkbasic: I confirm @linkthai is right about `hasNextPage`/`hasPrevPage` having to depend on which one between `before` or `after` is being used. We discussed it here: https://github.com/mikro-orm/mikro-orm/discussions/5154#discussioncomment-8232645

darkbasic (approved):
I've looked mostly at the bidirectional-cursor test suite because the data set was smaller and it was easier to follow the flow, but this seems right to me.
If anything I would add some tests showing how it behaves once you keep asking for data past the boundaries, like having `hasNextPage` to `false` and still asking more data `after` the last cursor. IIRC Github behavior [puzzled](https://github.com/mikro-orm/mikro-orm/discussions/5154#discussioncomment-8232645) me a bit last time I've seen it, so it would be interesting to test how MikroORM behaves.

B4nan (approved):

PR #5329, opened 2024-03-12
feat(core): add onlyOwnProperties option to assign (#5327)

### LINKED ISSUE #5327: Add an option to ignore nested entities when using assign
When assigning content to new/existing entities, nested entities are exposed (see exemple below)

An option passed to the assign fonction that make it ignore nested entities (no creation/update of relations). However it would allow us to change the ID (when updating the owning side).

We can validate the payload that is assigned to the entity, however it is not secure by default if a decorator is forgotten. I would like to disable that feature entirely.

**Additional context**
This exemple (with `updateNestedEntities:false`) create two nail object in DB. I would like the new nails to be ignored.

```typescript
@Entity()
export class Box {
    @PrimaryKey({ name: '_id'})
    id: ObjectId;

    @Property()
    name: string;

    @OneToMany(() => Nail, mel => mel.box)
    nails = new Collection<Picture>(this);
}

@Entity()
export class Nail {
    @PrimaryKey({ name: '_id'})
    @IsString()
    @IsOptional()
    id: ObjectId;

    @ManyToOne(() => Box)
    box: Box | string;

    @Property()
    size: number;
}
```

```typescript
    let em = entityManager.fork();
    const payload = {
      id: new ObjectId() as any,
      name: "My nail box",
      nails: [
        {
          id: new ObjectId() as any,
          size: 10
        }
      ]
    }
    const entity = new Box();
    wrap(entity).assign(payload as any, { em, updateNestedEntities:false, mergeObjectProperties: true, onlyProperties: true });

    const res = await em.persistAndFlush(entity);

    const payload2 = {
      id: payload.id,
      name: "Renamed Box",
      nails: [
        {
          id: new ObjectId() as any,
          size: 99
        },
      ]
    }

    em = entityManager.fork();

    const dbEntity = await em.getReference(Box, payload.id);

    wrap(dbEntity).assign(payload2 as any, { em, updateNestedEntities:false, mergeObjectProperties: true, onlyProperties: true });
    await em.flush();
```

    "@mikro-orm/core": "^6.1.6",
    "@mikro-orm/mongodb": "^6.1.6",
    
    node version : 18.15.0

acrosett:
add option, ignore entities but allow primary keys

Closes #5327

### DIFF
--- a/packages/core/src/entity/EntityAssigner.ts
+++ b/packages/core/src/entity/EntityAssigner.ts
@@ -71,7 +71,7 @@ export class EntityAssigner {
     let value = data[propName];
     const prop = { ...props[propName], name: propName } as EntityProperty<T>;
 
-    if (options.onlyOwnProperties && prop.entity && !Utils.isPrimaryKey(value)) {
+    if (options.onlyOwnProperties && (prop as any).entity && !Utils.isPrimaryKey(value)) {
       return;
     }
 
--- a/packages/core/src/entity/EntityAssigner.ts
+++ b/packages/core/src/entity/EntityAssigner.ts
@@ -71,6 +71,10 @@ export class EntityAssigner {
     let value = data[propName];
     const prop = { ...props[propName], name: propName } as EntityProperty<T>;
 
+    if (options.onlyOwnProperties && prop.entity && !Utils.isPrimaryKey(value)) {
+      return;
+    }
+
     if (propName in props && !prop.nullable && value == null) {
       throw new Error(`You must pass a non-${value} value to the property ${propName} of entity ${(entity as Dictionary).constructor.name}.`);
     }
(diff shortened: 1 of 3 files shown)


packages/core/src/entity/EntityAssigner.ts
```diff
@@ -71,6 +71,10 @@ export class EntityAssigner {
     let value = data[propName];
     const prop = { ...props[propName], name: propName } as EntityProperty<T>;
 
+    if (options.onlyOwnProperties && (prop as any).entity && !Utils.isPrimaryKey(value)) {
```
B4nan, line 74: you need to check `prop.kind`, not `prop.entity`

B4nan: also the value could be entity for to-one relations, and you should use it only as a FK, this is probably wrong place to handle this

acrosett: ok, so something like :

```typescript
[ReferenceKind.ONE_TO_ONE, ReferenceKind.ONE_TO_MANY, 
ReferenceKind.MANY_TO_ONE, ReferenceKind.MANY_TO_MANY].includes(prop?.kind)
```

PR #5343, opened 2024-03-15
fix(migrator): type mismatch between Migration interface and class

### LINKED ISSUE #5340: Type mismatch between class Migration and interface Migration
Migrations are generated using the Migration class from mikro-orm/migrations. This class defines these two methods:

```ts
abstract up(): Promise<void> | void;
down(): Promise<void> | void;
```

However, when initializing MikroORM and specifying migrations using the migrationsList property, it expects a MigrationObject. This object utilizes the Migration interface for its class definition, which is defined as follows:

```ts
up(): Promise<void>;
down(): Promise<void>;
```

While the Migration class allows both asynchronous and synchronous implementations for up and down methods, the interface used during initialization does not permit synchronous implementations. Is this intended?

montycb:
Closes #5340

### DIFF
--- a/packages/core/src/typings.ts
+++ b/packages/core/src/typings.ts
@@ -888,8 +888,8 @@ export interface IMigrationGenerator {
 }
 
 export interface Migration {
-  up(): Promise<void>;
-  down(): Promise<void>;
+  up(): Promise<void> | void;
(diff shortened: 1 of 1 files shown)


B4nan (approved):
thanks!

PR #5359, opened 2024-03-19
fix(entity-generator): allow arbitrary class and prop names as identifiers

boenrobot:
fix(entity-generator): allow arbitrary class and prop names as identifiers

AbstractNamingStrategy is updated to handle class conflicts, and inherently,
any naming strategy that inherits from it.

The "word boundary" character in entity and property names is now correctly replaced,
even in cases where multiple sequential separators are present.

Any table that starts with a character that is invalid as a JS start character (not just digits)
gets prefixed with "E".
Characters that are not valid in any position as a class identifier (or the character "$")
get turned into code points, prefixed with $, making them valid (though inconvenient to type)
identifiers and file names.

In AbstractNamingStrategy.columnNameToProperty(),
prop names that would be in conflict with PopulatePath values get prefixed with "$".

Prop names that are not valid identifiers get string quoted.
This is a SourceFile/EntitySchemaSourceFile feature, not a naming strategy feature,
so it will get applied regardless of naming strategy.

The MySqlSchemaHelper doesn't special case index expression based on column name.
Lack of column name or presence of an "expression" column with "where" is still considered.

The PostgreSqlSchemaHelper properly escapes all identifiers when looking them up.
Due to the relative complexity of PostgreSQL's expressions vs column names,
index properties are still not picked up for odd identifiers,
though an index expression is generated.

Also made the imports of types tied to the "entity" prop option, rather than the prop declaration.
This ensures proper imports when using mapToPk in extension hooks.

Also moved the schema de-duplication code to happen before bi-directional and many-to-many
detection, to ensure less work and error prone-ness for the other stages in updating references.

Also fixed quoting of strings in generated columns.

### DIFF
--- a/packages/core/src/naming-strategy/AbstractNamingStrategy.ts
+++ b/packages/core/src/naming-strategy/AbstractNamingStrategy.ts
@@ -1,10 +1,13 @@
 import type { NamingStrategy } from './NamingStrategy';
+import { PopulatePath } from '../enums';
+
+const populatePathMembers = Object.values(PopulatePath);
 
 export abstract class AbstractNamingStrategy implements NamingStrategy {
 
   getClassName(file: string, separator = '-'): string {
     const name = file.split('.')[0];
-    const ret = name.replace(new RegExp(`${separator}+(\\w)`, 'g'), m => m[1].toUpperCase());
+    const ret = name.replace(new RegExp(`(?:${separator})+(\\w)`, 'ug'), (_, p1) => p1.toUpperCase());
 
     return ret.charAt(0).toUpperCase() + ret.slice(1);
   }
@@ -46,11 +49,16 @@ export abstract class AbstractNamingStrategy implements NamingStrategy {
    * @inheritDoc
    */
   getEntityName(tableName: string, schemaName?: string): string {
-    return this.getClassName(tableName, '_');
+    const name = tableName.match(/^[^$_\p{ID_Start}]/u) ? `E_${tableName}` : tableName;
+    return this.getClassName(name.replaceAll(/[^\u200C\u200D\p{ID_Continue}]+/ug, r => r.split('').map(c => `$${c.codePointAt(0)}`).join('')), '_');
   }
 
   columnNameToProperty(columnName: string): string {
-    return columnName.replace(/[_\- ](\w)/g, m => m[1].toUpperCase()).replace(/[_\- ]+/g, '');
+    const propName = columnName.replace(/[_\- ]+(\w)/ug, (_, p1) => p1.toUpperCase());
+    if (populatePathMembers.includes(propName.replace(/^\${2,}/u, '$$').replace(/^\$\*$/u, '*') as PopulatePath)) {
+      return `$${propName}`;
+    }
+    return propName;
   }
 
   aliasName(entityName: string, index: number): string {
--- a/tests/features/entity-generator/__snapshots__/OddTableNames.mysql.test.ts.snap
+++ b/tests/features/entity-generator/__snapshots__/OddTableNames.mysql.test.ts.snap
@@ -1,0 +1,236 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`odd_table_names_example:100% entitySchema=false: mysql 1`] = `
+[
+  "import { Entity, OneToOne, PrimaryKeyProp } from '@mikro-orm/core';
+import { E50$37$32of$32stuff } from './E50$37$32of$32stuff';
+
+@Entity({ tableName: '*misc' })
+export class E_$42misc {
+
+  [PrimaryKeyProp]?: '@ref';
+
+  @OneToOne({ entity: () => E50$37$32of$32stuff, fieldName: '@ref', primary: true })
+  '@ref'!: E50$37$32of$32stuff;
+
+}
+",
+  "import { Entity, Index, ManyToOne, type Opt, PrimaryKey, PrimaryKeyProp, Property } from '@mikro-orm/core';
+import { Table$39s$32name$32has$32apostrophe$44$32also$32$96$32this } from './Table$39s$32name$32has$32apostrophe$44$32also$32$96$32this';
+
+@Entity({ tableName: '123_table_name' })
+@Index({ name: 'dollar\\'s index', properties: ['$$', '$$infer'] })
+export class E123TableName {
+
+  [PrimaryKeyProp]?: '$';
+
+  @PrimaryKey({ unsigned: false, autoincrement: false })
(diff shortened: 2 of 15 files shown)


tests/features/entity-generator/__snapshots__/OddTableNames.mysql.test.ts.snap
```diff
@@ -0,0 +1,186 @@
...
+      fieldName: '$',
+      unsigned: false,
+      autoincrement: false,
+    },
+    $$$: { type: 'bigint', fieldName: '$$' },
+    prototype: {
+      kind: 'm:1',
+      entity: () => Table$39s$32name$32has$32apostrophe$44$32also$32$96$32this,
+      fieldName: 'prototype',
+      index: 'fk_123_table_name_table's name has apostrophe, also\` this_idx',
```
B4nan, line 135: missing escaping for `'`

tests/features/entity-generator/__snapshots__/OddTableNames.mysql.test.ts.snap
```diff
@@ -0,0 +1,186 @@
...
+
+  [PrimaryKeyProp]?: '+20%';
+
+  @PrimaryKey({ fieldName: '+-20%' })
+  '+20%'!: number;
+
+  @Property({ length: 45 })
+  'my@odd.column'!: string;
+
+  @Property({ fieldName: 'column with apostrophe in it's name', length: 45 })
```
B4nan, line 59: same here

packages/core/src/naming-strategy/AbstractNamingStrategy.ts
```diff
@@ -46,11 +49,17 @@ export abstract class AbstractNamingStrategy implements NamingStrategy {
...
   getEntityName(tableName: string, schemaName?: string): string {
-    return this.getClassName(tableName, '_');
+    const name = tableName.match(/^[^$_\p{ID_Start}]/u) ? `E_${tableName}` : tableName;
+    return this.getClassName(name.replaceAll(/[^\u200C\u200D\p{ID_Continue}]+/ug, r => r.split('').map(c => `$${c.codePointAt(0)}`).join('')), '_');
   }
 
   columnNameToProperty(columnName: string): string {
-    return columnName.replace(/[_\- ](\w)/g, m => m[1].toUpperCase()).replace(/[_\- ]+/g, '');
+    const propName = columnName.replace(/[_\- ](\w)/g, m => m[1].toUpperCase()).replace(/[_\- ]+/g, '');
+    // If the column resolves to something defined in LoadedReference, prefix with "$" to avoid conflicts.
```
B4nan, line 58: i dont understand why you would care about the `Reference` properties, they can overlap with entity property names, the class is used explicitly and needs unwrapping, it never contains more than the helper methods plus the primary key of the target/wrapped entity.

B4nan:
> Prop names that conflict with the prototype of Reference (which all loaded entities inherit from)

This is not true. Why did you think so? Maybe we need to improve the docs somewhere. 

`Reference` is an optional wrapper class, nowadays represented by the `Ref` type. 

https://mikro-orm.io/docs/guide/type-safety#reference-wrapper

B4nan:
>  point is that you have a point at which you have accessible both the properties and extra identifiers like "$", and "load".

Nope, that's not how it works, the wrapped entity doesn't have anything in common with the `Reference` class. 

> So... when you have a wrapped object, be it because of Ref, or because it is the direct result of find()... Is there a way to read or write a prop with an arbitrary name? If there is, it is not documented.

I don't understand your question. It's a simple wrapper class, if you want to access anything else than the primary key, you need to unwrap with one of the helper methods.

packages/core/src/naming-strategy/AbstractNamingStrategy.ts
boenrobot: ok, after playing around a bit, I see what my big misunderstanding is...

I thought the results of find() are always Reference wrapped, and that from any Reference, you have either the props as well as "$", but I see it's actually that you have a partial of the entity, and that the wrapper doesn't provide direct access to the entity's props, except through "$" and "get", so indeed, no conflict.

I've removed this bit now.

tests/features/entity-generator/__snapshots__/OddTableNames.mysql.test.ts.snap
boenrobot: Fixed now. Also realized I was missing tests and escapes for composite indexes. Now added too, along with a fix for indexes with props that have odd identifiers.

tests/features/entity-generator/__snapshots__/OddTableNames.postgres.test.ts.snap
```diff
@@ -0,0 +1,216 @@
...
+  [PrimaryKeyProp]?: '+20%';
+
+  @PrimaryKey({ fieldName: '+-20%' })
+  '+20%'!: number;
+
+  @Property({ length: 45 })
+  'my@odd.column'!: string;
+
+  @Property({ fieldName: 'column with apostrophe in it\\'s name', length: 45 })
+  'columnWithApostropheInIt\\'sName'!: string;
```
B4nan, line 55: wrong escaping

tests/features/entity-generator/__snapshots__/OddTableNames.postgres.test.ts.snap
```diff
@@ -0,0 +1,216 @@
...
+
+  @Property({ fieldName: 'column with backtick in it\`\`s name', length: 45 })
+  'columnWithBacktickInIt\`\`sName'!: string;
```
B4nan, line 58: unnecessary escaping of `

B4nan (commented):
the escaping is apparently not working correctly

tests/features/entity-generator/__snapshots__/OddTableNames.postgres.test.ts.snap
boenrobot: If you mean the double slash... this is an artifact of the fact that jest snapshots aren't the literal file, but a template string inside a js file.

If you check the actual file produced, you'll see it has only one slash inside it.

That said, I should probably make it create the files, as in the mysql test, so... I'll change that part. But the escaping here is correct.

tests/features/entity-generator/__snapshots__/OddTableNames.postgres.test.ts.snap
boenrobot: I _think_ this is the same problem here too...

The snapshot is inside a template string... that's just how .snap files work...

So it escapes the `, in order to stay into the template string.

Issue #4678, opened 2023-09-06
Add support for psql JSON operators


Fr1m3n:
Hello

Currently, only 2 JSON operators are [supported](https://mikro-orm.io/docs/query-conditions/#comparison) in mikro-orm - `@>` and `<@` (exclusively for psql). Postgresql has a large set of [operators](https://www.postgresql.org/docs/10/functions-json.html) (Table 9.44) that are not supported in mikro-orm. In particular, I was faced with the need to use the operator `?|`.

I assume a solution similar to the existing one with the operators `$contains` and `$contained`.

B4nan:
PR welcome

sslava:
Hey @B4nan, i'd like to implement support of `?`, `?|` and `?&` operators. 
What names do you think would suit best?
- `?` — `$haskey`
- `?&` — `$haskeys`
- `?|`  — `$hassomekeys` 

```ts
  await orm.em.findAll(User, {
    where: {
      books: { parameters: { $haskey: 'seasons' } },
    },
    populate: ["books"],
  });

  await orm.em.findAll(User, {
    where: { books: { parameters: { $haskeys: ['seasons', 'pages'] } } },
  });

  await orm.em.findAll(User, {
    where: { books: { parameters: { $hassomekeys: ['seasons', 'pages'] } } },
  });
  
```

Maybe you could propose better names?

B4nan:
Let's at least use `pascalCase`, so `$hasKey` etc.

sslava:
Cool, will fix the PR soon

sslava:
@B4nan thanks for the reply. Fixed my [PR](https://github.com/mikro-orm/mikro-orm/pull/5366) with new naming

B4nan:
I am closing this one since it's too broad and the mentioned operator is now included thanks to #5366.

If there are more operators missing, PR for those is still welcome, it should be hopefully pretty straightforward.

PR #5366, opened 2024-03-22
feat(postgres): add `?`, `?|` and `?&` json operators

sslava:
This PR will introduce few missing PostgreSQL [JSON operators](https://www.postgresql.org/docs/10/functions-json.html) In response to https://github.com/mikro-orm/mikro-orm/issues/4678
`?` - Does the string exist as a top-level key within the JSON value?
`?|` - Do any of array strings exist as top-level keys?
`?&`- Do all of array strings exist as top-level keys?

```ts
  // ? operator
  await orm.em.findAll(User, {
    where: {
      books: { parameters: { $hasKey: 'seasons' } },
    },
  });
```
```ts
  // ?& operator
  await orm.em.findAll(User, {
    where: { books: { parameters: { $hasKeys: ['seasons', 'pages'] } } },
  });
```
```ts
  // ?| operator
  await orm.em.findAll(User, {
    where: { books: { parameters: { $hasSomeKeys: ['seasons', 'pages'] } } },
  });
```

Related: #4678

### DIFF
--- a/docs/docs/query-conditions.md
+++ b/docs/docs/query-conditions.md
@@ -51,23 +51,26 @@ const res = await orm.em.find(Author, [1, 2, 7]);
 
 ### Comparison
 
-| operator     | name             | description                                                                                 |
-| ------------ | ---------------- | ------------------------------------------------------------------------------------------- |
-| `$eq`        | equals           | Matches values that are equal to a specified value.                                         |
-| `$gt`        | greater          | Matches values that are greater than a specified value.                                     |
-| `$gte`       | greater or equal | Matches values that are greater than or equal to a specified value.                         |
-| `$in`        | contains         | Matches any of the values specified in an array.                                            |
-| `$lt`        | lower            | Matches values that are less than a specified value.                                        |
-| `$lte`       | lower or equal   | Matches values that are less than or equal to a specified value.                            |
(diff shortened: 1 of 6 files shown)


ladal1:
Thanks, I was looking into implementing this as well. Looks good to me, I would just add a test for checking keys in arrays, as these operators will inevitably end up being used that way

sslava:
> Thanks, I was looking into implementing this as well. Looks good to me, I would just add a test for checking keys in arrays, as these operators will inevitably end up being used that way

Do you mean with ArrayTypes? these three `?` `?&` `?|` only supported for json, not for arrays :)

ladal1:
No, I mean for json arrays, the operator for keys also checks for string elements in json arrays.

```
jsonb ?| text[] → boolean
Do any of the strings in the text array exist as top-level keys or **array elements?**
'{"a":1, "b":2, "c":3}'::jsonb ?| array['b', 'd'] → t
```

sslava:
> No, I mean for json arrays, the operator for keys also checks for string elements in json arrays.
> 
> jsonb ?| text[] → boolean Do any of the strings in the text array exist as top-level keys or **array elements?** '{"a":1, "b":2, "c":3}'::jsonb ?| array['b', 'd'] → t

Ah, good catch. I just added few tests with json arrays. 
Thanks again

PR #5369, opened 2024-03-22
docs(core): added definition of an identity map

boenrobot:
Related to #5367 .

The sentence
> they don't really explain anything to someone not knowing what an Identity Map is.

kind of triggered me on this one, in that it is actually a fair point, despite not being the core of the issue. Not sure if my definition here makes sense, but I think _some_ definition of the term identity map being present is important.

### DIFF
--- a/docs/docs/identity-map.md
+++ b/docs/docs/identity-map.md
@@ -5,6 +5,12 @@ sidebar_label: Identity Map
 
 `MikroORM` uses identity map in background, so we will always get the same instance of one entity.
 
+## What is an "Identity Map"
+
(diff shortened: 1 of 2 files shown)


B4nan (approved):
nice, thank you!

i would have swore we at least had a link to wiki for this, a tiny bit is explained in the getting started guide: https://mikro-orm.io/docs/guide/first-entity#persist-and-flush (and in the following section), i guess we could also replicate this small paragraph into that place

the docs could indeed do a better job at explaining those things.

B4nan:
btw since we are discussing this, in the long run i am also considering better naming for this. the `RequestContext` helper is in fact a context provider, and generally speaking the work "request" would be better replaced with "session" I think. it especially makes little sense to have those decorators that create the context, they are called "create request context" etc, but their use case is where there is no request at all, e.g. for queues and cron jobs. at this point i just feel this is pretty confusing, so will welcome ideas on how to improve this in v7 (and we can do such changes earlier and non-breaking and only with v7 remove the old APIs)

PR #5384, opened 2024-03-27
feat(postgres): allow defining deferred FK constraints

### LINKED ISSUE #5306: Support deferrable constraints for SQL
I'm always frustrated when I try to insert a lot of data in a single transaction and the foreign key constraints are violated because the related row was not inserted yet in that statement, but will be inserted later as part of the same transaction. SQL has a solution for this: deferring constraint checks until the commit of the transaction.

In postgres this is done by setting [the constraint to be DEFERRABLE](https://www.postgresql.org/docs/7.1/sql-createtable.html#:~:text=This%20controls%20whether%20the%20constraint%20can%20be%20deferred%20to%20the%20end%20of%20the%20transaction.), and [setting the constraint check timing for the transaction to DEFERRED](https://www.postgresql.org/docs/current/sql-set-constraints.html). Or just setting the default behavior of the constraint to be DEFERRABLE INITIALLY DEFERRED to apply by default for all transactions.

It would be nice if we can set deferrablility and default check timing through the ORM, maybe as options in the relationship decorators (`@ManyToOne`, `@OneToOne`, etc).

For example:
```ts
@ManyToOne({entity:()=>EntityA, deferMode:"DEFERRABLE", defaultDeferTiming:"DEFERRED"})
entityA!: Ref<EntityA>;
```
Then the generated migration would be:
```ts
this.addSql('create table "entity_b" ("id" serial primary key, "entity_a_id" int not null);');
this.addSql('alter table "entity_b" add constraint "entity_b_entity_a_id_foreign" foreign key ("entity_a_id") references "entity_a" ("id") on update cascade deferrable initially deferred;');
```
And we could have a default global value for these options. I don't know if there's a place for driver-specific configs, or maybe we could put them with the orm options.

Also for the transaction, maybe we could have an option for the check timing in `em.begin()` or `em.transactional()`:
```ts
await em.begin({checkTiming: "DEFERRED"})
```

I could manually edit the migration files to make constraints deferrable, and run a raw query to `SET CONSTRAINTS DEFERRED` 

Or alternatively I could insert my data in an order that guarentees no conflicts even within the same transaction.

**Additional context**
Somewhat related issue #1175  because I saw a test for immediate constraints in the issue tests.

DASPRiD:
This PR adds the option to specify constraints as deferrable for relations.

On OneToOne and ManyToOne relations you can now specify a `deferMode` property, which can be either `not deferrable` (default), `immediate` or `deferred`. Those are also covered by a new `DeferMode` enum.

Closes #5306

### DIFF
--- a/packages/knex/src/schema/SchemaComparator.ts
+++ b/packages/knex/src/schema/SchemaComparator.ts
@@ -9,7 +9,6 @@ import {
   type Dictionary,
   type EntityProperty,
   type Logger,
-  DeferMode,
 } from '@mikro-orm/core';
 import type { Column, ForeignKey, IndexDef, SchemaDifference, TableDifference } from '../typings';
 import type { DatabaseSchema } from './DatabaseSchema';
@@ -435,7 +434,7 @@ export class SchemaComparator {
       return true;
     }
 
-    if ((key1.deferMode ?? DeferMode.NOT_DEFERRABLE) !== (key2.deferMode ?? DeferMode.NOT_DEFERRABLE)) {
+    if (key1.deferMode !== key2.deferMode) {
       return true;
     }
 
--- a/packages/knex/src/schema/SchemaComparator.ts
+++ b/packages/knex/src/schema/SchemaComparator.ts
@@ -9,6 +9,7 @@ import {
   type Dictionary,
   type EntityProperty,
   type Logger,
+  DeferMode,
 } from '@mikro-orm/core';
 import type { Column, ForeignKey, IndexDef, SchemaDifference, TableDifference } from '../typings';
 import type { DatabaseSchema } from './DatabaseSchema';
@@ -431,6 +432,10 @@ export class SchemaComparator {
     }
 
     if (key1.referencedTableName !== key2.referencedTableName) {
+      return true;
+    }
+
+    if ((key1.deferMode ?? DeferMode.NOT_DEFERRABLE) !== (key2.deferMode ?? DeferMode.NOT_DEFERRABLE)) {
       return true;
     }
 
--- a/tests/features/deferrable-constraints/deferrable-constraints.postgres.test.ts
+++ b/tests/features/deferrable-constraints/deferrable-constraints.postgres.test.ts
@@ -21,13 +21,12 @@ class Child {
 
 describe('deferrable constraints in postgres', () => {
 
-  let orm: MikroORM<PostgreSqlDriver>;
+  let orm: MikroORM;
 
   beforeAll(async () => {
     orm = await MikroORM.init({
       entities: [Parent, Child],
       dbName: `mikro_orm_test_deferrable`,
-      driver: PostgreSqlDriver,
     });
     await orm.schema.refreshDatabase();
   });
--- a/tests/features/deferrable-constraints/deferrable-constraints.postgres.test.ts
+++ b/tests/features/deferrable-constraints/deferrable-constraints.postgres.test.ts
@@ -1,6 +1,4 @@
-import { MikroORM, Ref, Reference } from '@mikro-orm/core';
-import { DeferMode, Entity, ManyToOne, PrimaryKey } from '@mikro-orm/core';
-import { PostgreSqlDriver } from '@mikro-orm/postgresql';
+import { DeferMode, Entity, ManyToOne, PrimaryKey, MikroORM, Ref, Reference } from '@mikro-orm/postgresql';
 
 @Entity()
 class Parent {
--- a/tests/features/deferrable-constraints/deferrable-constraints.postgres.test.ts
+++ b/tests/features/deferrable-constraints/deferrable-constraints.postgres.test.ts
@@ -1,0 +1,58 @@
+import { MikroORM, Ref, Reference } from '@mikro-orm/core';
+import { DeferMode, Entity, ManyToOne, PrimaryKey } from '@mikro-orm/core';
+import { PostgreSqlDriver } from '@mikro-orm/postgresql';
+
+@Entity()
+class Parent {
+
+  @PrimaryKey()
+  id!: number;
+
+}
+
+@Entity()
+class Child {
+
+  @PrimaryKey()
+  id!: number;
+
+  @ManyToOne(() => Parent, { ref: true, deferMode: DeferMode.INITIALLY_DEFERRED })
+  parent!: Ref<Parent>;
+
+}
+
+describe('deferrable constraints in postgres', () => {
+
+  let orm: MikroORM<PostgreSqlDriver>;
+
+  beforeAll(async () => {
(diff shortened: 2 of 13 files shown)


tests/features/deferrable-constraints/deferrable-constraints.postgres.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  beforeAll(async () => {
+    orm = await MikroORM.init({
+      entities: [Parent, Child],
+      dbName: `mikro_orm_test_deferrable`,
+      driver: PostgreSqlDriver,
+    });
+    await orm.schema.ensureDatabase();
+    await orm.schema.execute('drop table if exists child');
+    await orm.schema.execute('drop table if exists parent');
+    await orm.schema.createSchema();
```
B4nan, line 37:
```suggestion
    await orm.schema.refreshDatabase();
```

tests/features/deferrable-constraints/deferrable-constraints.postgres.test.ts
```diff
@@ -0,0 +1,58 @@
+import { MikroORM, Ref, Reference } from '@mikro-orm/core';
+import { DeferMode, Entity, ManyToOne, PrimaryKey } from '@mikro-orm/core';
+import { PostgreSqlDriver } from '@mikro-orm/postgresql';
```
B4nan, line 3:
```suggestion
import { DeferMode, Entity, ManyToOne, PrimaryKey, MikroORM, Ref, Reference } from '@mikro-orm/postgresql';
```

tests/features/deferrable-constraints/deferrable-constraints.postgres.test.ts
```diff
@@ -0,0 +1,58 @@
...
+describe('deferrable constraints in postgres', () => {
+
+  let orm: MikroORM<PostgreSqlDriver>;
+
+  beforeAll(async () => {
+    orm = await MikroORM.init({
+      entities: [Parent, Child],
+      dbName: `mikro_orm_test_deferrable`,
+      driver: PostgreSqlDriver,
+    });
```
B4nan, line 33:
```suggestion
  let orm: MikroORM;

  beforeAll(async () => {
    orm = await MikroORM.init({
      entities: [Parent, Child],
      dbName: `mikro_orm_test_deferrable`,
    });
```

B4nan (changes requested):
nice, thanks!

can you please also test schema diffing? either in the test you added or by updating this one:

https://github.com/mikro-orm/mikro-orm/blob/master/tests/features/schema-generator/fk-diffing.postgres.test.ts

packages/knex/src/schema/SchemaComparator.ts
```diff
@@ -434,6 +435,10 @@ export class SchemaComparator {
       return true;
     }
 
+    if ((key1.deferMode ?? DeferMode.NOT_DEFERRABLE) !== (key2.deferMode ?? DeferMode.NOT_DEFERRABLE)) {
```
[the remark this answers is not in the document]
DASPRiD: I do, either could be undefined, and undefined wouldn't equal "not deferrable".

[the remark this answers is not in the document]
DASPRiD: Actually nevermind, I can get rid of it.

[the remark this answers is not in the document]
B4nan: if you add the schema diffing tests, it should be clear whether this is needed or not. 

i can imagine some cases where it might be - e.g. when you have the `deferMode` set but the driver does not support it (so one will be undefined and the other will be set, but there is still no point in marking such change as the feature is not supported)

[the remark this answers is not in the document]
DASPRiD: Am on it ;)

tests/features/schema-generator/fk-diffing.postgres.test.ts
```diff
@@ -154,3 +187,39 @@ describe('dropping tables with FKs in postgres', () => {
...
+    await orm.schema.ensureDatabase();
+    await orm.schema.execute('drop table if exists author cascade');
+    await orm.schema.execute('drop table if exists book cascade');
+    await orm.schema.createSchema();
+
+    orm.getMetadata().reset('Book3');
+    orm.discoverEntity([Book41]);
+    const diff1 = await orm.schema.getUpdateSchemaSQL({ wrap: false });
+    expect(diff1).toMatchSnapshot();
+    await orm.schema.execute(diff1);
```
B4nan, line 208: lets add some assertions to ensure there wont be any extra queries when the schema is already in sync, basically after each of the `schema.execute` calls

B4nan: no need for snapshot, just compare to empty string

B4nan: something like `await expect(orm.schema.getUpdateSchemaSQL({ wrap: false })).toBe('');`, check the existing tests, they do it too

tests/features/schema-generator/fk-diffing.postgres.test.ts
```diff
@@ -154,3 +187,42 @@ describe('dropping tables with FKs in postgres', () => {
...
+    expect(diff2).toMatchSnapshot();
+    await orm.schema.execute(diff2);
+    expect(await orm.schema.getUpdateSchemaSQL({ wrap: false })).toBe('');
+
+    orm.getMetadata().reset('Book42');
+    orm.discoverEntity([Book4]);
+    const diff3 = await orm.schema.getUpdateSchemaSQL({ wrap: false });
+    expect(diff3).toMatchSnapshot();
+    await orm.schema.execute(diff3);
+    expect(await orm.schema.getUpdateSchemaSQL({ wrap: false })).toBe('');
```
B4nan, line 223:
```suggestion
    await expect(orm.schema.getUpdateSchemaSQL({ wrap: false })).resolves.toBe('');

    orm.getMetadata().reset('Book41');
    orm.discoverEntity([Book42]);
    const diff2 = await orm.schema.getUpdateSchemaSQL({ wrap: false });
    expect(diff2).toMatchSnapshot();
    await orm.schema.execute(diff2);
    await expect(orm.schema.getUpdateSchemaSQL({ wrap: false })).resolves.toBe('');

    orm.getMetadata().reset('Book42');
    orm.discoverEntity([Book4]);
    const diff3 = await orm.schema.getUpdateSchemaSQL({ wrap: false });
    expect(diff3).toMatchSnapshot();
    await orm.schema.execute(diff3);
    await expect(orm.schema.getUpdateSchemaSQL({ wrap: false })).resolves.toBe('');
```

B4nan:
all right, thanks again!

PR #5397, opened 2024-03-31
docs: fix typo in getting started guide

phuffer:
Assuming this should read `userRef` instead of `user` based on previous const name.


rubiin (commented):
looks like that was a typo

B4nan (approved):

B4nan:
thanks!

PR #5416, opened 2024-04-04
feat(core): add support for serialization groups

B4nan:
Every property can specify its serialization groups, which are then used with explicit serialization.

> Properties without the `groups` option are always included.

Let's consider the following entity:

```ts
@Entity()
class User {

  @PrimaryKey()
  id!: number;

  @Property()
  username!: string;

  @Property({ groups: ['public', 'private'] })
  name!: string;

  @Property({ groups: ['private'] })
  email!: string;

}
```

Now when you call `serialize()`:
- without the `groups` option, you get all the properties
- with `groups: ['public']` you get `id`, `username` and `name` properties
- with `groups: ['private']` you get `id`, `username`, `name` and `email` properties
- with `groups: []` you get only the `id` and `username` properties (those without groups)

```ts
const dto1 = serialize(user);
// User { id: 1, username: 'foo', name: 'Jon', email: 'jon@example.com' }

const dto2 = serialize(user, { groups: ['public'] });
// User { id: 1, username: 'foo', name: 'Jon' }
```


jnbenoit-oandb:
Hello,
This is a really good enhancement. 
But, can we make it work with Embedabbles properties  too ?

PR #5435, opened 2024-04-09
feat(entity-generator): allow custom types for scalar relations

boenrobot:
This is done by fully refactoring type inference in general,
to take advantage of "runtimeType" in the prop,
combined with more cross checks to add IType where needed.

Also removed the accidental Embedded decorator in EntitySchema.

---

Using custom types for scalar relations was sort of possible even before this, but it has always been more hacky than anything else. With this, it is actually encoded as an expected possibility, and added to the test for hooks. Users who want to use custom types will need to handle their resolution in the fileName function. Same as if those identifiers were referring to an entity.


tests/features/entity-generator/__snapshots__/MetadataHooks.mysql.test.ts.snap
```diff
@@ -20,6 +20,7 @@ export class Address2 {
 ",
   "import { Cascade, Collection, EagerProps, Embedded, Entity, Formula, type Hidden, Index, ManyToMany, ManyToOne, OneToMany, type Opt, PrimaryKey, Property, Unique } from '@mikro-orm/core';
 import { Book2 } from './Book2';
+import { CustomBooleanType } from './CustomBooleanType';
```
B4nan, line 23: where is this `CustomBooleanType` defined?

B4nan: can we at least generate a stub for this? like `export type CustomBooleanType = boolean`, which users can replace?

as this feels like the only thing that the generator would output and wouldn't exist

boenrobot: Hm... now that you showed this specific stub, I'm thinking there's actually a different can of worms I haven't dealt with... In fact, it's fair to say I didn't think the whole thing through...

We have...

1. Type only imports for non-core imports in the prop declaration... The one in the test was meant to be imported and used as such. Custom shapes for plain JSON values (as opposed to embedded JSON values) are perhaps a more realistic example of something that users would want to do, and would still end up having to import as a type import. My `CustomBooleanType` example was instead an example of the second use case for type only imports - a branded type (in this case, a branded boolean, though branded integers are perhaps a more frequent use case).
2. IType annotated custom types. The above does not qualify as such, but what about types that do? Things like "an email type" for example that validate on input. The meta data would have the type class, but the type class is not what should be in the prop definition (just the prop options instead).

The scope of this PR suddenly increased a bit, but I guess better to address at least one of those two things, if not both...

I'm thinking of trying to do a dynamic import of the referenced identifier (when the ref is a scalar non-enum of a non-lib type...) after letting fileName dictate the import path... if the file does not exist, or the identifier does not exist in it, or exists, but is not a mapped type, assume it is a type only import, and output it as such. Otherwise, check `runtimeType` and `compareAsType()` to determine the value of the prop type. If they are the same, use that. If not, use `IType<${runtimeType}, ${compareAsType}>`, and add an import for `runtimeType` (if it's a non-lib type).

B4nan: > after letting fileName dictate the import path

This bit was also pretty confusing to me, if we expect people to use this to override import paths, we need to document it somewhere. And I am not so sure it's a good idea.

I feel like you are building something without a clear understanding of how the end users would do stuff. As I mentioned many times, entity generator wasn't meant as a schema-first approach (it was more of a helper instead), which I think you are trying to build. So I would first think about how are people supposed to combine it with their own code, and only then think about those cases. And of course, this all needs to be clearly documented, making code changes without that won't help much. Feels like you are going in the wrong direction, I would design this from the top to bottom.

PR #5451, opened 2024-04-14
docs: fix typo in 01-first-entity.md

Jedliu:
Fix the entity name of `Author` was a typo as `Book`.

### DIFF
--- a/docs/docs/guide/01-first-entity.md
+++ b/docs/docs/guide/01-first-entity.md
@@ -482,9 +482,9 @@ Now back to the validation error about global context. With the freshly gained k
 
    Now when someone requests the same article via both of those endpoints, we could end up with both returning the same output:
 
-    1. `GET /article/1` returns `Book` without populating its property `author` property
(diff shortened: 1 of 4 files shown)


docs/docs/guide/01-first-entity.md
```diff
@@ -482,9 +482,9 @@ Now back to the validation error about global context. With the freshly gained k
 
    Now when someone requests the same article via both of those endpoints, we could end up with both returning the same output:
 
-    1. `GET /article/1` returns `Book` without populating its property `author` property
-    2. `GET /article-with-author/1` returns `Book`, this time with `author` populated
-    3. `GET /article/1` returns `Book`, but this time also with `author` populated
+    1. `GET /article/1` returns `Author` without populating its property `author` property
```
B4nan, line 485: article endpoint is supposed to return article entity, not author entity. should be especially clear in the 2. and 3. items here - author entity with author property populated does not make much sense

PR #5464, opened 2024-04-17
fix(core): fix folder-based discovery for multiple entities in single file

xmajzel:
### Describe the bug

After upgrading from `v5` to `v6`, initializing ORM doesn't work, throwing following stack trace:

```bash
TypeError: Cannot read properties of undefined (reading 'length')
    at MetadataDiscovery.defineBaseEntityProperties (.../node_modules/@mikro-orm/core/metadata/MetadataDiscovery.js:733:75)
    at .../node_modules/@mikro-orm/core/metadata/MetadataDiscovery.js:90:39
    at MetadataDiscovery.processDiscoveredEntities (.../node_modules/@mikro-orm/core/metadata/MetadataDiscovery.js:90:18)
    at MetadataDiscovery.discover (.../node_modules/@mikro-orm/core/metadata/MetadataDiscovery.js:47:14)
    at async MikroORM.discoverEntities (.../node_modules/@mikro-orm/core/MikroORM.js:172:25)
    at async Function.init (.../node_modules/@mikro-orm/core/MikroORM.js:44:9)
    at async Function.handleMigrationCommand (.../node_modules/@mikro-orm/cli/commands/MigrationCommandFactory.js:83:21)
```
We went together with @Ruby184 through code and found out that that it was caused by `this.metadata.set(name, Utils.copy(MetadataStorage.getMetadata(name, path), false));` in `MetadataDiscovery.discoverDirectories` function.
- This is problematic when there are multiple entities in one file.
- As a key in local `this.metadata` is `const name = this.namingStrategy.getClassName(filename);` used, which overwrites local metadata, when iterating multiple entities in file.
  - Overriding applies if the name derived from the file name is the same as the name of the class entity.

**Resolution:**

Global metadata initialisation (`MetadataStorage.getMetadata(entity.meta.className, filepath)`) is important if the user uses `EntitySchema` and `TsMorphMetadataProvider ` for entities definition, so as a solution we moved this code to the `MetadataDiscovery.getSchema` function and refactored the `MetadataDiscovery.prepare` function to return `EntitySchema` even if it is located in `EntitySchema.REGISTRY`.
- **See the code for details**

However, we are not sure, that setting metadata locally is necessary

```ts
  private getSchema<T>(entity: Constructor<T> | EntitySchema<T>, filepath?: string): EntitySchema<T> {
    if (entity instanceof EntitySchema) {
      if (filepath) {
        const meta = Utils.copy(MetadataStorage.getMetadata(entity.meta.className, filepath), false);
        // Is following line neccessary?
        this.metadata.set(entity.meta.className, meta);
      }
      return entity;
    }
...
```
because they will be overwritten latter in the `MetadataDiscovery.discoverDirectories` function as follows:

```ts
...

const meta = schema.init().meta;
this.metadata.set(meta.className, meta);

...
```

### Reproduction

- We added clear test for reproduction

None

### MikroORM version

6.2.1

### Node.js version

v20.11.1

### Operating system

Mac OS Sonoma 14.4.1 (23E224)

### Validations

### DIFF
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -267,9 +267,8 @@ export class MetadataDiscovery {
           continue;
         }
 
-        this.metadata.set(name, Utils.copy(MetadataStorage.getMetadata(name, path), false));
         const entity = this.prepare(target) as Constructor<AnyEntity>;
-        const schema = this.getSchema(entity);
+        const schema = this.getSchema(entity, path);
         const meta = schema.init().meta;
         this.metadata.set(meta.className, meta);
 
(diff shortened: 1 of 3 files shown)


B4nan (approved):
thanks!

Issue #5460, opened 2024-04-17
Mysql auto_increment_increment setting is not work.


pius712:
### Describe the bug

If you set mysql variable auto_increment_increment to 2, 
When saving multiple entities, pk should be incremented by 2. 
However, if you use the following code in the MySqlDriver file, it will be set to 1.

```
async getAutoIncrementIncrement(ctx) {
        if (this.autoIncrementIncrement == null) {
            // the increment step may differ when running a cluster, see https://github.com/mikro-orm/mikro-orm/issues/3828
            const res = await this.connection.execute(`show variables like 'auto_increment_increment'`, [], 'get', ctx, { enabled: false });
            /* istanbul ignore next */
            this.autoIncrementIncrement = res?.auto_increment_increment ? +res?.auto_increment_increment : 1;
        }
        */ return this.autoIncrementIncrement;
    }
```

when I debugging, the res value is like below. 
```
{
  "Variable_name": "auto_increment_increment",
  "Value": "2"
}
```

I want to pull request this, but I don't know why this code is written. So I report this on issue.

### Reproduction

version Info: 
- mysql: 8.0.33   
- @mikro-orm/core": "^6.2.1
- @mikro-orm/mysql": "^6.2.1
- @mikro-orm/nestjs": "^5.2.3

```
import { Entity, PrimaryKey, Property } from '@mikro-orm/core';

export abstract class BaseEntity {
  @PrimaryKey({ type: 'bigint', autoincrement: true })
  id?: string;

  @Property({ type: 'timestamp',  onCreate: () => new Date() })
  createTime: Date = new Date();

  @Property({
    type: 'timestamp',
    onUpdate: () => new Date(),
    nullable: true,
  })
  updateTime?: Date = new Date();
}

@Entity({tableName: 'post'})
export class PostEntity extends BaseEntity {

    @Property({type:'varchar'})
    title:string
    
    constructor(title: string) {
        super();
        this.title = title;
    }
}
```

```
const result = await this.postRepository.save([
      new PostEntity('title'),
      new PostEntity('title2'),
      new PostEntity('title3'),
    ]);
    return result;
  }

@Injectable()
export class PostRepository {
  constructor(
    @InjectRepository(PostEntity)
    private readonly postRepository: EntityRepository<PostEntity>,
  ) {}

  async save(entity: PostEntity | PostEntity[]): Promise<PostEntity | PostEntity[]> {
[...]

PR #5465, opened 2024-04-17
chore: add comment on auto increment fix

pius712:
I reported this bug (https://github.com/mikro-orm/mikro-orm/issues/5460) and commented on the issue for the line that was fixed.


B4nan:
we dont need a comment pointing to an issue - this is all present in git history anyway, so closing

PR #5470, opened 2024-04-17
fix(migrator): store snapshot only after migration is generated

Ruby184:
### The problem
When using custom migration fileName function which forces using name (throws) as in the example in documentaion https://mikro-orm.io/docs/migrations#using-custom-migration-names and having snapshots enabled, causes the update of snapshot. Second retry with using name is not generating migration, because the snapshot has already been updated and no changes are found.

 ```typescript
 migrations: {
  fileName: (timestamp: string, name?: string) => {
    // force user to provide the name, otherwise you would end up with `Migration20230421212713_undefined`
    if (!name) {
      throw new Error('Specify migration name via `mikro-orm migration:create --name=...`');
    }

    return `Migration${timestamp}_${name}`;
  },
  snapshot: true,
},
 ```

### Description of the solution
Delaying update of snapshot after migration file is actually generated. Thus, updating it only after successful call to `this.generator.generate` which, under the hood, calls custom `fileName` function which is throwing error.

### DIFF
--- a/packages/migrations/src/Migrator.ts
+++ b/packages/migrations/src/Migrator.ts
@@ -80,8 +80,9 @@ export class Migrator implements IMigrator {
       return { fileName: '', code: '', diff };
     }
 
-    await this.storeCurrentSchema();
     const migration = await this.generator.generate(diff, path, name);
+    await this.storeCurrentSchema();
+
     return {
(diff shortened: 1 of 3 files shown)


B4nan (approved):
thanks, good catch

PR #5473, opened 2024-04-18
feat(entity-generator): extend filtering options for EntityGenerator

### LINKED ISSUE #5469: feat: extended filtering options for EntityGenerator
I have to create entity models for small subset of tables.
Currently I'd have to use `skipTables` with very long array of table names to get only the entities I need.
Also some of those table have constant prefix which can be matched with the pattern.

I'd like to have another option `takeTables` to specify tables to generate entities for.
I'd also like to have the ability to specify `RegExp` to match table names to take/skip.

Getting list of all tables, filtering them out and using in `skipTables`.

**Additional context**
I could implement this feature.

patrykbojczuk:
This adds the `takeTables` option to `EntityGenerator`'s `GenerateOptions` and support for matching table and column names using regex.

This allows to generate entity models only for tables specified in the `takeTables` as well as use an array of `string`s and `RegExp`s in the `takeTables`, `skipTables` and `skipColumns`.
 
 Resolves #5469


tests/features/entity-generator/EntityGenerator.mysql.test.ts
```diff
@@ -27,6 +27,43 @@ describe('EntityGenerator', () => {
     });
     expect(dump).toMatchSnapshot('mysql-entity-dump-skipTables');
     await expect(pathExists('./temp/entities/Author2.ts')).resolves.toBe(true);
+    await expect(pathExists('./temp/entities/Test2.ts')).resolves.toBe(false);
```
B4nan, line 30: let's revert those and keep only the postgres tests, the changes are not driver-specific and this will only slow down tests

B4nan (changes requested):
thanks, just a few nits

packages/entity-generator/src/EntityGenerator.ts
```diff
@@ -132,6 +132,20 @@ export class EntityGenerator {
     return metadata;
   }
 
+  private matchName(name: string, nameToMatch: string | RegExp) {
+  return typeof nameToMatch === 'string'
+    ? name.toLocaleLowerCase() === nameToMatch.toLocaleLowerCase()
+    : nameToMatch.test(name);
+}
+
+
```
B4nan, line 141:
```suggestion
  private matchName(name: string, nameToMatch: string | RegExp) {
    return typeof nameToMatch === 'string'
      ? name.toLocaleLowerCase() === nameToMatch.toLocaleLowerCase()
      : nameToMatch.test(name);
  }

```

PR #5480, opened 2024-04-19
docs: Fix typo in 00-introduction.md

nicorobo:
Removing "provide" here as it seems to have snuck in on accident.


B4nan (approved):
thanks

PR #5481, opened 2024-04-19
fix(core): ignore serialization options when using `toPOJO()`

### LINKED ISSUE #5479: toPOJO() should ignore serializedName option in @Property decorator
### Describe the bug

If we use the `serializedName` option when describing the entity field, then we cannot use the toPOJO method to save the entity to the cache and then restore it via repository.merge. It is placed in the cache with the changed field name and the relation is discarded when merging.

### Reproduction

```ts
   @OneToMany(() => PlaceUser, 'user', {
        serializedName: 'places',
        serializer: (t: Collection<PlaceUser>) => {
            if (!t.isInitialized()) {
                return undefined;
            }
            return t.map(p => ({
                ...p.place.toObject(),
                verified: p.verified,
            }));
        },
    })
    placesPivot = new Collection<PlaceUser>(this);

    @ManyToMany(() => Place, p => p.users, { hidden: true })
    places = new Collection<Place>(this);
```

@mikro-orm/postgresql

### MikroORM version

6.2.1

### Node.js version

20.12.1

### Operating system

Windows

### Validations

volk234:
Closes #5479

### DIFF
--- a/packages/core/src/serialization/EntityTransformer.ts
+++ b/packages/core/src/serialization/EntityTransformer.ts
@@ -88,7 +88,7 @@ export class EntityTransformer {
         return [prop, val] as const;
       })
       .filter(([, value]) => typeof value !== 'undefined')
-      .forEach(([prop, value]) => ret[this.propertyName(meta, prop!, wrapped.__platform) as any] = value as any);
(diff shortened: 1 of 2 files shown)


volk234:
I think we should also skip using custom serializers when using toPOJO()

B4nan (approved):
agreed we should ignore both of those options, so did the changes on top of yours in this PR

volkov-aleksandr:
Thx,
You are the best

PR #5502, opened 2024-04-24
fix(query-builder): enhance  query builder's where type

pius712:
like code below makes mistake. So I suguesst typings omit where after already using where

```
const qb = createQueryBuilder().where(...somecondition)

if(dynamicCondition) { 
  qb.where(...somecondition)
}
```

### DIFF
--- a/packages/knex/src/query/QueryBuilder.ts
+++ b/packages/knex/src/query/QueryBuilder.ts
@@ -345,9 +345,9 @@ export class QueryBuilder<T extends object = AnyEntity> {
     return this;
   }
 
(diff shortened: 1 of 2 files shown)


pius712:
no operation is specified, latter where cluase overwrite former.

B4nan:
I actually consider the change you are proposing as breaking. Changing the implementation would be even more breaking, somebody might rely on this for whatever reason.

This is not something specific to `qb.where`, many other methods like `qb.select` or `qb.orderBy` behave exactly the same.

B4nan:
Nope, right now I don't plan to limit the API like this. Maybe in v7 we could do it and add an explicit method to reset the values.

PR #5514, opened 2024-04-26
fix(cli): use `module: 'nodenext'` when registering ts-node

vinialbano:
Update configuration to allow `mikro-orm-esm` to work with newer versions of tsconfig.

#5427

### DIFF
--- a/packages/core/src/utils/ConfigurationLoader.ts
+++ b/packages/core/src/utils/ConfigurationLoader.ts
@@ -1,14 +1,14 @@
 import dotenv from 'dotenv';
 import { pathExistsSync, readJSONSync, realpathSync } from 'fs-extra';
-import { isAbsolute, join } from 'path';
 import { platform } from 'os';
+import { isAbsolute, join } from 'path';
 import { fileURLToPath } from 'url';
+import type { EntityManager } from '../EntityManager';
 import type { EntityManagerType, IDatabaseDriver } from '../drivers';
+import { colors } from '../logging/colors';
+import type { Dictionary } from '../typings';
 import { Configuration, type Options } from './Configuration';
 import { Utils } from './Utils';
-import type { Dictionary } from '../typings';
-import { colors } from '../logging/colors';
-import type { EntityManager } from '../EntityManager';
 
 /**
  * @internal
@@ -144,7 +144,7 @@ export class ConfigurationLoader {
       project: tsConfigPath,
       transpileOnly: true,
       compilerOptions: {
-        module: 'commonjs',
+        module: 'nodenext',
       },
     }).config;
 


B4nan:
are you sure this doesn't break CJS projects? i think we could maybe remove the overrides completely, as this should be now all taken from the project's tsconfig. maybe we could keep the old behavior when there is no tsconfig found.

B4nan:
Ok, now I am finally able to reproduce this. You confused me with the `mikro-orm-esm`, to me that works just fine with ESM projects (as the https://github.com/mikro-orm/guide). I had to use `mikro-orm` command instead and remove `type: 'module'` from the package.json file to see that error.

vinialbano:
Awesome! Thanks so much, @B4nan. That will help a lot in a project that I'm working on. We were manually fixing in node_modules so far haha

Far4way:
This change breaks CLI when using mikro-orm as part of a project that still uses "moduleResolution": "node".

Shouldn't a breaking change be part of a major update ?

B4nan:
It works fine on my end, can you provide a complete repro? I wouldn't do it if I would consider it breaking, I am sure we can make this work conditionally if needed.

Far4way:
I can't provide a link to the repo where I have the issue, but I tried the following steps (10min to complete a reproduction) and had the same issue :

1. Create a monorepo with nx
2. setup an angular app
3. setup a nestjs app
4. install mikro/orm 
5. 'try' to use cli

I did end up finding a solution where I indeed modified the main tsconfig and overwrote the main file in my nestjs and angular apps.

But again, this change is breaking, and I shouldn't have to modify config files because I updated to the last minor of mikro-orm.

I don't think that I'm familiar enough to help you make this whole thing conditional, but that sure would be great!

B4nan:
> I can't provide a link to the repo where I have the issue

I don't want your repository, I don't care about your app, I want a minimal example that shows the problem.

> Create a monorepo with nx

I never used nx and I am not interested in learning it now. 

> But again, this change is breaking, and I shouldn't have to modify config files because I updated to the last minor of mikro-orm.

Without a repro this is just a claim, it works just fine on my end 🤷 

So again, please provide a reproduction if you think it's breaking, then we can do something about it.

Far4way:
Here is a fork of the nestjs-example-app (quickest way to reproduce the problem) :  [nestjs-example-app_6.2.9](https://github.com/Far4way/nestjs-example-app_6.2.9).

The only change made to the forked repo is adding `"moduleResolution": "node"` to `compilerOptions` in the `tsconfig.json`. 

In 6.2.8, using `mikro-orm debug` we get the following output : 
```Node
$ npx mikro-orm debug
Current MikroORM CLI configuration
 - dependencies:
   - mikro-orm 6.2.8
   - node 20.11.0
   - knex not-found
   - mysql2 not-found
   - typescript not-found
 - package.json found
 - ts-node enabled
 - searched config paths:
   - /nestjs-example-app_6.2.9/src/mikro-orm.config.ts (found)
   - /nestjs-example-app_6.2.9/dist/mikro-orm.config.js (not found)
   - /nestjs-example-app_6.2.9/mikro-orm.config.ts (not found)
   - /nestjs-example-app_6.2.9/src/mikro-orm.config.js (not found)
   - /nestjs-example-app_6.2.9/mikro-orm.config.js (not found)
 - configuration found
 - database connection failed (connect ECONNREFUSED 127.0.0.1:3307)
 - will use `entities` array (contains 5 references and 0 paths)
```

In 6.2.9 : 
```Node
$ npx mikro-orm debug
Current MikroORM CLI configuration
 - dependencies:
   - mikro-orm 6.2.9
   - node 20.11.0
   - typescript not-found
 - package.json found
 - ts-node enabled
 - searched config paths:
   - /nestjs-example-app_6.2.9/src/mikro-orm.config.ts (found)
   - /nestjs-example-app_6.2.9/dist/mikro-orm.config.js (not found)
   - /nestjs-example-app_6.2.9/mikro-orm.config.ts (not found)
   - /nestjs-example-app_6.2.9/src/mikro-orm.config.js (not found)
   - /nestjs-example-app_6.2.9/mikro-orm.config.js (not found)
- configuration not found (⨯ Unable to compile TypeScript:
error TS5109: Option 'moduleResolution' must be set to 'NodeNext' (or left unspecified) when option 'module' is set to 'NodeNext'.
)
```

Hoping this much is good enough !

B4nan:
This should help d5d7f381363ab776b3c3af04a925a5547dbcdb65

I really don't understand what happened, it was working just fine and now I can reproduce in the same repo where it worked before.

PR #5537, opened 2024-05-04
feat(postgres): allow specifying deferred mode on unique constraints

B4nan:
Allows specifying `deferMode` on a unique constraint to get around issues with batch updates mainly.

### DIFF
--- a/docs/docs/decorators.md
+++ b/docs/docs/decorators.md
@@ -166,11 +166,12 @@ Use `@Index()` to create an index, or `@Unique()` to create unique constraint. Y
 
 See [Defining Entities](./defining-entities.md#indexes).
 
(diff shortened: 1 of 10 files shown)


DASPRiD:
Only after working on https://github.com/mikro-orm/mikro-orm/pull/5641 I spotted that you worked on this PR. Looking at the source, it seems we have some overlap, though this is specifically handling manually created unique constraints, while my PR handles unique constraints created automatically for 1:1 relations. How should we proceed on this? :)

B4nan:
i can merge yours first and release to stable, since its a fix. i am keeping this one for v6.3, which i'd like to start working on soon (this weekend probably), i also have #5482 which is waiting for the exact same reason, and have some other small things i want in there that will deserve a blog post

PR #5545, opened 2024-05-06
fix(core): detect path from decorator for each class only once

Ruby184:
### Fix for bun path detection
Related discussion here: https://github.com/mikro-orm/mikro-orm/discussions/5496
In [this commit](https://github.com/mikro-orm/mikro-orm/commit/271ff7dfe95ded8f2821b7a97a10ddfe409b1513) you changed the condition because of coverage. This is incorrect because it guards against going out of stack array bounds - I have added the test for this case. Also, I removed no longer needed condition.

### Performance
Before the changes, the path was detected for every used decorator on the class and overwritten. This is not needed because the path used as a key should be stable and should not change for a given class. So I introduced the new symbol `MetadataStorage.PATH_SYMBOL`, which will hold the path and check if it exists for a given class using `Object.hasOwn` (this also works for class inheritance so the path is not taken from the base class but defined for each child).

Moreover, as a side effect, this also allows defining a path on the class by the developer (maybe in cases like when Bun was not working), but it should not be needed and used only as a workaround:

```typescript
import { Entity, MetadataStorage } from '@mikro-orm/core';

@Entity()
class User {
  static [MetadataStorage.PATH_SYMBOL] = import.meta.filename // or __filename for commonjs

  @PrimaryKey()
  id!: number;

  @Property()
  username!: string;
}
```

### DIFF
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -329,7 +329,7 @@ export class MetadataDiscovery {
     return entity as EntityClass<T>;
   }
 
-  private getSchema<T>(entity: Constructor<T> | EntitySchema<T>, filepath?: string): EntitySchema<T> {
+  private getSchema<T>(entity: (Constructor<T> & { [MetadataStorage.PATH_SYMBOL]?: string }) | EntitySchema<T>, filepath?: string): EntitySchema<T> {
     if (entity instanceof EntitySchema) {
       if (filepath) {
         // initialize global metadata for given entity
@@ -339,7 +339,7 @@ export class MetadataDiscovery {
       return entity;
     }
 
-    const path = (entity as Dictionary).__path;
+    const path = entity[MetadataStorage.PATH_SYMBOL];
 
     if (path) {
       const meta = Utils.copy(MetadataStorage.getMetadata(entity.name, path), false);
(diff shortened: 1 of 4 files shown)


B4nan:
> Moreover, as a side effect, this also allows defining a path on the class by the developer

This is great, I've been thinking about introducing this for ages. We should document this somewhere.

Ruby184:
> This is great, I've been thinking about introducing this for ages. We should document this somewhere.

Sure, but there is a caveat. I think people will expect to define the path in the base class and inherit it. However, they need to copy-paste the code for each class, which is slightly counter-intuitive. This comes from the fact that `import.meta.filename` or `__filename` is scoped to a module (file), and we expect the path to be defined for each class because it usually differs. It's not worth making logic to detect if we can inherit it from the parent, maybe if it is defined as a getter, but we can document that. I am open to suggestions.

B4nan:
Yes, that is surely something to document. I don't think it makes sense to try to validate this, as having the entities in a single file is still a valid case.

B4nan:
I guess this could go to https://mikro-orm.io/docs/metadata-providers.

Also, do you think this should be a `perf` commit? Does it have a measurable difference? Maybe a `fix` would be better, or even a `feat` since we are adding a new feature that will have its documentation.

B4nan:
Btw does this help with the production build issue in bun mentioned in that discussion?

B4nan:
I'll merge this in as a fix, as I don't want to hold it for no reason, but I would still appreciate response to my previous comment as well as a PR mentioning this in the docs somewhere.

PR #5566, opened 2024-05-13
fix(query-builder): don't remove joins used by other joins during pagination

### LINKED ISSUE #5565: Join gets removed when the populate doesn't include the nested orderBy relation
### Describe the bug

When the orderBy is nested and the populate does not include the nested entities it will throw out the joins needed for the orderBy to work.
This is likely due to https://github.com/mikro-orm/mikro-orm/commit/d228976d9ca4b8a2e69361ad31f63ef88977ee9e

### Reproduction

https://github.com/mikro-orm/mikro-orm/pull/5566

@mikro-orm/postgresql

### MikroORM version

6.2.5

### Node.js version

18.x

### Operating system

MacOS

### Validations

AdriVanHoudt:
Fixes  #5565

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -3,3 +3,5 @@ compressionLevel: mixed
 enableGlobalCache: false
 
 nodeLinker: node-modules
+
+yarnPath: .yarn/releases/yarn-4.2.2.cjs
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -3,5 +3,3 @@ compressionLevel: mixed
 enableGlobalCache: false
 
 nodeLinker: node-modules
-
-yarnPath: .yarn/releases/yarn-4.2.2.cjs
--- a/packages/knex/src/query/QueryBuilder.ts
+++ b/packages/knex/src/query/QueryBuilder.ts
@@ -1583,7 +1583,7 @@ export class QueryBuilder<T extends object = AnyEntity> {
     const orderByAliases = this._orderBy
       .flatMap(hint => Object.keys(hint))
       .map(k => k.split('.')[0]);
-      const ownerAliases = Object.values(this._joins).map(join => join.ownerAlias);
+    const ownerAliases = Object.values(this._joins).map(join => join.ownerAlias);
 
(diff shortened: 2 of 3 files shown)


.yarnrc.yml
```diff
@@ -3,3 +3,5 @@ compressionLevel: mixed
 enableGlobalCache: false
 
 nodeLinker: node-modules
+
+yarnPath: .yarn/releases/yarn-4.2.2.cjs
```
B4nan, line 7:
```suggestion
```

packages/knex/src/query/QueryBuilder.ts
```diff
@@ -1583,6 +1583,7 @@ export class QueryBuilder<T extends object = AnyEntity> {
     const orderByAliases = this._orderBy
       .flatMap(hint => Object.keys(hint))
       .map(k => k.split('.')[0]);
+      const ownerAliases = Object.values(this._joins).map(join => join.ownerAlias);
```
B4nan, line 1586:
```suggestion
    const ownerAliases = Object.values(this._joins).map(join => join.ownerAlias);
```

.yarnrc.yml
B4nan: you should be using corepack, i guess you skipped this step if this was added automatically

packages/knex/src/query/QueryBuilder.ts
```diff
@@ -1596,11 +1596,26 @@ export class QueryBuilder<T extends object = AnyEntity> {
...
+
+    for (const orderByAlias of orderByAliases) {
+      addParentAlias(orderByAlias);
+    }
+
+    for (const [key, join] of joins) {
       const path = join.path?.replace(/\[populate]|\[pivot]|:ref/g, '').replace(new RegExp(`^${meta.className}.`), '');
 
-      if (!populate.has(path ?? '') && !orderByAliases.includes(join.alias)) {
+      if (!populate.has(path ?? '') && !orderByAliases.includes(join.alias) && !ownerAliases.includes(join.alias)) {
```
B4nan, line 1618:
```suggestion
      if (!populate.has(path ?? '') && !orderByAliases.includes(join.alias)) {
```

PR #5615, opened 2024-05-28
refactor(knex): moved better-sqlite, libsql and mariadb dialects into their packages

boenrobot:
This allows tree shaking bundlers (e.g. Vite) to exclude those dialects' clients, when they are not in use, without needing an explicit config.

### DIFF
--- a/packages/better-sqlite/src/BetterSqliteConnection.ts
+++ b/packages/better-sqlite/src/BetterSqliteConnection.ts
@@ -1,4 +1,5 @@
-import { BetterSqliteKnexDialect, BaseSqliteConnection } from '@mikro-orm/knex';
+import { BaseSqliteConnection } from '@mikro-orm/knex';
+import { BetterSqliteKnexDialect } from './BetterSqliteKnexDialect';
 
 export class BetterSqliteConnection extends BaseSqliteConnection {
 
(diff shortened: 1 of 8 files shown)


B4nan:
Hmm, the motivation behind this was that I wanted to introduce a new driver based on kysely, and there I would rather want to have a single package for all the supported dialects (as that would mean doubling the number of the packages basically, making everything harder to document and reason about).

I would rather try to find another way around this than moving things around. In the end, we already have the `MonkeyPatchable` export which is touching all the dialects and that is clearly not a problem.

B4nan:
Right, being able to extend the dialect was another motivation, so moving them elsewhere actually feels breaking.

PodaruDragos:
> Hmm, the motivation behind this was that I wanted to introduce a new driver based on kysely, and there I would rather want to have a single package for all the supported dialects (as that would mean doubling the number of the packages basically, making everything harder to document and reason about).
> 
> I would rather try to find another way around this than moving things around. In the end, we already have the `MonkeyPatchable` export which is touching all the dialects and that is clearly not a problem.

wow this would be such a great improvement over `knex`. In this scenario how would this have worked ? 
did you plan to remove the usage of `knex` from each package that uses it and replace it  ?

B4nan:
The new package (probably `@mikro-orm/kysely`) would have all the dialect specific code as well as the ORM drivers, and you would install the underlying libraries as peer deps just like you do with kysely/knex when used directly.

But it would be quite a lot of work, I might just swap knex with kysely in v7 and keep the structure as is. I am more and more tired with knex, its not really maintained anymore, and their dialects are often not well written and are missing features. I keep polyfilling what knex is supposed to be doing, and I already reached the level of "I am doing more than knex" instead of "I am getting a lot out of knex". I am just a bit afraid that some people might want to stay with knex for whatever reason.

I guess I will open RFC for this to see what the community thinks, as for me it would be much simpler to swap knex with kysely instead of trying to provide support for both.

Issue #3743, opened 2022-11-15
Replace direct require() calls with an indirection function


parisholley:
**Is your feature request related to a problem? Please describe.**
Many new JS stacks use some type compiler (swc, esbuild, babel) and will often have some setting enabled that will attempt to in-line all code that is included via require(). That is problematic for this project since things like database drivers are only included if needed.

**Describe alternatives you've considered**
You can try to figure out how to exclude or mock the module name but for some projects (such as remix), the compiler stack isn't exposed to you unless you patch the project.

B4nan:
I don't think we have require calls that cannot be already gotten around (e.g. for drivers you can pass the driver class via `driver` option instead of the `type` option). You can already use MikroORM with esbuild/swc, as long as you are fine with not using decorators for metadata.

Or is this about removing the require calls, so you don't need special handling for them (e.g. as in webpack)? We need defaults that work universally, so I don't think that's doable in near future.

B4nan:
Looking at the PR, I believe that is the exact way to break things. This won't work in ESM project at all, if I'm not mistaken - will try that tomorrow, I have one ESM example project.

FYI we also have `Utils.requireFrom` and `Utils.tryRequire`, those are tested in various types of project.

parisholley:
> as long as you are fine with not using decorators for metadata

decorators work with swc, you just have to define `type` for every property and i have a `registerMikro` function for gathering classes in an array and passing to my mikro init logic

> that is the exact way to break things

i'm not familiar enough with ESM to say

> we also have `Utils.requireFrom` and `Utils.tryRequire`

alright, i can change if you think this isn't a breaking change

parisholley:
take this comment with a grain of salt as I am biased :) but using webpack to bundle server-side node_modules seems like a smell? i could get trying tree-shake dependencies or maybe having a single file to say, deploy to a serverless function provider, but even in SSR situations node_modules i find to be separate most of the time and a lot of projects meant for node aren't tree-shakable to begin with. 

@B4nan on flip side, obviously a breaking change, could we not just have the user pass an instance/class/etc to config instead of driver "names" (same thing with migrations), and this problem goes away?

B4nan:
[...]

PR #5618, opened 2024-05-28
refactor(knex): moved all knex patches into a dedicated package @mikro-orm/knex-patches

boenrobot:
This enables users to be independent of driver, by using types from @mikro-orm/knex, and do so even in the presence of bundlers like Vite.

Custom driver implementations meanwhile will need to import from @mikro-orm/knex classes, as before, and if the custom driver requires a new knex dialect, also import from @mikro-orm/knex-patches.


B4nan:
I would call this even more breaking than the other PR, since you are moving more exports around.

B4nan:
Also it feels weird that this would fix it, I would expect the very same problem. Moving things to another package you will depend on directly shouldn't change anything about the import calls being present.

B4nan:
> This one in theory should, given that knex is a down dependency that also ultimately does a require() for all drivers.

Yeah and this was always a problem, at least with webpack, you always had to explicitly say those are external dependencies (those you don't use, so all drivers except the one you have installed). Maybe things changed or its different with vite, but it sounds quite weird to me.

Anyway, I would appreciate a repro for this so I can take a look too, maybe we can find some other way around it - like using dynamic strings in the imports instead of static ones which vite tries to bundle automatically.

B4nan:
> For all its smarts, I can't believe Vite isn't smart enough to pick that one up.

Maybe it's even intentional.

Note that even [wrapping the require/import call with a custom method will help](https://github.com/parisholley/mikro-orm/commit/e633a83966d684f4d126109ff090aad67bcd7cd5), bundlers literally seek the static require calls and skip anything tiny bit dynamic. There was some discussion around this topic in #3743.

PR #5630, opened 2024-05-30
fix(core): ensure correct mapping of related columns to owning entity fields for complex relations

### LINKED ISSUE #5629: array out of bounds error with columns shared in composite PK and FK
### Describe the bug

related to #5622
columns shared in composite PK/FK are breaking `getResultMapper` in EntityComparator.ts

the shorthand explanation is that the line `return propName(fieldNames[idx]);` is causing an array OOB error

image
in the screenshot above, it's trying to access index 3 in an array of length 3. i'll put up a repro test and see if I can figure out the bug. this bug causes querybuilder and EntityRepository.find to break when trying to retrieve an entity by its PK

### Reproduction
https://github.com/mikro-orm/mikro-orm/pull/5630/files

@mikro-orm/postgresql

### MikroORM version

6.2.8

### Node.js version

20.11

### Operating system

macos

### Validations

russell-dot-js:
This change fixes #5629, and no tests are failing, but I don't understand enough of what's happening here to be sure it didn't break something else

I also fixed a race condition where multiple test suites were contending for the same temp files by updating the contending test suites to use their own temp directories


packages/core/src/utils/EntityComparator.ts
```diff
@@ -268,14 +268,19 @@ export class EntityComparator {
     // respects nested composite keys, e.g. `[1, [2, 3]]`
     const createCompositeKeyArray = (prop: EntityProperty, fieldNames = prop.fieldNames, idx = 0): string => {
       if (!prop.targetMeta) {
-        return propName(fieldNames[idx++]);
+        return propName(fieldNames[idx]);
```
russell-dot-js, line 271: incrementing index after returning did nothing

packages/core/src/utils/EntityComparator.ts
```diff
@@ -268,14 +268,19 @@ export class EntityComparator {
...
+      const foundFields = new Set<string>();
 
       for (const pk of prop.targetMeta.getPrimaryProps()) {
+        const oldSize = foundFields.size;
+
+        pk.fieldNames.forEach(field => foundFields.add(field));
+        const added = foundFields.size - oldSize;
         parts.push(createCompositeKeyArray(pk, fieldNames, idx));
-        idx += pk.fieldNames.length;
+        idx += added;
```
russell-dot-js, line 283: this seems to fix the issue... but so did changing the line to `idx ++;`
need someone smarter to check this out and help determine the "right" fix. even though the test passes, I don't have enough context (why was it important to increment idx by the # of fieldNames?) to feel totally safe

overall it would be good to see more unit tests because this code is pretty scary

tests/features/composite-keys/GH5629.test.ts
```diff
@@ -0,0 +1,100 @@
...
+@Entity()
+class SomethingThatBelongsToSomething extends BaseEntity {
+
+  [PrimaryKeyProp]?: ['tenant', 'something', 'id'];
+
+  @ManyToOne(() => Tenant, { fieldName: 'tenant_id', primary: true })
+  tenant!: Tenant;
+
+  @ManyToOne(() => Something, {
+    fieldName: 'something_id',
```
B4nan, line 34: FYI this is a no-op and could potentially confuse either you or the ORM. Composite relation properties never have 1 field name but multiple join columns, you always have multiple field names here. This option is most probably ignored, otherwise it would error completely. Just keep the join columns, as that is what is used for owning sides anyway.

packages/core/src/utils/EntityComparator.ts
B4nan: this looks "good" to me, support for sharing columns in FKs is done exactly this way - we ignore the ones that are duplicated. 

i think it was working for simple cases, but yours is different - you try to have actual nested composite keys, while up until now it was only meant for composite keys that share the tenant and not the rest.

i agree more tests would be great, PR welcome for that. if you want to be sure, just change this test to be more complex - have two owned PKs on each entity, then you should see that doing `++` wont be enough.

even better, we could try to find a proper support for this. i would love to simplify the internals of handing nested composite keys.

tests/features/composite-keys/GH5629.test.ts
russell-dot-js: thank you for the clarity. it's nice to be able to see what the fields are when using them for FK's, but pointless if redunant. 

we can probably update the type for the annotation arg so that joinColumns and fieldName are mutually exclusive then?

packages/core/src/utils/EntityComparator.ts
russell-dot-js: adding a more complex relationship caused another issue:

```
 select `s0`.*
 from `something_that_belongs_x4` as `s0`
 where `s0`.`tenant_id` = '1'
 and (`s0`.`tenant_id`, `s0`.`something_id`) in ( values ('1', '2'))
 and (`s0`.`tenant_id`, `s0`.`something_id`, `s0`.`x1_id`) in ( values ('1', '1', '2', '3'))
 and (`s0`.`tenant_id`, `s0`.`something_id`, `s0`.`x1_id`, `s0`.`x2_id`) in ( values ('1', '1', '2', '1', '1', '2', '3', '4'))
 and (`s0`.`tenant_id`, `s0`.`something_id`, `s0`.`x1_id`, `s0`.`x2_id`, `s0`.`x3_id`) in ( values ('1', '1', '2', '1', '1', '2', '3', '1', '1', '2', '1', '1', '2', '3', '4', '5'))
 limit 1
```

the # of columns in values doesn't match the number of columns in the where. i fixed that but broke other tests... will keep noodling on it

russell-dot-js: it seems that createCompositeKeyArray is important enough that it should be removed from the getResultMapper closure and tested independently

tests/features/composite-keys/__snapshots__/GH5629.test.ts.snap
```diff
@@ -0,0 +1,80 @@
...
+      "something_id",
+    ],
+  },
+  {
+    "compositeKey": "[result.tenant_id, [result.something_id, result.something_id], result.x1_id]",
+    "fieldNames": [
+      "tenant_id",
+      "something_id",
+      "x1_id",
+    ],
```
russell-dot-js, line 58: @B4nan all the tests are passing now, but check this out
see the generated composite key is  `[result.tenant_id, [result.something_id, result.something_id], result.x1_id]`
I'd expect `[result.tenant_id, [result.tenant_id, result.something_id], result.x1_id]` instead

it just so happens that the values align with what's generated here, e.g. `[1, [2, 2], 3]` so this being broken aligns with the broken where clause generated by find

forgive my ignorance, but I feel like we should be able to simplify it to unique keys -> unique values. if tenant_id is used in multiple foreign keys, it doesn't really matter, because the fact that it matches for the first one implies that it matches for the rest

so i'd think we could simplify our WHERE clauses to
`WHERE tenant_id = 1 AND something_id = 2 AND x1_id = 3` 
rather than
` WHERE tenant_id = 1 AND (tenant_id, something_id) IN (1, 2) AND (tenant_id, something_id, x1_id) IN (1,2,3)`

where does this assumption break?

russell-dot-js: ah, in its current state it's actually generating in invalid mapper - the values in the nested entities are wrong, e.g.
`ret.x1 = [result.tenant_id, [result.something_id, result.something_id], result.x1_id];`
for "something", it's using "something_id" for the tenant_id. that's why a strict comparison e.g. `expect(x4.x2).toBe(x2)` was failing

tests/features/composite-keys/GH5629.test.ts
B4nan: that would be a breaking change, i am sure more people are doing this already, and since its not really breaking anything, why break their code now. also i am sure there is much more cases like this, i am not sure if its feasible to deal with this via discriminated unions (or do you have some other solution in mind?) 

note that there is `fieldName` and `fieldNames` and `name`, all doing the same (one for support of composite keys, one as alias because people kept trying it). and you can use that just fine instead of `joinColumns`, it works both ways. i think i would rather try to think of how to simplify this in v7 and add proper runtime validations.

tests/features/composite-keys/__snapshots__/GH5629.test.ts.snap
B4nan: > [result.something_id, result.something_id]

Yeah, this looks indeed wrong.

> so i'd think we could simplify our WHERE clauses to

Agreed. Let's say that the code was never really written to support this case (sharing columns in multiple properties), it was added later thanks to some small patches. I would greatly appreciate if you could help me make this better/simpler/more correct in any way.

packages/core/src/utils/EntityComparator.ts
```diff
@@ -250,6 +250,32 @@ export class EntityComparator {
     return snapshotGenerator;
   }
 
+  propName = (name: string, parent = 'result') => parent + this.wrap(name);
+
+  // respects nested composite keys, e.g. `[1, [2, 3]]`
+  createCompositeKeyArray = (prop: EntityProperty, fieldNames = prop.fieldNames, idx = 0, foundFields = new Set<string>()): string => {
```
B4nan, line 256: please use regular functions, and make them private or internal

tests/features/composite-keys/__snapshots__/GH5629.test.ts.snap
russell-dot-js: i’m happy to take this on, but would appreciate a little directionality. since the only document on the method is “supports composite keys eg [1,[2,3]]” it’s not clear how the method is supposed to behave

given a prop on entity A with fieldNames [A1,A2,A3] that reference entity B where A1=>B1 and A2=>B2, it appears the response is supposed to be [A1, A2] because that it what will be used in the mapper to set the values for B’s PK… is that correct? or is it supposed to be [B1, B2]

getting this working properly is one part, once the expected behavior is documented with unit tests i’ll crack on getting the mappers and find method playing along. removing redundancy from the where clause is a nice to have IMO but doesn’t fix anything, just makes it “more right”

russell-dot-js: figured it out, should have a working fix tomorrow & tests. spent too long trying to decode the “right” way to use the index on a potentially nested inner entity to find the column name on the root entity when there’s no relationship between the field index on the root and the nested index. really just needed to use fieldNames and referencesFieldNames instead of idx

russell-dot-js: @B4nan could use a little help on the last detail. it seems like based on the type of relationship, i need to find my field in either referencedPKs or referencedColumnNames. it's not clear to me when to look at each one.

at first it appears that for "m:n" relations, i should be using referencedPKs instead of referencedColumnNames:
image

("fieldName" is the field I'm searching for).  test: 'composite entity in m:n relationship, one of entity is composite' in tests/features/composite-keys/composite-keys.mysql.test.ts

however, this doesn't always work:


here's an m:n where I should be looking in referencedColumnNames instead of referencedPK's. it seems like there's some kind of combination of "owner" and potentially "ref" on m:n relations that inform where I should be looking. test 'shared column as composite PK and FK in M:N' in tests/features/sharing-column-in-composite-pk-fk.test.ts

the poor man's version is to just look in both and take which one has a match, but that doesn't seem correct.

there's a chance that posting this comment with the screenshots will suffice as a rubber duck where I can see the pattern, but if this is obvious to you, please share

russell-dot-js: ok, I got it. the use of a custom pivot entity is the differentiator. my fix is pushed on this branch, all tests are passing

this doesn't fix the issue with queries selecting the wrong # of columns e.g. in tests/features/composite-keys/GH5629.test.ts

i'd like to be able to write 
```
  const resolved = (await orm.em.findOne(SomethingThatBelongsX4, {
    tenant,
    something,
    x1,
    x2,
    x3,
  }))!;
```

and believe the orm is supposed to support that, but it currently results in an issue so I use
```
  const resolved = (await orm.em.findOne(SomethingThatBelongsX4, {
    tenant: { id: '1' },
    something: { id: '2' },
    x1: { id: '3' },
    x2: { id: '4' },
    x3: { id: '5' },
  }))!;
```

packages/core/src/utils/EntityComparator.ts
```diff
@@ -250,6 +251,55 @@ export class EntityComparator {
     return snapshotGenerator;
   }
 
+  propName(name: string, parent = 'result') {
+    return parent + this.wrap(name);
+  }
+
+  // respects nested composite keys, e.g. `[1, [2, 3]]`
+  createCompositeKeyArray(prop: EntityProperty, parents: EntityProperty[] = []): string {
```
russell-dot-js, line 264: since this area is highly performance sensitive and node doesn't support TCO, I'd like to do away with the recursion  but the dynamic approach was very difficult to read, and this should outperform the original method considering that createCompositeKeyArray and propName are no longer recompiled for each frame

tests/features/composite-keys/__snapshots__/GH5629.test.ts.snap
B4nan: >  i need to find my field in either referencedPKs or referencedColumnNames

`referencedPKs` points to target entity PK properties (and each of them can be potentially represented by multiple columns if it's a composite FK on its own), while `referencedColumnNames` is a flat array of the final columns in the target table

don't be too attached to what you see in the metadata, the problem could be there as well (not saying it is the case now, just a general warning)

packages/core/src/utils/EntityComparator.ts
B4nan: the perf consideration is more about the compiled code, its fine to use recursion in methods that generate it. that's also the reason why having the helper method inlined or not is not really relevant, this code gets executed exactly once per entity type and cached.

packages/core/src/utils/EntityComparator.ts
```diff
@@ -250,6 +251,62 @@ export class EntityComparator {
...
+    if (!prop.targetMeta) {
+      let fieldName = prop.fieldNames[0];
+      // traverse all parents, mapping my field name to each parent's field name until we reach the root
+      for (let i = parents.length - 1; i >= 0; i--) {
+        const parent = parents[i];
+
+        // look for the field name in the parent's referencedColumnNames, except for M:N without a custom pivot entity
+        //   (in that case, the field names are the referenced PK's )
+        const target = parent.kind === ReferenceKind.MANY_TO_MANY && parent.pivotEntity === parent.pivotTable ?
+          parent.referencedPKs : parent.referencedColumnNames;
```
B4nan, line 274: this feels weird, there shouldn't be any difference because of the custom pivot entity (also technically it can have the same name as the table name, so your condition wouldn't catch everything). 

sounds like we don't handle the metadata for it correctly, would be better to fix that instead of doing things conditionally later.

B4nan (approved):
i will be happy to merge this as is, but ideally we would also iron out the metadata difference for pivot entities, there shouldn't be a need for doing things conditionally because of them

russell-dot-js:
> i will be happy to merge this as is, but ideally we would also iron out the metadata difference for pivot entities, there shouldn't be a need for doing things conditionally because of them

👍 sounds great. I'd rather look on that stuff on a separate branch if it's cool with you, this bug was blocking us from using mikro for queries & we fellback to using knex directly for queries while using mikro to generate schema

my overall thought is that it would help a lot if more of this information was cached in the metadata instead of computed as-needed. especially since fixing this in one place [doesn't fix others](https://github.com/russell-dot-js/mikro-orm/pull/1/files#r1625274788) it's kind of like playing whack-a-mole right now. being able to look to the metadata as the shared source of truth, rather than implementing different logic for how to traverse props in different places, should help tidy it up

i know this can't be done at declaration time, especially considering circular dependencies, but it might be good to replace the usages of getPrimaryProps / targetMetadata / etc with a getter than lazily computes & stores all the necessary information about an entity / prop. don't have a conception for exactly what that would look like yet but my hunch is that it would improve readability & maintainability

B4nan:
> my overall thought is that it would help a lot if more of this information was cached in the metadata instead of computed as-needed

Absolutely agreed.

> rather than implementing different logic for how to traverse props in different places

The thing is that different places often require different logic and a bit different shapes. But that's a topic on its own, ideally such places should be refactored and we should unify those approaches. 

> i know this can't be done at declaration time, especially considering circular dependencies, but it might be good to replace the usages of getPrimaryProps / targetMetadata / etc with a getter than lazily computes & stores all the necessary information about an entity / prop's keys & relationships. don't have a conception for exactly what that would look like yet but my hunch is that it would improve readability & maintainability

I don't see any issues with doing things during discovery. We actually do a lot of what you are saying here, the EntityMetadata object is computing a lot of things on its own and caches them, discovery is also doing a lot of recursive stuff.

B4nan:
Anyway, thanks a lot for the PR, people are usually not into fixing bugs in the compiled functions and it needs to wait for me, really appreciate you were not afraid of touching this :]

PR #5637, opened 2024-05-30
fix(mssql): account for quotes in table names

patrykbojczuk:
This fixes invalid sql produced by `MsSqlSchemaHelper` for tables with quote char in the name.


B4nan (approved):
looking good, thanks!

PR #5641, opened 2024-05-30
fix(postgres): respect `deferMode` option in 1:1 relations

DASPRiD:
In my initial implementation on `deferMode` for 1:m and 1:1 I overlooked the unique constraint which is created for 1:1 relations. This PR corrects this and also applies the selected `deferMode` to the unique constraint.

This PR also lays the framework for potentially using `deferMode` on custom unique constraints, though that is not critical since those could already be created through the `expression` prop.

If this PR looks fine, I'd appreciate a quick bugfix release after you had time to merge it, since this is currently a showstopper in one of my projects and I can't find a workaround :)

### DIFF
--- a/packages/knex/src/schema/SqlSchemaGenerator.ts
+++ b/packages/knex/src/schema/SqlSchemaGenerator.ts
@@ -463,11 +463,15 @@ export class SqlSchemaGenerator extends AbstractSchemaGenerator<AbstractSqlDrive
 
         if (foreignKey && this.options.createForeignKeyConstraints) {
           delete diff.addedForeignKeys[foreignKey.constraintName];
-          col.references(foreignKey.referencedColumnNames[0])
+          const builder = col.references(foreignKey.referencedColumnNames[0])
             .inTable(this.getReferencedTableName(foreignKey.referencedTableName))
             .withKeyName(foreignKey.constraintName)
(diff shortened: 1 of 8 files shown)


packages/knex/src/schema/SqlSchemaGenerator.ts
```diff
@@ -463,11 +463,15 @@ export class SqlSchemaGenerator extends AbstractSchemaGenerator<AbstractSqlDrive
...
         if (foreignKey && this.options.createForeignKeyConstraints) {
           delete diff.addedForeignKeys[foreignKey.constraintName];
-          col.references(foreignKey.referencedColumnNames[0])
+          const builder = col.references(foreignKey.referencedColumnNames[0])
             .inTable(this.getReferencedTableName(foreignKey.referencedTableName))
             .withKeyName(foreignKey.constraintName)
             .onUpdate(foreignKey.updateRule!)
             .onDelete(foreignKey.deleteRule!);
+
+          if (foreignKey.deferMode) {
```
DASPRiD, line 472: This is actually an unrelated bug fix to the schema diffing for deferMode which slipped through our previous unit tests – that one took me probably 2 hours to find :D

B4nan (approved):

PR #5647, opened 2024-05-31
fix(knex): explicitly declare all the extended drivers as optional peer dependencies

boenrobot:
This help avoid bundlers like Vite trying to pre-package them, and instead leaves them as externals, without the need of an explicit config change.

Note that Knex itself declares all of the drivers it supports as optional peer dependencies. This package just defines its extensions as optional peer dependencies in addition to that.

### DIFF
--- a/packages/knex/package.json
+++ b/packages/knex/package.json
@@ -66,6 +66,20 @@
     "@mikro-orm/core": "^6.2.8"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


B4nan (approved):
Thanks, let's see if this won't break something else (e.g. webpack), I guess it makes sense to add it and feels better to the other solutions.

PR #5651, opened 2024-06-01
feat(core): check for ORM extensions dynamically

B4nan:
The extensions are now again checked dynamically if not explicitly registered, so it should be enough to have the given package (e.g. `@mikro-orm/seeder`) installed as in v5.

### DIFF
--- a/packages/mongodb/src/MongoPlatform.ts
+++ b/packages/mongodb/src/MongoPlatform.ts
@@ -34,13 +34,14 @@ export class MongoPlatform extends Platform {
   /** @inheritDoc */
   override getExtension<T>(extensionName: string, extensionKey: string, moduleName: string, em: EntityManager): T {
     if (extensionName === 'EntityGenerator') {
-      throw new Error('EntityGenerator is not support for this driver.');
+      throw new Error('EntityGenerator is not supported for this driver.');
     }
 
     if (extensionName === 'Migrator') {
       return super.getExtension('Migrator', '@mikro-orm/migrator', '@mikro-orm/migrations-mongodb', em);
     }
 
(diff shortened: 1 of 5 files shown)


packages/mongodb/src/MongoPlatform.ts
```diff
@@ -31,6 +31,19 @@ export class MongoPlatform extends Platform {
     MongoSchemaGenerator.register(orm);
   }
 
+  /** @inheritDoc */
+  override getExtension<T>(extensionName: string, extensionKey: string, moduleName: string, em: EntityManager): T {
+    if (extensionName === 'EntityGenerator') {
+      throw new Error('EntityGenerator is not support for this driver.');
```
boenrobot, line 37:
```suggestion
      throw new Error('EntityGenerator is not supported for this driver.');
```

tests/features/entity-generator/EntityGenerator.test.ts
```diff
@@ -7,7 +7,7 @@ describe('EntityGenerator', () => {
 
   test('not supported [mongodb]', async () => {
     const orm = await MikroORM.init({ driver: MongoDriver, dbName: 'mikro-orm-test', discovery: { warnWhenNoEntities: false }, connect: false });
-    expect(() => orm.entityGenerator).toThrow('EntityGenerator extension not registered.');
+    expect(() => orm.entityGenerator).toThrow('EntityGenerator is not support for this driver.');
```
boenrobot, line 10:
```suggestion
    expect(() => orm.entityGenerator).toThrow('EntityGenerator is not supported for this driver.');
```

PR #5669, opened 2024-06-03
feat(entity-generator): add a coreImportsPrefix option

boenrobot:
This option allows the user to specify a prefix to add to imports from @mikro-orm/core, to ensure they will not have conflicts with application identifiers.

The special casing of custom base entity called "BaseEntity" is removed in favor of this general purpose option.


packages/entity-generator/src/EntitySchemaSourceFile.ts
```diff
@@ -46,20 +42,18 @@ export class EntitySchemaSourceFile extends SourceFile {
     }
 
     if (primaryProps.length > 0) {
-      this.coreImports.add('PrimaryKeyProp');
       const primaryPropNames = primaryProps.map(prop => `'${prop.name}'`);
 
       if (primaryProps.length > 1) {
-        classBody += `${' '.repeat(2)}[PrimaryKeyProp]?: [${primaryPropNames.join(', ')}];\n`;
+        classBody += `${' '.repeat(2)}[${this.referenceCoreImport('PrimaryKeyProp')}]?: [${primaryPropNames.join(', ')}];\n`;
```
B4nan, line 48: using `' '.repeat(2)` instead of just `'  '` sounds like proper overengineering, lol

B4nan (approved):

packages/entity-generator/src/EntitySchemaSourceFile.ts
boenrobot: IIRC (and I may be misremembering...), it was like that when I first saw it... And I was considering whether I should "optimize" it into a constant, but then I thought "well... The _intention_ that there needs to be indents is a bit clearer this way, so let's keep it I guess".

If I'm the one who first introduced it, I for sure had my own second thoughts before pushing. I only remember the doubt, not the "creation event" 😄

PR #5670, opened 2024-06-03
fix(entity-generator): support complex enum names and values

boenrobot:
Enum options that are invalid JS identifiers get the property treatment (independent of naming strategy), and the enum name itself gets the entity name treatment.

Two new NamingStrategy methods have been added to optionally adjust the default behaivour. getEnumClassName() and enumValueToEnumProperty().

### DIFF
--- a/packages/core/src/naming-strategy/AbstractNamingStrategy.ts
+++ b/packages/core/src/naming-strategy/AbstractNamingStrategy.ts
@@ -61,6 +61,20 @@ export abstract class AbstractNamingStrategy implements NamingStrategy {
     return propName;
   }
 
+  /**
+   * @inheritDoc
+   */
+  getEnumClassName(columnName: string, tableName: string, schemaName?: string): string {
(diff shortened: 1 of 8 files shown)


packages/core/src/naming-strategy/AbstractNamingStrategy.ts
```diff
@@ -61,6 +61,14 @@ export abstract class AbstractNamingStrategy implements NamingStrategy {
     return propName;
   }
 
+  getEnumClassName(columnName: string, tableName: string, schemaName?: string): string {
+    return this.getEntityName(`${tableName}_${columnName}`, schemaName);
+  }
+
+  enumValueToEnumProperty(enumValue: string, columnName: string, tableName: string, schemaName?: string): string {
```
B4nan, line 68: lets add `@inheritDoc` for both of them so we get the docs from the interface on it

boenrobot: Done

B4nan (approved):

PR #5674, opened 2024-06-04
fix(entity-generator): include all entity options in EntitySchema definitions

boenrobot:
The getEmbeddableDeclOptions() and getEntityDeclOptions() methods from SourceFile are now reused in EntitySchemaSourceFile, thus ensuring they remain in sync.

Serialization of EntitySchema options had to be adjusted a bit for this reuse, which results in different decisions regarding whitespace in EntitySchema.

Specifically, previously, all options would always be on their own line, while property definitions would always be on their own line each, and then word wrapped at 80 characters each.
Now, instead, 80 character limit is applied on the entirety of the entity options, at every level, with the initial call taking into account the line where the object starts. Whenever a level as a whole would be more than 80 characters, each value is put on its own line, and the limit is applied recursively, taking into account the space before the value.


tests/features/entity-generator/__snapshots__/GeneratedColumns.postgresql.test.ts.snap
```diff
@@ -159,9 +157,7 @@ export class AllowedAgesAtCreation {
 
 export const AllowedAgesAtCreationSchema = new EntitySchema({
   class: AllowedAgesAtCreation,
-  properties: {
-    age: { primary: true, type: 'smallint', autoincrement: false },
-  },
+  properties: { age: { primary: true, type: 'smallint', autoincrement: false } },
```
B4nan, line 160: ideally we would keep this as it was, the line breaks for the properties are a good idea, but having this inline with the properties object itself feels rather confusing

boenrobot: Fair, I guess...

I've added a bit of a "hack" whereby if an object has the "Config" symbol added (with whatever value), it will force indents, even if the line could otherwise fit into less than 80 characters. I _could_ make it so that depending on the value, we may instead force no indenting even if the value is over 80 characters (rather than unconditionally force indent), but we probably don't need that anywhere for now.

packages/entity-generator/src/EntitySchemaSourceFile.ts
```diff
@@ -63,21 +66,16 @@ export class EntitySchemaSourceFile extends SourceFile {
...
+      ...(this.meta.embeddable ? this.getEmbeddableDeclOptions() : (this.meta.collection ? this.getEntityDeclOptions() : {})),
+    };
+    const declLine = `export const ${this.meta.className}Schema = new ${this.referenceCoreImport('EntitySchema')}(`;
+    ret += declLine;
 
     if (this.meta.indexes.length > 0) {
-      ret += `  indexes: [\n`;
-      this.meta.indexes.forEach(index => {
+      entitySchemaOptions.indexes = [];
+      entitySchemaOptions.indexes = this.meta.indexes.map(index => {
```
B4nan, line 78: whats the point of the empty array initializer when you override the value on the next line?

tests/features/entity-generator/__snapshots__/ConflictingEntityNames.mysql.test.ts.snap
```diff
@@ -151,8 +151,18 @@ export const ConfigSchema = new MikroORM_EntitySchema({
       persist: false,
       hydrate: false,
     },
-    test2: { kind: 'embedded', entity: () => Entity, persist: false, hydrate: false },
-    test3: { kind: 'embedded', entity: () => EntitySchema, persist: false, hydrate: false },
+    test2: {
+      kind: 'embedded',
+      entity: () => Entity,
+      persist: false,
+      hydrate: false,
```
B4nan, line 158: where is that `hydrate: false` coming? i dont think the entity generator should produce that anywhere really unless specified by the user. also it makes no sense when combined with `persist: false`, i don't think it has any effect

boenrobot: The test features `onInitialMetadata` hook to add a few more conflicting identifiers like `Embedded`.

[Here](https://github.com/mikro-orm/mikro-orm/blob/master/tests/features/entity-generator/ConflictingEntityNames.mysql.test.ts#L117)

packages/entity-generator/src/EntitySchemaSourceFile.ts
boenrobot: Oops. Oversight from an earlier, never pushed version... I thought I cleaned that up.

PR #5675, opened 2024-06-04
chore: add failing test for findOne by entity with composite PK's & FK's

russell-dot-js:
results in NPE, see [this comment](https://github.com/russell-dot-js/mikro-orm/pull/1#discussion_r1625274788)


B4nan:
can you adjust the PR title and description? it should reflect what the PR is doing since it dictates the commit message

russell-dot-js:
> can you adjust the PR title and description? it should reflect what the PR is doing since it dictates the commit message

sorry, having a little bit of a brain fart. the PR title says I added a failing test for findOne, code looks like it's doing that, what am I missing

B4nan:
oh its just a repro, i though its like the other PR where you also had the exact same PR title, tests were green and it was because you actually fixed that already

B4nan:
should be resolved via 6f95f7255aa1b620372cff036e29cc39df078ca0

PR #5684, opened 2024-06-06
feat(core): implement "character" type (DB type "char")

boenrobot:
The name "character" was picked to be different from the regular "char" type,
while still having precedent (that being alias of "char" in postgresql).

The char length is set to default of 1, to avoid excessive memory usage,
while still providing a default, as required.

For sqlite, this type is aliased to "text", just like "string".
For mssql, "character" is aliased to "nchar", to be consistent with "string" being "nvarchar",
while "char" means the native "char" type,
to be consistent with "varchar" meaning the native "varchar".
For postgresql, the "bpchar" type is also supported as an alias to "char"/"character".

### DIFF
--- a/docs/docs/custom-types.md
+++ b/docs/docs/custom-types.md
@@ -311,10 +311,13 @@ export const types = {
   double: DoubleType,
   boolean: BooleanType,
   decimal: DecimalType,
+  character: CharacterType,
   string: StringType,
   uuid: UuidType,
   text: TextType,
-};
+  interval: IntervalType,
+  unknown: UnknownType,
+} as const;
 ```
 
 ### ArrayType
--- a/packages/core/src/platforms/Platform.ts
+++ b/packages/core/src/platforms/Platform.ts
@@ -7,8 +7,30 @@ import type { EntityManager } from '../EntityManager';
 import type { Configuration } from '../utils/Configuration';
 import type { IDatabaseDriver } from '../drivers/IDatabaseDriver';
 import {
-  ArrayType, BigIntType, BlobType, Uint8ArrayType, BooleanType, DateType, DecimalType, DoubleType, JsonType, SmallIntType, TimeType,
-  TinyIntType, Type, UuidType, StringType, IntegerType, FloatType, DateTimeType, TextType, EnumType, UnknownType, MediumIntType, IntervalType,
+  ArrayType,
+  BigIntType,
+  BlobType,
+  Uint8ArrayType,
+  BooleanType,
+  CharacterType,
+  DateType,
+  DecimalType,
+  DoubleType,
+  JsonType,
+  SmallIntType,
+  TimeType,
+  TinyIntType,
+  Type,
+  UuidType,
+  StringType,
+  IntegerType,
+  FloatType,
+  DateTimeType,
+  TextType,
+  EnumType,
+  UnknownType,
+  MediumIntType,
+  IntervalType,
 } from '../types';
 import { parseJsonSafe, Utils } from '../utils/Utils';
 import { ReferenceKind } from '../enums';
@@ -131,6 +153,10 @@ export abstract class Platform {
     return 255;
   }
 
+  getDefaultCharLength(): number {
+    return 1;
+  }
+
   getDateTypeDeclarationSQL(length?: number): string {
     return 'date' + (length ? `(${length})` : '');
   }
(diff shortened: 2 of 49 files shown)


B4nan:
Why not just `char` when you have that alias for the same already (and that is actually a recognized column type)?

B4nan:
> Why pick "string" for the types map instead of "varchar"? 😄

`string` is a runtime JS type, we need to map it to something. Unlike your proposal which came out of nowhere :]

B4nan:
> meaning the only way to actually use MSSQL's "varchar" is with {columnType: 'varchar(255)'}

Just like the only way to use `timestamp` instead of `timestamptz` in postgres is via `columnType`. I don't really see anything wrong with that, I never wanted to have every possible column type available in there, using `columnType` is more than fine, it's not a code smell.

B4nan:
> But this is inconsistent

Which part is inconsistent? 

> and only gets worse with "char" in the mix, where {type: 'char'} would be equivalent to {columnType: 'nchar'}, and to use native char, you need {columnType: 'nchar(1)'}.

I guess you wanted to say `{columnType: 'char(1)'}`. Again, there is nothing wrong with using `columnType`, its there exactly for this use case - overriding what the ORM would infer otherwise (e.g. based on `type` which is optional and the value is inferred from TS is possible, either via reflect metadata or ts-morph - that is the original purpose of `type` - to reflect the runtime type from the metadata provider).

> In the case of "timestamp" vs "timestamptz", I guess there's no way to rectify this without breaking BC for most Postgresql users, since both identifiers are used as DB types.

I don't think we should "fix" anything here, I am more than fine with how this works (and I haven't received any reports about this either, it was always done this way).

It feels like you dislike `columnType` for some reason and expect to be able to do everything with `type`. That is surely not the case, at least not with the string value (but you can put a custom type there and that can have its column type set explicitly). I don't consider that an issue at all.

> But in the case of varchar/nvarchar, there is a way... {type:'varchar'} could come to mean varchar in all dialects that have varchar, while {type: 'string'} stays on "nvarchar" in MSSQL for BC. And then "char" just becomes consistent with it to mean "char everywhere where char exists", and "fixedstring" (or let's say "character" if you want to avoid introducing a name out of nowhere? For pgsql, that is an alias of char anyway, and doesn't mean anything in other dialects...), it would mean "char where supported, except in MSSQL, where it means char".

This part sounds good to me. Although I still don't understand why wouldn't `type: 'char'` be enough, what would be the difference from `type: 'fixedstring'`?

> "char where supported, except in MSSQL, where it means char"

That's a typo again I guess? char where supported but in mssql its (again) char? 

--- 

Don't get me wrong, I welcome improvements of any kind, but keep in mind we can't do breaking stuff randomly, and changing the mapping is breaking too. We need to be careful here, and I also don't want to add some new confusing types unless we actually need them.

docs/docs/custom-types.md
```diff
@@ -311,10 +311,13 @@ export const types = {
   double: DoubleType,
   boolean: BooleanType,
   decimal: DecimalType,
+  character: CharacterType,
```
B4nan, line 314: can we use `char` instead of `character`? same for the class name?

or does this need to have a different name because of the `nchar` in mssql?

docs/docs/custom-types.md
```diff
@@ -311,10 +311,13 @@ export const types = {
...
   string: StringType,
   uuid: UuidType,
   text: TextType,
-};
+  interval: IntervalType,
+  unknown: UnknownType,
```
B4nan, line 319: do we need this? technically the unknown type is not something users should use anywhere, its implementation detail

(not a big deal, we can keep it if it makes our code simpler)

docs/docs/custom-types.md
boenrobot: > or does this need to have a different name because of the nchar in mssql?

Yes... That's exactly it. Anything other than "char" will do, but "character" seems the most natural alternative.

B4nan (approved):

docs/docs/custom-types.md
boenrobot: Since this is docs, and they say the map is defined like that... I thought it might as well reflect reality... And the reality is that "unknown" is also part of it.

Whether people should use it is a separate matter, which could perhaps be clarified with a doc comment above the map entry (both in code and in the docs)?

B4nan: hmm right, i thought you added it to the map itself, if it was there, lets mention it in the docs too.

PR #5707, opened 2024-06-14
feat(postresql): add support for varchar with unlimited length

boenrobot:
Implemented by specifying length to -1, analogously to MSSQL.

Not specifying a length still defaults to 255, as before.


tests/features/embeddables/embeddable-custom-types.postgres.test.ts
```diff
@@ -14,7 +14,7 @@ export class AlwaysConvertsToAbc extends Type<string, string> {
   }
 
   override getColumnType(): string {
-    return 'varchar';
+    return 'varchar(255)';
```
B4nan, line 17: was this required for things to work? if so, it feels a bit breaking.

boenrobot: A change of some sort is required... either that, or adding -1 in the entity definition.

This is effectively broken wothout this, due to the default being 255 rather than "undefined" which PG treats as unlimited.

B4nan: Well, column type needs to be always used as is, if you have a custom type that implements the method, it cannot require changes in the property definition. In other words, this shoudl stay as is without any changes and should be interpretted as length: -1 in pg.

boenrobot: ok, I think I managed to find a way to keep that test fully BC, while still supporting -1 where needed.

packages/knex/src/schema/DatabaseTable.ts
```diff
@@ -110,7 +112,7 @@ export class DatabaseTable {
       }
 
       if (prop.length == null && prop.columnTypes[idx]) {
-        prop.length = this.platform.getSchemaHelper()!.inferLengthFromColumnType(prop.columnTypes[idx]);
+        prop.length = this.platform.getSchemaHelper()!.inferLengthFromColumnType(prop.columnTypes[idx].trimStart());
```
B4nan, line 115: in which case is this part needed? feels weird to have spaces in the column type, also why only trim from start and not both ends?

packages/postgresql/src/PostgreSqlSchemaHelper.ts
```diff
@@ -628,4 +630,18 @@ export class PostgreSqlSchemaHelper extends SchemaHelper {
     return res[tableName];
   }
 
+  override inferLengthFromColumnType(type: string): number | undefined {
+    const match = type.match(/^(varchar|character varying)\s*(?:\(\s*(\d+)\s*\))?/);
```
B4nan, line 634: what about `char`, that can also have length i think

tests/features/entity-generator/__snapshots__/index-expressions.postgres.test.ts.snap
```diff
@@ -17,21 +17,21 @@ export class DcimDevice {
   @PrimaryKey({ autoincrement: false })
   id!: bigint;
 
-  @Property({ nullable: true })
+  @Property({ length: -1, nullable: true })
```
B4nan, line 20: do we really need to emit this? postgres default mapping for string is `text`, so the length: -1 is implicit there, right?

B4nan (commented):
just two questions, not really blocking

packages/postgresql/src/PostgreSqlSchemaHelper.ts
boenrobot: I thought I'd deal with char will be dealt with in the char PR 😅

packages/knex/src/schema/DatabaseTable.ts
boenrobot: Trimming, because I thought this place is also used with schema generator, right? That is, I wanted to handle code like
```
columnType: `
sometype::someclass::varchar::yadayada('I don't really know the full postgresql syntax for types')
`
```

and I wanted to anchor what was previously an unanchored pattern, to more specifically handle the known cases without accidentally bumping into the unknown more expressive ones.

That said, I guess it should be safe to trim from both ends, or maybe trim elsewhere if this line is only relevant for the entity generator.

tests/features/entity-generator/__snapshots__/index-expressions.postgres.test.ts.snap
boenrobot: But this is varchar, not text. If this was `type: 'text'`, yes. And even now, there's no length for that one.

varchar with no length is not an alias to text.

packages/knex/src/schema/DatabaseTable.ts
B4nan: Why would anyone write a code like that explicitly? If we want to trim, it should happen during metadata discovery most likely, but I am not sure if this is really needed. If someone wants to write it like that, they can just append the `.trim()` at the end of their multiline string. To me it's wrong, the value shouldn't be multiline, and formatting cannot dictate values.

tests/features/entity-generator/__snapshots__/index-expressions.postgres.test.ts.snap
B4nan: for some reason i thought we are mapping to `text` by default since v6, my bad

packages/knex/src/schema/DatabaseTable.ts
boenrobot: ok, fair enough. Removed this trimStart().

PR #5714, opened 2024-06-15
fix: prevent lazy properties from being undefined at runtime if they were loaded with explicit populate

ropbastos:
First of all, thank you so much for the awesome ORM! ❤️

Now for the motivation for this PR:
The types for `const loaded = someReference.load({ populate: ['someProperty'] })` guarantee that the `loaded` object will have `someProperty`'s content in `loaded.someProperty.$`. However, I seem to have found a situation where that is not true at runtime. 

If the entity has already been initialized, then the `.load()` call ignores any options passed in and just returns what is in the ORM's cache. It is possible, however, that the entity in the cache has lazy properties which haven't been loaded at its first initialization and are present in the `populate` option of the latest `.load()` call. In that way, there seems to be a mismatch between the types and runtime.

This PR is my attempt at making the type guarantee match the runtime, by making sure .load() refreshes the entity if some property is present in the populate array but hasn't been loaded yet to the cache. I tried following the commit guidelines, and have also ran the tests. Im pretty certain I added tests for the new behavior in the wrong place though, but I couldn't quite figure out where to put it, so I put it next to the only other place where the method I modified was being tested 😅. Please let me know of any changes I need to make 🙏.

Thanks in advance, I'm very grateful for this project!

### DIFF
--- a/packages/core/src/entity/Reference.ts
+++ b/packages/core/src/entity/Reference.ts
@@ -1,4 +1,6 @@
 import { inspect } from 'util';
+import type { FindOneOptions, FindOneOrFailOptions } from '../drivers/IDatabaseDriver';
+import { DataloaderType } from '../enums';
 import type {
   AddEager,
   Dictionary,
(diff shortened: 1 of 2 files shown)


B4nan:
Oops, I just fixed this a few minutes ago :]

04fb826ba96426eb700284d923887594f3f7f09e

ropbastos:
Just saw it when I was notified of the conflict haha, thanks!

ropbastos:
@B4nan Sorry to bother, but when are fixes like this released? 👉 👈

B4nan:
Every commit to master is released as dev version.

PR #5722, opened 2024-06-18
fix(core): respect nullability of wrapped types for ScalarReference's

ropbastos:
If one is to use a ScalarReference to get type-safe access to a lazily loaded scalar value it should be possible to encode in the types that the scalar may be null in the database. Currently, ScalarReferences, when de-referenced through `.get()` or `.$` are typed with "NonNullable" variants of their wrapped types. That non-nullability is not reflected at runtime, which opens up code that perfectly type-checks to runtime exceptions because of null reference accesses. This is demonstrated in this reproduction: https://github.com/ropbastos/mikro-orm-reproduction

Briefly linking the "fixed" version of the library with my companies project I didn't see any types that looked wrong, and all tests still pass, so I'm creating this PR. However, I'm aware the NonNullable helpers are pretty explicit about their intent 😅. I'm not sure what I'm trying to do makes sense/why were they added in the first place, but in the off-chance that they are there by mistake (or they ceased to make sense), I've already done the small work of removing them 🙏

Here is a discussion I opened when I first encountered an error of the type represented in the reproduction repository: https://github.com/mikro-orm/mikro-orm/discussions/5711

### DIFF
--- a/docs/docs/defining-entities.md
+++ b/docs/docs/defining-entities.md
@@ -1112,7 +1112,7 @@ const user = await em.findOne(User, 1);
 const passwordHash = await user.passwordHash.load();
 ```
 
-For object-like types, if you choose to use the reference wrappers, you should use the `ScalarReference<T>` type explicitly. For example, you might want to lazily load a large JSON value:
+For object-like types, if you choose to use the reference wrappers, you should use the `ScalarRef<T>` type explicitly. For example, you might want to lazily load a large JSON value:
 
 ```ts
 @Property({ type: 'json', nullable: true, lazy: true, ref: true })
--- a/docs/docs/defining-entities.md
+++ b/docs/docs/defining-entities.md
@@ -1135,10 +1135,6 @@ const populatedReport = await em.populate(report, ['reportParameters']);
 // Logs `null`
 console.log(populatedReport.reportParameters.$); 
 ```
-
-
-
-
 ## Virtual Properties
 
 We can define our properties as virtual, either as a method, or via JavaScript `get/set`.
--- a/docs/docs/defining-entities.md
+++ b/docs/docs/defining-entities.md
@@ -1100,19 +1100,41 @@ const b2 = await em.find(Book, 1, { populate: ['text'] }); // this will load the
 
 ### `ScalarReference` wrapper
 
-Similarly to the `Reference` wrapper, we can also wrap lazy scalars with `Ref` into a `ScalarReference` object.
+Similarly to the `Reference` wrapper, we can also wrap lazy scalars with `Ref` into a `ScalarReference` object. The `Ref` type automatically resolves to `ScalarReference` for non-object types, so the below is correct:
 
 ```ts
 @Property({ lazy: true, ref: true })
 passwordHash!: Ref<string>;
 ```
 
-The `Ref` type automatically resolves to `ScalarReference` for non-object types. You can use it explicitly if you want to wrap an object scalar property (e.g. JSON value).
-
 ```ts
 const user = await em.findOne(User, 1);
 const passwordHash = await user.passwordHash.load();
 ```
+
+For object-like types, if you choose to use the reference wrappers, you should use the `ScalarReference<T>` type explicitly. For example, you might want to lazily load a large JSON value:
+
+```ts
+@Property({ type: 'json', nullable: true, lazy: true, ref: true })
+reportParameters!: ScalarReference<ReportParameters | null>; // ReportParameters is an example class, imagine it defined elsewhere.
+```
+
+Keep in mind that once a scalar value is managed through a ScalarReference, accessing it through MikroORM managed objects will always return the Reference wrapper. That can be confusing in case the property is also `nullable`, since the Reference will always be truthy. In such cases, you should inform the type system of the nullability of the property through `ScalarReference<T>`'s type parameter as demonstrated above. Below is an example of how it works:
+
+```ts
+// Say Report of id "1" has no reportParameters in the Database.
+const report = await em.findOne(Report, 1);
+if (report.reportParameters) {
+  console.log(report.reportParameters); // Logs Ref<?>, not the actual value. **Would always run***.
+  //@ts-expect-error $/.get() is not available until the reference has been loaded.
+  // const mistake = report.reportParameters.$
+}
+const populatedReport = await em.populate(report, ['reportParameters']);
+console.log(populatedReport.reportParameters.$); // Logs `null`
+```
+
+
+
 
 ## Virtual Properties
 
--- a/docs/docs/defining-entities.md
+++ b/docs/docs/defining-entities.md
@@ -1135,6 +1135,7 @@ const populatedReport = await em.populate(report, ['reportParameters']);
 // Logs `null`
 console.log(populatedReport.reportParameters.$); 
 ```
+
 ## Virtual Properties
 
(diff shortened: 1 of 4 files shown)


B4nan:
Looks good, can you also mention this in the docs?

https://mikro-orm.io/docs/defining-entities#scalarreference-wrapper

ropbastos:
> Looks good, can you also mention this in the docs?
> 
> https://mikro-orm.io/docs/defining-entities#scalarreference-wrapper

Sure! Will be free in couple minutes, 1 hour tops, then I'll edit the docs to try to dispel what confused me :)

docs/docs/defining-entities.md
```diff
@@ -1100,20 +1100,45 @@ const b2 = await em.find(Book, 1, { populate: ['text'] }); // this will load the
...
 ```
 
-The `Ref` type automatically resolves to `ScalarReference` for non-object types. You can use it explicitly if you want to wrap an object scalar property (e.g. JSON value).
-
 ```ts
 const user = await em.findOne(User, 1);
 const passwordHash = await user.passwordHash.load();
 ```
 
+For object-like types, if you choose to use the reference wrappers, you should use the `ScalarReference<T>` type explicitly. For example, you might want to lazily load a large JSON value:
```
B4nan, line 1115: There is an alias for this called `ScalarRef`, maybe it would be better to suggest that. Unlike with entity references (where you get the PK directly on `Ref` but not on `Reference` type), there is no difference from the `ScalarReference`, it's just a bit shorter to write, so not really important.

docs/docs/defining-entities.md
```diff
@@ -1100,20 +1100,45 @@ const b2 = await em.find(Book, 1, { populate: ['text'] }); // this will load the
...
+
+```ts
+@Property({ type: 'json', nullable: true, lazy: true, ref: true })
+// ReportParameters is an example class, imagine it defined elsewhere.
```
B4nan, line 1119: This is a bit weird, for mapping to classes you should use a custom type, I would just say its an object type instead of a class.

docs/docs/defining-entities.md
```diff
@@ -1100,20 +1100,45 @@ const b2 = await em.find(Book, 1, { populate: ['text'] }); // this will load the
...
+reportParameters!: ScalarReference<ReportParameters | null>; 
+```
+
+Keep in mind that once a scalar value is managed through a `ScalarReference`, accessing it through MikroORM managed objects will always return the `Reference` wrapper. That can be confusing in case the property is also `nullable`, since the `Reference` will always be truthy. In such cases, you should inform the type system of the nullability of the property through `ScalarReference<T>`'s type parameter as demonstrated above. Below is an example of how it all works:
```
B4nan, line 1123: It will return `ScalarReference` wrapper, its not the same as `Reference`, those are two different classes.

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/entity/Reference.ts#L199

docs/docs/defining-entities.md
```diff
@@ -1100,20 +1100,45 @@ const b2 = await em.find(Book, 1, { populate: ['text'] }); // this will load the
...
+  // const mistake = report.reportParameters.$
+}
+const populatedReport = await em.populate(report, ['reportParameters']);
+// Logs `null`
+console.log(populatedReport.reportParameters.$); 
+```
+
+
+
+
```
B4nan, line 1141:
```suggestion

```

B4nan (changes requested):
great, thanks! left a few ideas on how to improve the docs

docs/docs/defining-entities.md
```diff
@@ -1100,20 +1100,41 @@ const b2 = await em.find(Book, 1, { populate: ['text'] }); // this will load the
...
+if (report.reportParameters) {
+  // Logs Ref<?>, not the actual value. **Would always run***.
+  console.log(report.reportParameters); 
+  //@ts-expect-error $/.get() is not available until the reference has been loaded.
+  // const mistake = report.reportParameters.$
+}
+const populatedReport = await em.populate(report, ['reportParameters']);
+// Logs `null`
+console.log(populatedReport.reportParameters.$); 
+```
```
B4nan, line 1137: lets keep one blank line here :]

docs/docs/defining-entities.md
ropbastos: Oops, tried to fix here: [2189bcc](https://github.com/mikro-orm/mikro-orm/pull/5722/commits/2189bccc80f800da07ab884fd4af4c05ce4570cf)

LMK if it's ok then 🙏

docs/docs/defining-entities.md
ropbastos: Agreed! Changed it: [82dd273](https://github.com/mikro-orm/mikro-orm/pull/5722/commits/82dd273c9f8a6f950bf70846156de94ff1bb2191)

docs/docs/defining-entities.md
ropbastos: Done 🤝

docs/docs/defining-entities.md
ropbastos: Sorry, what do you mean, where exactly? 👉👈

B4nan: before the next heading, after this code block

docs/docs/guide/05-type-safety.md
```diff
@@ -124,20 +124,44 @@ await article2.author.load(); // no additional query, already loaded
...
 const user = await em.findOne(User, 1);
 const passwordHash = await user.passwordHash.load();
 ```
 
+For object-like types, if you choose to use the reference wrappers, you should use the `ScalarRef<T>` type explicitly. For example, you might want to lazily load a large JSON value:
+
+```ts
+@Property({ type: 'json', nullable: true, lazy: true, ref: true })
+// ReportParameters is an object type, imagine it defined elsewhere.
+reportParameters!: ScalarReference<ReportParameters | null>;
```
B4nan, line 146: i would use the `ScalarRef` in all the other places too

docs/docs/defining-entities.md
```diff
@@ -1100,20 +1100,41 @@ const b2 = await em.find(Book, 1, { populate: ['text'] }); // this will load the
...
 const user = await em.findOne(User, 1);
 const passwordHash = await user.passwordHash.load();
 ```
 
+For object-like types, if you choose to use the reference wrappers, you should use the `ScalarRef<T>` type explicitly. For example, you might want to lazily load a large JSON value:
+
+```ts
+@Property({ type: 'json', nullable: true, lazy: true, ref: true })
+// ReportParameters is an object type, imagine it defined elsewhere.
+reportParameters!: ScalarReference<ReportParameters | null>;
```
B4nan, line 1120: same here

B4nan (changes requested):
few more nits, looking great otherwise

docs/docs/guide/05-type-safety.md
ropbastos: Dooone [d7505a7](https://github.com/mikro-orm/mikro-orm/pull/5722/commits/d7505a7d2fa70fdd314ece16544d2b0c48ed59e0)

docs/docs/defining-entities.md
ropbastos: [d7505a7](https://github.com/mikro-orm/mikro-orm/pull/5722/commits/d7505a7d2fa70fdd314ece16544d2b0c48ed59e0)

B4nan (approved):

PR #5732, opened 2024-06-20
docs: fix the dataloader reference in usage

ssut:
Screenshot 2024-06-20 at 19 03 33

I just found that `Dataloader` should be `DataloaderType` in the docs example.


B4nan:
thanks!

PR #5749, opened 2024-06-25
feat(core): addz `Platform.getDefaultVarcharLength` and optional `Type.getDefaultLength`

boenrobot:
The existing types are converted to rely on this method's presence.

The entity generator now omits the length property when it is equal to the default.

The schema generator will not detect a difference between not having length (when that method is defined on the type) and having the length with the default value. As always, specifying columnType bypasses all of that.


docs/docs/custom-types.md
```diff
@@ -39,6 +39,10 @@ You can define custom types by extending `Type` abstract class. It has several o
 - `compareValues(a, b): boolean`
 
   Allows to override the internal comparison logic. Works with the database values (results of `convertToDatabaseValue` method). This can be helpful when the database value is not stable.
+-
```
B4nan, line 42:
```suggestion

```

B4nan (approved):

docs/docs/custom-types.md
```diff
@@ -39,6 +39,9 @@ You can define custom types by extending `Type` abstract class. It has several o
 - `compareValues(a, b): boolean`
 
   Allows to override the internal comparison logic. Works with the database values (results of `convertToDatabaseValue` method). This can be helpful when the database value is not stable.
+- `getDefaultLength?(platform: Platform): number`
```
B4nan, line 42:
```suggestion

- `getDefaultLength?(platform: Platform): number`
```

PR #5771, opened 2024-07-03
fix(entity-generator): when using esmImport, FKs are now wrapped with Rel

boenrobot:
Default values with rel() or ref() are also now added for FKs.

Composite FKs with default values are ignored, and a new ignoreSchemaChanges option is added to enable safe migration generation when such entities are present in the database.


tests/features/entity-generator/__snapshots__/MetadataHooks.mysql.test.ts.snap
```diff
@@ -1651,7 +1651,7 @@ export class Author2 {
 
   @Index({ name: 'author2_terms_accepted_index' })
   @Property({ type: 'boolean', lazy: true, default: false })
-  termsAccepted: Ref<boolean>;
+  termsAccepted: Ref<boolean> & Opt = new ScalarReference(false);
```
B4nan, line 1654: I would use `ref()` instead of `new ScalarReference()`.

boenrobot: Done 👍

PR #5772, opened 2024-07-03
feat(entity-generator): enable the generator to dictate import specs via `extraImport`

boenrobot:
Entities located in different subfolders are also properly resolved,
whether they use fileName or the new extraImport.

When an IType wrapper is introduced, the default value is put in the option,
and omitted from the definition line.


B4nan:
So what exactly is the minor breaking change? The option was added in v6.0.0, so it's not great to change how it behaves, it's been around for quite some time. At the same time, I still consider (and the docs are saying that as well) the entity generation as one-time process, so I wouldn't be that afraid of breaking some very specific implementation of it. But if we can find a way to make this non breaking completely, I would be in favor of that.

B4nan:
Yes, I get that, it's still a breaking change, you can't change the behavior of any option, it's not just about the defaults.

It feels like something we can work around to me, if the previous variant wasn't returning an extension and now it should be, we can simply use that to make it work conditionally, or am I missing something?

Also, it feels weird that it should even return the extension, we have the `emit` option for that.

B4nan (commented):
Wouldn't it be better to have a separate option for the side effect imports? 

I won't mind merging this in as is, it just feels a bit weird to do this from inside an option that has completely different purpose.

tests/features/entity-generator/MetadataHooks.mysql.test.ts
```diff
@@ -285,14 +296,35 @@ class EmailType extends Type<Email, string> {
...
-    JSONObject: '../runtimeTypes/JSONObject',
-    Email: '../runtimeTypes/Email',
-  })[name] ?? name;
+    Book2: { path: `${basePath}/${name}${extension}`, name: 'Book2' },
+    CustomBooleanType: { path: `${basePath}/../types/MyBoolean`, name: 'MyBoolean' },
+    UrlTypeLike: `../types/UrlTypeLike`,
+    CustomBooleanRuntimeType: { path: '', name: `${basePath}/../runtimeTypes/BrandedTypes` },
+    JSONObject: { path: `${basePath}/../runtimeTypes/JSONObject`, name: '' },
+    Email: { path: `${basePath}/../runtimeTypes/Email`, name: 'default' },
+    URL: { path: '', name: '' },
```
B4nan, line 326: This is the bit I am talking about btw, it feels weird to do this from inside `fileName` option, a separate option sounds like a better way to me, could be a simple map of symbol name <-> import path. Something like `extraImports: { UrlTypeLike: '../types/UrlTypeLike', ... }`.

I just feel like this mixture will be hard to explain, we don't have a single example on this topic, so its not very likely people would use it for this purpose.

boenrobot: Ah, so you mean the overall approach of using fileName for imports handling, not specifically just side effects imports?

I could add some example in the docs, if complexity is the primary concern.

Or... hm... I suppose a separate option could work, if it is a function (taking in the path of the entity file as input) that returns the map, which would be consulted first, and fileName is merely used as a fallback when the map lacks an entry.

B4nan: Yeah, it feels weird to me to tell users to do this in their projects if they want to be able to import some existing classes in their generated entities. The `fileName` option is there to be able to set a filename for the entity, all of this is a side effect to me. I don't mind that it can be done this way, but it feels more like hacking the ORM than a public API for defining external imports.

boenrobot: Ok, how about this thing now then? A separate callback that may return the object spec or undefined, at which point, fileName is used.

In this way, fileName becomes sort of like the less complex version of extraImport, and users are encouraged to use fileName when relative path from the output folder is enough.

I also added some examples in the doc as part of the custom types one.

boenrobot: Btw, I also added as part of the metadata hooks test a save test... and turned out I also had to add some ensureDir for the subfolders... which is now in.

packages/entity-generator/src/SourceFile.ts
```diff
@@ -139,9 +140,9 @@ export class SourceFile {
   }
 
   protected generateImports() {
-    const imports = [];
+    const imports = new Set<string>;
```
B4nan, line 143: lets be consistent

```suggestion
    const imports = new Set<string>();
```

B4nan (approved):
just one nit

B4nan:
Have you considered adding some guide about schema first approach at some point? Now we only have docs for the various options, but I mean something more in line with the getting started guide (but smaller and specifically about using those new features of the entity generator).

You have made a huge progress in there, but I feel like many people won't use those new options since they might not know about them. Or maybe you could write a shorter section I could include the in the upcoming blog post about v6.3? I would like to ship that next week, probably at the end, something like thursday.

B4nan:
I would start with a single "schema first guide" page, once we have enough content, we can split that into multiple pages/chapters.

PR #5779, opened 2024-07-04
fix(entity-generator): generate all bidirectional relations in case of conflicts

### LINKED ISSUE #5738: Issues with bidirectionalRelations Generation in Entity Models
### Describe the bug

I am encountering an issue with Mikro-ORM where the entity generator does not automatically create the relationship properties between Users and Licenses. Although I have defined the relationship in the database schema, the entities generated do not reflect this relationship as expected

### Reproduction

```ts
import { MsSqlDriver, Options } from '@mikro-orm/mssql';
import { TsMorphMetadataProvider } from '@mikro-orm/reflection';
import dotenv from 'dotenv';
dotenv.config();

const config: Options = {
    driver: MsSqlDriver,
    host: process.env.DB_HOST,
    port: 1433,
    user: process.env.DB_USER,
    password: process.env.DB_PASS,
    dbName: process.env.DB_NAME,
    entities: ['dist/entities/**/*.js'],
    entitiesTs: ['src/entities/**/*.ts'],
    discovery: {
        warnWhenNoEntities: false,
    },
    metadataProvider: TsMorphMetadataProvider,
    baseDir: process.cwd(),
    debug: true,
};
export default config;
```
```ts
// scripts/generate-entities.ts
import { EntityGenerator } from '@mikro-orm/entity-generator';
import config from '../src/config/mikro-orm.config';

const generateEntities = async () => {
        ...config,
        discovery: { warnWhenNoEntities: false },
        extensions: [EntityGenerator],
    });
    const dump = await orm.entityGenerator.generate({
        save: true,
        path: process.cwd() + '/src/entities',
        bidirectionalRelations: true,
    });
    await orm.close(true);
};

generateEntities().catch((err) => {
    console.error(err);
    process.exit(1);
});
```

```ts
@Entity({ tableName: 'Users' })
export class Users {
  @PrimaryKey({ fieldName: 'UserID' })
  UserID!: number;

  @Property({ fieldName: 'Username', length: 255 })
  Username!: string;

}

@Entity({ tableName: 'Licenses' })
export class Licenses {
  @PrimaryKey({ fieldName: 'LicenseID' })
  LicenseID!: number;

 @ManyToOne({ entity: () => Users, fieldName: 'UserID', index: '<Index_License_UserID>' })
  UserID!: Users;

}

```

@mikro-orm/mssql

### MikroORM version

^6.2.9

### Node.js version

v22.3.0

### Operating system

Windows 11

### Validations

B4nan:
Closes #5738

### DIFF
--- a/tests/features/entity-generator/__snapshots__/ConflictingEntityNames.mysql.test.ts.snap
+++ b/tests/features/entity-generator/__snapshots__/ConflictingEntityNames.mysql.test.ts.snap
@@ -48,7 +48,7 @@ export class Entity {
   manyToMany = new MikroORM_Collection<Property>(this);
 
   @MikroORM_OneToMany({ entity: () => ManyToMany, mappedBy: 'entity' })
-  entityInverse = new MikroORM_Collection<ManyToMany>(this);
+  manyToManys = new MikroORM_Collection<ManyToMany>(this);
 
 }
 
@@ -70,7 +70,7 @@ export class Enum {
   opt!: string;
 
   @MikroORM_OneToMany({ entity: () => Property, mappedBy: 'oneToMany' })
-  oneToManyInverse = new MikroORM_Collection<Property>(this);
+  propertys = new MikroORM_Collection<Property>(this);
 
 }
 ",
@@ -116,7 +116,7 @@ export class Property {
   manyToManyInverse = new MikroORM_Collection<Entity>(this);
 
   @MikroORM_OneToMany({ entity: () => ManyToMany, mappedBy: 'property' })
-  propertyInverse = new MikroORM_Collection<ManyToMany>(this);
+  manyToManys = new MikroORM_Collection<ManyToMany>(this);
 
 }
 ",
@@ -174,7 +174,7 @@ export class Entity {
   id!: number;
   type!: EntityType;
   manyToMany = new MikroORM_Collection<Property>(this);
-  entityInverse = new MikroORM_Collection<ManyToMany>(this);
+  manyToManys = new MikroORM_Collection<ManyToMany>(this);
 }
 
 export enum EntityType {
@@ -197,7 +197,7 @@ export const EntitySchema = new MikroORM_EntitySchema({
       fixedOrder: true,
       fixedOrderColumn: 'creation_order',
     },
-    entityInverse: { kind: '1:m', entity: () => ManyToMany, mappedBy: 'entity' },
+    manyToManys: { kind: '1:m', entity: () => ManyToMany, mappedBy: 'entity' },
   },
 });
 ",
@@ -207,7 +207,7 @@ import { Property } from './Property';
 export class Enum {
   id!: number;
   opt!: string;
-  oneToManyInverse = new MikroORM_Collection<Property>(this);
+  propertys = new MikroORM_Collection<Property>(this);
 }
 
 export const EnumSchema = new MikroORM_EntitySchema({
@@ -215,11 +215,7 @@ export const EnumSchema = new MikroORM_EntitySchema({
   properties: {
     id: { primary: true, type: 'smallint' },
     opt: { type: 'string' },
-    oneToManyInverse: {
-      kind: '1:m',
-      entity: () => Property,
-      mappedBy: 'oneToMany',
-    },
+    propertys: { kind: '1:m', entity: () => Property, mappedBy: 'oneToMany' },
   },
 });
 ",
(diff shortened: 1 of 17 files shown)


tests/features/entity-generator/__snapshots__/ConflictingEntityNames.mysql.test.ts.snap
```diff
@@ -70,7 +70,7 @@ export class Enum {
   opt!: string;
 
   @MikroORM_OneToMany({ entity: () => Property, mappedBy: 'oneToMany' })
-  oneToManyInverse = new MikroORM_Collection<Property>(this);
+  propertys = new MikroORM_Collection<Property>(this);
```
boenrobot, line 73: Wouldn't it be better to have `${entityName.substring(0,1).toLowerCase()}${entityName.substring(1)}Inverse`, due to cases like this one here?

English can be wild... There is this package called [pluralize](https://www.npmjs.com/package/pluralize) that deals with a lot of the rules in English to make proper plural rules, but maybe adding it by default is too much... So... Maybe the above scheme is a good compromise? It's still a tiny bit different from before - i.e. m:n are property + inverse, and 1:m are entity + inverse... plus the numbered suffix in cases of duplicates...

B4nan: > Wouldn't it be better to have ${entityName.substring(0,1).toLowerCase()}${entityName.substring(1)}Inverse, due to cases like this one here?

But it's not `propertyInverse`, it's inverse of the `oneToMany`, not an inverse of `property`. I still see this as something for users to polish.

I don't want to include stuff like `pluralize` because of this, it's up to users to bring such dependencies if they want to, that's why I made this configurable. The numbered suffix is already implemented here.

boenrobot: Ah... it's the owner prop, not the inverse, got it... so... `${entityName.substring(0,1).toLowerCase()}${entityName.substring(1)}Owner` (in this case "propertyOwner") then perhaps? Still side stepping the plurality issue...

B4nan: Hmm, maybe. But it sounds weird to have the suffix `Inverse` for one and `Owner` for other reference kinds :] 

>  i.e. m:n are property + inverse, and 1:m are entity + inverse...

Btw the M:N properties are always called based on the pivot table name, so there the `inverse` suffix works quite nicely, but it's not the same case with the rest of the kinds.

B4nan: Worth noting that I was always considering this as something for downstream to rename, I don't think we can find a good solution here, but we are looking for something better than the one we have now, since that seems pretty bad.

I would actually not even try to fall into this if fixing #5738 wouldn't end up breaking on its own (since the generated inverse was the last one with the computed name, if we add a suffix, it will change the meaning of the generated property name, so I decided to try to make this naming better allow easier distinction.

B4nan: Maybe we can make the pluralization a tiny bit smarter and handle the `-ys` vs `-ies` at least?

Or in the end, the `pluralize` dependency is really small, so we could include it too. But I would rather think of a way to include it only for the entity generator and not for the core package.

image

boenrobot: > Worth noting that I was always considering this as something for downstream to rename, I don't think we can find a good solution here

No _perfect_ solution, sure. _Good enough_ at best... I myself would probably use pluralize in my projects, but that's for each downstream to decide itself. The provided default should be sane enough without it.

> But it sounds weird to have the suffix Inverse for one and Owner for other reference kinds :]

Maybe, but "English oversimplified" grammar sounds even weirder... "propertys", "moneys", "peoples", "datas"...

Between that and "propertyOwner", "moneyOwner", "peopleOwner", "dataOwner", that second set is... less weird.

B4nan: Yeah, when you pick a set of edge cases, you are right the second approach is less weird (but I am not sure if its more correct). When you pick the 90% (?) cases where it works nicely, it's not.

What about suffixing with `Collection` instead?

boenrobot: Sounds good too.

Whether it's "propertyCollection", "moneyCollection", "peopleCollection", "dataCollection"... or the "90%" examples like "userCollection", "fileCollection", "bookCollection", "authorCollection", "carCollection"... it all works.

B4nan: Yeah, I will go with that.

PR #5786, opened 2024-07-08
fix(mssql): only escape strings and unicode strings when necessary

### LINKED ISSUE #5811: MsSQL driver stores newline (\n) as plaintext (\\n)
### Describe the bug

When persisting a text property in MsSQL, any `\n` gets encoded as `\\n`. I can't figure out whether this is happening in MikroORM or Tedious.

### Reproduction

I will supply a PR with a failing test.

@mikro-orm/mssql

### MikroORM version

6.2.9

### Node.js version

v20.15.1

### Operating system

PopOS 22.04

### Validations

SteadEXE:
This PR tries to fix an issue with string being escaped when it's not needed.
For dialects other than MySQL, a path like `\\path\to\directory` will get inserted as `\\\\path\\to\\directory`.

Closes #5811


tests/platforms/SqlPlatform.mssql.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  @Property({ type: 'string', length: 255 })
+  path?: string;
+
+}
+
+describe('MsSqlPlatform', () => {
+  let orm: MikroORM;
+
+  beforeAll(async () => {
+    orm = await initORMMsSql({
```
B4nan, line 21: please dont use that method and use the `init` call directly. also you import from sqlite package, that's why the compilation error

B4nan: also i wanted you to test other drivers too, as well as the non unicode strings in mssql, since this all feels like something that needs handling only in MySQL actually (and the escaping library is basically a copy paste of the MySQL version)

SteadEXE: Sorry I think I copied way too many lines :(

tests/platforms/SqlPlatform.mssql.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  @Property({ type: 'varchar', columnType: 'varchar(255)' })
+  nonUnicode?: string;
+
+}
+
+describe('MsSqlPlatform', () => {
+  let orm: MikroORM;
+
+  beforeAll(async () => {
+    const dbName = `mikro_orm_test_${(Math.random() + 1).toString(36).substring(2)}`;
```
B4nan, line 23: no need for dynamic db names here

SteadEXE:
@boenrobot I added this test aswell, it looks like it works as intended

tests/platforms/strings-escape.test.ts
```diff
@@ -0,0 +1,95 @@
...
+class Test {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property({ type: 'string', length: 255 })
+  unicode?: string;
+
+  @Property({ type: 'varchar', columnType: 'varchar(255)' })
+  nonUnicode?: string;
```
B4nan, line 16: you can drop the `type` option from both

tests/platforms/strings-escape.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    test.nonUnicode = '\\\\path\\to\\directory';
+
+    // Should be escaped.
+    const test2 = new Test();
+    test2.unicode = `It's sunny today`;
+    test2.nonUnicode = `It's raining today`;
+
+    // Should be escaped twice.
+    const test3 = new Test();
+    test3.unicode = `It''s already escaped`;
```
boenrobot, line 58: I meant two apostrophes with something in between... so that you trigger a syntax error if it's not escaped.

As it stands here... if this was an SQL injection point, you're creating a string concatenation... At least that's the case in MySQL and some other dialects (not sure if T-SQL also works like this, or if it would error...).

B4nan: I am not following, how could that trigger a syntax error? They should be both escaped, regardless of where they are placed.

B4nan: Note that this whole problem is about MySQL not following the SQL specs here, all the other dialects don't do that (and are failing to handle backslashes correctly, that's what this PR is trying to fix).

(not saying that the fix we have now is great, maybe we should keep using the escape function and only un-escape the backslashes as it was originally implemented - this was actually my suggestion from discord)

boenrobot: > I am not following, how could that trigger a syntax error? They should be both escaped, regardless of where they are placed.

In MySQL
```INSERT INTO my_table SET my_col = 'It''s already escaped';```
is the same as
```INSERT INTO my_table SET my_col = CONCAT('It', 's already escaped');```
or
```INSERT INTO my_table SET my_col = 'Its already escaped';```

Like I said, I don't know for sure about other dialects.

But given the above, if instead you were to have
```INSERT INTO my_table SET my_col = 'It's already escaped and it's ok';```
that would be the equivalent of
```INSERT INTO my_table SET my_col = CONCAT('It', s already escaped and it, 's ok');```
which would cause a syntax error at `s already escaped and it`, since that is not valid SQL

Maybe that's ok in SQLite and MSSQL... I honestly don't know... which is why I'm asking for the test... It's just that the idea of unescaping SQL is very much sending me into paranoid mode 😄

B4nan:
Gave this a look and this is only mssql issue, its not surfacing in the sqlite drivers, neither in the postgres one. Also changed the fix to be applied to the `tsqlstring` library instead of manually escaping here, should be less error prone.

PR #5787, opened 2024-07-08
docs(guide): added schema first guide

boenrobot:


B4nan:
Created https://github.com/mikro-orm/schema-first-guide for you, so we can have the final project at hand like with the https://github.com/mikro-orm/guide

B4nan:
I am thinking about shipping 6.3 this evening or tomorrow, I guess it would be better to merge this (and populate the new repo), or would you rather want to merge this once it's complete? 

Still need to write the blog post, but I hope that won't be too time-consuming.

B4nan:
Recipes sound good, I would also mention this in the very first block of the current getting started guide.

docs/docs/schema-first-guide.md
```diff
@@ -0,0 +1,2969 @@
...
+npm install @mikro-orm/core \
+            @mikro-orm/mysql \
+            @mikro-orm/migrations \
+            fastify
+```
+
+and some dev dependencies
+
+```bash npm2yarn
+npm install --save-dev @mikro-orm/cli \
```
B4nan, line 192: hmm i thought `--dev` works the same, if not, we should alter the getting started guide too

also the indent is broken because of this

docs/docs/schema-first-guide.md
```diff
@@ -0,0 +1,2969 @@
...
+                  typescript \
+                  ts-node \
+                  @types/node \
+                  rimraf \
+                  vitest
+```
+
+### ECMAScript Modules
+
+Just [as in the "code first" guide](./guide/01-first-entity.md#ecmascript-modules), we'll be using ECMAScript modules. Make sure you have
```
B4nan, line 203: i would maybe explicitly mention that people can still use CJS, since this sounds a bit like you need to use ESM

in the long run, i guess we should also have CJS versions of both guides, or maybe just a CJS version of the final project in a branch.

docs/docs/schema-first-guide.md
```diff
@@ -0,0 +1,2969 @@
...
+  "ts-node": {
+    "esm": true,
+    "transpileOnly": true
+  }
+}
+```
+
+### Configuring the CLI
+
+Configuring the MikroORM CLI tools is essential to a "schema first" approach.  We need the migrator to run our initial and later migration, as well as the entity generator, to create our entities out of the schema state.
```
B4nan, line 244:
```suggestion
Configuring the MikroORM CLI tools is essential for the "schema first" approach.  We need the migrator to run our migrations, as well as the entity generator to create our entities out of the schema state.
```

docs/docs/schema-first-guide.md
boenrobot: I am afk right now, so I can't do that change right now...

But yes, in many places in the docs, this is used wrong😅

PR #5793, opened 2024-07-09
feat(entity-generator): support adding groups through metadata hooks

boenrobot:


B4nan (approved):
lgtm, maybe lets mention the `groups` option here too?

https://mikro-orm.io/docs/next/entity-generator#processing-of-generated-metadata

PR #5802, opened 2024-07-11
feat(migrations): allow initial migration to be blank if no entities are defined

boenrobot:
This enables easier "schema first" approach, whereby users first create the schema through an initial migration, before generating the entities out of it.

The "No entities found" error is still present when the "--blank" flag is absent.

When using "--blank" and "--initial" together, the new blank migration is not logged.

### DIFF
--- a/packages/migrations/src/Migrator.ts
+++ b/packages/migrations/src/Migrator.ts
@@ -70,7 +70,7 @@ export class Migrator implements IMigrator {
    */
   async createMigration(path?: string, blank = false, initial = false, name?: string): Promise<MigrationResult> {
     if (initial) {
-      return this.createInitialMigration(path, name);
+      return this.createInitialMigration(path, name, blank);
     }
 
     await this.ensureMigrationsDirExists();
@@ -99,14 +99,14 @@ export class Migrator implements IMigrator {
   /**
    * @inheritDoc
    */
-  async createInitialMigration(path?: string, name?: string): Promise<MigrationResult> {
+  async createInitialMigration(path?: string, name?: string, blank = false): Promise<MigrationResult> {
     await this.ensureMigrationsDirExists();
-    const schemaExists = await this.validateInitialMigration();
-    const diff = await this.getSchemaDiff(false, true);
+    const schemaExists = await this.validateInitialMigration(blank);
+    const diff = await this.getSchemaDiff(blank, true);
     const migration = await this.generator.generate(diff, path, name);
     await this.storeCurrentSchema();
 
-    if (schemaExists) {
+    if (schemaExists && !blank) {
       await this.storage.logMigration({ name: migration[1], context: null });
     }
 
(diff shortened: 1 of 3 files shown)


packages/migrations/src/Migrator.ts
```diff
@@ -189,7 +189,7 @@ export class Migrator implements IMigrator {
       }
     });
 
-    if (expected.size === 0) {
+    if (expected.size === 0 && !blank) {
```
B4nan, line 192: the method will most probably return true since you have no tables or entities, making the migration logged automatically, that wouldn't really work for the use case you described. in fact in that particular case, you should just drop the `--initial` flag and it would work as just fine, the `--initial` flag is needed only when you have a schema and entities and you want to start using migrations, for anything else you don't need it

B4nan: i think we should either drop this completely or rename it, this flag is really just about this extra automatic migration logging (marking them as executed), so having an explicit `--log` (or maybe `--mark-executed`) flag would be more understandable.

boenrobot: Hm... but using only `--blank` without `--initial` would mean that there won't be an error if you _do_ have entities and/or a schema defined, right?

So, while yes, `--blank` would work to create the initial blank migration, `--initial` is still a safeguard to prevent creating that migration when you previously executed migrations on the target database.

`--log`/`--mark-executed` is a separate feature which I agree would be useful, beyond the initial migration.

All of that said, I see your point... maybe the `--blank` + `--initial` combination can be special cased to not log the initial migration?

B4nan: > Hm... but using only --blank without --initial would mean that there won't be an error if you do have entities and/or a schema defined, right?

There is no validation there, you want to create blank migration, and that is what you get. Only the initial flag is validating whether it looks like it's safe to log the migration right ahead.

> --initial is still a safeguard to prevent creating that migration when you previously executed migrations on the target database.

You are asking the CLI to create blank migration, why add a validation, its completely fine to create a blank migration at any point of time. 

> maybe the --blank + --initial combination can be special cased to not log the initial migration?

Yes, I would just return `false` when it's a blank migration. I would also fine if we just ignored the validation completely for blank migrations, but if you see a value in this added validation, I guess there is some :]

boenrobot: > You are asking the CLI to create blank migration, why add a validation, its completely fine to create a blank migration at any point of time.

When not using `--initial`, but only `--blank`... Makes perfect sense. `--blank` on its own is just "make me the file... I'll fill it afterwards". No mention of intent. Safe at any time.

The `--initial` option, to me suggests the intent... The first migration that denotes the current state between entities and schema.

`--initial` without `--blank`, still, the behavior makes perfect sense... The generated initial migration is expected to not be blank, because you should be having some entities to make a migration for, and you should not have logged other migrations either. Not having entities should cause an error here.

> I would also fine if we just ignored the validation completely for blank migrations, but if you see a value in this added validation, I guess there is some :]

Yes, I do see some value :)

When using `--initial` with `--blank`... To me, this says "I want a blank migration that will later be the first executed in this schema"... So it would make sense for the migrator to error if I do have entities (because my "initial" migration would not match my entities) or if I have other migrations logged (because my migration would not be "initial"). One can always use `--blank` without `--initial` if they know better.

PR #5812, opened 2024-07-12
test: add unit test for MsSQL newline encoding issue #5811

DASPRiD:


B4nan:
I'll close this one right ahead since the PR trying to address this already has tests.

PR #5817, opened 2024-07-15
fix(entity-generator): output all DB related info even for virtual properties

boenrobot:
The extra information may be useful to custom types.
In the case of the "type" option, it's also required to skip runtime type inference.


B4nan:
I don't mind merging this, but it feels like a fix on the wrong place, if this is needed in some case, it would also force users doing the regular code first approach to type their virtual properties, which feels quite weird. 

Would be better to do this during discovery automatically.

PR #5828, opened 2024-07-18
ci: fix husky hooks

boenrobot:


B4nan:
i am actually thinking of getting rid of the whole eslint in favor of biome, now its so slow that most of the time i just skip the commit hooks completely

PR #5849, opened 2024-07-22
feat(entity-generator): added support for forceUndefined being false

boenrobot:
There's a separate option "undefinedDefaults" about whether the reported null default values should be changed to "undefined", and those columns be made optional, as opposed to require null.

Also fixed empty string defaults in either mode.

The Opt type modifier is no longer added to properties with defaults in the property declaration. It is still added when the "default" is only at the option, and not at the property declaration. Also not added to properties with "?" type annotation.


tests/features/entity-generator/__snapshots__/ScalarPropsForFks.mysql.test.ts.snap
```diff
@@ -1054,7 +1054,7 @@ export class Test2 {
   parent?: Test2;
 
   @Property({ type: 'integer' })
-  version: number & Opt = 1;
+  version: number = 1;
```
boenrobot, line 1058: I am not 100% sure about the rules on these... is Opt needed here?

tests/features/entity-generator/__snapshots__/MetadataHooks.mysql.test.ts.snap
```diff
@@ -1,6 +1,3327 @@
...
+
+  @Property({ type: 'integer', nullable: true, concurrencyCheck: true })
+  age: number | null = null;
+
+  @Index({ name: 'author2_terms_accepted_index' })
+  @Property({ type: 'boolean', lazy: true })
+  termsAccepted: boolean = false;
+
+  @Property({ type: CustomBooleanType, columnType: 'tinyint(1)', nullable: true })
+  optional: IType<CustomBooleanRuntimeType, boolean> | null = null;
```
boenrobot, line 64: I am also not sure if this is ok...

Can custom types accept a DB null? Can they return null? What about serializers?

tests/features/entity-generator/__snapshots__/ScalarPropsForFks.mysql.test.ts.snap
B4nan: the `Opt` type is needed exactly for this use case.

tests/features/entity-generator/__snapshots__/ScalarPropsForFks.mysql.test.ts.snap
```diff
@@ -773,7 +773,7 @@ export class Book2 {
   publisher?: Publisher2;
 
   @Property({ type: 'string', nullable: true })
-  foo?: string & Opt = 'lol';
+  foo?: string = 'lol';
```
boenrobot, line 776: These ones are ok without Opt though, right? Nullable by prop option, and accepting undefined on type level...

tests/features/entity-generator/__snapshots__/MetadataHooks.mysql.test.ts.snap
boenrobot: OK, I now played around, and I see that indeed, custom types may accept null and may return not null when given null or vice versa...

But also... if the "| null" is not outside of IType, it seems no TS error is detected, which I think is a separate bug perhaps.

i.e.
```
  @Property({ type: CustomBooleanType, columnType: 'tinyint(1)', nullable: true })
  optional: IType<CustomBooleanRuntimeType, boolean> | null = null;
```

is preferable, where possible (i.e. if both the column type and custom type indicate they are nullable), and

```
  @Property({ type: CustomBooleanType, columnType: 'tinyint(1)', nullable: true, default: null })
  optional: IType<CustomBooleanRuntimeType | null, boolean | null> & Opt;
```
doesn't work to detect the cases where the runtime type is null.

and in fact, when combined with other type modifiers, they need to wrap the nullable, i.e.
```
  @Property({ type: CustomBooleanType, columnType: 'tinyint(1)', nullable: true, default: null })
  optional: Opt<IType<CustomBooleanRuntimeType, boolean> | null>;
```
is actually the form that works. If combined with Hidden, it doesn't matter which one is outside and which one is in, but the type arg form needs to be used rather than the "& Hidden" form.

I'll adjust this PR in accordance to that.

boenrobot: Related to this...

Should
```
prop!: string | null;
```

be considered optional for the purposes of create() calls?

Right now, it seems to be, while

```
prop!: Hidden<string | null>;
```

is not, and I'm not entirely sure whether I should be fixing the latter to be optional, or the former to be required.

B4nan: > be considered optional for the purposes of create() calls?

Yes, it should work like this now, I was fixing this recently.

> prop!: Hidden<string | null>;

I'd say it should be `prop!: Hidden<string> | null`, which is technically the same as `prop?: Hidden<string>` if you swap `null` with `undefined`.

Re the custom types and `null`, there might be some cases where we feed `null` into the custom type (and based on what you say, you found some - can you point to some places?), but I don't think we do that from inside hydration. Maybe we should unify this, I am not sure if it makes sense to deal with `null` in custom types in general, I would rather not do that.

boenrobot: > I'd say it should be `prop!: Hidden<string> | null`, which is technically the same as prop?: Hidden<string> if you swap null with undefined.

Oh... Yeah... That's a clever workaround I didn't think of. Great. Shouldn't require type fixes for this case either.

> Re the custom types and null, there might be some cases where we feed null into the custom type (and based on what you say, you found some - can you point to some places?)

I didn't "find" any, I just tried to implement such a custom type (specifically, I played around with the "Password" type from the schema first guide...), and it worked as you'd think... It does get the DB null, which it can convert it to a non-null runtime value that gets hydrated and can be outputted... It can be fed runtime null at create(), which it can convert it to a non-null DB value and save it as the non-null value (different from omitting the value; omitting the value doesn't produce a runtime value, which leads the DB to use its default)... Which is all what I would expect with a custom type involved. Built in types all just leave null untouched on both ends, which is also what you'd expect.

So... To outline it all for myself... in the above case, I guess
```
  @Property({ type: CustomBooleanType, columnType: 'tinyint(1)', nullable: true, default: null })
  optional!: IType<CustomBooleanRuntimeType, boolean> | null;
```

should do (I.e. move the default up into the option, to give custom types a chance to create a non-null runtime value)... when combined with hidden, I guess
```
  @Property({ type: CustomBooleanType, columnType: 'tinyint(1)', nullable: true, hidden: true, default: null })
  optional!: Hidden<IType<CustomBooleanRuntimeType, boolean>> | null;
```
or
```
  @Property({ type: CustomBooleanType, columnType: 'tinyint(1)', nullable: true, hidden: true, default: null })
  optional!: (IType<CustomBooleanRuntimeType, boolean> & Hidden) | null;
```

or in cases of mixed nullability where the runtime type is overridden to be not nullable...
```
  @Property({ type: CustomBooleanType, columnType: 'tinyint(1)', nullable: true, hidden: true, default: null })
  optional!: IType<CustomBooleanRuntimeType, (boolean & Hidden) | null>;
```
or if a custom nullable string serializer type is added, I guess
```
  @Property({ type: CustomBooleanType, columnType: 'tinyint(1)', nullable: true, hidden: true, default: null })
  optional!: IType<CustomBooleanRuntimeType, boolean | null, (string & Hidden) | null>;
```

or by explicitly overriding optional to be true, we'd have
```
  @Property({ type: CustomBooleanType, columnType: 'tinyint(1)', nullable: true, hidden: true, default: null })
  optional?: IType<CustomBooleanRuntimeType, boolean | null, (string & Hidden) | null>;
```

And because the default will be always moved into the option and optionality can be determined with "!" or "?", there shouldn't be a case where Opt, Hidden and IType are all present.

B4nan (changes requested):

docs/versioned_docs/version-6.3/entity-generator.md
```diff
@@ -84,6 +84,8 @@ Available options:
...
...ld be configured to match the configuration you intend to use at runtime. When set to `true` (the default of this option for backwards compatibility; not default for the runtime option), nullable properties get treated as if `null` is not a possible value, with type hints reflecting that, and `null` defaults being omitted. When set to `false`, `\| null` is added to the type annotations based on the `nullable` option, and if `default` is set to `null`, it will be outputted.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  |
```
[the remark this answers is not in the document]
boenrobot: 👍

packages/entity-generator/src/EntityGenerator.ts
```diff
@@ -283,6 +287,8 @@ export class EntityGenerator {
         } else if (prop.kind === ReferenceKind.ONE_TO_ONE && !prop.mappedBy) {
           newProp.kind = ReferenceKind.ONE_TO_ONE;
           newProp.nullable = true;
+          newProp.default = null;
+          newProp.defaultRaw = 'null';
```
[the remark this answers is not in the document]
boenrobot: It's here to keep the meta data consistent and extensible by hooks, while and the serialization logic in SourceFile remains straight forward.

The property generated here is optional, so it needs to be kept that way regardless of settings, unless otherwise overwritten by hooks.

Based on the setting `undefinedDefaults`, the "default" added here may be turned into "undefined" later on, but either way, it stays optional.

Example from `MetadataHooks.mysql.test.ts.snap`, with `forceUndefined: false` and `undefinedDefaults: false`:

```typescript
  @OneToOne({ entity: () => Test2, mappedBy: 'book' })
  test2: Test2 | null = null;
```

With `undefinedDefaults: true`, the above would be serialized as

```typescript
  @OneToOne({ entity: () => Test2, mappedBy: 'book' })
  test2?: Test2 | null;
```

And with `forceUndefined: true` (the current default), it would be serialized as:

```typescript
  @OneToOne({ entity: () => Test2, mappedBy: 'book' })
  test2?: Test2;
```

So... optional regardless.

If the above lines were absent (or removed through the hooks...), that would be serialized as

```typescript
  @OneToOne({ entity: () => Test2, mappedBy: 'book' })
  test2!: Test2 | null;
```

which would make it required on a type level (even if em.create() is ok with that...), and that's probably wrong in most cases. If users decide to override this in hooks (to get this last result), they'll probably do so with some hacky logic (defined at a custom base class perhaps) in mind.

packages/core/src/utils/Configuration.ts
```diff
@@ -126,6 +126,8 @@ export class Configuration<D extends IDatabaseDriver = IDatabaseDriver, EM exten
       prefixMode: 'absolute',
     },
     entityGenerator: {
+      forceUndefined: true,
```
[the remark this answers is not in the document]
boenrobot: Very much agree.

Considering how the entity generator is used in practice, IMO, it's OK to have the default be "false" to match the runtime default and potentially break custom entity generation scripts. If users really need the old behavior due to fragile entity generation scripts, it's trivial to adjust the setting back to `true`...

But you wanted to not have a BC break in v6 for this, so... we're stuck with this mismatched default at least for v6.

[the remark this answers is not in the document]
B4nan: yeah, lets keep it the way you have it, and normalize in v7

PR #5856, opened 2024-07-23
docs: fixed some grammar errors and added some notices and warnings in the guides

boenrobot:


docs/docs/guide/01-first-entity.md
```diff
@@ -50,7 +50,7 @@ npm install --save-dev @mikro-orm/cli \
 
 ## ECMAScript Modules
 
-You probably heard about ECMAScript Modules (ESM), but this might easily be the first time you try them. Now keep in mind - the whole ecosystem is far away from ready, and this guide is using the ESM mainly to show how it is possible. Many dependencies are not ESM ready, and often there are weird workarounds needed. MikroORM is no exception to this - there are quirks as well, namely in dynamic imports of TypeScript files under test setup.
+You probably heard about ECMAScript Modules (ESM), but this might easily be the first time you try them. Now keep in mind - the whole ecosystem is far away from ready, and this guide is using the ESM mainly to show how it is possible. You do not have to use ECMAScript Modules to use MikroORM. Many dependencies are not ESM ready, and often there are weird workarounds needed. MikroORM is no exception to this - there are quirks as well, namely in dynamic imports of TypeScript files under test setup.
```
B4nan, line 53: I can see this in bold or even a `> quote` to make it more visible.

docs/docs/schema-first-guide.md
```diff
@@ -42,7 +42,9 @@ To take a peek at the final project we will be building, try cloning the [`mikro
 git clone https://github.com/mikro-orm/schema-first-guide.git
 ```
 
-We will use MySQL for this project. Other database engines follow the same process. If you're building an application from scratch (as opposed to migrating an existing application), you can use GUI tools (in the case of MySQL, this includes f.e. MySQL Workbench) to make this part of the process easier. We are also assuming you already have MySQL itself installed locally and can connect to it via the username "root" and no password.
+We will use MySQL for this project. Other database engines follow the same process. We are also assuming you already have MySQL itself installed locally and can connect to it via the username "root" and no password.
+
+In general, if you're building an application from scratch (as opposed to migrating an existing application), you can use GUI tools (in the case of MySQL, this includes f.e. MySQL Workbench) to make this part of the process easier.
```
B4nan, line 47: Btw we use `e.g.` instead of `f.e.` in here, I believe that's the standard in English, it's actually the first time  I see `f.e.` written somewhere.

```suggestion
In general, if you're building an application from scratch (as opposed to migrating an existing application), you can use GUI tools (in the case of MySQL, this includes e.g. MySQL Workbench) to make this part of the process easier. 
```

boenrobot: When in the middle of a sentence, "f.e." is more appropriate.

But if you insist on "e.g.", it would be correct to say
```(e.g. In the case of MySQL, this includes MySQL Workbench)```
(putting "e.g." at the beginning makes it correct...)

docs/docs/guide/01-first-entity.md
boenrobot: Done. Moved that into a quote.

PR #5857, opened 2024-07-23
refactor: use "node:" protocol prefix for all nodejs imports

boenrobot:


B4nan (approved):
can we lint this somehow? i can already see myself using auto import which wont use the `node:` prefix :]

B4nan:
Hmm, I would prefer something else than a package with a huge dependency list (I am not in "team Sindre", I actually hate the way his packages are designed, bloating the dependencies crazily). But I can live with this since it's just a dev dep.

I would very much like to drop eslint completely in favor of something like `oxlint` or `biome`, not sure if they have this rule implemented already.

B4nan:
Nice, I'll try to find some time to check them out again, last time I played with oxlint and it was really nice. Any preferences from your end (or any problems with the switch you could see)?

PR #5860, opened 2024-07-23
fix(mssql): do not escape as many charecters

lart2150:
The fix for mikro-orm#5811 solved new line but there are other escape characters that when escaped don't persist correctly in mssql.

### DIFF
--- a/packages/mssql/src/MsSqlPlatform.ts
+++ b/packages/mssql/src/MsSqlPlatform.ts
@@ -35,7 +35,7 @@ export class MsSqlPlatform extends AbstractSqlPlatform {
   override init(orm: MikroORM): void {
     super.init(orm);
     // do not double escape backslash inside strings
-    SqlString.CHARS_GLOBAL_REGEXP = /[\0\b\f\t\r\v\x1a']/g; // eslint-disable-line no-control-regex
+    SqlString.CHARS_GLOBAL_REGEXP = /[\0\b\x1a']/g; // eslint-disable-line no-control-regex
   }
 
   override usesOutputStatement(): boolean {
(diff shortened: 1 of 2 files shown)


lart2150:
I've updated the pr to test all characters that were in the regex.

packages/mssql/src/MsSqlPlatform.ts
```diff
@@ -35,7 +35,7 @@ export class MsSqlPlatform extends AbstractSqlPlatform {
   override init(orm: MikroORM): void {
     super.init(orm);
     // do not double escape backslash inside strings
-    SqlString.CHARS_GLOBAL_REGEXP = /[\0\b\f\t\r\v\x1a']/g; // eslint-disable-line no-control-regex
+    SqlString.CHARS_GLOBAL_REGEXP = /[']/g;
```
B4nan, line 38: if all that needs escaping is really the quotation mark, maybe we should remove the use of that https://github.com/kylefarris/tsqlstring library completely with custom implementation, since this is really bad finding. looking at the implementation, its full of regexps, so it wont be very performant.

B4nan (approved):

packages/mssql/src/MsSqlPlatform.ts
lart2150: This change passes the test.  Do you want me to open another PR with this change?

`    if (value instanceof UnicodeString) {
      return `N'${value.value
        .replace(/'/g, `''`)
        .replace(/\\\\/, '\\\\')}'`;
    }`

Changing the escape at the fallthrough seems to cause all kinds of fun breaking.  I tried adding a case for typeof number but it was still exploding when trying to inspect the current database schema.

B4nan: nono lets keep it this way, i meant we could completely remove that dependency and implement it ourselves, that will take time and i have better things to work on right now :]

PR #5875, opened 2024-07-27
docs: fix `make` method usage example in seeding section

epaulet:
Hello, this is just fixing a typo in the docs. The `make` method requires a number as its first argument, but the documents do not include that.

I added correct examples for `make` and `makeOne`, similar to the existing examples for `create` and `createOne` below it.


B4nan (approved):
thanks

PR #5888, opened 2024-08-01
docs: improve grammar and typos of docs

sergioalmela:
Docs of the latest version revised to improve English grammar and fix typos

### DIFF
--- a/docs/versioned_docs/version-6.3/defining-entities.md
+++ b/docs/versioned_docs/version-6.3/defining-entities.md
@@ -331,7 +331,7 @@ properties: {
 
 We can set default value of a property in 2 ways:
 
-1. Use runtime default value of the property. This approach should be preferred as long as we are not using any native database function like `now()`. With this approach our entities will have the default value set even before it is actually persisted into the database (e.g. when we instantiate new entity via `new Author()` or `em.create(Author, { ... }))`.
+1. Use runtime default value of the property. This approach should be preferred as long as we are not using any native database function like `now()`. With this approach our entities will have the default value set even before it is actually persisted into the database (e.g. when we instantiate new entity via `new Author()` or `em.create(Author, { ... })`).
 
 <Tabs
 groupId="entity-def"
--- a/docs/versioned_docs/version-6.3/defining-entities.md
+++ b/docs/versioned_docs/version-6.3/defining-entities.md
@@ -331,7 +331,7 @@ properties: {
 
 We can set default value of a property in 2 ways:
 
-1. Use runtime default value of the property. This approach should be preferred as long as we are not using any native database function like `now()`. With this approach our entities will have the default value set even before it is actually persisted into the database (e.g. when we instantiate new entity via `new Author()` or `em.create(Author, { ... })`.
+1. Use runtime default value of the property. This approach should be preferred as long as we are not using any native database function like `now()`. With this approach our entities will have the default value set even before it is actually persisted into the database (e.g. when we instantiate new entity via `new Author()` or `em.create(Author, { ... }))`.
 
 <Tabs
 groupId="entity-def"
@@ -437,7 +437,7 @@ properties: {
   </TabItem>
 </Tabs>
 
-Note that we use the `Opt` type to intersect with the property type to tell the ORM (on type level) that the property should be considered optional for input types (e.g. in `em.create()`)), but will be present for managed entities (e.g. `EntityDTO` type).
+Note that we use the `Opt` type to intersect with the property type to tell the ORM (on type level) that the property should be considered optional for input types (e.g. in `em.create()`), but will be present for managed entities (e.g. `EntityDTO` type).
 
 ## Enums
 
@@ -554,7 +554,7 @@ properties: {
 
 ### PostgreSQL native enums
 
-By default, the PostgreSQL driver, represents enums as a text columns with check constraints. Since v6, you can opt-in for a native enums by setting the `nativeEnumName` option.
+By default, the PostgreSQL driver, represents enums as a text columns with check constraints. Since v6, you can opt in for a native enums by setting the `nativeEnumName` option.
 
 <Tabs
 groupId="entity-def"
@@ -1041,7 +1041,7 @@ We can define custom types by extending `Type` abstract class. It has 4 optional
 
 - `toJSON(value: any, platform: Platform): any`
 
-  Converts a value from its JS representation to its serialized JSON form of this type. By default converts to the database value.
+  Converts a value from its JS representation to its serialized JSON form of this type. By default, converts to the database value.
 
 - `getColumnType(prop: EntityProperty, platform: Platform): string`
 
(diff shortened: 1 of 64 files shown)


boenrobot (approved):

B4nan (changes requested):
thanks, but please make all those changes to the next version too, you only modified the v6.3 snapshot

docs/versioned_docs/version-6.3/defining-entities.md
```diff
@@ -331,7 +331,7 @@ properties: {
 
 We can set default value of a property in 2 ways:
 
-1. Use runtime default value of the property. This approach should be preferred as long as we are not using any native database function like `now()`. With this approach our entities will have the default value set even before it is actually persisted into the database (e.g. when we instantiate new entity via `new Author()` or `em.create(Author, { ... })`.
+1. Use runtime default value of the property. This approach should be preferred as long as we are not using any native database function like `now()`. With this approach our entities will have the default value set even before it is actually persisted into the database (e.g. when we instantiate new entity via `new Author()` or `em.create(Author, { ... }))`.
```
B4nan, line 334:
```suggestion
1. Use runtime default value of the property. This approach should be preferred as long as we are not using any native database function like `now()`. With this approach our entities will have the default value set even before it is actually persisted into the database (e.g. when we instantiate new entity via `new Author()` or `em.create(Author, { ... })`).
```

docs/versioned_docs/version-6.3/deployment.md
```diff
@@ -118,7 +118,7 @@ await MikroORM.init({
 
 This will make use of a Webpack feature called [dynamic imports](https://webpack.js.org/guides/code-splitting/#dynamic-imports). This way you can import dependencies as long as part of the path is known.
 
-In following example [`require.context`](https://webpack.js.org/guides/dependency-management/#requirecontext) is used. This 'function' is only usable during the building process from Webpack so therefore there is an alternative solution provided that will as long as the environment variable WEBPACK is not set (e.g. during development with `ts-node`).
+In following example [`require.context`](https://webpack.js.org/guides/dependency-management/#requirecontext) is used. This 'function' is only usable during the building process from Webpack so there is an alternative solution provided that will as long as the environment variable WEBPACK is not set (e.g. during development with `ts-node`).
```
B4nan, line 121: this sounds weird, maybe better to remove `so` instead of `therefore`?

docs/versioned_docs/version-6.3/filters.md
```diff
@@ -128,7 +128,7 @@ em.find(Book, {}, { filters: false }); // disabled all filters, so truly `{}`
...
 
-Since v6, filters are applied to the relations too, as part of `JOIN ON` condition. If a filter exists on a M:1 or 1:1 relation target, such an entity will be automatically joined, and when the foreign key is defined as `NOT NULL`, it will result in an `INNER JOIN` rather than `LEFT JOIN`. This is especially important for implementing soft deletes via filters, as the foreign key might point to a soft-deleted entity. When this happens, the automatic `INNER JOIN` will result in such a record not being returned at all. You can disable this behavior via `autoJoinRefsForFilters` ORM option.
+Since v6, filters are applied to the relations too, as part of `JOIN ON` condition. If a filter exists on an M:1 or 1:1 relation target, such an entity will be automatically joined, and when the foreign key is defined as `NOT NULL`, it will result in an `INNER JOIN` rather than `LEFT JOIN`. This is especially important for implementing soft deletes via filters, as the foreign key might point to a soft-deleted entity. When this happens, the automatic `INNER JOIN` will result in such a record not being returned at all. You can disable this behavior via `autoJoinRefsForFilters` ORM option.
```
B4nan, line 131: to me this reads as `on a many to one`, i hope nobody reads `m:1` as `em to one`

```suggestion
Since v6, filters are applied to the relations too, as part of `JOIN ON` condition. If a filter exists on a M:1 or 1:1 relation target, such an entity will be automatically joined, and when the foreign key is defined as `NOT NULL`, it will result in an `INNER JOIN` rather than `LEFT JOIN`. This is especially important for implementing soft deletes via filters, as the foreign key might point to a soft-deleted entity. When this happens, the automatic `INNER JOIN` will result in such a record not being returned at all. You can disable this behavior via `autoJoinRefsForFilters` ORM option.
```

docs/versioned_docs/version-6.3/inheritance-mapping.md
```diff
@@ -118,7 +118,7 @@ export class Employee extends Person {
 
 ### Explicit discriminator column
 
-The `discriminatorColumn` specifies the name of special column that will be used to define what type of class should given row be represented with. It will be defined automatically for us and it will stay hidden (it won't be hydrated as a regular property).
+The `discriminatorColumn` specifies the name of the special column that will be used to define what type of class a given row should be represented with. It will be defined automatically for us, and it will stay hidden (it won't be hydrated as a regular property).
```
B4nan, line 121:
```suggestion
The `discriminatorColumn` specifies the name of a special column that will be used to define what type of class a given row should be represented with. It will be defined automatically for us, and it will stay hidden (it won't be hydrated as a regular property).
```

sergioalmela:
> thanks, but please make all those changes to the next version too, you only modified the v6.3 snapshot

Where is the next folder?

docs/versioned_docs/version-6.3/relationships.md
```diff
@@ -18,7 +18,7 @@ Relations can be unidirectional and bidirectional. Unidirectional are defined on
 
 > Many instances of the current Entity refer to One instance of the referred Entity.
 
-There are multiple ways how to define the relationship, all of following is equivalent:
+There are multiple ways how to define the relationship, all following is equivalent:
```
B4nan, line 21: this feels wrong to me, if something, i would add a definite article

```suggestion
There are multiple ways how to define the relationship, all of the following is equivalent:
```

docs/versioned_docs/version-6.3/relationships.md
```diff
@@ -45,7 +45,7 @@ You can also specify how operations on given entity should [cascade](./cascading
 
 > One instance of the current Entity has Many instances (references) to the referred Entity.
 
-Again, all of following is equivalent:
+Again, all following is equivalent:
```
B4nan, line 48: same

B4nan (changes requested):
found few more weird things. i guess you ran this through Grammarly, right? most of the places were indeed worth improving, i just feel like you picked the first suggestion and that's not always the correct one. 

not a native speaker myself, so all of my comments are up for a discussion 🙃

B4nan:
> Where is the next folder?

in the root https://github.com/mikro-orm/mikro-orm/tree/master/docs

sergioalmela:
@B4nan Hi, just checked all changes again, with more natural meaning :)

B4nan:
gave it a quick look and i dont see anything wrong right now, so lets port the changes to the next docs

sergioalmela:
> docs

Done :)

B4nan (approved):

PR #5892, opened 2024-08-02
fix(knex): update `PoolConfig` interface to match what knex supports

ladal1:
Knex moved to tarn.js way back (https://github.com/knex/knex/issues/4395), this type hasn't kept up

The current type includes options which will prevent the pool from starting up as tarn validates the provided options (`evictionRunIntervalMillis`) and is missing options which can be useful

### DIFF
--- a/packages/core/src/utils/Configuration.ts
+++ b/packages/core/src/utils/Configuration.ts
@@ -509,7 +509,7 @@ export interface SeederOptions {
 
 export interface PoolConfig {
   name?: string;
-  afterCreate?: (...a: unknown[]) => unknown;
+  afterCreate?: Function;
   min?: number;
   max?: number;
   refreshIdle?: boolean;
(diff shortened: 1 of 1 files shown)


packages/core/src/utils/Configuration.ts
```diff
@@ -509,7 +509,7 @@ export interface SeederOptions {
 
 export interface PoolConfig {
   name?: string;
-  afterCreate?: (...a: unknown[]) => unknown;
+  afterCreate?: Function;
```
B4nan, line 512: i would keep this one as it was

B4nan (approved):

packages/core/src/utils/Configuration.ts
ladal1: I did it for parity, either way makes sense to me

B4nan: its not that important, we had eslint rule to disallow `Function` but i got rid of that some time ago since i got tired of the eslint ignore comments :D

PR #5911, opened 2024-08-07
docs(example): add missing mikro-orm section to package.json

yeonuk-hwang:
The documentation mentions that adding the mikro-orm section to package.json enables TypeScript support for the CLI. However, the actual example code does not include this section.

Based on the information provided in [Checkpoint 1](https://stackblitz.com/edit/mikro-orm-getting-started-guide-cp-1?file=package.json), I have added the necessary mikro-orm configuration to the package.json example to ensure consistency and correctness.

### DIFF
--- a/docs/docs/guide/01-first-entity.md
+++ b/docs/docs/guide/01-first-entity.md
@@ -164,7 +164,10 @@ Save this file into `src/mikro-orm.config.ts`, so it will get compiled together 
 {
   "type": "module",
   "dependencies": { ... },
-  "devDependencies": { ... }
+  "devDependencies": { ... },
+  "mikro-orm": {
+    "useTsNode": true
+  }
 }
 ```
 


B4nan:
this is on purpose, it was removed with v6.3 since its no longer needed. we need to update the text instead of the code

yeonuk-hwang:
Oh, I see. Thanks for the information. I'm just starting to learn about MikroORM by reading through the documentation. I've noticed there are similar inconsistencies between the text and code examples in the Getting Started guide, which can be quite confusing.

I'm not entirely sure how the text should be updated, so I'm unable to create a PR for that specific change. Could you please close this PR?

By the way, I think the documentation and this package are excellent. Thank you!

B4nan:
Can you point to such examples? I'll be happy to fix them myself.

yeonuk-hwang:
I saw your comment a bit late. While I might not be able to pinpoint exactly what seemed odd, I have tried to understand and address any issues as best as I could. If I come across any more issues in the future, I will add comments or create new issues.

In the meantime, I have identified and addressed the parts that were definitely not working before seeing your comment, and I have submitted the changes.

Here is the PR: [#5913](https://github.com/mikro-orm/mikro-orm/pull/5913)

If there are any parts in the PR that need adjustments, please feel free to modify them, or close the PR and make the changes you see fit.

Feel free to adjust any parts of the message as needed!

yeonuk-hwang:
https://mikro-orm.io/docs/guide/relationships#creating-entity-graph

In this document, the following content is provided as an example code:

```ts
// create new user entity instance via constructor
const user = new User();
user.email = 'foo@bar.com';
user.fullName = 'Foo Bar';
user.password = '123456';

// fork first to have a separate context
const em = orm.em.fork();

// first mark the entity with `persist()`, then `flush()`
await em.persist(user).flush();

// clear the context to simulate fresh request
em.clear();

// create the article instance
const article = em.create(Article, {
  title: 'Foo is Bar',
  text: 'Lorem impsum dolor sit amet',
  author: user.id,
});

// `em.create` calls `em.persist` automatically, so flush is enough
await em.flush();
console.log(article);
```

And below that, it mentions that if you run this code, an error will occur stating that `createdAt` and `updatedAt` are missing.

However, in reality, the error that occurs includes not only `createdAt` and `updatedAt`, but also `slug` and `description` are missing. At this point, since there is no code to calculate slug and description using the constructor yet, and no optional handling has been done, you need to include slug and description when creating.

I would like to inform you that this part also needs to be modified. Thank you.

yeonuk-hwang:
https://mikro-orm.io/docs/guide/project-setup#seeding-the-database

In the document above,

```ts
expect(res.json()).toMatchObject({
  items: [
    { author: 1, slug: 'title-13', title: 'title 1/3', text: 'text text text 1/3' },
    { author: 1, slug: 'title-23', title: 'title 2/3', text: 'text text text 2/3' },
    { author: 1, slug: 'title-33', title: 'title 3/3', text: 'text text text 3/3' },
  ],
  total: 3,
});
```

You have written the test matcher code like this, but the test actually fails. This is because the expected value includes the text property, which is a lazy scalar type. However, in the example code of app.ts written earlier, the code for populating is missing as shown below.

Therefore, we should either exclude the text property from the expected value in the test, or populate it in the app.ts code.

```ts
import { RequestContext } from '@mikro-orm/core';
import { fastify } from 'fastify';
import { initORM } from './db.js';

export async function bootstrap(port = 3001) {
  const db = await initORM();
  const app = fastify();

  // register request context hook
  app.addHook('onRequest', (request, reply, done) => {
    RequestContext.create(db.em, done);
  });

  // shut down the connection when closing the app
  app.addHook('onClose', async () => {
    await db.orm.close();
  });

  // register routes here
  app.get('/article', async request => {
    const { limit, offset } = request.query as { limit?: number; offset?: number };
    const [items, total] = await db.article.findAndCount({}, {
      limit, offset,
    });

    return { items, total };
  });

  const url = await app.listen({ port });

  return { app, url };
}
```

yeonuk-hwang:
https://mikro-orm.io/docs/guide/advanced#validation-via-zod

In this chapter, the process of installing Zod is missing.

B4nan:
> The documentation mentions that adding the mikro-orm section to package.json enables TypeScript support for the CLI.

I was checking the original problem again, and it feels like you were looking at older docs version, there is no place mentioning the `useTsNode` in the getting started guide in v6.3 or next, it is only in the older versions (and there it needs to stay). I will update the stackblitz projects, that's a good point.

B4nan:
Addressed the other issues via #5920 so closing this one. Thanks a lot!

yeonuk-hwang:
Thanks a lot, Have a good one!

PR #5913, opened 2024-08-07
docs(guide): resolve 'No entities found' error in test setup

yeonuk-hwang:
The initial test setup was missing the `entities` option in the ORM configuration, leading to a 'No entities found' error. Updated `initTestApp` function to include the existing configuration from `mikro-orm.config.js` and override specific settings for testing.

- Imported the existing ORM configuration.
- Added `entities` option to the configuration for test environment.


B4nan (commented):
adjust + port to v6.3

B4nan (approved):

PR #5915, opened 2024-08-07
refactor(core): adjusted the types to properly extract the hidden and optional indicators

boenrobot:


B4nan:
I think you are too concerned with edge cases nobody even reported, I wouldn't bother with custom types that can remap null to a value now. What else is this trying to solve? The changes are quite complex (which I am not very happy to see, as it can lead to perf issues).

B4nan:
> propWithHiddenUnion: (string | null) & Hidden;

I wouldn't bother with that.

> propWithHiddenUnion: (string & Hidden) | null;

And if this doesn't work for the `Hidden` type, that is what we should try to fix.

> And IType is merely complicating stuff further.

Only if you consider null as a valid value for it, which I wouldn't care about really. I believe we never hydrate nulls into custom types.

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/hydration/ObjectHydrator.ts#L91-L93

PR #5916, opened 2024-08-07
fix(postgres): respect `timezone` option and interpret `timestamp` columns in UTC by default

B4nan:
Postgres driver always used to write dates as UTC, but it was parsing them back in local timezone by default. Now it properly respects the `forceUtcTimezone` and `timezone` options. The postgres driver defaults to `forceUtcTimezone: true` since it used to be writing everything in UTC anyway. 

Other drivers behave differently, notably MySQL and MSSQL use `forceUtcTimezone: false` by default, changing that would be breaking, but we might want to normalize this in v7.

This also improves the processing of properties with a custom type and a runtime type `Date`, those used to be always converted during result mapping, now we only do such conversion if the custom type is not a numeric type like int/bigint.

Closes #5591

### DIFF
--- a/packages/core/src/hydration/ObjectHydrator.ts
+++ b/packages/core/src/hydration/ObjectHydrator.ts
@@ -100,7 +100,7 @@ export class ObjectHydrator extends Hydrator {
         );
       } else if (prop.runtimeType === 'boolean') {
         ret.push(`    entity${entityKey} = !!data${dataKey};`);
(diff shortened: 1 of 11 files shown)


B4nan:
@boenrobot what are your thoughts on this? Should we unify it in v7 in MySQL/MSSQL so they also default to storing dates in UTC?

PR #5927, opened 2024-08-09
fix(entity-generator): unknown type defaults always use default/defaultRaw, never runtime

boenrobot:
The type may be ignored for schema changes if it features length, since we can not guarantee that the "length" is the first argument (if at all present), yet is what the generated columnType assumes.
If the type is ignored for that reason, the "extra" (if present) will also be ignored. Metadata hooks may remove the ignoreSchemaChanges option as a way to confirm it is OK.

The default values of unknown types are left unwrapped in MySQL and MSSQL, which leaves them written as defaultRaw.

### DIFF
--- a/packages/core/src/types/UnknownType.ts
+++ b/packages/core/src/types/UnknownType.ts
@@ -1,8 +1,8 @@
-import { StringType } from './StringType';
 import type { EntityProperty } from '../typings';
 import type { Platform } from '../platforms/Platform';
+import { Type } from './Type';
 
-export class UnknownType extends StringType {
+export class UnknownType extends Type<unknown | null | undefined, unknown | null | undefined> {
 
   override getColumnType(prop: EntityProperty, platform: Platform) {
     return prop.columnTypes?.[0] ?? platform.getVarcharTypeDeclarationSQL(prop);
@@ -14,6 +14,10 @@ export class UnknownType extends StringType {
 
   override compareAsType(): string {
     return 'unknown';
+  }
+
+  override ensureComparable(): boolean {
+    return false;
   }
 
 }
--- a/packages/entity-generator/src/SourceFile.ts
+++ b/packages/entity-generator/src/SourceFile.ts
@@ -258,28 +258,11 @@ export class SourceFile {
               return prop.runtimeType;
             }
 
-            const mappedDeclaredType = this.platform.getMappedType(prop.type);
-            const mappedRawType = (prop.customTypes?.[0] ?? ((prop.type !== 'unknown' && mappedDeclaredType instanceof UnknownType)
-              ? this.platform.getMappedType(prop.columnTypes[0])
-              : mappedDeclaredType));
-            const rawType = mappedRawType.runtimeType;
+            const breakdownOfIType = this.breakdownOfIType(prop);
 
-            const serializedType = (prop.customType ?? mappedRawType).runtimeType;
-
-            // Add non-lib imports where needed.
-            for (const typeSpec of [prop.runtimeType, rawType, serializedType]) {
-              const simplePropType = typeSpec.replace(/\[]+$/, '');
-              if (!primitivesAndLibs.includes(simplePropType)) {
-                this.entityImports.add(simplePropType);
-              }
-            }
-
-            if (prop.runtimeType !== rawType || rawType !== serializedType) {
+            if (typeof breakdownOfIType !== 'undefined') {
               hasITypeWrapper = true;
-              if (rawType !== serializedType) {
-                return `${this.referenceCoreImport('IType')}<${prop.runtimeType}, ${rawType}, ${serializedType}>`;
-              }
-              return `${this.referenceCoreImport('IType')}<${prop.runtimeType}, ${rawType}>`;
+              return `${this.referenceCoreImport('IType')}<${breakdownOfIType.join(', ')}>`;
             }
 
             return prop.runtimeType;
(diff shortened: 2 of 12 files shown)


B4nan:
I don't recall if the extension is needed, but the source of this is that we do consider unknown types as strings for schema purposes. This is because of reflect metada sometimes fails to produce the right type, we initially had the default to `object` but that was picked up as `json` and converting JSON columns to something else was a mess by that time (now it should be hopefully all cleared and working).

If it doesn't break any tests, we can surely remove this extension, we are not using anything from the parent class apparently.

packages/entity-generator/src/SourceFile.ts
```diff
@@ -570,7 +570,7 @@ export class SourceFile {
         const serializedType = (prop.customType ?? mappedRawType).runtimeType;
 
         return prop.runtimeType !== rawType || rawType !== serializedType;
-      })()))) {
+      })())))) {
```
B4nan, line 573: this would really deserve extraction to a private method :D

B4nan (approved):

packages/entity-generator/src/SourceFile.ts
boenrobot: Now extracted... and reused already at the other spot where this callback was originally derived from.

To avoid repeated recomputations, I also added a WeakMap. Granted it would usually be accessed only twice per property, but it's still a bit of a boost. And it being WeakMap, I hope for it to be GC-ed by the end of the SourceFile.generate() call, thus not resulting in too much memory overhead.

boenrobot: I also made UnknownType not derive out of StringType. Added ensureComparable false, just in case (performance and what not).

packages/core/src/types/UnknownType.ts
```diff
@@ -16,4 +16,8 @@ export class UnknownType extends StringType {
     return 'unknown';
   }
 
+  override ensureComparable(): boolean {
+    return false;
+  }
```
boenrobot, line 21: ok, so I guess I could remove this ensureComparable maybe, but... is there a test case we could add that would probe whether it needs to be true or false? I'm not quite sure in what scenario this would play out.

packages/entity-generator/src/SourceFile.ts
```diff
@@ -582,6 +555,48 @@ export class SourceFile {
...
+          const r = [prop.runtimeType, rawType, serializedType];
+          this.propTypeBreakdowns.set(prop, r);
+          return r;
+        }
+        const r = [prop.runtimeType, rawType];
+        this.propTypeBreakdowns.set(prop, r);
+        return r;
+    }
+    this.propTypeBreakdowns.set(prop, undefined);
+    return;
```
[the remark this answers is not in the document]
boenrobot: TS complains about "Not all paths return a value" without this line.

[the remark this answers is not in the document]
B4nan: you should specify the return type then, this line doesn't make any sense

packages/core/src/types/UnknownType.ts
boenrobot: ok, so adding such a declaration would make it covered, but... How would you check the effect of this setting being `true` vs it being `false`?

I see `integer` has `ensureComparable` be `false`, while `double` has it to `true`... Which seems to suggest that if `unknownType` has `ensureComparable` be `false`

```@Property({ type: 'unknown', columnType: 'double' })```

would behave differently from

```@Property({ type: 'double' })```

in some circumstances... But what are the circumstances?

File: packages/entity-generator/src/SourceFile.ts (hunk 3 above)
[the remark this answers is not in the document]
Comment (boenrobot) [reply packages/entity-generator/src/SourceFile.ts:599]: Still doesn't work (gives out the same error...).

```return undefined;```

does work.

File: packages/core/src/types/UnknownType.ts (hunk 2 above)
Comment (B4nan) [reply packages/core/src/types/UnknownType.ts]: > How would you check the effect of this setting being true vs it being false?

There is no effect really. I am find with just a coverage ignore comment too, but it feels better to keep it here.

Comment (boenrobot) [reply packages/core/src/types/UnknownType.ts]: > There is no effect really, this is specifically designed to handle JSON columns.

Does that include embeddable properties? If I have an embeddable class referenced with object: true (i.e. a typed JSON column...), and it contains properties like
```@Property({ type: 'unknown', columnType: 'double' })```
and
```@Property({ type: 'double' })```

should there be any observable differences when handling these?

> but it feels better to keep it here.

I agree there... It's just that it feels dirty to have a test that isn't testing anything... and yet is there purely for coverage... I guess I could look for an existing test to add such a prop to... At least then, the test is testing the other thing, and the coverage here is just a side effect.

Comment (B4nan) [reply packages/core/src/types/UnknownType.ts:21]: > `@Property({ type: 'unknown', columnType: 'double' })`

Is this something you put in there or the generator does it? There is no point in explicit `type: 'unknown'`, we shouldn't generate that. There shouldn't be any different I think, its about mapping JSON columns, since drivers like sqlite will give you JSON string and we need to convert it so it can be diffed later when flushing.

Comment (boenrobot) [reply packages/core/src/types/UnknownType.ts:21]: > Is this something you put in there or the generator does it?

I was just following up on your example about

```@Property({ type: UnknownType, columnType: 'int' })```

but replacing "int" with "double" in the discussion, since DoubleType has ensureComparable = true, meaning that manually writing out one or the other might mean different things, unlike "int" where either form would give you a class with ensureComparable = false.

In terms of the entity generator, yes, it would never generate any of the above on its own... The only way it could is if the user explicitly changes `type` from `integer`/`double` to "unknown" via the hooks... And I don't see why would anyone do that.

File: tests/issues/GH910.test.ts [hunk 4]
```diff
@@ -68,9 +68,17 @@ export class CartItem {
   @Property()
   quantity: number;
 
+  @Property({ type: UnknownType, columnType: 'double' })
+  singularPriceUnensured: number;
```
boenrobot, line 72: I've added this prop here for the sake of the coverage...

B4nan (approved):

PR #5930, opened 2024-08-11
fix(entity-generator): ensure `columnType` is emitted correctly and when necessary

B4nan:
Closes #5928


B4nan:
This fixes several things in the end:
- postgres/mysql generates `columnType: 'timestamp'`
- mssql generates `type: 'double'` for `float(53)`
- mssql default lengths are not printed
- mssql lengths are only assigned to whitelisted types

tests/features/entity-generator/__snapshots__/EntityGenerator.mssql.test.ts.snap
```diff
@@ -151,13 +151,13 @@ export class Book2 {
...
 
-  @Property({ type: 'text', length: 2147483647, nullable: true })
+  @Property({ type: 'text', nullable: true })
   perex?: string;
 
   @Property({ type: 'decimal', precision: 8, scale: 2, nullable: true })
   price?: string;
 
-  @Property({ type: 'float', nullable: true })
+  @Property({ type: 'double', nullable: true })
```
boenrobot, line 160: What about float/real with precisions other than 53? Or, I should say 24, since according to [the docs on float/real](https://learn.microsoft.com/en-us/sql/t-sql/data-types/float-and-real-transact-sql?view=sql-server-ver16), any value other than those two is converted to those.

Maybe convert those with length 24 to `type: 'float'`?

boenrobot (approved):

tests/features/entity-generator/__snapshots__/EntityGenerator.mssql.test.ts.snap
B4nan: that's what already happens i believe, i only remapped `float(53)` to double, the rest is still just a `float`. so if something, i would expect its the opposite, only length 53 is considered a double, the rest is all just a float. will add more tests around this i guess.

B4nan: yeah they normalize the precision to either 24 or 53, so even with `float(36)` i get back `float(53)` from the info schema. improved the support for this a bit.

PR #5932, opened 2024-08-11
docs: added authors.yml and changed inline author references to use that instead

boenrobot:


B4nan (approved):
Hmm, those are just warnings, I don't think that's what is breaking the build, they messed up something with imports I believe. I tried v3.5.0 on another repo and it failed, they released 3.5.1 which fixed one import issue but we now see another one here. Could be some component extension we have, in the other repo is was something else in their own code.

Anyway, thanks :]

B4nan:
This time it might be indeed something in our code that needs changes, I can see the `@theme/Unlisted` import which is probably no longer available. It still feels breaking from them :]

B4nan:
They changed the name of that here [#10376](https://github.com/facebook/docusaurus/pull/10376)

B4nan:
Heh, those component overrides were always dangerous. Worst case we can recreate the components (swizzle them again), the changes there are not huge I think, I think its just the giscus comments in there.

PR #5973, opened 2024-08-27
fix(sql): correct cursor conditions application on virtual entities

Kazma830124:
Now we cannot use cursor-based pagination on virtual entities because  `qb.where` will override the `processCursorOptions` return where.

So I move the `qb.where(where)` to beginning.


B4nan (changes requested):

B4nan (approved):
thanks!

PR #5979, opened 2024-08-29
fix(core): check proptotype chain while cloning

icecreamparlor:
I'm using MikroORM 6.3.7 with Postgresql 14.

I'm using MikroORM with ts-jenum.
When I use [ts-jenum](https://github.com/reforms/ts-jenum) in MikroORM's where condition, following error occurs

```
Cannot set property enumName of [object Object] which has only a getter
TypeError: Cannot set property enumName of [object Object] which has only a getter
```

This issue was caused by the [following code](https://github.com/reforms/ts-jenum/blob/e1a45f54537ec5bd6abc258391e656e8c6339595/src/ts/jenum.ts#L181-L184). In the ts-jenum library, the “enumName” of the parent enum type has a getter but no setter implemented. Therefore, when [clone.ts tries to call the setter](https://github.com/mikro-orm/mikro-orm/blob/c0dacd8964786118d094fb6f88c3c136bcfa7115/packages/core/src/utils/clone.ts#L123), an error occurs.

I explored why this issue occurs. The problem arises in MikroORM’s “clone” function when using the `Object.getOwnPropertyDescriptor` function. `Object.getOwnPropertyDescriptor` inspects the properties of a JavaScript object but does not check the properties inherited from its prototype. Therefore, I removed the `Object.getOwnPropertyDescriptor` function in the clone operation and created a function that inspects the properties of the object as well as those inherited from the connected prototype.

### DIFF
--- a/tests/Utils.test.ts
+++ b/tests/Utils.test.ts
@@ -133,27 +133,69 @@ describe('Utils', () => {
     expect(Utils.diff({ a: new ObjectId('00000001885f0a3cc37dc9f0') }, { a: new ObjectId('00000001885f0a3cc37dc9f0') })).toEqual({});
   });
 
-  test('copy', () => {
-    const a = { a: 'a', b: 'c' };
-    const b = Utils.copy(a);
-    b.a = 'b';
-    expect(a.a).toBe('a');
-    expect(b.a).toBe('b');
-    expect(Utils.copy(new Error('foo'))).toEqual(new Error('foo'));
-    expect(Utils.copy(/abc/gim)).toEqual(/abc/gim);
+  describe('copy', () => {
+    test('should copy simple object', () => {
+      const a = { a: 'a', b: 'c' };
+      const b = Utils.copy(a);
+      b.a = 'b';
+      expect(a.a).toBe('a');
+      expect(b.a).toBe('b');
+      expect(Utils.copy(new Error('foo'))).toEqual(new Error('foo'));
+      expect(Utils.copy(/abc/gim)).toEqual(/abc/gim);
 
-    const re = /a/;
-    re.lastIndex = 1;
-    expect(Utils.copy(re)).toEqual(re);
-    expect(Utils.copy(re).lastIndex).toEqual(re.lastIndex);
+      const re = /a/;
+      re.lastIndex = 1;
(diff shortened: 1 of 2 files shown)


B4nan:
Please add a test case, I dont want to touch internals without seeing what you are trying to fix.

icecreamparlor:
@B4nan 
Thank you for letting me know about test cases.
I added test cases. 
I reproduced similar circumstances like problems that occurs in ts-jenums library.

B4nan (approved):
thanks!

tests/Utils.test.ts
```diff
@@ -133,27 +133,69 @@ describe('Utils', () => {
...
+      const child = new Child(expected);
+
+      // when
+      const copied = Utils.copy(child);
+
+      // then
+      expect(copied).toBeInstanceOf(Child);
+      expect(copied.name).toBe(expected);
+
+    })
```
B4nan, line 197:
```suggestion
    });
```

tests/Utils.test.ts
```diff
@@ -133,27 +133,69 @@ describe('Utils', () => {
...
+  })
```
B4nan, line 198:
```suggestion
  });
```

yakpoong:
@B4nan 
I intended to migrate to MikroORM version 6, but due to scheduling constraints, I won't be able to proceed with that migration. Instead, I would like to upgrade to the latest version in the 5.x series. However, during testing with @mikro-orm/core@5.9.8, I encountered the same bug described in this PR.

Would it be possible to backport the changes from this PR to the 5.x series of MikroORM? It would be incredibly helpful if you could include the changes from this PR in a new version of the 5.x series (e.g., 5.9.9).

Thank you so much in advance!

icecreamparlor:
@B4nan 

I intended to migrate to MikroORM version 6, but due to scheduling constraints, I won't be able to proceed with that migration. Instead, I would like to upgrade to the latest version in the 5.x series. However, during testing with @mikro-orm/core@5.9.8, I encountered the same bug described in this PR.

Would it be possible to backport the changes from this PR to the 5.x series of MikroORM? It would be incredibly helpful if you could include the changes from this PR in a new version of the 5.x series (e.g., 5.9.9).

Thank you so much in advance!

PR #5980, opened 2024-08-30
refactor: group tables by schema to load information schema

patrykbojczuk:
This changes `MsSqlSchemaHelper` and `PostgreSqlSchemaHelper` to group tables by schema name and use `IN` operator for table names instead of using `OR` for every schema/table combo, e.g.

```sql
WHERE (
  (ccu.table_name = 'seals' and ccu.table_schema = 'mammals')
OR (ccu.table_name = 'owls' and ccu.table_schema = 'birds')
OR (ccu.table_name = 'parrots' and ccu.table_schema = 'birds')
)

```

becomes

```sql
WHERE (
  (ccu.table_name IN ('seals') and ccu.table_schema = 'mammals')
OR (ccu.table_name IN('owls','parrots') and ccu.table_schema = 'birds')
)

```

This doesn't change the current behavior, yet it makes the entity generator faster (at least for large databases) when multiple tables belong to the same schema (which is almost always the case).

### DIFF
--- a/packages/knex/src/schema/SchemaHelper.ts
+++ b/packages/knex/src/schema/SchemaHelper.ts
@@ -511,7 +511,7 @@ export abstract class SchemaHelper {
   }
 
   getTablesGroupedBySchemas(tables: Table[]): Map<string | undefined, Table[]> {
-    const tablesBySchema = tables.reduce((acc, table) => {
+    return tables.reduce((acc, table) => {
       const schemaTables = acc.get(table.schema_name);
       if (!schemaTables) {
         acc.set(table.schema_name, [table]);
@@ -520,8 +520,6 @@ export abstract class SchemaHelper {
       schemaTables.push(table);
       return acc;
     }, new Map<string | undefined, Table[]>());
-
-    return tablesBySchema;
   }
 
   async getAlterTable?(changedTable: TableDifference, wrap?: boolean): Promise<string>;
--- a/packages/knex/src/schema/SchemaHelper.ts
+++ b/packages/knex/src/schema/SchemaHelper.ts
@@ -510,6 +510,20 @@ export abstract class SchemaHelper {
     return builder;
   }
 
+  getTablesGroupedBySchemas(tables: Table[]): Map<string | undefined, Table[]> {
+    const tablesBySchema = tables.reduce((acc, table) => {
+      const schemaTables = acc.get(table.schema_name);
+      if (!schemaTables) {
+        acc.set(table.schema_name, [table]);
+        return acc;
+      }
+      schemaTables.push(table);
+      return acc;
+    }, new Map<string | undefined, Table[]>());
+
+    return tablesBySchema;
+  }
+
   async getAlterTable?(changedTable: TableDifference, wrap?: boolean): Promise<string>;
 
   get knex(): Knex {
--- a/packages/mssql/src/MsSqlSchemaHelper.ts
+++ b/packages/mssql/src/MsSqlSchemaHelper.ts
@@ -301,15 +301,7 @@ export class MsSqlSchemaHelper extends SchemaHelper {
       return;
     }
 
-    const tablesBySchema = tables.reduce((acc, table) => {
-      const schemaTables = acc.get(table.schema_name);
-      if (!schemaTables) {
-        acc.set(table.schema_name, [table]);
-        return acc;
-      }
-      schemaTables.push(table);
-      return acc;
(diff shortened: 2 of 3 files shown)


packages/mssql/src/MsSqlSchemaHelper.ts
```diff
@@ -301,10 +301,20 @@ export class MsSqlSchemaHelper extends SchemaHelper {
...
+      if (!schemaTables) {
+        acc.set(table.schema_name, [table]);
+        return acc;
+      }
+      schemaTables.push(table);
+      return acc;
+    }, new Map<string | undefined, Table[]>());
+
     const columns = await this.getAllColumns(connection, tables);
     const indexes = await this.getAllIndexes(connection, tables);
```
boenrobot, line 315: Shouldn't these use `tablesBySchema` too?

patrykbojczuk:
@boenrobot 
> Could you also add the same for MySQL and MariaDB?

MySQL/MariaDB already uses `IN` for table names (and, maybe incorrectly, always uses `database()` as a schema name).

packages/mssql/src/MsSqlSchemaHelper.ts
patrykbojczuk: Currently indices and columns aren't selected by schema name, perhaps incorrectly.
I can modify the SQL to use the same logic as with checks/foreign keys.

boenrobot: Go for it 👍 

If any tests break, there's a possibility they were broken in the first place.

Maybe add a test case that fails under the current logic for MSSQL, while it works fine for PostgreSQL, due to PostgreSQL using the schema name?

packages/postgresql/src/PostgreSqlSchemaHelper.ts
```diff
@@ -80,10 +80,20 @@ export class PostgreSqlSchemaHelper extends SchemaHelper {
...
+      if (!schemaTables) {
+        acc.set(table.schema_name, [table]);
+        return acc;
+      }
+      schemaTables.push(table);
+      return acc;
+    }, new Map<string | undefined, Table[]>());
+
+    const columns = await this.getAllColumns(connection, tablesBySchema, nativeEnums);
     const indexes = await this.getAllIndexes(connection, tables);
```
boenrobot, line 94: This one too...

patrykbojczuk: This actually uses `indrelid`, which is `schema_name`.`table_name`, so it still uses `IN`.

packages/mssql/src/MsSqlSchemaHelper.ts
patrykbojczuk: Tests don't break, schema name is returned with the rest of the metadata, so I think it's actually matched to the table in correct schema later on.

boenrobot (approved):

packages/mssql/src/MsSqlSchemaHelper.ts
```diff
@@ -301,10 +301,20 @@ export class MsSqlSchemaHelper extends SchemaHelper {
       return;
     }
 
-    const columns = await this.getAllColumns(connection, tables);
-    const indexes = await this.getAllIndexes(connection, tables);
-    const checks = await this.getAllChecks(connection, tables);
-    const fks = await this.getAllForeignKeys(connection, tables);
+    const tablesBySchema = tables.reduce((acc, table) => {
```
B4nan, line 304: lets move this to the base schema helper to reduce duplication

B4nan (approved):
thanks, just one nit

packages/knex/src/schema/SchemaHelper.ts
```diff
@@ -510,6 +510,20 @@ export abstract class SchemaHelper {
...
+      const schemaTables = acc.get(table.schema_name);
+      if (!schemaTables) {
+        acc.set(table.schema_name, [table]);
+        return acc;
+      }
+      schemaTables.push(table);
+      return acc;
+    }, new Map<string | undefined, Table[]>());
+
+    return tablesBySchema;
```
B4nan, line 524:
```suggestion
    return tables.reduce((acc, table) => {
      const schemaTables = acc.get(table.schema_name);
      if (!schemaTables) {
        acc.set(table.schema_name, [table]);
        return acc;
      }
      schemaTables.push(table);
      return acc;
    }, new Map<string | undefined, Table[]>());
```

PR #6020, opened 2024-09-12
fix(core): fix dynamic imports on Windows

boenrobot:
This most notably affects the migration generator.

### DIFF
--- a/packages/core/src/utils/Utils.ts
+++ b/packages/core/src/utils/Utils.ts
@@ -2,7 +2,7 @@ import { createRequire } from 'node:module';
 import globby, { type GlobbyOptions } from 'globby';
 import { extname, isAbsolute, join, normalize, relative, resolve } from 'node:path';
 import { platform } from 'node:os';
-import { fileURLToPath, pathToFileURL, type URL } from 'node:url';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 import { pathExistsSync } from 'fs-extra';
 import { createHash } from 'node:crypto';
 import { tokenize } from 'esprima';
@@ -1095,6 +1095,10 @@ export class Utils {
         id = pathToFileURL(id).toString();
       } catch {
         // ignore
+      }
+      // For some reason, ".ts" files do not work with "file://" URLs on Windows.
+      if (id.endsWith('.ts')) {
+        id = fileURLToPath(id);
       }
     }
 


B4nan:
> Should we add github actions to run the tests with Windows and MAC OS runners, to prevent this problem from accidentally popping again? Now that the tests are fully green on Windows after this, this seems like the perfect time.

The main reason they are not here is that windows is terribly slow, I do maintain several projects where we have it setup for all the platforms, windows tests are always super flaky and do take twice the time. Feel free to PR this so we can see actual numbers relevant for this project, but I don't want to make the CI checks take twice the time and be 100 times more flaky. We could have a periodic check once a day, no problems with that.

packages/core/src/utils/Utils.ts
```diff
@@ -1091,10 +1091,12 @@ export class Utils {
   static async dynamicImport<T = any>(id: string): Promise<T> {
     /* istanbul ignore next */
     if (platform() === 'win32') {
-      try {
-        id = pathToFileURL(id).toString();
-      } catch {
-        // ignore
+      if (id.startsWith('file://')) {
```
B4nan, line 1094: no need for nesting here

packages/core/src/utils/Utils.ts
```diff
@@ -1091,10 +1091,12 @@ export class Utils {
...
+        try {
+            id = fileURLToPath(id);
```
B4nan, line 1096: how come this was not working when the tests were passing already? the code here is quite old, it wasn't touched for years i think, feels weird that this wouldn't be working before (and the fix is a bit scary since it does the opposite of the old code, right?).

i guess this part is relevant only to ESM projects?

boenrobot: There was a point in time... I'm not sure when exactly... some time after my previous Windows test fixing PR... where the tests were all green... But then something happened and 3 of the Migrator tests started failing, but only on my (Windows) machine.

Because I never touched the migrator before, and we still have the CI, I didn't look into it, until now.

It's possible this has something to do with ESM vs. CJS, but if that's the case, is there a way to run the test suite in both modes?

packages/core/src/utils/Utils.ts
boenrobot: Ah... I see what you mean... ok, moved it into the parent if now.

packages/core/src/utils/Utils.ts
boenrobot: I suppose I could make it try to load the file URL unmodified, and on failure, try to change it to path, and try again, before giving up? That would make Windows + ESM faster than Windows + CJS, while still keeping Windows + CJS working.

I don't really like this solution, but it is one that for sure works.

PR #6021, opened 2024-09-12
ci: added Windows and MacOS runners for the test suite

boenrobot:


B4nan:
Maybe we should run just sqlite tests on windows, I think it should be fine, we can test things that are windows specific regardless of what driver we use, and with sqlite things will be reasonably fast, so the windows job wont take ages.

B4nan:
Closing in favor of #6849, I'll probably merge that later when v7 branch is in master.

PR #6030, opened 2024-09-16
fix: using raw fragment with bigint type (postgres/msql)

linkthai:
When using atomic raw update with bigint column, it will cause error.
```ts
   // BigIntType.js
    convertToDatabaseValue(value) {
        if (value == null) {
            return value;
        }
        return '' + value;
    }
```
The `return '' + value;` trigger `valueOf` on raw so it throws an error, the fix will check if the value is an instance of RawFragment then it will directly return that value without trigger `valueOf`
This PR fix only works with PostgresQL and MySQL.

### DIFF
--- a/docs/docs/entity-manager.md
+++ b/docs/docs/entity-manager.md
@@ -440,6 +440,8 @@ user.lastOrderNumber = order.number; // throws, it could resolve to a different 
 JSON.stringify(order); // throws, raw value cannot be serialized
 ```
 
+Using `raw()` with BigInt type currently only works with PostgresQL and MySQL.
+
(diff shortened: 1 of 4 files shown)


B4nan:
I believe I already told you on discord that this is just a workaround and not a proper fix, the check should be implemented higher in the callstack, not in a type implementation itself.

linkthai:
@B4nan I must have misunderstood the wording in the discord message, you said that it was a good workaround in my app so I thought that was ok to make a PR, oops 😬

B4nan:
well, at least we have the test case, that's surely welcome :]

PR #6053, opened 2024-09-22
fix(core): support dataloader for M:N without inverse side

### LINKED ISSUE #6031: Dataloader failure with ManyToMany relationship
### Describe the bug

With a simple many to many relationship on an entity like this:

```
  @ManyToMany(() => Country)
  countries = new Collection<Country>(this);
```

 Enabling dataloader to load this collection with e.g. `await user.countries.load({dataloader: true})` fails with the following error:

```
TypeError: Cannot read properties of undefined (reading '__helper')

      at helper (node_modules/@mikro-orm/core/entity/wrap.js:24:19)
      at node_modules/@mikro-orm/core/utils/DataloaderUtils.js:155:42
          at Array.filter (<anonymous>)
      at node_modules/@mikro-orm/core/utils/DataloaderUtils.js:179:33
          at Array.map (<anonymous>)
      at DataLoader._batchLoadFn (node_modules/@mikro-orm/core/utils/DataloaderUtils.js:171:34)
```

### Reproduction

https://github.com/tackley/mikro-orm-reproduction

test that illustrates the issue in https://github.com/tackley/mikro-orm-reproduction/blob/master/src/dataloader-m-to-m.test.ts

@mikro-orm/postgresql

### MikroORM version

6.3.10

### Node.js version

20.17.0

### Operating system

Linux

### Validations

B4nan:
Closes #6031


darkbasic:
>> looking at the implementation, it feels like it cannot really work correctly with the various options, namely I can't see any place where we would actually try to handle different where conditions or child populate hints, it feels like the options are internally used only for the keys and not really for what they are meant for

> That's expected, the initial mainlined implementation was supposed to handle just the simplest use cases. Of course the typings should be updated to reflect that.

I was wrong: I in fact did end up implementing filtering/populate in the reference/collection dataloader and it works.
I've updated the tests to reflect that (and I've commented your fix to show that it works without it). Applying your fix again will break filters and tests will fail.

Unfortunately while writing the new tests I've also found one case where the dataloader leads to incorrect results: when you are applying different filters on the same collection instance (you can uncomment some lines in the tests to see how it behaves).

The root of the issue is the fact that if you call the EntityManager and ask for an entity with a specific ID twice, it will return the same instance and thus any associated collection will share the same instance as well.
Without the dataloader if you call `loadItems()` on the same collection with different filters it will fire different queries and in return the timing will be different enough to be able to return the results for each of them before the next one overwrites the internal `.items` property.
With the dataloader on the other hand everything happens on the same tick and so you will get the same result for each of the shared instances.

Obviously even without the dataloader accessing `Collection.getItems()` will share the same issue: the returned results will be the ones of the very last filter you've used (but that's by design I guess).

Right now if you use a dataloader I'm not even sure if it makes sense to have loaded collections considering you can specify different filters and everything happens in the same tick. The same issue could potentially be reproduced without dataloader as well: consider the case where you `load()` a collection with certain filters, then you do something async and finally you run `.getItems()`. If in the meantime another query ended up sharing the same instance and loaded it with different filters by the time you run `.getItems()` you would have returned different results than the ones you previously loaded.

I'm starting to think that `load()` shouldn't accept filters, while running `loadItems()` with any filter shouldn't load the collection: instead it should just return the results. That way a loaded collection would only contain results with the default filters. We could potentially extend the concept by creating a way to instantiate new collections with different sets of filters, which will be stored inside the collection. At this point `.load()` and `.loadItems()` won't need to have filters because it's the collection instance that carries the filters and the problem won't exist. I like this solution more.

B4nan:
Thanks for the additional tests, that's super handy. 

> Without the dataloader if you call loadItems() on the same collection with different filters it will fire different queries

If it behaves like that, it's most likely a bug, the method shouldn't change the collection contents if it's already loaded.

> I'm starting to think that load() shouldn't accept filters, while running loadItems() with any filter shouldn't load the collection: instead it should just return the results.

That's a highway to hell, I don't want different behavior based on filters being present or not. How could you even reason about that (from the user point of view), when filters can be enabled/controlled globally, not just via the options of these methods? I believe we already had a very similar discussion earlier, it's fine that users can do things that make no sense, we don't need to automagically solve all their needs when they don't make sense. The only way to not store the collection items is `Collection.matching`, I want to keep things that way.

darkbasic:
> If it behaves like that, it's most likely a bug, the method shouldn't change the collection contents if it's already loaded.

That makes more sense now, but you will still have to remember to check if the collection is loaded before loading it otherwise you might end up with different results loaded previously by another copy of the same instance.

> That's a highway to hell, I don't want different behavior based on filters being present or not

Why would the behavior be different? If the collection instance carries the filters then in load you won't need to specify them and the behavior will be consistent everywhere. We could even specify which filters we want to instantiate the collection with via additional populate options.

> How could you even reason about that (from the user point of view), when filters can be enabled/controlled globally, not just via the options of these methods?

And what' the problem about it? The collection can be instantiated with global filters enabled or disabled and it will stay consistent in its behavior. If you have filters on a certain entity enabled globally then their collections will be instantiated with such global filters enabled by default, unless you specify differently via the populate options.

> I believe we already had a very similar discussion earlier

I think so as well.

B4nan:
> That makes more sense now, but you will still have to remember to check if the collection is loaded before loading it otherwise you might end up with different results loaded previously by another copy of the same instance which used different filters. 

That's exactly the user's problem, a misusage if you want. You should not use `loadItems` if you want to have new values regardless of the loaded state, just use `init` instead if you want to always run the query and rehydrate.

> Or maybe you are suggesting that load() should throw an error if the collection is already loaded?

Not at all, that is the exact opposite of what the method is/should be doing. Its purpose is to load items if they are not loaded already, period.

> Why would the behavior be different? If the collection instance carries the filters then in load you won't need to specify them and the behavior will be consistent everywhere. We could even specify which filters we want to instantiate the collection with via additional populate options.

Collection instance does not carry any filters (also it is instantiated exactly once, when the entity instance is created, not during population or hydration), filters are on entity metadata plus globally on the EM. And they can be controlled via find options or globally. Also one filter can affect multiple properties.

darkbasic:
> That's exactly the user's problem, a misusage if you want. You should not use loadItems if you want to have new values regardless of the loaded state, just use init instead if you want to always run the query and rehydrate.

But I don't see how that would solve the problem I mentioned before.
Consider the following case:
You have a loaded collection with your desired filters. You wait for an external api call or anything async. When the async function returns you call `.getItems()` on the collection. If in the meantime any other code path (another GraphQL resolver which happened to share the same collection instance for example) initialized the collection using `.init()` with a different set of filters it will return the wrong results.

> Collection instance does not carry any filters

I was referring to what would hypothetically happen ** IF ** collections carried the filters.

B4nan:
That sounds like a use case for the `matching` method instead, which does return the items and won't modify the collection state by default.

darkbasic:
Sure, but the point is that you will **have** to define your own set of rules to protect yourself against collection items changing against your will because otherwise you will have no control against other code paths of your application tampering with your collection states.

If collections stay that way these will be the rules that I will personally enforce (either via linting or something else):
- Never load anything but the default filters (the globally enabled ones plus nothing else) into a collection.
- Thus always assume that if a collection is loaded it has been loaded with the default filters.
- If you need non-default filters always use `matching` to prevent tampering with other instances of the same collection.

That way it is indeed possible to have a consistent state, but if I can't manage to enforce this via linting there won't be any guarantee.

B4nan:
Also, I feel like you are mixing terms, you keep talking about [filters](https://mikro-orm.io/docs/filters) but it feels like you are talking about regular conditions. 

>  other instances of the same collection.

What other instances? One entity = one collection, there are no other instances of the owning entity.

darkbasic:
Before tackling the `matching` dataloader I started to further look into your `MANY_TO_MANY` implementation with `findChildrenFromPivotTable`.
Currently while it supports default filters at the entity level it doesn't support custom where filters in `load`/`loadItems`.
This is a required feature to be able to use it as a base for the `matching` dataloader.
I've added a test (`Collection.load with where (Many to Many + Inverse side + custom where filters)`) to reflect that.
I still didn't look into it, but do you think it would be possible to modify your implementation to support them?

B4nan:
First, we need to fix bugs, then we can add more features, so let's not focus on the matching dataloader now. I am quite close to removing the tests you added here, that only show things not relevant to what I was trying to fix, and just merge it. It's been more than half a year since the initial report, and I only see more unrelated tests instead of a fix. We already have more dataloader m:n bug reports, and some are blocked by the fix for this. The issue here has nothing to do with filters or the matching method, let's not try to solve things people didn't even run into, while we have limitations like this one for no apparent reason.

> I've added a test (Collection.load with where (Many to Many + Inverse side + custom where filters)) to reflect that.

This is not a "fix everything in dataloader" PR, it's yet another test that has nothing to do with the issue the PR is trying to close. Also it feels like you rather delegate fixing bugs in your code to me by only pushing more tests 😅

darkbasic:
> First, we need to fix bugs, then we can add more features

I agree (that's why I said "**Before** tackling the matching dataloader"), but while this fixes one bug it introduces a regression.

> let's not try to solve things people didn't even run into

They didn't run into them simply because they are not a problem with the existing implementation.

> and I only see more unrelated tests instead of a fix

Why do you call them unrelated? findChildrenFromPivotTable does not work with custom `where` conditions while my implementation does. You cannot expect to merge it and not have people complaining about that because it will be a regression.

> I am quite close to removing the tests you added here, that only show things not relevant to what I was trying to fix, and just merge it.

I am not sure what to say about that. You are of course free to do whatever you want since the project is yours, I've already told you why I have not been able to work on this in the past few months. While this PR fixes a problem it introduces another one, if you think that's ok I don't know what to say.

darkbasic:
> The failing test you added fails on master too (the one added months ago, not the recent one)

The old failing test (`Collection.load with where (Many to Many + Inverse side)`) does **not** fail on master: if you comment `findChildrenFromPivotTable` out it works:

image

> I already pointed out that you were testing stuff that was never supposed to be working in the first place, there can't be multiple instances of a collection owned by a single entity, and you cannot have different states of a single collection instance.

These corner cases never really materialized in a test. I've **tried** to test them and we **discussed** about that, but they always stayed commented out:

image

> Which one?

`Collection.load with where (Many to Many + Inverse side + custom where filters)`

With my most recent commit I've just splitted the original `Collection.load with where (Many to Many + Inverse side)` test into three tests of increasing complexity:

- `Collection.load with where (Many to Many + Inverse side)`: it works with both implementations
- `Collection.load with where (Many to Many + Inverse side + custom where filters)`: it works with my implementation but not yours
- `Collection.load with where (Many to Many + Inverse side + different custom where filters for each collection)`: it works with my implementation but not yours (it's basically a more complex variant of the previous test and you can ignore it for the moment).

Your implementation:
image

My implementation:
image

> Maybe I missed that, or maybe I forgot

I've had a big disaster recovery to deal with following a flood which messed up with my backlog for the past 6 months.

B4nan:
I am really sorry to hear that, I must have missed that :/ 

And I surely do not want to ignore tests that are passing on master either. I will try to have another look at those.

PR #6056, opened 2024-09-22
feat(collections): add first and last

rubiin:
Adds first and last methods on collection which is just a syntatic sugars


packages/core/src/entity/ArrayCollection.ts
```diff
@@ -337,6 +337,10 @@ export class ArrayCollection<T extends object, O extends object> {
     return true;
   }
 
+  first(): T | undefined {
+    return this.getItems()[0];
```
B4nan, line 341: this could be done more effectively, now you convert the internal items set to an array just to select its first value, you can do `this[0]` to read the first value instead (plus check the loaded state first)

packages/core/src/entity/ArrayCollection.ts
```diff
@@ -345,6 +349,10 @@ export class ArrayCollection<T extends object, O extends object> {
     return this.count() === 0;
   }
 
+  last(): T | undefined {
+    return this.getItems()[this.items.size - 1];
```
B4nan, line 353: same here, you can just check the loaded state and return `this[this.length - 1]`. in fact `this.length` is a getter that will check the loaded state, so here it should be enough to just do this
```suggestion
    return this[this.length - 1];
```

packages/core/src/entity/ArrayCollection.ts
rubiin:
```
  first(): T | undefined {
    return this.items.values().next().value;
  }
```
How about this one. This doesnt convert it into array

B4nan: yes, but why? your code feels more complex than my proposal (and you still need to check the loaded state).

packages/core/src/entity/ArrayCollection.ts
rubiin: i guess you meant
```ts

  first(): T | undefined {
    return this[0];
  }

```

packages/core/src/entity/ArrayCollection.ts
rubiin:
```ts

first(): T | undefined {
    return this[0];
  }

```
I guess you meant this

B4nan: yes, plus `this.checkInitialized();` before that

rubiin: sure thing. Will push the updates

packages/core/src/entity/ArrayCollection.ts
```diff
@@ -337,6 +337,13 @@ export class ArrayCollection<T extends object, O extends object> {
     return true;
   }
 
+  first(): T | undefined {
+    if (this.isInitialized()) {
+      return this[0];
+    }
+    return undefined;
```
B4nan, line 344: the method should throw if the collection is not initialized, same for the `last` one. i would move the `checkInitialized` to ArrayCollection (and mark it as protected) and do this instead
```suggestion
    this.checkInitialized();
    return this[0];
```

then the `Collection` implementations can be removed completely

B4nan:
fyi i will most likely remove the whole ArrayCollection since that abstraction makes very little sense and only makes things harder to write/test. but that's something for v7, now we need to deal with it

rubiin:
> fyi i will most likely remove the whole `ArrayCollection` (and inline it to the `Collection`) since that abstraction makes very little sense and only makes things harder to write/test. but that's something for v7, now we need to deal with it

Yeah its kind of confusing but then again laravels elequoent orm as well as doctrine seems to follow this pattern so mixed thoughts

packages/core/src/entity/ArrayCollection.ts
rubiin: `checkInitialized` doesnt exist on ArrayCollection class.  I have the checkInitialized on the collection one though
```ts
  override first(): T | undefined {
    this.checkInitialized();
    return super.first();
  }

```

B4nan:
do you plan to finish this?

rubiin:
> @rubiin do you plan to finish this?

Currently occupied with some work and couldnt push changes . Will update that sometime soon

B4nan:
Closing as stale

Issue #5996, opened 2024-09-04
Wrong type of PK in mysql when using tinyint as type


tomaszwylezek:
### Describe the bug

Driver: mysql
Version: Checked on 6.1.9, 6.3.6 and 6.3.7 and next

When we have two entities like

```
@Entity({ schema: 'foo', tableName: 'with_tiny' })
export class WithTiny {
  @PrimaryKey({ name: 'tiny_id', type: TinyIntType })
  id!: number

  @Property()
  name!: string
}

@Entity({ schema: 'foo', tableName: 'not_tiny' })
export class CasualEntity {
  @PrimaryKey({ name: 'not_tiny_id' })
  id!: number

  @ManyToOne(() => WithTiny, { fieldName: 'tiny_id' })
  tiny!: WithTiny
}
```

Then during initialisation of db it creates them like

```
  CREATE TABLE `with_tiny` (
    `tiny_id` int unsigned NOT NULL AUTO_INCREMENT, <---- NOTE: That should be tinyint
    `name` varchar(255) NOT NULL,
    PRIMARY KEY (`tiny_id`)
  ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci

  CREATE TABLE `not_tiny` (
    `not_tiny_id` int unsigned NOT NULL AUTO_INCREMENT,
    `tiny_id` tinyint unsigned NOT NULL,  <---- NOTE: There is correct one
    PRIMARY KEY (`not_tiny_id`),
    KEY `not_tiny_tiny_id_index` (`tiny_id`)
  ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci
  ```
  
  Where the `with_tiny`.`tiny_id` should be `tinyint` and it's `int` (but in `not_tiny` table the type is correct) what cause the constraint to fail
  
 ```
 DriverException: alter table `foo`.`not_tiny` add constraint `not_tiny_tiny_id_foreign` foreign key (`tiny_id`) references `foo`.`with_tiny` (`tiny_id`) on update cascade; - Referencing column 'tiny_id' and referenced column 'tiny_id' in foreign key constraint 'not_tiny_tiny_id_foreign' are incompatible.
 ```
 
 Tried to find if I'm missing anything but I failed to find any information about it :(

### Reproduction

https://github.com/tomaszwylezek/micro-orm-pk-tinyint
[...]

PR #6063, opened 2024-09-24
fix(schema): ensure `unsigned: false` works for primary keys

### LINKED ISSUE #6057: `unsigned: false` not being applied on `@PrimaryKey`
### Report related to the Schema Generator

In my case, there were several projects created before using MikroORM, and the primary keys in the mysql databases of these existing projects have the autoincrement bigint attribute.
I decided to switch to MikroORM and used the Entity Generator to import the schema from the existing database and generate entity code.

Then, it is generated as `@PrimaryKey({ unsigned: false })` like this.
```javascript
@Entity()
export class Member {
  [PrimaryKeyProp]?: 'memberId';

  @PrimaryKey({ unsigned: false })
  memberId!: bigint;
  
  ...
}
```

Upon reviewing the generated entities, they were created without any issues.
In other words, the entity generation worked perfectly fine.

And now I have used the Schema Generator to create the DDL.
(`orm.schema.getCreateSchemaSQL()` or `schema:create --dump` or `schema:update --dump`)

The generated DDL always returns unsigned, regardless of the unsigned property.
And `number`(int) type pk has same issue.

```sql
-- if memberId type is bigint and { unsigned: false }
create table `member` (`member_id` bigint unsigned not null auto_increment primary key) -- unexpected unsigned property

-- if memberId type is number and { unsigned: false }
create table `member` (`member_id` int unsigned not null auto_increment primary key) -- unexpected unsigned property
```

I checked the source code and found that it was designed to always output DDL strings with unsigned included, by distinguishing between `increments` and `bigincrements` types.
I modified it to pass the unsigned attribute to check whether it should be unsigned and generate the type accordingly, and it works well.
(In this case, I found that additional modifications are needed for case where the autoincrement attribute is missing in generated schema update DDL.)

I think that, except for special cases, there are people who would like to use the bigint type in databases.
This is probably due to the value types of other programming languages.
Therefore, I believe it would be better to make the unsigned option optional rather than mandatory.

If this is accepted as an issue, I would like to create a PR for it. :)

### Reproduction

https://github.com/lsh4711/reproduction

@mikro-orm/mysql

### MikroORM version

6.3.10

### Node.js version

20.17.0

### Operating system

Windows11

### Validations

lsh4711:
Support for the unsigned attribute on auto-incrementing primary keys seems to be a special case that is likely only supported by MySQL.
Previous source code seemed to be designed to generate DDL for MySQL without considering unsigned property, similar to how it handles schema generation for other databases.

In scenarios where the unsigned option needs to be passed, the code relied on the external knex package's types.
To address this, I modified the code to extend these types.
Initially, I attempted to override interface types within namespace using `declare module`, but encountered type conflicts.
As a solution, I chose to inherit from the existing TableBuilder interface.

Closes #6057

### DIFF
--- a/packages/knex/src/schema/DatabaseTable.ts
+++ b/packages/knex/src/schema/DatabaseTable.ts
@@ -127,11 +127,7 @@ export class DatabaseTable {
         extra: prop.extra,
         ignoreSchemaChanges: prop.ignoreSchemaChanges,
       };
-
-      if (this.columns[field].unsigned === undefined) {
-        this.columns[field].unsigned = this.columns[field].autoincrement;
-      }
-
+      this.columns[field].unsigned ??= this.columns[field].autoincrement;
       const defaultValue = this.platform.getSchemaHelper()!.normalizeDefaultValue(prop.defaultRaw!, prop.length);
       this.columns[field].default = defaultValue as string;
     });
--- a/packages/knex/src/schema/DatabaseTable.ts
+++ b/packages/knex/src/schema/DatabaseTable.ts
@@ -127,7 +127,11 @@ export class DatabaseTable {
         extra: prop.extra,
         ignoreSchemaChanges: prop.ignoreSchemaChanges,
       };
-      this.columns[field].unsigned ||= this.columns[field].autoincrement;
+
+      if (this.columns[field].unsigned === undefined) {
+        this.columns[field].unsigned = this.columns[field].autoincrement;
+      }
+
       const defaultValue = this.platform.getSchemaHelper()!.normalizeDefaultValue(prop.defaultRaw!, prop.length);
       this.columns[field].default = defaultValue as string;
     });
--- a/tests/issues/GH6057.test.ts
+++ b/tests/issues/GH6057.test.ts
@@ -1,0 +1,165 @@
+import type { EntitySchemaMetadata } from '@mikro-orm/core';
+import { MikroORM as MySqlORM, EntitySchema as MySqlSchema } from '@mikro-orm/mysql';
+import { MikroORM as SqliteORM, EntitySchema as SqliteSchema } from '@mikro-orm/sqlite';
+import { MikroORM as PostgreSqlORM, EntitySchema as PostgreSqlSchema } from '@mikro-orm/postgresql';
+
+type Key = 'bigint1' | 'bigint2' | 'bigint3' | 'number1' | 'number2' | 'number3';
+type Property = { id: number };
+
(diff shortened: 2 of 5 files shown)


tests/issues/GH6057.test.ts
```diff
@@ -0,0 +1,165 @@
...
+    name: 'entity',
+    properties: { id: { primary: true, type: 'number', unsigned: true } },
+  },
+  number3: {
+    name: 'entity',
+    properties: { id: { primary: true, type: 'number' } },
+  },
+};
+
+describe('MySql support unsigned increment pk.', () => {
```
B4nan, line 37: i would drop the sqlite/postgres tests, what are you trying to test there? that the feature that was never relevant to those will not be applied? we already have tests for that, lets simplify and instead, add test for mariadb which is relevant for this.

lsh4711: Sure, I'll make the changes as you suggested. :)
The purpose of the SQLite and PostgreSQL tests was to ensure that the generated DDL does not include the unsigned attribute for databases that do not support unsigned incrementing primary keys.

packages/knex/src/schema/DatabaseTable.ts
```diff
@@ -127,7 +127,11 @@ export class DatabaseTable {
         extra: prop.extra,
         ignoreSchemaChanges: prop.ignoreSchemaChanges,
       };
-      this.columns[field].unsigned ||= this.columns[field].autoincrement;
+
+      if (this.columns[field].unsigned === undefined) {
+        this.columns[field].unsigned = this.columns[field].autoincrement;
+      }
```
B4nan, line 133: if this change was really needed, just do `??=` instead of `||=`, no need for explicit if statement here

tests/issues/GH6057.test.ts
B4nan: there are lots of snapshot tests asserting the schema diffs for all drivers already, they would fail on their own

B4nan (approved):
thanks!

B4nan:
if you are up for another PR, there is #5996 which i think needs fixes in the very same places as you were touching in this pr

lsh4711:
Thank you very much as well!
Initially, my plan was to address the issue of 'missing autoincrement in alter' that I briefly mentioned in the issue after this PR is merged.
I will address the issue you pointed out first and then create a new issue.
I am truly delighted to contribute to this amazing project. :)

B4nan:
great, feel free to prioritize whatever is itching you :]

PR #6073, opened 2024-09-26
fix(schema): resolve missing auto_increment for PK in `MySQL/MariaDB` update DDL

### LINKED ISSUE #6072: Missing auto_increment in Update DDL for primary key changes in MySQL..
### Describe the bug

1. Modify primary key properties of existing entity.
2. Generate update DDL using the schema generator.
3. Generated alterDDL is missing auto_increment.

### Reproduction

https://github.com/lsh4711/reproduction

@mikro-orm/mysql

### MikroORM version

next(after PR#6065)

### Node.js version

20.17.0

### Operating system

Windows 11

### Validations

lsh4711:
Closes #6072

### DIFF
--- a/packages/knex/src/schema/SchemaHelper.ts
+++ b/packages/knex/src/schema/SchemaHelper.ts
@@ -174,7 +174,7 @@ export abstract class SchemaHelper {
   createTableColumn(table: ExpandedTableBuilder, column: Column, fromTable: DatabaseTable, changedProperties?: Set<string>, alter?: boolean): Knex.ColumnBuilder | undefined {
     const compositePK = fromTable.getPrimaryKey()?.composite;
 
-    if (column.autoincrement && !column.generated && !compositePK && (!changedProperties || changedProperties.has('autoincrement') || changedProperties.has('type'))) {
+    if (column.autoincrement && !column.generated && !compositePK && column.primary) {
       const primaryKey = !changedProperties && !this.hasNonDefaultPrimaryKeyName(fromTable);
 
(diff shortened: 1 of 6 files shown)


B4nan:
I can't wait for the time when we get rid of knex in there, the code is so convoluted because of their weird `increments/bigIncrements` abstraction. I hope this wont break anything...

lsh4711:
It seems that issues arise from knex's inflexible `increments` abstraction, making it difficult to fix externally...
I will study more to resolve other issues as well.
I hope to continue contributing to this great project!
Specifically, I want to make the schema generator more robust. :)

B4nan:
Ok, I would like to see knex gradually removed from pretty much everywhere, so if you see more issues like this, consider doing it right ahead instead of hacking your way with knex. If its a one-liner to fix with the current abstractions, let's do that, but if you need more control and you see it's gonna be more about fighting knex, it's often better to just get around it completely. 

I hope we could get rid of knex in v7 (and let users pick whatever QB they want to from outside).

lsh4711:
Got it.
I will look into Knex in more detail and work towards gradually replacing it if possible.
I also think about abstraction for external query builders.
Thank you! :)

B4nan:
> I also think about abstraction for external query builders.

Note that this is only for runtime, we don't need it internally for schema queries. No need to worry about that, now we just need to get rid of the tight coupling, the goal is to use knex only as a query runner first, which can be done in a non breaking way.

lsh4711:
Well noted with thanks.
I also hope for the day we are free from Knex!

PR #6084, opened 2024-09-29
fix(query-builder): ensure cached `execute('get')` returns a single result instead of an array

### LINKED ISSUE #6083: `createQueryBuilder` `cache()` with `execute("get")` wrong return type after first execute
### Describe the bug

`createQueryBuilder` `cache()` with `execute("get")` wrong return type after first execute:
```
// .execute("get") for all cases
userRank { rank: '3' } // First run return object with `rank` key
userRank [ { rank: '3' } ] // Cache hit, return array of objects with `rank` ket
userRank [ { rank: '3' } ] // Cache hit, return array of objects with `rank` ket
```

### Reproduction

```ts
const userRank: { rank: number } = await em.createQueryBuilder(User)
        .select([
            raw("count(*) + 1 AS rank"),
        ])
        .where({ level, profitPerHour: { $gt: user.profitPerHour } })
        .cache([`user_${user.id}_rank`, 600_000])
        .execute("get");
    console.log("userRank", userRank);
```

@mikro-orm/postgresql

### MikroORM version

6.3.10

### Node.js version

v22.9.0

### Operating system

_No response_

### Validations

lsh4711:
The result of the first cached execute('get') is guaranteed to be a single object,
so I modified it to ensure that the data stored in or retrieved from the cache is a single object.
When there is no result, it will be `null` rather than an array containing `null`.

Closes #6083

### DIFF
--- a/tests/issues/GH6083.test.ts
+++ b/tests/issues/GH6083.test.ts
@@ -1,0 +1,47 @@
+import { Entity, MikroORM, PrimaryKey } from '@mikro-orm/postgresql';
+
+@Entity()
+class User {
+
+  @PrimaryKey()
+  userId!: string;
+
+  constructor(userId: string) {
+    this.userId = userId;
+  }
+
+}
+
+let orm: MikroORM;
+
+beforeAll(async () => {
+  orm = await MikroORM.init({ entities: [User], dbName: '6083' });
+  await orm.schema.refreshDatabase();
+});
+
+afterAll(async () => {
+  await orm.schema.dropDatabase();
+  await orm.close(true);
+});
+
+test('check schema', async () => {
+  await orm.em.insertMany([
+    new User('6083-1'),
+    new User('6083-2'),
+    new User('6083-3'),
+  ]);
+
+  await orm.em.createQueryBuilder(User)
+    .select('userId')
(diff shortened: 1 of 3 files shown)


tests/issues/GH6083.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    .cache([`6083`, 6083])
+    .execute('get');
+
+  const result = await orm.em.createQueryBuilder(User)
+    .select('userId')
+    .where({ userId: '6083-1' })
+    .cache([`6083`, 6083])
+    .execute('get');
+
+  expect(result).toMatchSnapshot();
```
B4nan, line 46: lets inline this, no need for such a small snapshot

B4nan (changes requested):
thanks, just one nit

tests/issues/__snapshots__/GH6083.test.ts.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
B4nan, line 1: you will need to remove this file too to make the tests pass

PR #6085, opened 2024-09-29
fix(core): accept `readonly` arrays in operator where conditions

trim21:


B4nan (approved):
thanks!

packages/core/src/typings.ts
```diff
@@ -132,11 +132,11 @@ export type ExpandScalar<T> = null | (T extends string
 
 export type OperatorMap<T> = {
   $and?: ExpandQuery<T>[];
```
boenrobot, line 134: Is there a reason `$and` was left out?

B4nan: damn, i overlooked the first line. i guess it was omitted because it broke compilation on one line, fixed it here d33a707235f7fa50036cfec584253684b3175aa8

trim21: yes, what's the reason. And I don't want to touch other files

B4nan: next time at least mention it

B4nan:
what was the exact motivation for those changes @trim21? they are actually breaking on type level (since you no longer can dynamically compose the conditions without casting), i will most likely revert them.

trim21:
> what was the exact motivation for those changes @trim21? they are actually breaking on type level (since you no longer can dynamically compose the conditions without casting), i will most likely revert them.

This make it possible to use pre-defined readonly array list in `$in` operator

```ts
const some_kind_of_job_types = [0,1,2,3] as const;

repo.select().where({ $in: some_kind_of_job_types }).execute()
```

trim21:
and

image

trim21:
unless orm will mutate then internally, I think they are correct at type level as API

trim21:
> please provide actual code i can try myself. i know what readonly keyword is for, i am asking for actual repro because what you say works fine with `em.find` (and i doubt this could be an issue for QB, where the query is basically untyped to allow passing aliased keys).

```ts
import { Entity, PrimaryKey, MikroORM, Property } from '@mikro-orm/sqlite';

@Entity()
class Item {
  @PrimaryKey({ type: Number })
  id!: number;

  @Property({ type: Number })
  category!: number;

  @Property({ type: String })
  name!: string;
}

let orm: MikroORM = await MikroORM.init({
  dbName: ':memory:',
  entities: [Item],
});

const CATEGORY_US = Object.freeze([1, 2, 3]);
const CATEGORY_UK = [4, 5, 6] as const;
const CATEGORY_FR = [7, 8, 9] as const;

const r1 = await orm.em.getRepository(Item).find({
  category: { $in: CATEGORY_US },
});

await orm.close();
```

trim21:
do you stil consider revert this?

B4nan:
the problem this introduced is composing and/or conditions, which no longer works, since the arrays are now readonly

```ts
const where = {} as ObjectQuery<T>;
where.$and = where.$and ? [...where.$and] : [];
where.$and.push({ ... });
```

i am considering reverting just the `$and/$or` operators, where this makes the most sense

trim21:
maybe only add readonly before `ExpandScalar<T>[]`?

PR #6093, opened 2024-10-01
feat(core): prevent conversion of `many` entities in `oneToMany` from `extractConstructorParams`

### LINKED ISSUE #6092: Issue with bigint primary key and forceEntityConstructor
### Describe the bug

There appears to be an issue when using forceEntityConstructor on entities which have collections that use bigint for their primary key.

The issue happens when retrieving the parent object and loading it's collection.  It seems like each object in the collection is trying to be parsed as a bigint and this caues the error:

```js
 SyntaxError: Cannot convert [object Object] to a BigInt
        at BigInt (<anonymous>)

      at BigIntType.convertToJSValue (node_modules/@mikro-orm/core/types/BigIntType.js:32:24)
      at node_modules/@mikro-orm/core/entity/EntityFactory.js:308:62
          at Array.map (<anonymous>)
      at EntityFactory.extractConstructorParams (node_modules/@mikro-orm/core/entity/EntityFactory.js:278:39)
      at EntityFactory.createEntity (node_modules/@mikro-orm/core/entity/EntityFactory.js:193:33)
      at EntityFactory.create (node_modules/@mikro-orm/core/entity/EntityFactory.js:58:39)
      at SqlEntityManager.find (node_modules/@mikro-orm/core/EntityManager.js:152:45)
      at Object.<anonymous> (src/example.test.ts:75:17)
```

Outline of the reproduction 

```ts
@Entity()
export class User {
  // Unrealistic example, but it's just to show use case of forceEntityConstructor
  private role = "ADMIN";

  @PrimaryKey()
  id!: bigint;

  @Enum()
  name!: string;

  @OneToMany(() => UserComments, (x) => x.user)
  comments = new Collection<UserComments>(this);

  constructor(props: Pick<User, "name">) {
    this.name = props.name;
  }

  getRole() {
    return this.role;
  }
}

@Entity()
export class UserComments {
  @PrimaryKey()
  id!: bigint; // change to number to fix the issue

  @ManyToOne(() => User)
  user!: User;

  @Property()
  text!: string;
}

let orm: MikroORM;

beforeAll(async () => {
  orm = await MikroORM.init({
    dbName: ":memory:",
    entities: [User, UserComments],
    debug: ["query", "query-params"],
    forceEntityConstructor: true,
    allowGlobalContext: true,
  });
});

afterAll(async () => {
  await orm.close(true);
});

test("can load entity with bigint pk and access private variables", async () => {
  const user = new User({ name: "John Doe" });

  const comment = new UserComments();
  comment.text = "a comment";
  user.comments.add(comment);

  await orm.em.persistAndFlush(user);
  orm.em.clear();

  // error : SyntaxError: Cannot convert [object Object] to a BigInt
  const users = await orm.em.findAll(User, { populate: ["comments"] });
  expect(users[0].getRole()).toBe("ADMIN");
});

```

### Reproduction

https://github.com/cpt-pete/mikro-orm-bigint-bug/

@mikro-orm/sqlite

### MikroORM version

6.3.11

### Node.js version

20

### Operating system

macos

### Validations

lsh4711:
Return value of `extractConstructorParams` appears to be passed only as an argument to the Entity constructor.
In typical use cases, entities in many relationships should only be accessed through a Collection.
However, in special situations where `forceEntityConstructor: true`, users can handle it directly in the constructor.

Closes #6092

### DIFF
--- a/packages/core/src/entity/EntityFactory.ts
+++ b/packages/core/src/entity/EntityFactory.ts
@@ -407,12 +407,14 @@ export class EntityFactory {
         const tmp = { ...data };
 
         for (const prop of meta.props) {
-          if (prop.kind !== ReferenceKind.ONE_TO_MANY && options.convertCustomTypes && prop.customType && tmp[prop.name] != null) {
-            if ([ReferenceKind.MANY_TO_ONE, ReferenceKind.ONE_TO_ONE].includes(prop.kind) && Utils.isPlainObject(tmp[prop.name]) && !Utils.extractPK(tmp[prop.name], meta.properties[prop.name].targetMeta, true)) {
-              tmp[prop.name] = Reference.wrapReference(this.create(meta.properties[prop.name].type, tmp[prop.name]!, options), prop);
-            } else {
-              tmp[prop.name] = prop.customType.convertToJSValue(tmp[prop.name], this.platform) as any;
-            }
+          if (!options.convertCustomTypes || !prop.customType || tmp[prop.name] == null) {
+            continue;
+          }
+
+          if ([ReferenceKind.MANY_TO_ONE, ReferenceKind.ONE_TO_ONE].includes(prop.kind) && Utils.isPlainObject(tmp[prop.name]) && !Utils.extractPK(tmp[prop.name], meta.properties[prop.name].targetMeta, true)) {
+            tmp[prop.name] = Reference.wrapReference(this.create(meta.properties[prop.name].type, tmp[prop.name]!, options), prop);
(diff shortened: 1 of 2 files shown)


B4nan (approved):
i was just looking into this too, so replaced your fix (which would fail the same for M:N collections) with the proper one, which calls the convertor only on scalars or to-one properties.

thanks anyway :]

B4nan:
uh, why was this marked as a feature, i need to pay more attention to the PR titles :D

lsh4711:
Oh... It was a mistake.
I'm sorry, and I'll make sure to be more careful in the future.

PR #6094, opened 2024-10-01
fix(core): fix same-second dates diffing

alexandre-abrioux:
This small PR fixes a minor regression that I think was introduced in https://github.com/mikro-orm/mikro-orm/commit/55df57ff1aa84e5d45188d849ba09e91ae6d3642.

## Issue
When updating an entity and changing a date property to another one very close to it (by a difference in microseconds), the change is not persisted in the database. Indeed, diffing dates by comparing the `toString()` output doesn't give enough precision to compare microseconds shifts. Example:
- `toString()` output: `Tue Oct 01 2024 08:54:48 GMT+0000 (Coordinated Universal Time)`
- `toISOString()` output: `2024-10-01T08:54:48.651Z`

Two dates can have the same `toString()` output even though their `toISOString()` output is different.

The impact is low and probably only concerns the test scenarios: It's rare to persist the same entity twice in the same second without changing anything other than a date.

## Fix
In this PR I'm switching the dates comparison to use `toISOString()` and adding a new unit test.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -61,6 +61,8 @@ services:
       MSSQL_SA_PASSWORD: Root.Root
       MSSQL_TELEMETRY_ENABLED: 0
       ACCEPT_EULA: 1
+    healthcheck:
+      test: /opt/mssql-tools/bin/sqlcmd -S mssql -U sa -P 'Root.Root' -Q 'select 1'
 
 volumes:
   mysql8:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -157,6 +157,7 @@ jobs:
       - name: Init docker
         run: |
           docker compose up -d
+          sleep 5
 
       - name: Set CC Required env vars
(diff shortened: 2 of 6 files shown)


docker-compose.yml
```diff
@@ -61,6 +61,8 @@ services:
       MSSQL_SA_PASSWORD: Root.Root
       MSSQL_TELEMETRY_ENABLED: 0
       ACCEPT_EULA: 1
+    healthcheck:
+      test: /opt/mssql-tools/bin/sqlcmd -S mssql -U sa -P 'Root.Root' -Q 'select 1'
```
B4nan, line 65: @boenrobot looks like this solves the flaky CI runs

B4nan: Hmm or maybe not :D but it feels better, lets see

boenrobot: Hm, spoke too soon, judging by the latest commit. But I'm sure this does help _reduce_ the flaky-ness.

PR #6096, opened 2024-10-01
ci: try to fix flaky mssql tests

B4nan:


B4nan:
I was trying out the setup from knex/kysely for the CI yesterday and that seems to work. They don't use the edge image (which is no longer maintained and will lose the ARM support, meh), maybe the new tedious version won't work properly with that. Weird that it works just fine locally, never had this issue there.

B4nan:
lol, i also tried to add docker cache, and it only made things slower (30s -> 60s).

PR #6098, opened 2024-10-02
refactor(schema): move increment related changes to `MySqlSchemaHelper`

### LINKED ISSUE #6095: regression, NotNullConstraintViolationException using better-sqlite since 6.3.11
### Describe the bug

I have simple entity like

```ts
@Entity({ tableName: 'some_table' })
class SomeEntity {
  @PrimaryKey({ autoincrement: true, type: 'bigserial' })
  id!: string;
  
  @Property()
  prop!: string;
}
```

And on `flush` error happens:

_NotNullConstraintViolationException: insert into `some_table` (`prop`) values ('test') returning `id`, `prop` - NOT NULL constraint failed: some_table.id_

On 6.3.10 was all fine

### Reproduction

```ts
import { BetterSqliteDriver } from '@mikro-orm/better-sqlite';
import { Entity, MikroORM, PrimaryKey, Property } from '@mikro-orm/core';

@Entity({ tableName: 'some_table' })
class SomeEntity {
  constructor(prop: string) {
    this.prop = prop;
  }

  @PrimaryKey({ autoincrement: true, type: 'bigserial' })
  id!: string;

  @Property()
  prop!: string;
}

(async () => {
  const db = await MikroORM.init({
    dbName: ':memory:',
    driver: BetterSqliteDriver,
    entities: [SomeEntity],
  });

  await db.getSchemaGenerator().refreshDatabase();
  await db.getSchemaGenerator().clearDatabase();

  // on 6.3.11, 6.3.12 error happens; 6.3.10 and below is fine
  await db.em.fork().persistAndFlush(new SomeEntity('test'));

  console.log('done');

  process.exit(0);
})();

```

@mikro-orm/better-sqlite

### MikroORM version

6.3.11, 6.3.12

### Node.js version

22

### Operating system

_No response_

### Validations

lsh4711:
Closes #6095

### DIFF
--- a/packages/knex/src/dialects/mysql/MySqlColumnCompiler.ts
+++ b/packages/knex/src/dialects/mysql/MySqlColumnCompiler.ts
@@ -1,6 +1,6 @@
 // @ts-ignore
 import BaseMySqlColumnCompiler from 'knex/lib/dialects/mysql/schema/mysql-columncompiler';
-import type { MySqlIncrementOptions } from '../..//typings';
+import type { MySqlIncrementOptions } from '../../typings';
 
 export class MySqlColumnCompiler extends BaseMySqlColumnCompiler {
 
--- a/packages/knex/src/dialects/mysql/MySqlColumnCompiler.ts
+++ b/packages/knex/src/dialects/mysql/MySqlColumnCompiler.ts
@@ -1,20 +1,20 @@
 // @ts-ignore
 import BaseMySqlColumnCompiler from 'knex/lib/dialects/mysql/schema/mysql-columncompiler';
-import type { IncrementOptions } from '../../typings';
+import type { MySqlIncrementOptions } from '../..//typings';
 
 export class MySqlColumnCompiler extends BaseMySqlColumnCompiler {
 
   // we need the old behaviour to be able to add auto_increment to a column that is already PK
-  increments(options: IncrementOptions) {
+  increments(options: MySqlIncrementOptions) {
     return this.generateDDL(options);
   }
 
   /* istanbul ignore next */
-  bigincrements(options: IncrementOptions) {
+  bigincrements(options: MySqlIncrementOptions) {
     return this.generateDDL(options);
   }
 
-  private generateDDL(this: any, options: IncrementOptions = {}) {
+  private generateDDL(this: any, options: MySqlIncrementOptions = {}) {
     const { primaryKey = true, unsigned = true, type = 'int' } = options;
     return type
       + (unsigned ? ' unsigned' : '')
--- a/tests/features/embeddables/GH5847.test.ts
+++ b/tests/features/embeddables/GH5847.test.ts
@@ -18,7 +18,7 @@ class EmbeddableEntity {
 class EntityA {
 
   @PrimaryKey({ autoincrement: true })
-  readonly id!: bigint;
+  readonly id!: string;
 
   @Embedded(() => EmbeddableEntity)
   reward = new EmbeddableEntity();
@@ -41,5 +41,5 @@ afterAll(async () => {
 
 test('correctly build migration for enum type', async () => {
   const up = await orm.schema.getCreateSchemaSQL({ wrap: false });
-  expect(up.trim()).toEqual('create table `entity_a` (`id` integer not null primary key autoincrement, `reward_active` integer not null default false, `reward_email` json not null, `reward_subject` text not null default \'Here is your reward\');');
+  expect(up.trim()).toEqual('create table `entity_a` (`id` integer not null primary key autoincrement, `reward_active` integer not null default false, `reward_email` json not null, `reward_subject` text not null default \'Here is your reward\', unique (`id`));');
 });
--- a/tests/features/embeddables/GH5847.test.ts
+++ b/tests/features/embeddables/GH5847.test.ts
@@ -17,7 +17,7 @@ class EmbeddableEntity {
 @Entity()
 class EntityA {
 
-  @PrimaryKey({ autoincrement: true })
+  @PrimaryKey({ autoincrement: true, type: 'integer' })
   readonly id!: string;
 
   @Embedded(() => EmbeddableEntity)
@@ -41,5 +41,5 @@ afterAll(async () => {
 
 test('correctly build migration for enum type', async () => {
   const up = await orm.schema.getCreateSchemaSQL({ wrap: false });
-  expect(up.trim()).toEqual('create table `entity_a` (`id` integer not null primary key autoincrement, `reward_active` integer not null default false, `reward_email` json not null, `reward_subject` text not null default \'Here is your reward\', unique (`id`));');
+  expect(up.trim()).toEqual('create table `entity_a` (`id` integer not null primary key autoincrement, `reward_active` integer not null default false, `reward_email` json not null, `reward_subject` text not null default \'Here is your reward\');');
 });
--- a/tests/features/embeddables/GH5847.test.ts
+++ b/tests/features/embeddables/GH5847.test.ts
@@ -18,7 +18,7 @@ class EmbeddableEntity {
 class EntityA {
 
   @PrimaryKey()
-  readonly id!: bigint;
+  readonly id!: number;
 
   @Embedded(() => EmbeddableEntity)
   reward = new EmbeddableEntity();
--- a/tests/features/embeddables/GH5847.test.ts
+++ b/tests/features/embeddables/GH5847.test.ts
@@ -17,8 +17,8 @@ class EmbeddableEntity {
 @Entity()
 class EntityA {
 
-  @PrimaryKey({ autoincrement: true, type: 'integer' })
-  readonly id!: string;
+  @PrimaryKey()
+  readonly id!: bigint;
 
   @Embedded(() => EmbeddableEntity)
   reward = new EmbeddableEntity();
--- a/tests/issues/GH6095.test.ts
+++ b/tests/issues/GH6095.test.ts
@@ -1,7 +1,7 @@
 import { Entity, MikroORM, PrimaryKey } from '@mikro-orm/better-sqlite';
 
 @Entity()
-class Entity1 {
+class User {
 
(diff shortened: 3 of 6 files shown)


packages/knex/src/dialects/mysql/MySqlColumnCompiler.ts
```diff
@@ -1,20 +1,20 @@
 // @ts-ignore
 import BaseMySqlColumnCompiler from 'knex/lib/dialects/mysql/schema/mysql-columncompiler';
-import type { IncrementOptions } from '../../typings';
+import type { MySqlIncrementOptions } from '../..//typings';
```
B4nan, line 3:
```suggestion
import type { MySqlIncrementOptions } from '../../typings';
```

tests/features/embeddables/GH5847.test.ts
```diff
@@ -18,7 +18,7 @@ class EmbeddableEntity {
 class EntityA {
 
   @PrimaryKey({ autoincrement: true })
-  readonly id!: bigint;
+  readonly id!: string;
```
B4nan, line 21: why this change?

B4nan (changes requested):
lets add the test case from #6095 here

tests/features/embeddables/GH5847.test.ts
lsh4711: This part was the code before the changes, and the tests were passing.
I was curious why no error occurred when `autoIncrement` was `true` for a string type without explicitly specifying the `type`, but the existing `SchemaHelper` works with both `string` and `bigint` types, and the tests pass as well.
Based on your suggestion, I can change it to `number` or` bigint`, as this is indeed the correct way to use it. :)

B4nan: i was just curious, all of those variants should work.

tests/issues/GH6095.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    entities: [Default, Bigserial, Bigint, Custom],
+  });
+  await orm.schema.refreshDatabase();
+});
+
+afterAll(async () => {
+  await orm.close(true);
+});
+
+test('6095', async () => {
```
B4nan, line 49: lets also test this is not producing extra schema diffs. just check if `orm.schema.getUpdateSchemaSQL()` returns empty string.

tests/issues/GH6095.test.ts
```diff
@@ -0,0 +1,69 @@
...
+  const sql = await orm.schema.getUpdateSchemaSQL();
+  expect(sql).toMatchSnapshot();
```
B4nan, line 51: again, lets not use snapshots for this, checking empty strings with snapshots make no sense

tests/issues/__snapshots__/GH6095.test.ts.snap
```diff
@@ -0,0 +1,25 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
B4nan, line 1: seeing this, it sounds like my hunch was right :D

lsh4711: In SQLite, getUpdateSchemaSQL() always generates SQL in a way that replaces the existing table,
so I used a snapshot to verify if the format is correct in createTable.
I'm not sure if this is the appropriate method 😢

B4nan: there should be no diff if the schema is up to date, we have many tests checking that. so if you always see something generated, it means the changes are not correct.

B4nan: maybe its again caused by the driver mismatch, since the test is using pg specific types which are only recognized by the pg platform, but it uses sqlite platform instead

lsh4711: There seems to be a misunderstanding.. 😢 
I checked your suggestion before using createSchemaSQL() and tried using createUpdateSQL(),
but as you can see from https://github.com/mikro-orm/mikro-orm/commit/56f11810076d3f5ea8d319075090140f220940be, a DDL for creating a new schema was generated.

And, I just checked again, and even in version 6.3.10, entities other than the number type or { type: 'Integer' } are generating DDLs for creating new tables instead of an empty diff.

Could this be a bug?

B4nan: what should be happening behind the scenes:
- you use sqlite platform, so the column types should be resolved based on that
- diffing works based on the column types
- sqlite uses TEXT for pretty much everything which is not a number, so this should result in comparing TEXT with TEXT or NUMBER with NUMBER
- no diffs should be produced

i can have a closer look to see whats happening, i would only consider this as a bug if it happens with the platform that knows those types, using sqlite for other types is an option for tests, but its not relevant that we produce extra diffs there. its important not to produce those for your actual database/when using from the CLI, e.g. for migrations

lsh4711: With the PostgreSQL driver, an empty diff is generated, meaning it works correctly, and an empty diff, as expected, also appears with MySQL.
It seems that only with SQLite, when trying to use types other than integer, a new schema is being created.

lsh4711: Exactly, in the above test, PostgreSQL works correctly with both default entities and custom entities using varchar type and custom types, respectively. For the remaining entities, an empty diff is generated. It seems that PostgreSQL is functioning properly.

lsh4711: I've reconsidered, and I think PostgreSQL should also have empty diffs for all entities.
However, it keeps generating alter statements by comparing anew each time.
I'll take a closer look at this.

B4nan: indeed, let me say it for the third time, there should be no queries if the schema is up to date.

B4nan: i would first check the logs from schema comparator, they should give you a good understanding why a diff is created

https://github.com/mikro-orm/mikro-orm/blob/master/packages/knex/src/schema/SchemaComparator.ts#L698

the logs are disabled by default, you need to enable `verbose` option to see them, or just replace this method with `console.log` in your fork to see

B4nan: i checked the test once again, and not sure what you are trying to show with the `type: 'custom'`? its weird that it resolves to integer to be honest, but its rather undefined behaviour, people should use `columnType` here if they want that as a type in their schema. i would personally expect that to resolve to string/text, since that is the default type mapping iirc

B4nan: The default mapping is string, not number, so fallback to integer is not normal, definitely not what I would expect here. I can feel its caused by knex and its `increments` helper. Which we should really get rid of in the first place, but we don't have to deal with that in this PR, it might require quite a lot of changes in all the drivers.

lsh4711: I identified the cause and understood how the code operates.
I discovered that using `autoIncrement: true` with `text` type or `custom-naming` type can lead to unintended diffs.
```
@Entity()
class Bigserial {

  @PrimaryKey({ autoincrement: true, type: 'bigserial' })
  id!: string;

}
```
If autoIncrement is explicitly set to true like in the example above, create DDL is always generated as `Integer auto_increment.`.
This leads to a diff between the table type `Integer` and the entity type `bigserial` during updates or migrations.
The issue is that the altered schema also becomes `Integer auto_increment`, causing diffs to occur continuously.

Additionally, I found that `autoIncrement: true` can be problematic even when the table already exists.
Typically, SQLite is used for testing purposes on in-memory, but if we are using a file-based mode and the existing schema uses (id bigserial not null, primary key (id)), setting `autoIncrement: true` in the entity code will cause an autoIncrement diff.
Up to this point, it might not seem problematic.
However, it always changes to `Integer auto_increment`, regardless of `bigserial`, leading to diffs continuously...

If auto increment is necessary, we must use it carefully as follows:
Use only `number` type or specify `{ type: 'int|integer|number' }`.

In issue #6095, should remove `autoIncrement: true` part.
If autoIncrement is needed, ensure that type is explicitly set to `int, integer, or number` to prevent unintended diffs by comparing correct values. :)

tests/features/embeddables/GH5847.test.ts
```diff
@@ -17,8 +17,8 @@ class EmbeddableEntity {
 @Entity()
 class EntityA {
 
-  @PrimaryKey({ autoincrement: true })
-  readonly id!: bigint;
+  @PrimaryKey({ autoincrement: true, type: 'integer' })
+  readonly id!: string;
```
B4nan, line 21: i would remove this change completely. what you did there now is wrong, integers are always mapped to numbers. only bigints support mapping to number/string/bigint, not integers.

lsh4711: I agree with what you said.
I will only keep the part about bigserial.
However, it is true that using autoincrement as true in the use case of the related issue causes continuous diff errors,
so I am curious about how this part should be improved in the future.

B4nan:
```suggestion
  @PrimaryKey()
  readonly id!: bigint;
```

tests/features/embeddables/GH5847.test.ts
```diff
@@ -17,7 +17,7 @@ class EmbeddableEntity {
 @Entity()
 class EntityA {
 
-  @PrimaryKey({ autoincrement: true })
+  @PrimaryKey()
   readonly id!: bigint;
```
B4nan, line 21:
```suggestion
  @PrimaryKey()
  readonly id!: number;
```

lsh4711:
I apologize.
It was a review for test 5847, but I only made changes to the recent test.
And now, both `schema first` approach and `entity first` approach seem to be working well in MySQL. :)
Thank you for your help.

PR #6102, opened 2024-10-02
feat(migrations): allow passing list of classes to `migrationsList`

### LINKED ISSUE #6099: `migrationsList` as plain array of migrations
Currently when I create migrations I end up adding it to the list as usualy but its always a bit useless code since it will look like this:
```ts
migrations: [
  // [...],
  {
    class: MigrateXyz127,
    name: MigrateXyz127.name
  }
]
```
A simpler list where I could only provide the class would really great.

So for example the new list would look like this:
```ts
migrations: [
  // [...],
  MigrateXyz127
]
```
In this case i need 3 less lines of code and my migration is simplified.  
In the beackground it could probably automatically converted into the old format.

Another solutions could also be a simple mapping function like this:
```ts
function mapClassesToList(migrations: Array<Constructor<Migration>>): MigrationObject[] {
  return migrations.map((migration): MigrationObject => ({
    class: migration,
    name: migration.name
  }));
}
```
That would map a migration class array to the needed list format with just one small function the package can provide

**Additional context**
This would be just like the migrations of TypeORM as well. Wich probably makes the understanding of the differences for people who wants to switch easier.

DerHerrGammler:
closes #6099

### DIFF
--- a/docs/docs/migrations.md
+++ b/docs/docs/migrations.md
@@ -257,8 +257,7 @@ await orm.em.transactional(async em => {
 
 ## Importing migrations statically
 
-If you do not want to dynamically import a folder (e.g. when bundling your code with webpack) you can import migrations directly.
-You can do that with an explicit migration name or the implicit filename as migration name.
+If you do not want to dynamically import a folder (e.g. when bundling your code with webpack) you can import migrations directly. You can do that with an explicit migration name or the implicit filename as migration name.
 
 ```ts
 import { MikroORM } from '@mikro-orm/core';
--- a/docs/docs/migrations.md
+++ b/docs/docs/migrations.md
@@ -258,20 +258,25 @@ await orm.em.transactional(async em => {
 ## Importing migrations statically
 
 If you do not want to dynamically import a folder (e.g. when bundling your code with webpack) you can import migrations directly.
+You can do that with an explicit migration name or the implicit filename as migration name.
 
 ```ts
 import { MikroORM } from '@mikro-orm/core';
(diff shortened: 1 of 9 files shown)


DerHerrGammler:
I am currently not sure what your convention on testing is.
I would like to add a test for the new syntax but where should I add it and what is the naming conmvention on it?

B4nan:
some tests for this are here

https://github.com/mikro-orm/mikro-orm/blob/master/tests/features/migrations/Migrator.test.ts#L483

you could duplicate that file and use the new approach. same for mongo migrator, either move the changes a level above so they are covered, or add a test specific for mongo.

B4nan (changes requested):
thanks, last missing piece is updating the docs

DerHerrGammler:
I added the docs directly to the `Importing migrations statically` section with a comment. If thats ok.

docs/docs/migrations.md
```diff
@@ -258,20 +258,25 @@ await orm.em.transactional(async em => {
 ## Importing migrations statically
 
 If you do not want to dynamically import a folder (e.g. when bundling your code with webpack) you can import migrations directly.
+You can do that with an explicit migration name or the implicit filename as migration name.
```
B4nan, line 261:
```suggestion
If you do not want to dynamically import a folder (e.g. when bundling your code with webpack) you can import migrations directly. You can do that with an explicit migration name or the implicit filename as migration name.
```

B4nan (approved):
yeah that will do it, thanks!

PR #6107, opened 2024-10-03
fix(core): ignore non-persisted properties on embeddables

### LINKED ISSUE #6105: Property with `persist: false` in embeddables tries to insert into non-existing column
### Describe the bug

In embeddables, a property with `persist: false` does not generate the column in DB, **but** is added to the _create_ statement (as if the column existed).

It happens with _sqlite_ and _postgres_, so I suppose that it is not driver-specific.

> I have not tested, but I assume that the problem also exists with update statements.

**Do I intent to submit a PR?**  
I've already forked the repo, so I'll give it a try.

### Reproduction

The code fails with the following error: 

```text
insert into `person` (`name_first`, `name_last`, `name_display`) values ('John', 'Doe', 'John Doe') returning `id` - table person has no column named name_display
```

```ts
@Embeddable()
class Name {
  @Property()
  first!: string;
  @Property()
  last!: string;

  @Property({ persist: false, getter: true })
  public get display(): Opt<string> {
    return `${this.first} ${this.last}`;
  }
}
@Entity()
class Person extends BaseEntity {
  @PrimaryKey()
  id!: number;
  @Embedded(() => Name)
  name!: Name;
}

describe('GH issue', () => {
  let orm: MikroORM;

  beforeAll(async () => {
    orm = await MikroORM.init({
      entities: [Person],
      dbName: ':memory:',
      strict: true,
      validate: true,
      validateRequired: true,
    });

  });

  afterAll(() => orm.close(true));
  afterEach(() => orm.em.clear());

  it('should create entity with non-persited data in embeddable', async () => {
    const repo = orm.em.getRepository(Person);

    // Fail at insertion (try to insert into non-existing column `name_display`)
    repo.create({ name: { first: 'John', last: 'Doe' } });
    await orm.em.flush();
    orm.em.clear();

    const [person] = await repo.findAll();
    expect(person.name.display).toBe('John Doe');
  });
});
```

The full test can be found [here](https://github.com/heap-code/mikro-orm/blob/9ee4a284eee27062b16c708a819cf2470de4c9d0/tests/features/embeddables/non-persisted-in-embeddables.test.ts).

> **Note about the test**:  
> At first I thought that the error only occurred with polymorphic embeddables with `ManyToOne` relations, but I later figured out that it also happens with a "simple" embeddable.
> That's why there's a second more "complex" test case.

None

### MikroORM version

6.3.12

### Node.js version

v20.17.0

### Operating system

Ubuntu

### Validations

HugoMendes98:
fixes #6105

### DIFF
--- a/packages/core/src/utils/EntityComparator.ts
+++ b/packages/core/src/utils/EntityComparator.ts
@@ -487,7 +487,11 @@ export class EntityComparator {
       return true;
     }
 
-    ret += meta.props.filter(p => p.embedded?.[0] === prop.name).map(childProp => {
+    ret += meta.props.filter(p =>
+      p.embedded?.[0] === prop.name
+      // object for JSON embeddable
+      && (p.object || (p.persist !== false)),
+    ).map(childProp => {
       const childDataKey = meta.embeddable || prop.object ? dataKey + this.wrap(childProp.embedded![1]) : this.wrap(childProp.name);
       const childEntityKey = [...path, childProp.embedded![1]].map(k => this.wrap(k)).join('');
       const childCond = `typeof entity${childEntityKey} !== 'undefined'`;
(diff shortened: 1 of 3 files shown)


B4nan (approved):
thanks

PR #6120, opened 2024-10-07
fix(core/utils): compare invalid dates

alexandre-abrioux:
This PR fixes an issue I introduced in https://github.com/mikro-orm/mikro-orm/pull/6094

Using `toISOString()` on an invalid date throws a [RangeError exception](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString#exceptions). We should test if the date is valid before comparing it with `toISOString()`.

### DIFF
--- a/tests/Utils.test.ts
+++ b/tests/Utils.test.ts
@@ -90,8 +90,8 @@ describe('Utils', () => {
     expect(compareObjects(sql`select ${1}`, sql`select ${2}`)).toBe(false);
     expect(compareObjects(new Date('2024-10-01T08:54:48.651Z'), new Date('2024-10-01T08:54:48.651Z'))).toBe(true);
     expect(compareObjects(new Date('2024-10-01T08:54:48.651Z'), new Date('2024-10-01T08:54:48.676Z'))).toBe(false);
-    expect(compareObjects(new Date('2024-10-01T08:54:48.651Z'), new Date('invalid'))).toBe(false);
-    expect(compareObjects(new Date('invalid one'), new Date('invalid two'))).toBe(true);
+    expect(compareObjects(new Date('2024-10-01T08:54:48.651Z'), new Date(NaN))).toBe(false);
+    expect(compareObjects(new Date(NaN), new Date(NaN))).toBe(true);
     expect(Utils.equals(NaN, NaN)).toBe(true);
   });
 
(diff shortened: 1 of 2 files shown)


tests/Utils.test.ts
```diff
@@ -88,7 +88,10 @@ describe('Utils', () => {
     expect(compareObjects(sql`select 1`, sql`select 1`)).toBe(true);
     expect(compareObjects(sql`select ${1}`, sql`select ${1}`)).toBe(true);
     expect(compareObjects(sql`select ${1}`, sql`select ${2}`)).toBe(false);
+    expect(compareObjects(new Date('2024-10-01T08:54:48.651Z'), new Date('2024-10-01T08:54:48.651Z'))).toBe(true);
     expect(compareObjects(new Date('2024-10-01T08:54:48.651Z'), new Date('2024-10-01T08:54:48.676Z'))).toBe(false);
+    expect(compareObjects(new Date('2024-10-01T08:54:48.651Z'), new Date(NaN))).toBe(false);
+    expect(compareObjects(new Date(NaN), new Date(NaN))).toBe(true);
```
B4nan, line 94: i would focus more on how this happened instead of supporting it on this low level helper. this feels like you are fixing it on wrong level.

alexandre-abrioux:
@B4nan You were right! We should not fix this the way I did, even not fix it at all.

I was initially trying to replicate the behavior of `v6.3.10`, but I think the new (current) behavior is safer. I've added an integration test to document the use-case.

In `v6.3.10`, when updating a JSON property in MongoDB from a valid date to an invalid date (it is very easy to make such a mistake by instantiating a date like so: `new Date(undefined)`), then the date was simply erased and replaced by the Unix time "1970-01-01T00:00:00.000Z". This is IMO dangerous and unwanted.

After more thought, I think it is best to throw an error and explain that this is behavior is unsupported. There are no "invalid dates" in the BSON format, so I don't think we should handle that at all.

In the last commit I pushed on this PR, I changed the function to throw an exception when we encounter such a case. This does not "fix" anything but makes it explicit that there is an issue with the user code. Investigating this made me find an issue in the codebase I'm working on, so this is a good thing 🙏

alexandre-abrioux:
> maybe we could revert those recent changes, since they were added because of how `structuredClone` works in combination with jest, and we no longer use `structuredClone` internally?

Yes probably! Let's settle on the desired behavior for updates with invalid dates, then with all the tests we've added recently, it should be safe to try 👌

B4nan:
I don't mind explicit exceptions as you did. Perf wise, I would rather store the `getTime()` values and compare those instead of using `toISOString` and comparing strings.

alexandre-abrioux:
Good point! done

PR #6121, opened 2024-10-07
chore: update virtual entities example query

AdriVanHoudt:
This better reflects the example.

I also changed it to use a string template to improve readability


B4nan (approved):

B4nan:
if you want to see the changes in already released versions, you'd have to update those snapshots too, but updating only the next docs (what you did) is enough (and the more important part)

PR #6123, opened 2024-10-08
refactor(cli): made the CLI definition type safe

boenrobot:


packages/cli/src/commands/GenerateEntitiesCommand.ts
```diff
@@ -32,18 +33,18 @@ export class GenerateEntitiesCommand<U extends Options = Options> implements Com
...
    * @inheritDoc
    */
-  async handler(args: ArgumentsCamelCase<U>): Promise<void> {
+  async handler(args: ArgumentsCamelCase<GenerateEntitiesArgs>): Promise<void> {
     if (!args.save && !args.dump) {
       return CLIHelper.showHelp();
     }
 
-    const orm = await CLIHelper.getORM(false);
+    const orm = await CLIHelper.getORM();
```
B4nan, line 47: why the `warnWhenNoEntities` change? is it handled in another way? explicit `false` is needed, with `undefined` you will end up with the `true` default from `CLIHelper.getConfiguration`

B4nan (commented):
thanks!

PR #6128, opened 2024-10-11
feat(mssql): add support for smalldatetime type

### LINKED ISSUE #5904: `unknown` type assigned for `smalldatetime` column
### Describe the bug

When generating entity models from a SQL Server, an `unknown` type is assigned for columns that have a `smalldatetime` type.
On top of that, if these columns have a default value of `current_timestamp`, it is set as a `default` string instead of `defaultRaw` (probably because it is not considered a date column).

Instead of image, image is generated.

I currently get around this by adding the following code in `onProcessedMetadata`:
```typescript
metadata.forEach((entityMetadata) => {
  entityMetadata.props.forEach((property) => {
    if (property.type === 'unknown') {
      if (
        property.columnTypes.some((columnType) =>
          columnType.startsWith('smalldatetime'),
        )
      ) {
         property.type = 'Date';
      }
      if (property.default === 'current_timestamp') {
        property.defaultRaw = `current_timestamp`;
        property.default = undefined;
      }
    }
  });
});
```

While fiddling around to better describe the problem, I noticed that the problem lies in the type mapping in `MsSqlPlatform` and adding `smalldatetime: 'datetime'` to `map` in `getDefaultMappedType` actually solves it, so I can quickly create a PR fixing this bug.

### Reproduction

Using SQL Server, add a new table with a smalldatetime column:
```sql
CREATE TABLE SmalldatetimeTest (
  SomeSmallDateTime SMALLDATETIME DEFAULT CURRENT_TIMESTAMP
)
```

Use MikroORM with mssql driver to generate entity model for the created table.

As a result, a file is generated with the contents shown below:
```typescript
import { Entity, Property } from '@mikro-orm/core';

@Entity({ tableName: 'SmalldatetimeTest' })
export class SmalldatetimeTest {
  @Property({
    fieldName: 'SomeSmallDateTime',
    type: 'unknown',
    columnType: 'smalldatetime(0)',
    nullable: true,
  })
  SomeSmallDateTime?: unknown & Opt = current_timestamp;
}
```

@mikro-orm/mssql

### MikroORM version

6.3.3

### Node.js version

v20.15.0

### Operating system

macOS

### Validations

patrykbojczuk:
This adds `SmallDateTimeType` extending `DateTimeType` and `smalldatetime` to `SmallDateTimeType` mapping to @mikro-orm/mssql package, as suggested by @boenrobot in [#5904](https://github.com/mikro-orm/mikro-orm/issues/5904#issuecomment-2270490466).

Closes #5904


tests/features/entity-generator/__snapshots__/SmallDateTimeType.mssql.test.ts.snap
```diff
@@ -0,0 +1,19 @@
...
+  "import { Entity, type IType, PrimaryKey, Property } from '@mikro-orm/core';
+
+@Entity()
+export class Test {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property({ type: 'Date', columnType: 'smalldatetime', nullable: true, defaultRaw: \`current_timestamp\` })
+  at?: IType<Date, string>;
```
B4nan, line 14: @boenrobot any idea why is this `IType` and not just `Date`?

packages/mssql/src/SmallDateTimeType.ts
```diff
@@ -0,0 +1,13 @@
+import { DateTimeType } from '@mikro-orm/core';
+
+export class SmallDateTimeType extends DateTimeType {
+
+  override getColumnType(): string {
+    return 'smalldatetime';
+  }
+
+  override getDefaultLength(): number {
+    return 0;
```
B4nan, line 10: not sure if this is needed, if yes, you will need to add some tests that use it

boenrobot: This line would be covered if "type" was correctly determined to be "smalldatetime"... The length in the metadata is compared against this value in the type class. MSSQL shows any smalldatetime column with 0 datetime precision (because it doesn't have option for sub second precision), and since this method also returns 0, no length option will ever be outputted. Without it, "length: 0" would be added to the output options.

tests/features/entity-generator/__snapshots__/SmallDateTimeType.mssql.test.ts.snap
patrykbojczuk: > I think maybe 'smalldatetime' needs to be added here

Adding it changes `columnType` like this:
 
```diff
-   @Property({ type: 'Date', columnType: 'smalldatetime', nullable: true, defaultRaw: `current_timestamp` })
+   @Property({ type: 'Date', columnType: 'smalldatetime(0)', nullable: true, defaultRaw: `current_timestamp` })
```

boenrobot: ok, so I was wrong, and it's not that then... The above would just opt "smalldatetime" into the length option, which means without it, SmallDateTimeType.getDefaultLength() is also not going to be needed.

Still, "type: 'Date'" remains wrong...

I'd suggest you run a step by step debugger on the whole flow and compare the paths taken between a "smalldatetime" column vs "datetime2" in the same table. Both are MSSQL specific types, and yet "datetime2" is picked up properly. Watch out for MsSqlSchemaHelper and MsSqlPlatform in particular, and any calls to those two from DatabaseTable and SourceFile.

It might be easiest to start with a console.log() in onInitialMetadata, to potentially either rule out or zero in on SourceFile as the culprit. If "type" is correct at this stage, the fault is somewhere in SourceFile. If it's "Date" at that stage, then it's not in SourceFile.

patrykbojczuk: The difference is that `datetime2` is mapped in `MsSqlPlatform.getDefaultMappedType` to `datetime`, so it's not a separate type.

```ts
if (normalizedType === 'smalldatetime') {
  return Type.getType(SmallDateTimeType);
}

const map = {
  int: 'integer',
  bit: 'boolean',
  real: 'float',
  uniqueidentifier: 'uuid',
  varbinary: 'blob',
  datetime2: 'datetime',
} as Dictionary;

return super.getDefaultMappedType(map[normalizedType] ?? type);
```

I've actually described this in the issue #5904, that adding `smalldatetime: 'datetime'` to this dictionary outputs, it seems to me, correct entity model, without the need to create a separate type.

In `DatabaseTable.getPropertyDeclaration`, type is determined using predefined types, where `datetime` exists, but `smalldatetime` obviously does not.

boenrobot: It's still odd to me that it didn't pick the type class, but whatever... It works with just a type map to an existing type class, so that's perfectly fine.

👍

boenrobot (approved):

B4nan (approved):

PR #6132, opened 2024-10-12
feat(core): add shortcut getter for `getEntityManager()`

lsh4711:


B4nan:
I would rather not add this, I am not very happy with the way people overuse this method in the first place.

Let's see what others thing, will leave it open for now.

lsh4711:
All right, I read a document expressing concerns about people misinterpreting the context of em and confirmed in advance that it was removed in e649480308cf367b73cda79ab5791a9dd3247fdd.
I mainly use the repository pattern(without em injection) with DI in NestJS, and I repeatedly access the getEntityManager() function to use transactions or `flush()`, which led me to attempt this PR.
Although I haven't tried it yet, what are your thoughts on the `@Transactional` decorator? :)

B4nan:
> without em injection

But why? You should inject it and use it where appropriate instead of flushing via repositories. They are extension points, thin wrappers around EM without any functionality, not a replacement for the EM.

The transactional decorator could be a nice way around the need for explicit flush (and in the end, you often want to have the transaction around the whole handler including the selecs). I'd welcome a PR for that.

lsh4711:
There is no problem with injecting both `EntityRepository` and `EntityManager` together, but I simply wanted to have only the dependency on repository.
I think this is probably due to my habit from when I used `JPA`.

Since I heard your positive opinion about `@Transactional`, I will give it a try. :)

PR #6138, opened 2024-10-14
perf(schema): filter tables before fetching table details

### LINKED ISSUE #5900: Filtering tables before fetching their details in Entity Generator
The database I am currently working on has a lot of tables. I am using the `takeTables` option to generate entity models for only some of them, but currently the filtering is done after retrieving the details of each table, making the generation take a very long time.

I would like it to be possible to filter tables before retrieving its details, i.e. columns, constraints etc.

I am currently using a patch that adds the `takeTables` and `skipTables` parameters to the `DatabaseSchema::create` method in the **@mikro-orm/knex** package and moves `isTableNameAllowed` from `EntityGenerator` there. This allows to load schema information only for the corresponding tables.
Then in `EntityGenerator::generate`, for references for which no entity model will exist, I substitute the type for the column type.

**Additional context**
I could work on PR implementing these changes if you think what I described is correct or change the approach if something doesn't seem right to you.

patrykbojczuk:
This moves table filtering from after retrieving all table details (i.e. columns, foreign keys, etc.) to after retrieving a list of tables as described in #5900.

Closes #5900

### DIFF
--- a/tests/features/entity-generator/__snapshots__/EntityGenerator.postgres.test.ts.snap
+++ b/tests/features/entity-generator/__snapshots__/EntityGenerator.postgres.test.ts.snap
@@ -36,6 +36,24 @@ export enum Publisher2Type2 {
 
 exports[`EntityGenerator generate entities from schema [postgres]: postgres-entity-dump 1`] = `
 [
+  "import { Entity, ManyToOne, PrimaryKey } from '@mikro-orm/core';
+import { Label2 } from './Label2';
+import { Test2 } from './Test2';
+
+@Entity({ tableName: 'label2_tests', schema: 'label_schema' })
+export class Label2Tests {
+
+  @PrimaryKey()
+  id!: number;
+
+  @ManyToOne({ entity: () => Label2, updateRule: 'cascade', deleteRule: 'cascade' })
+  label2!: Label2;
+
+  @ManyToOne({ entity: () => Test2, updateRule: 'cascade', deleteRule: 'cascade' })
+  test2!: Test2;
+
+}
+",
   "import { Entity, OneToOne, PrimaryKeyProp, Property } from '@mikro-orm/core';
 import { Author2 } from './Author2';
 
@@ -305,24 +323,6 @@ export class Label2 {
 
 }
 ",
-  "import { Entity, ManyToOne, PrimaryKey } from '@mikro-orm/core';
-import { Label2 } from './Label2';
-import { Test2 } from './Test2';
-
-@Entity({ tableName: 'label2_tests', schema: 'label_schema' })
-export class Label2Tests {
-
-  @PrimaryKey()
-  id!: number;
-
-  @ManyToOne({ entity: () => Label2, updateRule: 'cascade', deleteRule: 'cascade' })
-  label2!: Label2;
-
-  @ManyToOne({ entity: () => Test2, updateRule: 'cascade', deleteRule: 'cascade' })
-  test2!: Test2;
-
-}
-",
   "import { Entity, Enum, PrimaryKey, Property } from '@mikro-orm/core';
 
 @Entity()
@@ -431,6 +431,24 @@ export class Test2 {
 
 exports[`EntityGenerator generate entities from schema with forceUndefined = false [postgres]: postgres-entity-dump 1`] = `
 [
+  "import { Entity, ManyToOne, PrimaryKey } from '@mikro-orm/core';
+import { Label2 } from './Label2';
+import { Test2 } from './Test2';
+
+@Entity({ tableName: 'label2_tests', schema: 'label_schema' })
+export class Label2Tests {
+
+  @PrimaryKey()
+  id!: number;
+
+  @ManyToOne({ entity: () => Label2, updateRule: 'cascade', deleteRule: 'cascade' })
+  label2!: Label2;
+
+  @ManyToOne({ entity: () => Test2, updateRule: 'cascade', deleteRule: 'cascade' })
+  test2!: Test2;
+
+}
+",
   "import { Entity, OneToOne, PrimaryKeyProp, Property } from '@mikro-orm/core';
 import { Author2 } from './Author2';
 
@@ -700,24 +718,6 @@ export class Label2 {
 
 }
 ",
-  "import { Entity, ManyToOne, PrimaryKey } from '@mikro-orm/core';
-import { Label2 } from './Label2';
-import { Test2 } from './Test2';
-
-@Entity({ tableName: 'label2_tests', schema: 'label_schema' })
-export class Label2Tests {
-
-  @PrimaryKey()
-  id!: number;
-
-  @ManyToOne({ entity: () => Label2, updateRule: 'cascade', deleteRule: 'cascade' })
-  label2!: Label2;
-
-  @ManyToOne({ entity: () => Test2, updateRule: 'cascade', deleteRule: 'cascade' })
-  test2!: Test2;
-
-}
-",
   "import { Entity, Enum, PrimaryKey, Property } from '@mikro-orm/core';
 
 @Entity()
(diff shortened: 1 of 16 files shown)


patrykbojczuk:
I’ve found an issue with the `OddTableNames.mysql.test.ts`.

Previously, in the `E123TableName` entity, there was a nonexistent `Table$39s$32name$32has$32apostrophe$44$32also$32$96$32this` entity used (it’s part of another database) as a `prototype` property type, but as it does not exist in the metadata, it’s typed as a `string` instead.

I’m not sure how you’d want to address this.

patrykbojczuk:
> Same way as if the name was a completely typical name that doesn't need any escapes...

@boenrobot It could be different name (it just happened to be odd in this case), but it still wouldn't get any tables from other schemas (which in mysql are equivalent to databases), and therefore it won't add schema name to the entity options.

This was the case before too - the test snapshot does not contain the entity model of the aforementioned table.

patrykbojczuk:
@boenrobot Yes, I think it will be better if addressed in the future PR, perhaps by someone with more experience in mysql/mariadb realm.

tests/features/entity-generator/__snapshots__/EntityGenerator.postgres.test.ts.snap
```diff
@@ -1384,8 +1383,8 @@ export class Configuration2 {
   @PrimaryKey()
   property!: string;
 
-  @ManyToOne({ entity: () => Test2, updateRule: 'cascade', primary: true })
-  test!: Test2;
+  @PrimaryKey({ fieldName: 'test_id' })
+  test!: number;
```
B4nan, line 1387: this doesn't feel correct, neither the other change in this file

patrykbojczuk: I've assumed, that this snapshot was previously incorrect, as skipped or not taken tables' entity models were used.

tests/features/multiple-schemas/__snapshots__/multiple-schemas.postgres.test.ts.snap
```diff
@@ -15,8 +14,8 @@ export class Book {
   @Property({ nullable: true })
   name?: string;
 
-  @ManyToOne({ entity: () => Author, updateRule: 'cascade', deleteRule: 'cascade', nullable: true })
-  author?: Author;
```
boenrobot, line 19: Although the result is now correct, this correction is a hint that generating multiple schemas at once is now broken... And upon adding an extra test that generates the entities without a filter, that's what I'm seeing.

Breaking for MySQL is fine, since it was never properly supported anyway, but for Postgresql and MSSQL, it was. At least for those, the filter for tables should include all schemas, and when searching for FKs, the skipped ones should be for tables that are otherwise skipped for whatever reason (be it the use of "schema" or a skipTable match...).

patrykbojczuk: The test you're referring to - `multiple connected schemas in postgres generate entities for given schema only` - is specifically checking if entities from other schemas are filtered out.

I couldn't reproduce the issue with `Author` entity not existing even if not filtering by schema.
When generating the entities without the `schema` property specified, entity models for all tables are correctly generated.

boenrobot: Yes, all are generated, but the link between the schemas is missing. The very relation this comment is on shows with just author_id rather than author, when schema filter is not present.

patrykbojczuk: I've added a test to check entity model generation for all schemas (without `schema` option specified) to both mssql and postgres `multiple-schemas` test suites.

patrykbojczuk: And now I see that problem.

tests/features/multiple-schemas/__snapshots__/multiple-schemas.postgres.test.ts.snap
```diff
@@ -1,9 +1,8 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`multiple connected schemas in postgres generate entities for given schema only 1`] = `
+exports[`multiple connected schemas in postgres generate entities for all schemas 1`] = `
 [
   "import { Collection, Entity, ManyToMany, ManyToOne, PrimaryKey, Property } from '@mikro-orm/core';
-import { Author } from './Author';
 import { BookTag } from './BookTag';
 
 @Entity({ schema: 'n2' })
```
patrykbojczuk, line 8: @boenrobot Do you have any idea why would an entity model for `n2.book` be generated twice, once without the `N2` class prefix and `tableName` in the `EntityOptions` and the second time, under the `Author` entity model, with the class prefix and `tableName` option set?

patrykbojczuk: When debugging the issue, I've got correct result when other tests are skipped, but the output keeps changing - it's not sorted.

boenrobot: I haven't checked out your new implementation in detail, but... Do you have ORDER BY in all the final queries (for tables, columns, indexes, FKs...)? If not, that may be a source of order instability.

patrykbojczuk: Ok, the sorting issue should now be fixed.

boenrobot: > Do you have any idea why would an entity model for n2.book be generated twice, once without the N2 class prefix and tableName in the EntityOptions and the second time, under the Author entity model, with the class prefix and tableName option set?

And as for that part... IIRC, by default, the default schema ("public" in the case of Postgresql, "dto" in the case of MSSQL...) gets a special treatment... When an entity is defined there, it's not prefixed in cases of conflicts. Only all other schemas are. I believe there is one test (forgot which one) in which the naming strategy is overridden to prefix any conflicting entities with the schema name instead.

patrykbojczuk: The problem with doubled entity models for a single table still does occur.

patrykbojczuk: So in short the entities generated in the first test to run appear in the second test to run.

patrykbojczuk: Which is because the `EntityGenerator` instance is reused between the tests and therefore all of the `EntityGenerator`'s fields are shared between test runs.

patrykbojczuk: Now each test should have its own ORM instance.

patrykbojczuk:
> you've [indirectly](https://app.codecov.io/gh/mikro-orm/mikro-orm/pull/6138/indirect-changes) caused loss of coverage in [MsSqlQueryBuilder's getKnexQuery() when using it for truncate type queries](https://app.codecov.io/gh/mikro-orm/mikro-orm/pull/6138/blob/packages/mssql/src/MsSqlQueryBuilder.ts#L26) (like... how?)

It's probably because truncate is not tested anywhere, whereas earlier, for each test the table was truncated, which probably counted as a coverage 🤔

As for the `getTableName`, it is used in the truncate, so that was probably covered with truncate.

tests/features/truncate/truncate.mssql.test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  test('truncates table and resets identity value', async () => {
+    await orm.em.persistAndFlush([
+      orm.em.create(User, { name: 'u1' }),
+      orm.em.create(User, { name: 'u2' }),
+      orm.em.create(User, { name: 'u3' }),
+    ]);
+
+    const [{ identity: identityBefore }] = await orm.em
+      .getKnex()
```
B4nan, line 42: please dont use knex directly in tests, just do `em.execute()`

i already have a local branch that removes knex from the QB implementation, this would only add conflicts for no reason

B4nan (changes requested):

boenrobot (approved):

B4nan:
Looks like there is still one line that is no longer covered by tests, not a big deal in terms of coverage, I just wonder if it's needed or what happened.

https://app.codecov.io/gh/mikro-orm/mikro-orm/commit/74bf43fa34cc129e4561f6ad5596af9ecbba32a9/blob/packages/core/src/drivers/DatabaseDriver.ts#L486

B4nan:
That method was quite overcomplicated, we don't need 4 return statements and nested conditions at all, refactored here to get around this without more tests.

04321e06ce083ec129432cc3436846737382b006

PR #6139, opened 2024-10-14
docs: fix forceObject to true to enforce objects

thomasgoldstein:
The documentation for 6.x says `forceObject` should be `false` if you want to enforce objects. This is wrong, it should be `true`.

### DIFF
--- a/docs/docs/serializing.md
+++ b/docs/docs/serializing.md
@@ -157,7 +157,7 @@ Primary keys are automatically included. If you want to hide them, you have two 
 
 ### Foreign keys are `forceObject`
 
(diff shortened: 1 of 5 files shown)


B4nan (approved):
ouch, good catch!

B4nan (changes requested):
please change it in the next docs too, your changes would be forgotten on next feature release

https://github.com/mikro-orm/mikro-orm/tree/master/docs

thomasgoldstein:
The next docs are the ones in docs/docs, I assume. Done!

B4nan (approved):

PR #6143, opened 2024-10-15
feat(core): add `@Transactional()` decorator

lsh4711:
This is implementation of `@Transactional()` decorator, which internally uses `entityManager.transactional()`.
It works on async functions and can be used alongside `entityManager.transactional()`.

Closes https://github.com/mikro-orm/mikro-orm/issues/6183


packages/core/src/utils/DatasourceUtils.ts
```diff
@@ -0,0 +1,24 @@
+import type { EntityManager, MikroORM } from '../';
+
+export class DatasourceUtils {
```
B4nan, line 3: What is this for and why you think you need it? Because I don't believe you do. 

It looks like this is destined to leak memory. Also lets not bring typeorm terms in here, we don't call anything in the whole codebase a datasource.

packages/core/src/utils/DatasourceUtils.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  private static ormMap?: Map<string, MikroORM>;
+
+  static getEntityManager(name = 'default'): EntityManager | undefined {
+    return this.ormMap?.get(name)?.em;
+  }
+
+  /**
+   * To start first transaction using @Transactional(), need to set datasource.
```
B4nan, line 12: It should behave just like the existing `@CreateRequestContext` decorator. The user needs to inject something into the context, either the ORM instance or EM or some repository (or pass it in the decorator parameters), and we will use that. You can use `em.transactional` on a global EM too.

packages/core/src/decorators/Transactional.ts
```diff
@@ -0,0 +1,32 @@
+import { DatasourceUtils, RequestContext, TransactionContext, type TransactionOptions } from '../';
+import type { AsyncFunction } from '../typings';
+
+type TransactionalOptions = TransactionOptions & { name?: string };
+
+/**
+ * To use forked EntityManager, use TransactionContext.getEntityManager().
```
B4nan, line 7: `TransactionContext.getEntityManager()` is internal method, users should be using whatever they inject, e.g. `this.em`, not those.

B4nan (commented):
I'd like to see a different approach, this feels way too hacked. Try to think of this as an alternative to `@CreateRequestContext` decorator which also opens a transaction (and flushes at the end because of that).

packages/core/src/utils/DatasourceUtils.ts
lsh4711: When attempting to start a transaction with the `@Transactional()` decorator, if there is no valid EntityManager (in RequestContext or TransactionContext), a reference to the EntityManager was needed.
To address this, I created a class to store references to the ORM. Additionally, in scenarios where multiple ORMs are used and a reference is needed, it stores the reference and name of the ORM, allowing the user to specify which ORM to create a transaction with using `@Transactional({ name: 'ormName' })`.

Although I wrote the code with this thought process, after hearing your advice, I realized how to improve it. :)

packages/core/src/decorators/Transactional.ts
```diff
@@ -0,0 +1,72 @@
...
+      const context = typeof getContext === 'function'
+        ? await getContext(this)
+        : await getContext;
+      const em = getEntityManager(this, context);
+
+      if (!em) {
+        throw new Error(`@Transactional() decorator can only be applied to methods of classes with \`orm: MikroORM\` property, \`em: EntityManager\` property, or with a callback parameter like \`@Transactional(() => orm)\` that returns one of those types. The parameter will contain a reference to current \`this\`. Returning an EntityRepository from it is also supported.`);
+      }
+
+      return em.transactional(() => originalMethod.apply(this, args), options);
```
lsh4711, line 67: I thought it would be good to function the same as `em.transactional()`,
but should I improve it by using `em.begin()`, `flush()`, and `commit()` instead?

B4nan: Nope, that would also mean creating your own async context (to be able to use services from DI container), which `em.transactional` does for you, I don't see a reason to reimplement what the method is doing already.

packages/core/src/decorators/Transactional.ts
```diff
@@ -0,0 +1,72 @@
...
+  EntityRepository,
+  MikroORM,
+  RequestContext,
+  TransactionContext,
+  type TransactionOptions,
+  Utils,
+} from '../';
+import type { AsyncFunction, GetContext } from '../typings';
+
+function getEntityManager(caller: { orm?: MikroORM; em?: EntityManager }, context: unknown) {
```
B4nan, line 12: lets move this to some shared file and use the same in the other context decorators

packages/core/src/decorators/Transactional.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export function Transactional<T extends object>(): MethodDecorator;
+
+export function Transactional<T extends object>(getContext: GetContext<T>): MethodDecorator;
+
+export function Transactional<T extends object>(options: TransactionOptions): MethodDecorator;
+
+export function Transactional<T extends object>(getContext: GetContext<T>, options: TransactionOptions): MethodDecorator;
+
+export function Transactional<T extends object>(getContext?: GetContext<T> | TransactionOptions, options?: TransactionOptions): MethodDecorator {
```
B4nan, line 44: i would like to see a single signature, instead of mixing what the first parameter is doing, create extension of the `TransactionOptions` which also lets you pass the context getter as one of the options

B4nan (commented):
i will have a closer look later, but it looks quite good now. we are missing docs (as well as jsdoc)

packages/core/src/typings.ts
```diff
@@ -1217,3 +1218,11 @@ export interface Seeder<T extends Dictionary = Dictionary> {
 export type ConnectionType = 'read' | 'write';
 
 export type MetadataProcessor = (metadata: EntityMetadata[], platform: Platform) => MaybePromise<void>;
+
+export type SyncOrAsync<T> = T | Promise<T>;
```
B4nan, line 1222: we already have `MaybePromise` type

packages/core/src/decorators/Transactional.ts
```diff
@@ -0,0 +1,32 @@
+import { RequestContext, resolveGetContext, TransactionContext, type TransactionOptions } from '../';
```
B4nan, line 1: don't import from barrel files

packages/core/src/decorators/CreateRequestContext.ts
```diff
@@ -1,31 +1,24 @@
-import { MikroORM } from '../MikroORM';
+import { resolveGetContext } from '../';
```
B4nan, line 1: don't import from barrel files

B4nan (changes requested):
few nits

B4nan (changes requested):
lets document this

https://mikro-orm.io/docs/transactions#approach-2-explicitly
https://mikro-orm.io/docs/identity-map#ensurerequestcontext-decorator

packages/core/src/decorators/Transactional.ts
```diff
@@ -0,0 +1,35 @@
+import type { TransactionOptions } from '../enums';
+import type { AsyncFunction, GetContext } from '../typings';
+import { RequestContext } from '../utils/RequestContext';
+import { TransactionContext } from '../utils/TransactionContext';
+import { resolveGetContext } from '../utils/Utils';
+
+type TransactionalOptions<T> = TransactionOptions & { getContext?: GetContext<T> };
+
+/**
+ * you can specify context to open transaction by providing `getContext` property.
```
B4nan, line 10: lets make this more informative, starting a comment with "you can" feels weird. describe what the decorator is about, not just one of its options.

```suggestion
 * you can specify context to open transaction by providing `getContext` option.
```

packages/core/src/utils/Utils.ts
```diff
@@ -183,6 +186,38 @@ export function parseJsonSafe<T = unknown>(value: unknown): T {
...
+
+  return undefined;
+}
+
+/**
+ * Find entityManager in injected context, or else in class's orm or em properties.
+ */
+export async function resolveGetContext<T>(caller: T & { orm?: MaybePromise<MikroORM>; em?: EntityManager }, getContext?: GetContext<T>): Promise<EntityManager | undefined> {
+  const context = typeof getContext === 'function' ? await getContext(caller) : await getContext;
+  return getEntityManager({ orm: await caller.orm, em: caller.em }, context);
```
B4nan, line 218: the EM can be also async callback
```suggestion
  return getEntityManager({ orm: await caller.orm, em: await caller.em }, context);
```

packages/core/src/utils/Utils.ts
```diff
@@ -15,9 +15,12 @@ import type {
   EntityMetadata,
   EntityName,
   EntityProperty,
+  GetContext,
   IMetadataStorage,
+  MaybePromise,
   Primary,
 } from '../typings';
+import { EntityManager, EntityRepository, MikroORM } from '@mikro-orm/core';
```
B4nan, line 23: you cant import from `core` package like this from inside the `core` package, this needs to be relative import. and it feels like an import cycle, the Utils class is imported from the EM, lets move all those changes to a separate file.

B4nan (changes requested):
few more notes

packages/core/src/utils/resolveGetContext.ts
```diff
@@ -0,0 +1,36 @@
+ import { EntityRepository } from '../entity/EntityRepository';
```
B4nan, line 1:
```suggestion
import { EntityRepository } from '../entity/EntityRepository';
```

packages/core/src/typings.ts
```diff
@@ -1217,3 +1218,9 @@ export interface Seeder<T extends Dictionary = Dictionary> {
 export type ConnectionType = 'read' | 'write';
 
 export type MetadataProcessor = (metadata: EntityMetadata[], platform: Platform) => MaybePromise<void>;
+
+/**
+ * you can also provide Promise<MikroORM> or callbck which return Promise<MikroORM | EntityManager | EntityRepository>,
+ * and callback parameter is injected with `this`.
```
B4nan, line 1224: either improve this comment or mark it as `@internal`, otherwise, it will end up in docs and there we need something better than a comment starting with "you can also...".

not to mention there are some typos (`callbck`) and missing code blocks

packages/core/src/decorators/Transactional.ts
```diff
@@ -0,0 +1,38 @@
+import type { TransactionOptions } from '../enums';
+import type { AsyncFunction, GetContext } from '../typings';
+import { RequestContext } from '../utils/RequestContext';
+import { resolveGetContext } from '../utils/resolveGetContext';
+import { TransactionContext } from '../utils/TransactionContext';
+
+type TransactionalOptions<T> = TransactionOptions & { getContext?: GetContext<T> };
```
B4nan, line 7: I would maybe call this just `context` instead of `getContext`. `getContext` sounds like a getter name, not a parameter name. We already have `context` option in the ORM and this feels like a very similar use case.

docs/docs/identity-map.md
```diff
@@ -163,6 +163,10 @@ export class MyService {
 
 Sometimes you may prefer to just ensure the method is executed inside a request context, and reuse the existing context if available. You can use the `@EnsureRequestContext()` decorator here, it behaves exactly like the `@CreateRequestContext`, but only creates new context if necessary, reusing the existing one if possible.
 
+## `@Transactional()` decorator
+
+If you want to ensure that the method runs within a transaction scope, use `@Transactional()` decorator. The operations inside the function are performed within a new transaction context, and upon function completion, flush and commit are automatically executed. This decorator can also be used in a nested manner. For more details about transactions, check the document on [Transactions](./transactions).
```
B4nan, line 168: we should also mention the context option and that it works the same as the one for `@CreateRequestContext` decorator

B4nan (commented):
we are almost there, good job!

docs/docs/identity-map.md
```diff
@@ -163,6 +163,13 @@ export class MyService {
 
 Sometimes you may prefer to just ensure the method is executed inside a request context, and reuse the existing context if available. You can use the `@EnsureRequestContext()` decorator here, it behaves exactly like the `@CreateRequestContext`, but only creates new context if necessary, reusing the existing one if possible.
 
+## `@Transactional()` decorator
+If you want to ensure that the method runs within a transaction scope, use `@Transactional()` decorator.
+
+You can provide the same callback that can be used in `@CreateRequestContext()` decorator as `context` option, which means you can specify the context in which the transaction begins. The function are performed within a new transaction context, and upon function completion, flush and commit are automatically executed. This decorator can also be used in a nested manner.
```
B4nan, line 169:
```suggestion
You can provide the same callback that can be used in the `@CreateRequestContext()` decorator via the `context` option, allowing you to handle the injection of the EntityManager in non-standard scenarios. The function is executed within a new transaction context, and upon completion, flush and commit are automatically executed. This decorator can also be used in a nested manner.
```

docs/docs/identity-map.md
```diff
@@ -163,6 +163,13 @@ export class MyService {
...
+
+For more details about transactions, check the document on [Transactions](./transactions).
```
B4nan, line 171:
```suggestion
For more details about transactions, check the [Transactions section](./transactions).
```

docs/docs/transactions.md
```diff
@@ -57,9 +57,31 @@ try {
 }
 ```
 
+Another way to handle explicit transactions is by using `@Transactional()` decorator.
```
B4nan, line 60:
```suggestion
Another way to handle explicit transactions is by using the`@Transactional()` decorator.
```

docs/docs/transactions.md
```diff
@@ -57,9 +57,31 @@ try {
...
+
+}
+```
+
+This decorator wraps the method with `em.transactional()`, so you can provide `TransactionOptions` just like with `em.transactional()`. The difference is that you can specify the context in which the transaction begins by providing `context` option, and if omitted, the transaction will begin in the current context implicitly. It works on async functions and can be nested with `em.transactional()`.
+
 Explicit transaction demarcation is required when you want to include custom DBAL operations in a unit of work or when you want to make use of some methods of the EntityManager API that require an active transaction. Such methods will throw a `ValidationError` to inform you of that requirement.
 
-`em.transactional(cb)` will flush the inner `EntityManager` prior to transaction commit.
+`em.transactional(cb)` and `@Transactional()` will flush the inner `EntityManager` prior to transaction commit.
```
B4nan, line 84:
```suggestion
`em.transactional(cb)` and the `@Transactional()` will flush the inner `EntityManager` before transaction commit.
```

packages/core/src/typings.ts
```diff
@@ -1217,3 +1218,8 @@ export interface Seeder<T extends Dictionary = Dictionary> {
 export type ConnectionType = 'read' | 'write';
 
 export type MetadataProcessor = (metadata: EntityMetadata[], platform: Platform) => MaybePromise<void>;
+
+/**
+ * The type of context that the user intends to inject.
+ */
+export type Context<T> = MaybePromise<MikroORM> | ((type: T) => MaybePromise<MikroORM | EntityManager | EntityRepository<any> | { getEntityManager(): EntityManager }>);
```
B4nan, line 1225: I wouldn't call this "Context", since it does not really represent one, I said I would rename the option, not the type. I can't say I like the previous name either, but this one feels way too generic, we might want to use "Context" for something else in the long run.

A context is what the `@CreateRequestContext` provides, not its input. Maybe we could call this "context provider", I guess that sounds better.

B4nan (changes requested):

B4nan (changes requested):
one last bit before we merge. the test is currently quite slow, because you use the `init...` methods from the bootstrap file. especially with MySQL, this is terribly slow, since it creates 2 replicas, so it issues the create schema queries 3 times, not to mention those methods register quite a lot of entities, but you need just one in here.

i am also not sure if we really need to test this with different drivers, since you haven't touched anything driver specific, i would simplify the test to use sqlite in memory database only to make it fast.

B4nan (approved):
thanks!

PR #6147, opened 2024-10-16
feat(core): follow standard-schema

xcfox:
This PR makes Mikro's EntitySchema follow [standard-schema](https://github.com/standard-schema/standard-schema)

To implement `~validate` for [standard-schema](https://github.com/standard-schema/standard-schema?tab=readme-ov-file#the-interface), I also added the `validateSafely` method for simple validation in `EntityValidator`


packages/core/src/typings.ts
```diff
@@ -1217,3 +1217,117 @@ export interface Seeder<T extends Dictionary = Dictionary> {
...
 
 export type MetadataProcessor = (metadata: EntityMetadata[], platform: Platform) => MaybePromise<void>;
+
+
+// Standard Schema
+
+/**
+ * The Standard Schema v1 interface.
+ */
+export interface StandardSchema<Input = unknown, Output = Input> {
```
boenrobot, line 1227: In the FAQ, there is this part about [whether to add standard schema as a dependency](https://github.com/standard-schema/standard-schema?tab=readme-ov-file#do-i-need-to-include-standard-schemaspec-as-a-dependency), and they do encourage copy & paste-ing it, as you've done here, but...

IMO, it will be best to add it as a dev dependency (as also encouraged by the FAQ answer), as this means we get an early warning if they accidentally tweak v1 in an incompatible way, or decide to deprecate it. Or (more likely), if v2 comes along, we'll get news about it through dependabot / renovate updating the version, and can plan out if/how/when to use v2.

The downside is that if they do break it, MikroORM can't build a new version... But MikroORM is a lively enough project that if this happens, I'm sure we'll be able to quickly fix it _somehow_... Probably by making the adjustments needed. @B4nan ?

packages/core/src/entity/EntityValidator.ts
```diff
@@ -41,29 +41,94 @@ export class EntityValidator {
     });
   }
 
+  validateSafely<T>(payload: unknown, meta: EntityMetadata<T>): StandardOutput<T> {
```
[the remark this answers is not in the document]
xcfox: I have tried to implement `~validate` directly using the logic of `EntityFactory.create()`, but EntityFactory has more dependencies and is difficult to call in EntitySchema.

I will try to implement a new `SchemaValidator` to make it follow `standard-schema`, `SchemaValidator` is not used to validate POJOs, but to validate DTOs to make them strictly conform to `RequiredEntityData<TEntity>`. For simple validation, we can just use `SchemaValidator` instead of `zod`.

xcfox:
@B4nan I reimplemented the `~validate` method, which will try to get the EntityManager from the context at runtime, create the entity using the EntityFactory, and validate the created entity using the EntityValidator.

packages/core/src/metadata/EntitySchema.ts
```diff
@@ -62,6 +70,46 @@ export class EntitySchema<Entity = any, Base = never> {
   private internal = false;
   private initialized = false;
 
+
+  readonly '~standard' = 1;
+
+  readonly '~vendor' = 'mikro-orm';
+
+  '~validate'(input: StandardInput): StandardOutput<Entity> {
+    const em = RequestContext.getEntityManager();
```
B4nan, line 79: this also doesnt feel good, first of all, this is completely optional way to deal with the contexts, people might just fork explicitly and never create a context this way; second, its configurable via `getContext` option, people can bring their own ALS manager.

packages/core/src/metadata/EntitySchema.ts
```diff
@@ -62,6 +70,46 @@ export class EntitySchema<Entity = any, Base = never> {
...
+      const prop = this.meta.properties[key];
+
+      if (prop && prop.kind === ReferenceKind.SCALAR && SCALAR_TYPES.includes(prop.runtimeType)) {
+        try {
+          data[key] = validator.validateProperty(prop, data[key], data);
+        } catch (error) {
+          issues.push({ message: (error as Error).message });
+        }
+      }
+    });
```
B4nan, line 99: this is duplicating the work done by the EntityFactory, which already validates things. i doubt this code is reachable

packages/core/src/metadata/EntitySchema.ts
```diff
@@ -62,6 +70,46 @@ export class EntitySchema<Entity = any, Base = never> {
...
+
+  '~validate'(input: StandardInput): StandardOutput<Entity> {
+    const em = RequestContext.getEntityManager();
+    if (em == null) {
+      return { issues: [{ message: 'No EntityManager found in the context' }] };
+    }
+    const entityFactory = new EntityFactory(em);
+
+    const issues: StandardIssue[] = [];
+    const data = entityFactory.create(this._meta.className, input.value as any) as Entity & {};
```
B4nan, line 86: the factory returns entities, not "data"

B4nan (commented):
this is unfortunately not much better

lets first properly define what you want to do, since i still don't understand fully what this new method should be about. the ORM does validation for you, so i am not really sold on having it.

xcfox:
> this is unfortunately not much better
> 
> lets first properly define what you want to do, since i still don't understand fully what this new method should be about. the ORM does validation for you, so i am not really sold on having it.

The role of `~validate` is to validate user input (DTO), as [zod](https://mikro-orm.io/docs/guide/advanced#validation-via-zod) does.

The `~validate` method should tell the user whether the current input can create a legal entity. 
I think it should also auto-add the default fields and do some necessary type conversions and return an entity, if possible. (Looks like what `em.create()` does, but `em.create()` does't do validation.)

B4nan:
Yes, but what should that method be for? The ORM won't use it anyhow.

Also, you keep saying DTO, but what does it mean, raw DB results or something else? What about custom types? 

> but em.create() does't do validation.

Right, because the ORM will validate things on flush, the validation still happens.

xcfox:
Please forgive me for not being clear in my explanation. A [**data transfer object**](https://learn.microsoft.com/en-us/aspnet/web-api/overview/data/using-web-api-with-entity-framework/part-5) (DTO) is an object that defines how the data will be sent over the network. The [examples in nestjs](https://docs.nestjs.com/techniques/validation#mapped-types) may be more intuitive.

## Where will the `~validate` method be used?

For libraries that follow the `standard-schema`, the primary use of the `validate` function in a back-end APP is to validate server requests. Third-party libraries (like [tRPC](https://trpc.io/)) will call the `validate` function to validate that the [input](https://trpc.io/docs/server/validators#input-validators) or [output](https://trpc.io/docs/server/validators#output-validators) of the request is as expected.

But in most cases, we prefer to use libraries like [zod](https://zod.dev/) or [valibot](https://valibot.dev) that focus on validation. I think the cases where Mikro is used directly for request validation are very rare. Considering that Mikro is already well validated internally, `~validate` is probably useless for EntitySchema.

Do you accept a `~validate` method that has no actual functionality, but just matches `standard-schema` in type?

```ts
class EntitySchema<Entity> implements StandardSchema<RequiredEntityData<Entity>, Entity> {

  readonly '~standard' = 1;

  readonly '~vendor' = 'mikro-orm';

  ~validate(_: StandardInput): StandardOutput<Entity> {
    return { issues:[{ message:'Not implemented in EntitySchema' }] };
  }

}
```

Another option is to implement only [BaseSchema](https://github.com/standard-schema/standard-schema/blob/main/packages/spec/src/base.ts) instead of the full StandardSchema.

```ts
class EntitySchema<Entity> implements BaseSchema<RequiredEntityData<Entity>, Entity> {

  readonly '~standard' = 1;

  readonly '~vendor' = 'mikro-orm';

}
```

xcfox:
## In tRPC

If we have the following entities:

```ts
export const User = new EntitySchema<IUser>({
  name: "User",
  properties: {
    id: { primary: true, type: "number" },
    email: { type: "string" },
    name: { type: "string" },
    posts: { kind: "1:m", entity: () => Post },
  },
})

export const Post = new EntitySchema<IPost>({
  name: "Post",
  properties: {
    id: { primary: true, type: "number" },
    title: { type: "string" },
    content: { type: "string" },
    user: { kind: "m:1", entity: () => User },
  },
})
```

If the EntitySchema does not follow the `standard-schema`, then we need to prepare an additional function to convert the EntitySchema:

```ts
function asStandardSchema<TEntity>(
  schema: EntitySchema<TEntity>
): StandardSchema<RequiredEntityData<Entity>, Entity> {
  // return a standard schema
}
```

We would have routes like the following:

```ts
export const t = initTRPC.create()
const publicProcedure = t.procedure

export const appRouter = t.router({
  user: publicProcedure.output(asStandardSchema(User)).query(() => {
    return someUser
  }),

  post: publicProcedure.output(asStandardSchema(Post)).query(() => {
    return somePost
  }),
})
```

If the EntitySchema follows the `standard-schema`, then we can use schema directly:

```ts
export const appRouter = t.router({
  user: publicProcedure.output(User).query(() => {
    return someUser
  }),

  post: publicProcedure.output(Post).query(() => {
    return somePost
  }),
})
```

## In GQLoom

[GQLoom](https://gqloom.dev/) is a GraphQL framework that uses various schemas directly and weaves them into GraphQL Schema, avoiding the tedious process of writing additional GraphQL Schema. GQLoom will provide prioritized integration with Mikro ORM.

If the EntitySchema does not follow the `standard-schema`, then we need to convert it manually when defining the Schema:

```ts
import { mikroSilk } from "@gqloom/mikro-orm"

export const User = mikroSilk(
  new EntitySchema<IUser>({
    name: "User",
    properties: {
      id: { primary: true, type: "number" },
      email: { type: "string" },
      name: { type: "string" },
      posts: { kind: "1:m", entity: () => Post },
    },
  })
)

export const Post = mikroSilk(
  new EntitySchema<IPost>({
    name: "Post",
    properties: {
      id: { primary: true, type: "number" },
      title: { type: "string" },
      content: { type: "string" },
      user: { kind: "m:1", entity: () => User },
    },
  })
)
```

Using in the resolver:

```ts
export const userResolver = resolver.of(User, {
  user: query(User, () => {
    return someUser
  }),

  posts: field(silk.list(Post), (user) => {
    return em.find(Post, { user })
  }),
})
```

If the EntitySchema follows the `standard-schema`, then we don't need to convert it manually when defining the Schema:

```ts
export const User = new EntitySchema<IUser>({
  name: "User",
  properties: {
    id: { primary: true, type: "number" },
    email: { type: "string" },
    name: { type: "string" },
    posts: { kind: "1:m", entity: () => Post },
  },
})

export const Post = new EntitySchema<IPost>({
  name: "Post",
  properties: {
    id: { primary: true, type: "number" },
    title: { type: "string" },
    content: { type: "string" },
    user: { kind: "m:1", entity: () => User },
  },
})
```

xcfox:
In fact, the biggest reason I want Mikro Entity to follow the `standard schema` is that it allows us to use the Entity Schema directly in third-party libraries without wrapping any other functions, and to ensure type safety. For example, in the tRPC router:

```ts
export const postRouter = router({
  add1: authedProcedure
    .input(
      z.object({
        id: z.string().uuid().optional(),
        text: z.string().min(1),
      })
    )
    .output(Post) // tRPC will infer output by `Post` and TypeScript will check if the mutation's return value matches the Post.
    .mutation(async ({ input, ctx }) => {
      const { name } = ctx.user
      const post = await em.create(Post, { ...input, name })
      return post // the reutrn value is OK
    }),

  add2: authedProcedure
    .output(Post)
    .mutation(async () => {
      const user = await em.create(User, { ... })
      return user // the return value is not math `Post`, TypeScript will give you the alert.
    }),
})
```

In most cases, the Entity Schema will be the Output, and usually we don't need to validate the output at runtime, it's enough to validate the return value using TypeScript before compilation. Existing projects usually use `zod` to validate Input, I think it is not necessary to provide complex validation mechanism inside Mikro ORM.

packages/core/src/metadata/EntitySchema.ts
xcfox: You're right, I refactored `EntityValidator`, `ValidationError` so that they don't depend on entity.

packages/core/src/metadata/EntitySchema.ts
xcfox: Yes, the refactored version will not use `ef.create()` but will validate directly for data.

xcfox:
You're quite right that a wrapper function that provides extended functionality is quite necessary.

In addition to the relations' types you mentioned, more frequent usage scenes are arrays and nullable:

```ts
export const postRouter = router({
  add3: authedProcedure
    .output(helper(Post).list()) // the output type is `IPost[]`
    .mutation(async ({ input, ctx }) => {
      const { name } = ctx.user
      const post = await em.create(Post, { ...input, name })
      return post // TypeScript will error because it isn't list
    }),

  add2: authedProcedure.output(helper(Post).nullable()).mutation(async () => {
    return null // null is allowed
  }),
})
```

B4nan:
Sidenote: I don't like the idea of `helper()` supporting the class handle like this, its purpose is to provide a wrapped entity instance for a given entity instance, not entity type/class. This function was added as more of an internal helper, users should be using the `wrap` function instead. I don't want to add any conditional logic to `helper()` since I want those calls to be cheap/free perf-wise.

> I would add those as extra options in helper(entity).serializer()

We already have `serialize()` method there, so adding `serializer` sounds like a very confusing thing to do. Or did you mean the existing method? I thought we were talking about schema validation, not serialization. I don't have much time to follow this discussion at the moment, I guess I missed some details, but I don't see how this whole effort is about serialization.

B4nan:
> My bad... I thought the difference between wrap() and helper() was exactly that helper() takes in a type class, while wrap() takes in an entity instance, so that seemed like the place for this.

`wrap()` can return the entity itself if it extends the ORM `BaseEntity` which defines all the public helpers already, with the second parameter set to `true` it will always return the `WrappedEntity` instance. `helper()` is always doing the latter, which is rather internal, since the extra methods and properties are prefixed with `__`. 

Regarding the serialization, why should we have some wrappers for that when we already have `wrap(e).serialize()`? And how is this related to the "standard schema" exactly?

I see those examples above, but I don't understand why would anyone want to do that instead of just calling the `wrap(e).serialize()` directly. Those examples look like we should implement all those helpers like `.optional()` and whatnot, and I don't see a proper reason to do that. Also, such things can easily be implemented as a 3rd party library, proving their usability first before they make it to the ORM directly.

xcfox:
> Those examples look like we should implement all those helpers like .optional() and whatnot, and I don't see a proper reason to do that. Also, such things can easily be implemented as a 3rd party library, proving their usability first before they make it to the ORM directly.

I think it might be more appropriate to implement it in third-party libraries. There are actually `silk` utils in [GQLoom](https://gqloom.dev/) to help with schema transformations. `silk` utils accepts not only EntitySchema from Mikro, but also schema from `valibot`, and all schemas that match `standard-schema`:

```ts
import { silk, resolver, field, query } from "@gqloom/core"
import { EntitySchema } from "@mikro-orm/core"
import * as v from "valibot"

const User = new EntitySchema<IUser>({})

const Cat = v.object({})

const catResolver = resolver.of(Cat, {
  cats: query(silk.list(Cat), () => {
    return []
  }),

  keeper: field(silk.nullable(User), (cat) => {
    return em.findOne(User, cat.keeperId)
  }),
})

const userResolver = resolver.of(User, {
  users: query(silk.list(User), () => {
    return em.find(User)
  }),
})
```

For Mikro ORM, we just need to make the EntitySchema follow the `standard-schema`

PR #6149, opened 2024-10-16
fix(core): check that inverse side exists when mapping collection items

jukkahell:


tests/issues/GH6148.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  beforeAll(async () => {
+    orm = await MikroORM.init({
+      entities: [A, B, C],
+      dbName: ':memory:',
+      driver: SqliteDriver,
+      debug: true,
+      dataloader: true,
+      allowGlobalContext: true,
+      populateWhere: 'infer',
```
B4nan, line 67:
```suggestion
      dataloader: true,
```

tests/issues/GH6148.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    const b = orm.em.create(B, { name: 'b1', c: [{ name: 'c2' }] });
+    await orm.em.flush();
+    orm.em.clear();
+
+    const a1 = await orm.em.findOneOrFail(A, a.id);
+    const b1 = await orm.em.findOneOrFail(B, b.id);
+    a1.c.load();
+    const bc = await b1.c.load();
+    expect(bc.length).toBe(1);
+    expect(bc[0].name).toBe('c2');
```
jukkahell, line 82: Forgot to add this change in the commit:
```suggestion
    const c = await Promise.all([a1.c.load(), b1.c.load()]);
    expect(c.length).toBe(2);
    expect(c[0].length).toBe(1);
    expect(c[1].length).toBe(1);
```

PR #6151, opened 2024-10-16
refactor(cli): made the CLI explicitly pass the config path from the arg, or else infer it once

boenrobot:
The output of debug is adjusted as a consequence of the config being a passed dependency. Driver dependency versions are going to be shown if a config is found, independently of whether a package.json was found.

The config is always validated, even during entity generation. When warnings about missing entities are disabled, the entities options are also set to empty arrays, to avoid the old / additional entities' meta data influencing the entity generation.

If the config path is automatically inferred from a CLI argument during MikroORM.init(), a deprecation warning is logged.

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -581,6 +581,18 @@ MikroORM.init({
 });
 ```
 
+## Deprecation warnings
+
+By default, doing something that is deprecated will result in a deprecation warning being logged. The default logger will in turn show it on the console.
+
+You can ignore all or only specific deprecation warnings. See [Logging's section on deprecation warnings](./logging.md#deprecation-warnings) for details.
+
+The full list of deprecation errors:
+
+| label | message                                                                                                                                                                                                                                                                            |
+|-------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| D001  | Path for config file was inferred from the command line arguments. Instead, you should set the MIKRO_ORM_CLI_CONFIG environment variable to specify the path, or if you really must use the command line arguments, import the config manually based on them, and pass it to init. |
+
 ## Using environment variables
 
 Since v4.5 it is possible to set most of the ORM options via environment variables. By default `.env` file from the root directory is loaded - it is also possible to set full path to the env file you want to use via `MIKRO_ORM_ENV` environment variable.
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -591,7 +591,7 @@ The full list of deprecation errors:
 
 | label | message                                                                                                                                                                                                                                                                            |
 |-------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| D001  | Path for config file was inferred from the command line arguments. Instead, you should set the MIKRO_ORM_CLI_CONFIG environment variable to specify the path, or if you really must use the command line arguments, import the config manually based on them, and pass it to init. |
+| D001  | Path for config file was inferred from the command line arguments. Instead, you should set the `MIKRO_ORM_CLI_CONFIG` environment variable to specify the path, or if you really must use the command line arguments, import the config manually based on them, and pass it to init. |
 
 ## Using environment variables
 
--- a/docs/docs/logging.md
+++ b/docs/docs/logging.md
@@ -63,7 +63,7 @@ You can ignore all deprecation warnings by setting `ignoreDeprecations` to `true
 
 ```ts
 return MikroORM.init({
-    ignoreDeprecations: true, // now no deprecations will be logged, though you may be surprised when upgrading
+  ignoreDeprecations: true, // now no deprecations will be logged, though you may be surprised when upgrading
 });
 ```
 
@@ -71,7 +71,7 @@ When you are actively trying to remove deprecation warnings in preparation for a
 
 ```ts
 return MikroORM.init({
-    ignoreDeprecations: ['D0001'], // ignore deprecation with label "D0001", but show others if they pop up
+  ignoreDeprecations: ['D0001'], // ignore deprecation with label "D0001", but show others if they pop up
 });
 ```
 
--- a/docs/docs/logging.md
+++ b/docs/docs/logging.md
@@ -49,9 +49,33 @@ return MikroORM.init({
 });
 ```
 
-Currently, there are 5 namespaces – `query`, `query-params`, `schema`, `discovery` and `info`.
+Currently, there are 6 namespaces – `query`, `query-params`, `schema`, `discovery`, `info` and `deprecated`.
 
 If you provide `query-params` then you must also provide `query` in order for it to take effect.
+
+## Deprecation warnings
+
+Even without `debugMode` enabled, the default logger will show `deprecated` messages in console.
+
+When something is deprecated, it means there is an intention for it to be removed in a future version. The deprecation message should suggest alternatives for you that you should switch to before migrating to a major version.
+
+You can ignore all deprecation warnings by setting `ignoreDeprecations` to `true`
+
+```ts
+return MikroORM.init({
+    ignoreDeprecations: true, // now no deprecations will be logged, though you may be surprised when upgrading
+});
+```
+
+When you are actively trying to remove deprecation warnings in preparation for an upgrade, you would likely want to tackle them one at a time. You can ignore only specific deprecation warnings you can't deal with right now, while still being alerted of others you didn't know about. To do this, list the deprecation warnings you want to ignore, e.g.
+
+```ts
+return MikroORM.init({
+    ignoreDeprecations: ['D0001'], // ignore deprecation with label "D0001", but show others if they pop up
+});
+```
+
+You can see a list of deprecation errors in [Configuration's section on deprecated warnings](./configuration.md#deprecation-warnings).
 
 ## Highlighters
 
--- a/packages/cli/src/commands/ClearCacheCommand.ts
+++ b/packages/cli/src/commands/ClearCacheCommand.ts
@@ -12,7 +12,9 @@ export class ClearCacheCommand implements BaseCommand {
    * @inheritDoc
    */
   async handler(args: ArgumentsCamelCase<BaseArgs>) {
-    const config = await CLIHelper.getConfiguration();
+    const config = await CLIHelper.getConfiguration(
+      (typeof args.config !== 'undefined') ? [args.config] : undefined,
+    );
 
     if (!config.get('metadataCache').enabled) {
       CLIHelper.dump(colors.red('Metadata cache is disabled in your configuration. Set cache.enabled to true to use this command.'));
--- a/packages/cli/src/commands/ClearCacheCommand.ts
+++ b/packages/cli/src/commands/ClearCacheCommand.ts
@@ -12,9 +12,7 @@ export class ClearCacheCommand implements BaseCommand {
    * @inheritDoc
    */
   async handler(args: ArgumentsCamelCase<BaseArgs>) {
-    const config = await CLIHelper.getConfiguration(
-      (typeof args.config !== 'undefined') ? [args.config] : undefined,
-    );
+    const config = await CLIHelper.getConfiguration(args.config);
 
     if (!config.get('metadataCache').enabled) {
       CLIHelper.dump(colors.red('Metadata cache is disabled in your configuration. Set cache.enabled to true to use this command.'));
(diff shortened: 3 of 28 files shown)


B4nan:
> If the config path is automatically inferred from a CLI argument during MikroORM.init(), a deprecation warning is logged.

I am still not sure about this one, can you remind me why did you want to get rid of this so hard? :]

Or do you envision some more native solution for #5914 that would cover this use case too?

B4nan:
Ok, thanks for the write-up, agreed with the points you made.

> remove empty init altogether

I am not so sure about this one, mainly because of the use case with using env vars instead of the ORM config, but I am surely open to exploring this more.

packages/core/src/MikroORM.ts
```diff
@@ -34,7 +34,12 @@ export class MikroORM<D extends IDatabaseDriver = IDatabaseDriver, EM extends En
     const env = ConfigurationLoader.loadEnvironmentVars<D>();
 
     if (!options) {
-      options = (await ConfigurationLoader.getConfiguration<D, EM>()).getAll();
+      const configPathFromArg = ConfigurationLoader.configPathsFromArg();
+      const config = (await ConfigurationLoader.getConfiguration<D, EM>(configPathFromArg ?? ConfigurationLoader.getConfigPaths()));
+      options = config.getAll();
+      if (configPathFromArg) {
+        config.getLogger().warn('info', 'Path for config file was inferred from the command line arguments. This is deprecated. Set the MIKRO_ORM_CLI_CONFIG environment variable to specify the path, or if you really must use the command line arguments, import the config manually based on them, and pass it to init.');
```
B4nan, line 41: same here regarding the deprecation namespace.

packages/cli/src/commands/DebugCommand.ts
```diff
@@ -20,15 +20,21 @@ export class DebugCommand implements BaseCommand {
...
-    const configPaths = CLIHelper.getConfigPaths();
+    const configPaths = (typeof args.config !== 'undefined') ? [args.config] : CLIHelper.getConfigPaths();
     CLIHelper.dump(' - searched config paths:');
     await DebugCommand.checkPaths(configPaths, 'yellow');
 
     try {
-      const config = await CLIHelper.getConfiguration();
+      const config = await CLIHelper.getConfiguration(configPaths);
       CLIHelper.dump(` - configuration ${colors.green('found')}`);
+      const [drivers] = await Promise.all([CLIHelper.getDriverDependencies(config)]);
```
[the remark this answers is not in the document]
boenrobot: Messy refactoring on my part... async is also not required in the first place...

My thought process was that getting the dependencies probably involves some IO, and we don't want to wait for the IO for each dependency, but wait for them together... but that's not the case, and I forgot to go back to fix this...

Anyway... now cleaned up.

B4nan:
Btw I wanted to ship v6.4 later this week unless you have some plans to PR something worth including in a feature release?

packages/cli/src/commands/ClearCacheCommand.ts
```diff
@@ -12,7 +12,9 @@ export class ClearCacheCommand implements BaseCommand {
    * @inheritDoc
    */
   async handler(args: ArgumentsCamelCase<BaseArgs>) {
-    const config = await CLIHelper.getConfiguration();
+    const config = await CLIHelper.getConfiguration(
+      (typeof args.config !== 'undefined') ? [args.config] : undefined,
```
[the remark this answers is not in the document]
boenrobot: How about allowing --config to occur multiple times (make it an array)? If present multiple times, search all paths in the order they appear in CLI.

In addition to being a non-breaking new feature, it allows this ternary to collapse to

```typescript
const config =  await CLIHelper.getConfiguration(args.config)
```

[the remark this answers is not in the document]
B4nan: I can't say I understand what that would be good for, but I don't mind it either.

B4nan:
> Considering it's not a breaking change, and the modifications will be to "internal" classes, I think it's ok to have it even in a minor release like 6.4.1 or whatever version is next at the time.

I am fine waiting a bit more, this sounds like something worth including in a feature release to me. I'll drop a short blog post about that, and this one sounds like worth promoting too, so better wait for it unless it proves to be more complex/harder to implement.

B4nan:
> ok, made deprecated be outputted by default by the default logger... by changing the default of "debugMode" from "false" to "['deprecated']". An explicit "false" would still behave the same as always, i.e. not log even deprecated notices.

I am not sure about that, I can already see people using `['query']` or `['query', 'query-params']` explicitly, I think we should print the deprecations regardless of the log level. Disabling them with explicit `false` sounds good, but not showing them with explicit array value feels a bit weird.

I still hope not many people are using this `--config` on app level, this is a rather theoretical discussion since there will be more deprecations coming in the upcoming months.

B4nan:
Indeed, I said we need a way to opt-out. I just don't think we should mix this with the existing mechanism, as deprecations are important and they could be easily missed this way. I'd rather see a separate option for this like `ignoreDeprecations`.

packages/core/src/MikroORM.ts
```diff
@@ -34,7 +34,12 @@ export class MikroORM<D extends IDatabaseDriver = IDatabaseDriver, EM extends En
     const env = ConfigurationLoader.loadEnvironmentVars<D>();
 
     if (!options) {
-      options = (await ConfigurationLoader.getConfiguration<D, EM>()).getAll();
+      const configPathFromArg = ConfigurationLoader.configPathsFromArg();
+      const config = (await ConfigurationLoader.getConfiguration<D, EM>(configPathFromArg ?? ConfigurationLoader.getConfigPaths()));
+      options = config.getAll();
+      if (configPathFromArg) {
+        config.getLogger().warn('deprecated', 'Path for config file was inferred from the command line arguments. This is deprecated. Set the MIKRO_ORM_CLI_CONFIG environment variable to specify the path, or if you really must use the command line arguments, import the config manually based on them, and pass it to init.', { label: 'MikroORM:0001' });
```
B4nan, line 41: I am not sure about the `MikroORM:` prefix, I'd drop that. All deprecations would have it, users would have to repeat it for no apparent reason.

tests/features/cli/CLIHelper.test.ts
```diff
@@ -545,4 +535,39 @@ Maybe you want to check, or regenerate your yarn.lock or package-lock.json file?
...
+        discovery: {
+          warnWhenNoEntities: false,
+        },
+        debug: true,
+        logger: message => {
+          messages.push(message);
+        },
+      }))));
+    await MikroORM.init();
+    expect(messages[0]).toBe('[deprecated] (MikroORM:0001) Path for config file was inferred from the command line arguments. This is deprecated. Set the MIKRO_ORM_CLI_CONFIG environment variable to specify the path, or if you really must use the command line arguments, import the config manually based on them, and pass it to init.');
```
B4nan, line 556: seeing this, I would change the label to be just `D0001`

and I would personally also drop the `This is deprecated` bit, but no strong opinions about that

tests/features/cli/CLIHelper.test.ts
```diff
@@ -545,4 +535,39 @@ Maybe you want to check, or regenerate your yarn.lock or package-lock.json file?
...
+    messages.length = 0;
+
+    (global as any).process.argv = ['node', 'start.js', '--config=./override2/orm-config.ts'];
+    expect(ConfigurationLoader.configPathsFromArg()).toEqual(['./override2/orm-config.ts']);
+    await MikroORM.init();
+    expect(messages[0]).toBe('[deprecated] (MikroORM:0001) Path for config file was inferred from the command line arguments. This is deprecated. Set the MIKRO_ORM_CLI_CONFIG environment variable to specify the path, or if you really must use the command line arguments, import the config manually based on them, and pass it to init.');
+    messages.length = 0;
+
+    (global as any).process.argv = ['npx', 'mikro-orm', 'debug', '--config', './override3/orm-config.ts'];
```
B4nan, line 565: why is this printing the deprecation when its going through the CLI?

B4nan (commented):
i dont see any changes to the ORM options, that's where the `ignoreDeprecations` need to be. also this should be documented somewhere (the `ignoreDeprecations` option).

tests/features/cli/CLIHelper.test.ts
boenrobot: Done (both the label scheme and replacing "This is deprecated. Set" with "Instead,"set" instead.

tests/features/cli/CLIHelper.test.ts
boenrobot: The `ConfigurationLoader.configPathsFromArg()` doesn't care what the binary's name is, and the actual mikro-orm CLI never calls it.

I've now adjusted the test to call "npm start" instead of "npx mikro-orm debug" to avoid the confusion.

B4nan (approved):

B4nan:
One last thing is the docs for the `ignoreDeprecations` option, not sure where it fits better, either https://mikro-orm.io/docs/logging or https://mikro-orm.io/docs/configuration (or both)

docs/docs/configuration.md
```diff
@@ -581,6 +581,18 @@ MikroORM.init({
...
+
+You can ignore all or only specific deprecation warnings. See [Logging's section on deprecation warnings](./logging.md#deprecation-warnings) for details.
+
+The full list of deprecation errors:
+
+| label | message                                                                                                                                                                                                                                                                            |
+|-------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| D001  | Path for config file was inferred from the command line arguments. Instead, you should set the MIKRO_ORM_CLI_CONFIG environment variable to specify the path, or if you really must use the command line arguments, import the config manually based on them, and pass it to init. |
```
B4nan, line 594:
```suggestion
| D001  | Path for config file was inferred from the command line arguments. Instead, you should set the `MIKRO_ORM_CLI_CONFIG` environment variable to specify the path, or if you really must use the command line arguments, import the config manually based on them, and pass it to init. |
```

docs/docs/logging.md
```diff
@@ -49,10 +49,34 @@ return MikroORM.init({
...
+
+Even without `debugMode` enabled, the default logger will show `deprecated` messages in console.
+
+When something is deprecated, it means there is an intention for it to be removed in a future version. The deprecation message should suggest alternatives for you that you should switch to before migrating to a major version.
+
+You can ignore all deprecation warnings by setting `ignoreDeprecations` to `true`
+
+```ts
+return MikroORM.init({
+    ignoreDeprecations: true, // now no deprecations will be logged, though you may be surprised when upgrading
```
B4nan, line 66:
```suggestion
  ignoreDeprecations: true, // now no deprecations will be logged, though you may be surprised when upgrading
```

docs/docs/logging.md
```diff
@@ -49,10 +49,34 @@ return MikroORM.init({
...
+});
+```
+
+When you are actively trying to remove deprecation warnings in preparation for an upgrade, you would likely want to tackle them one at a time. You can ignore only specific deprecation warnings you can't deal with right now, while still being alerted of others you didn't know about. To do this, list the deprecation warnings you want to ignore, e.g.
+
+```ts
+return MikroORM.init({
+    ignoreDeprecations: ['D0001'], // ignore deprecation with label "D0001", but show others if they pop up
```
B4nan, line 74:
```suggestion
  ignoreDeprecations: ['D0001'], // ignore deprecation with label "D0001", but show others if they pop up
```

B4nan (approved):
nice, that looks much better than what I would write myself :]

B4nan:
Something from this PR is not closing the opened connections, I'd guess the test which calls `MikroORM.init`.

tests/EntityManager.mongo2.test.ts
```diff
@@ -29,6 +29,17 @@ describe('EntityManagerMongo2', () => {
...
+    const commandMock = jest
+      .spyOn(orm.config.getDriver().getConnection().getDb(), 'command')
+      .mockReturnValue(Promise.resolve({ error: 'boom!' }));
+    expect(await orm.isConnected()).toBe(false);
+    expect(await orm.checkConnection()).toEqual({
+      ok: false,
+      reason: 'Ping reply does not feature "ok" property, or it evaluates to "false"',
+    });
+    expect(commandMock).toHaveBeenCalledTimes(2);
+    commandMock.mockRestore();
```
B4nan, line 42: this is the part that needs some additional cleanup

> A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.

B4nan: ce2f19671efe7fdec961ccb3d7d57d973037d3f6

PR #6186, opened 2024-10-23
fix(core) : fix primary key serialization when a custom type is involved

### LINKED ISSUE #6179: Serialization with forceObject: true does not apply primary key custom type
### Describe the bug

Serialization using `forceObject: true` doesn't apply custom type on foreign key. 

### Reproduction

```ts
import { MikroORM, serialize, Entity, ManyToOne, Platform, PrimaryKey, Property, Ref, Type } from '@mikro-orm/sqlite';

export class IDStoredAsInteger extends Type<string, number> {

  convertToDatabaseValue(value: string): number {
    return parseInt(value);
  }

  convertToJSValue(value: number): string {
    return `${value}`;
  }

  getColumnType() {
    return `integer`;
  }

}

@Entity()
class User {

  @PrimaryKey({ type: IDStoredAsInteger })
  id!: string;

  @Property({ unique: true })
  email!: string;

  @ManyToOne(() => User, { nullable: true })
  bestFriend?: Ref<User>;

}

let orm: MikroORM;

beforeAll(async () => {
  orm = await MikroORM.init({
    dbName: ':memory:',
    entities: [User],
  });
});

afterAll(async () => {
  await orm.close(true);
});

it('should apply custom type when serializing primary key of non-populated relation', async () => {
  const totoUser = orm.em.create(User, {
    email: 'toto@test.com',
  });
  orm.em.create(User, {
    email: 'yolo@test.com',
    bestFriend: totoUser,
  });
  await orm.em.flush();

  const u1 = await orm.em.fork().findOneOrFail(User, { email: 'yolo@test.com' });
  const dto1 = serialize(u1);
  expect(typeof dto1.id).toBe('string');
  expect(dto1.bestFriend).toBeDefined();
  expect(typeof dto1.bestFriend).toBe('string');

  const u2 = await orm.em.fork().findOneOrFail(User, { email: 'yolo@test.com' }, { populate: ['bestFriend'] });
  const dto2 = serialize(u2);
  expect(typeof dto2.id).toBe('string');
  expect(dto2.bestFriend).toBeDefined();
  expect(typeof dto2.bestFriend).toBe('string');
});
```

@mikro-orm/postgresql

### MikroORM version

6.3

### Node.js version

v20.11.1

### Operating system

Macos

### Validations

leonardbinet:
Closes #6179


packages/core/src/serialization/EntitySerializer.ts
```diff
@@ -229,7 +229,7 @@ export class EntitySerializer {
       return this.serialize(child, childOptions) as EntityValue<T>;
     }
 
-    const pk = wrapped.getPrimaryKey(true)!;
+    const pk = wrapped.getPrimaryKey(false)!;
```
B4nan, line 232: the very same issue will be in implicit serialization, lets alter the test to try that too and fix that too, its in `EntityTransformer` class

```suggestion
    const pk = wrapped.getPrimaryKey()!;
```

B4nan:
It was indeed a bit more complicated 🙃

leonardbinet:
ahah indeed 😅 , thx a lot for the fix 🙏

PR #6188, opened 2024-10-24
feat(core): implement BaseSchema of standard-schema

xcfox:
This PR makes Mikro's EntitySchema implement [BaseSchema](https://github.com/standard-schema/standard-schema/blob/main/packages/spec/src/base.ts) of [standard-schema](https://github.com/standard-schema/standard-schema)

In the discussion in the [last PR](https://github.com/mikro-orm/mikro-orm/pull/6147), I realized that a full implementation of standard-schema is not necessary, and that in third-party libraries we only need to be able to infer the type of Entity Schema.
In most scenarios, the Entity Schema will be used as the output of the request, and validation of the output is not necessary, so we don't need to implement the `~validate` method.

In this PR, I just added a few lines of necessary metadata and types, and didn't add or change anything for functionality.

## Infer Entity

After implementing BaseSchema, a third library can integrate EntitySchema on types and infer the type of Entity through the `InferOutput` generic: 

```ts
import type { InferOutput } from '@standard-schema/spec'

export const User = new EntitySchema({ })

export interface IUser extends InferOutput<typeof User> {}
```

## In tRPC

A BaseSchema can be used directly in third-party libraries like tRPC without the need for wrappers like `wrap()`, which dramatically reduces boilerplate code.

```ts
export const postRouter = router({
  add1: authedProcedure
    .input(
      z.object({
        id: z.string().uuid().optional(),
        text: z.string().min(1),
      })
    )
    .output(Post) // tRPC will infer output by `Post` and TypeScript will check if the mutation's return value matches the Post.
    .mutation(async ({ input, ctx }) => {
      const { name } = ctx.user
      const post = await em.create(Post, { ...input, name })
      return post // the reutrn value is OK
    }),

  add2: authedProcedure
    .output(Post)
    .mutation(async () => {
      const user = await em.create(User, { ... })
      return user // the return value is not math `Post`, TypeScript will give you the alert.
    }),
})
```


xcfox:
I understand that a warp function would make Entity Schema more extensible.
At the same time, I also think it would be useful to be able to use EntitySchema directly.

```ts
export const postRouter = router({
  add1: authedProcedure
    .input(
      z.object({
        id: z.string().uuid().optional(),
        text: z.string().min(1),
      })
    )
    .output(Post)
    .mutation(async ({ input, ctx }) => {
      const { name } = ctx.user
      const post = await em.create(Post, { ...input, name })
      return post // the reutrn value is OK
    }),

  myPosts: authedProcedure.output(loaded(Post).with("author")).query(async ({ ctx }) => {
    const posts = await em.find(Post, { name: ctx.user.name, populate: ["author"] })
    return posts
  }),
})
```

> the code in this PR seems incomplete as it is now.

I think these few lines of code in the current PR are enough for a third library to easily integrate EntitySchema.

In the case of GQLoom, for example, it already has `silk` utils, accepts `BaseEntity` as a parameter, and in addition to query and mutation, EntitySchema can easily be used as a parameter to `resolver.of()` (which means that we can provide additional relationships and fields for that Entity in GraphQL)：

```ts
import { silk, resolver, field, query } from "@gqloom/core"
import { EntitySchema } from "@mikro-orm/core"
import * as v from "valibot"

const User = new EntitySchema<IUser>({})

const Cat = v.object({})

const catResolver = resolver.of(Cat, {
  cats: query(silk.list(Cat), () => {
    return []
  }),

  keeper: field(silk.nullable(User), (cat) => {
    return em.findOne(User, cat.keeperId)
  }),
})

const userResolver = resolver.of(User, {
  users: query(silk.list(User), () => {
    return em.find(User)
  }),
})
```

xcfox:
Thanks for your reminder, I've updated the code and now EntitySchema's `~types` complies with `wrap().serialize()`

The adaptation of tRPC to Standard Schema is still [in progress](https://github.com/trpc/trpc/pull/6079).

xcfox:
Fairly.
I think this will be quick, and it will be easy to implement `~validate` considering EntitySchema's `~types` follow the `wrap().serialize()` signature, and by then, we should implement the full standard-schema.

B4nan:
As with everything that is not very clear (and this thing is far away from it), I would like to see that born as a 3rd party package first, proving its design and usability. I really don't want to end up in a state where things will change somewhere (the spec itself, trpc, ...), and to adapt we would have to do something breaking.

PR #6201, opened 2024-10-30
feat(core): allow defining multiple named configurations in one config file

### LINKED ISSUE #6067: First class multi database setups

It's not uncommon for a single application to need to use multiple different database connections. Common use cases include using a different connection for executing DB migrations vs rest of the app and multi tenant setups (in the "one DB per tenant" setup that is). Perhaps less common, but still found in more complex apps are cases where certain separate connection is used for a subset of features (logging, event tracking, etc. ; specialized solutions exist, but can be implemented in DBs as well).

While MikroORM's RequestContext supports this, the mikro-orm.config.ts file does not, meaning such setups are not easily manageable by the CLI. And beyond that, framework integrations should also be allowing multi database setups, and ideally type the driver properly, but they can not easily do that without more user input, making integration more painful.

Allow mikro-orm.config.ts file to export an array of ORM config objects, in addition to a single ORM config object.

Let the CLI specify a config name in a separate option to operate it on, looking it up at the already existing "name" option. It would be error in CLI if names repeat in the array. Default to the name "default" for compatibility with the RequestContext.

Having this supported by the CLI and ConfigurationLoader, and documented as such will in turn enable framework integrations to rely on it for multi database setups. In frameworks that have separate type gen steps (like Nuxt and Astro), integrations with MikroORM can generate the appropriate types for all the entity managers based on that config. Framework integrations (in general, i.e. also including f.e. NestJS) may further expose a proxy of the config object to enable lazy loading of additional configs, which would be a big blessing in a multi tenant setup.

In the case of the CLI, we can already specify a custom config file path, which means one can have a config file per connection.

In the case of framework integrations, they need a path to a config file that may also be an array of paths to config files instead... or some custom map of config objects... but relative file path lookup in general is a vector of complexity for users ("what is that path relative to? does that file even exist at the time I'm trying to read it?"), so having a fixed location where all the config is found would be easier on everybody.

**Additional context**

I've been playing around with trying to make a sample app with Astro, and while I'm already successful, I wanted to make it as simple as adding an astro "integration" that searches for the config at build time, and takes care of providing a forked EM wherever the request is available. Further, I previously made a MikroORM Nuxt module, and I already got a feature request about it not requiring driver imports everywhere, which is a reasonable request that I can accomplish (in a similar fashion to Astro)... If it wasn't for the fact that multiple connections are a thing, meaning some user level complexity is still required.

I think I can do both Nuxt and Astro integrations that are able to infer more and lead to a more pleasant DX... But I do need the multi DB setup somehow, as I will need it in my own applications (in my case, we have a dedicated server for logging... and also dedicated credentials for migrations).

boenrobot:
A config file may now export a function or an array.

When a function is exported, it is called with a runtime supplied configuration name
("default" by default), and is expected to return a config object,
or a promise resolving to a configuration object.

When an array is supplied, it is searched for a config object
with a "name" option matching the user supplied one.
Not having that option set is the same as setting it to "default".

If the array does not contain a config with that name,
it is analyzed for functions,
and each is executed from top to bottom with the name as argument.
The first object returned is used as the final config.
If no function returns an object, or no functions are available,
an error is thrown.

Closes #6067


B4nan (commented):
I guess it looks fine (although I dislike the 3 level nesting in that huge condition).

> and both options feel wrong, for different reasons.

Can you list some of those? If there is a config file, it should return something, right?

tsconfig.build.json
```diff
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
     "module": "Node16",
-    "target": "es2022",
-    "lib": ["es2022"],
+    "target": "es2023",
+    "lib": ["es2023"],
```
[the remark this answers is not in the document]
boenrobot: For the sake of [Array.prototype.findLastIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLastIndex), which is available since Node 18. There's no major version bump required, so just bumping the library to what Node 18 already supports seems a no brainer.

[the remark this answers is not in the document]
B4nan: Yeah but for that you need the `lib`, not the `target`. When I check the docs, they say the target should be es2022, but lib can be es2023.

https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping#node-18

I am just a bit afraid about the change of the target.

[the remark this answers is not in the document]
B4nan: I guess it shouldnt really matter, there should be no syntactic changes between es2022 and es2023, but still, I'd rather change just the lib.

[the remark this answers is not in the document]
boenrobot: The wiki was probably written very early on... Because the es2023 lib was added first, and the target was only later added...

The target is available since TypeScript 5.5:
https://github.com/microsoft/TypeScript/commit/551a600e84a8be35b25a758119f86731463d6a53

While the lib existed before then. Either way, MikroORM is using TypeScript 5.6, so it doesn't have to worry about the target maybe not being present.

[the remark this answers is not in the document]
B4nan: I am not worried about our build not working, I am worried about what it means for downstream users (but as I said, it shouldn't really matter since there were no syntactic changes apparently).

Using `lib` is all you need to have the new methods, the target shouldn't be relevant here, right?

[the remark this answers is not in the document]
boenrobot: sure... ok, I'll change the target back, keeping just the lib.

B4nan (commented):
It would be nice if those PRs start with some docs (can be just a few examples in the PR description) instead of having docs as the last step, it's all still very abstract to me. It sounds like you consider allowing multiple config files for one context, which to me is a nongoal, and it wouldn't work well with the current setup, where you can/should provide multiple config paths, TS first, JS last, and we use the first valid one - thanks to this you don't have to compile your app to be able to use the CLI. We definitely cannot merge TS and JS configs, that could result in discovering the same entity from JS and TS files, ending up with validation errors on init, or worse, with weird behavior/errors on runtime. I am also still unsure how all of this would work for those who don't use config files and depend solely on the env vars (so those without a config file).

If those two problems are covered, I'll trust you on picking the right approach here. Allowing the functions everywhere sounds good to me.

packages/cli/src/CLIConfigurator.ts
```diff
@@ -36,6 +36,11 @@ export class CLIConfigurator {
         array: true,
         desc: `Set path to the ORM configuration file`,
       })
+      .option('configName', {
+        type: 'string',
+        desc: 'Set name of config to load out of the ORM configuration file. Used when config file exports an array or a function',
+        default: process.env.MIKRO_ORM_NAME ?? 'default',
```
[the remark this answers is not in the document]
boenrobot: I was following the convention of other env variables that override the config, where it's "MIKRO_ORM_{propNameInConfig}", but yeah, it felt weird to call it just "name" to me too... Changed to "MIKRO_ORM_CONFIG_NAME" now.

docs/docs/configuration.md
```diff
@@ -619,6 +619,7 @@ Full list of supported options:
 
 | env variable                                                | config key                               |
 |-------------------------------------------------------------|------------------------------------------|
+| `MIKRO_ORM_CONFIG_NAME`                                     | `name`                                   |
```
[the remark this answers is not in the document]
boenrobot: Huh... I was pretty sure it's also used at RequestContext, and yet now I see there is a separate "contextName" option that is actually what gets into the requst context map key. Maybe it was changed recently, and I didn't notice.

My thought was that there's no need for a new option, but the one in use for the request context at the root config is what should be used, so that ultimately, when the config is an array, it may implicitly give an EntityManager array, and thus fork all ORM instances in the array (other than the factory function ones I guess).

So yeah, maybe the config name can be taken from "contextName" instead, and still default to "default"?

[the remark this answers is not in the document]
B4nan: > Maybe it was changed recently, and I didn't notice.

It has been like this since the beginning.

> So yeah, maybe the config name can be taken from "contextName" instead, and still default to "default"?

Agreed, the context name was added for the same use case I believe.

docs/docs/quick-start.md
```diff
@@ -293,24 +299,100 @@ Another way to control these CLI-related settings is with the environment variab
 - `MIKRO_ORM_CLI_ALWAYS_ALLOW_TS`: enable `.ts` files to use without ts-node
 - `MIKRO_ORM_CLI_VERBOSE`: enable verbose logging (e.g. print queries used in seeder or schema diffing)
 
-Alternatively, you can also specify the config path via `--config` option:
+MikroORM will always try to load the first available config file, based on the order in `configPaths`. When you have `useTsNode` disabled or `ts-node` is not already registered nor detected, TS config files will be ignored.
```
B4nan, line 302:
```suggestion
MikroORM will always try to load the first available config file, based on the order in `configPaths`. When you have `useTsNode` explicitly disabled or `ts-node` is not already registered nor detected, TS config files will be ignored.
```

docs/docs/quick-start.md
```diff
@@ -293,24 +299,100 @@ Another way to control these CLI-related settings is with the environment variab
...
+You can also specify the config path via `--config` option:
 
 ```sh
 $ npx mikro-orm debug --config ./my-config.ts
 ```
 
-The `--config` flag will be respected also when you run your app (as long as it is part of `process.argv`), not just when you use the CLI. 
+Since v6.3, the CLI will always try to use TS config file, even without explicitly enabling it via `useTsNode` flag in your `package.json` file. You can still use it to disable the TS support explicitly. Keep in mind that having `ts-node` installed is still required for the TS support to work. The `useTsNode` has effect only on the CLI.
+
+Your configuration file may export multiple configuration objects in an array. The different configurations must heve a "name" in them. If no name is specified, it is treated as the name "default". You can use the `MIKRO_ORM_CONFIG_NAME` environment variable or the `--configName` command line option to pick a configuration with a particular name to use for the CLI. See [below](#configuration-file-structure) for details on the config object.
```
B4nan:
```suggestion
Your configuration file may export multiple configuration objects in an array. The different configurations must have a `contextName` option. If no context name is specified, it is treated as the name "default". You can use the `MIKRO_ORM_CONFIG_NAME` environment variable or the `--configName` command line option to pick a configuration with a particular name to use for the CLI. See [below](#configuration-file-structure) for details on the config object.
```

i also wonder if we should change/unify the `configName` and corresponding env var to use the existing `contextName` instead?

docs/docs/quick-start.md
```diff
@@ -293,24 +299,100 @@ Another way to control these CLI-related settings is with the environment variab
...
+
+Options:
+      --config      Set path to the ORM configuration file               [array]
+      --configName  Set name of config to load out of the ORM configuration
+                    file. Used when config file exports an array or a function
+                                                   [string] [default: "default"]
+  -v, --version     Show version number                                [boolean]
+  -h, --help        Show help                                          [boolean]
+  -v, --version  Show version number                                   [boolean]
+  -h, --help     Show help                                             [boolean]
```
B4nan: why is this part here twice?

docs/docs/quick-start.md
```diff
@@ -293,24 +299,100 @@ Another way to control these CLI-related settings is with the environment variab
...
+
+Examples:
+  mikro-orm schema:update --run  Runs schema synchronization
+```
+
+To verify your setup, you can use `mikro-orm debug` command.
+
+## Running MikroORM.init() without arguments
```
B4nan, line 363: IMO this really doesn't belong to quick start, i would leave here just the first sentence and refer to the dedicated CLI guide

docs/docs/quick-start.md
```diff
@@ -293,24 +299,100 @@ Another way to control these CLI-related settings is with the environment variab
...
+
+When you have CLI config properly set up, you can omit the `options` parameter when calling `MikroORM.init()` in your app. Configuration is loaded similarly to how it is loaded when using the MikroORM CLI.
```
B4nan:
```suggestion
When a CLI config is properly set up, you can omit the `options` parameter when calling `MikroORM.init()` in your app. Configuration is loaded similarly to how it is loaded when using the MikroORM CLI.
```

docs/docs/quick-start.md
```diff
@@ -293,24 +299,100 @@ Another way to control these CLI-related settings is with the environment variab
...
+
+The `--config` flag from the command line will be respected also when you run your app (as long as it is part of `process.argv`), not just when you use the CLI.
```
B4nan, line 367: we should note here that this is now deprecated

docs/docs/quick-start.md
```diff
@@ -293,24 +299,100 @@ Another way to control these CLI-related settings is with the environment variab
...
+
+```sh
+$ node ./dist/index.js -- --config ./my-orm-config.js
```
B4nan, line 370: do we really need the `--` in here? i know its not needed for `npx` or all other package managers except `npm`, and i would guess its not needed for running `node` directly either

docs/docs/quick-start.md
```diff
@@ -293,24 +299,98 @@ Another way to control these CLI-related settings is with the environment variab
...
+```
+
+This might introduce a conflict with other tools like `jest` that also support overriding the config path via `--config` argument. In those cases you can use the `MIKRO_ORM_CONFIG_ARG_NAME` environment variable to change the argument name to something other than `config`:
 
 ```sh
 $ MIKRO_ORM_CONFIG_ARG_NAME=mikro-orm-config \
-  npx mikro-orm debug --mikro-orm-config ./my-config.ts
+  node ./dist/index.js -- --mikro-orm-config ./my-orm-config.js
```
B4nan, line 377: same

docs/docs/quick-start.md
```diff
@@ -293,24 +299,98 @@ Another way to control these CLI-related settings is with the environment variab
...
 ```
 
-> `jest` does not allow unrecognised parameters, to run tests with a custom configuration you can use this together with `MIKRO_ORM_CLI_CONFIG` environment variable to point to a test config. 
+> `jest` does not allow unrecognised parameters, to run tests with a custom configuration you can use this together with `MIKRO_ORM_CLI_CONFIG` environment variable to point to a test config.
 
-MikroORM will always try to load the first available config file, based on the order in `configPaths`. When you have `useTsNode` disabled or `ts-node` is not already registered nor detected, TS config files will be ignored.
+> Currently, `process.argv` is analyzed for backwards compatibility. This will be removed in v7. Using `--config` with the MikroORM CLI will still be available.
+
+By default, TS config files are not considered when running your app. You can use the `alwaysAllowTs` option in your `package.json` file, which will enable checking the TS files even for your actual app, as well as in the MikroORM CLI. This can be handy if you run your app via [Bun](https://bun.sh).
```
B4nan, line 384: > By default, TS config files are not considered when running your app

this is no longer true, they are considered by default as long as you have `ts-node` installed

B4nan (commented):
few notes about the docs, haven't read the whole thing yet

docs/docs/configuration.md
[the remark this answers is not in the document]
boenrobot: OK, changed all the "name" and "configName" introduced here to "contextName", along with the env also now being "MIKRO_ORM_CONTEXT_NAME" to match the convention, while still being explicit.

docs/docs/quick-start.md
boenrobot: I just did that too in the latest commit, lol

docs/docs/quick-start.md
boenrobot: Fixed already in the latest commit... I was pasting that portion out of my CLI output, but I pasted wrong the first time around I guess.

docs/docs/quick-start.md
boenrobot: I guess we don't... We would if node adds a "--config" option of its own, but then again, that would be a problem with or without "--", since the entire argv is passed to the file, and there's no scanning for "--" in the currently deprecated logic.

docs/docs/quick-start.md
```diff
@@ -264,13 +264,19 @@ $ npx mikro-orm
...
-
-Since v6.3, the CLI will always try to use TS config file, even without explicitly enabling it via `useTsNode` flag in your `package.json` file. You can still use it to disable the TS support explicitly. Keep in mind that having `ts-node` installed is still required for the TS support to work. The `useTsNode` has effect only on the CLI. Alternatively, you can use the `alwaysAllowTs` option in your `package.json` file, which will enable checking the TS files even for your actual app and not just the CLI (in case you call `MikroORM.init()` without any parameters). This can be handly if you run your app via [Bun](https://bun.sh).
+By default, the following paths, relative to the current working directory, are searched in this order:
+```
+./src/mikro-orm.config.ts
+./mikro-orm.config.ts
+./dist/mikro-orm.config.js
+./build/mikro-orm.config.js
+./src/mikro-orm.config.js
+./mikro-orm.config.js
```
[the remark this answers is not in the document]
boenrobot: Changed to an ordered list.

And yeah, I guess adding cjs/mjs/cts/mts would be nice... It would be a relatively easy addition even in this PR, if you'd like...

IMHO, the order should probably be folder first, extension second, with extensions being in order of "esm", "cjs", "default", i.e.

```
./src/mikro-orm.config.mts
./src/mikro-orm.config.cts
./src/mikro-orm.config.ts
./mikro-orm.config.mts
./mikro-orm.config.cts
./mikro-orm.config.ts
./dist/mikro-orm.config.mjs
./dist/mikro-orm.config.cjs
./dist/mikro-orm.config.js
./build/mikro-orm.config.mjs
./build/mikro-orm.config.cjs
./build/mikro-orm.config.js
./src/mikro-orm.config.mjs
./src/mikro-orm.config.cjs
./src/mikro-orm.config.js
./mikro-orm.config.mjs
./mikro-orm.config.cjs
./mikro-orm.config.js
```

docs/docs/quick-start.md
boenrobot: Does that then mean

> The `useTsNode` has effect only on the CLI.

Is also no longer true? That nowadays, you can set `useTsNode` to also disable TS configs in your app, not only in CLI?

docs/docs/quick-start.md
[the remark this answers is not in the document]
B4nan: yeah agreed on the order. but cant say i would like to have this extensive list in the docs :]

docs/docs/quick-start.md
B4nan: The app itself will never register ts-node loader itself, only CLI is doing that, that's because people can leave the ts-node dependency in production builds too for whatever reason, and we don't want to use TS files in there.

boenrobot: Ah, so it's not about whether `ts-node` is _installed_, it's about whether its loader is _enabled_ into the current node process?

And it just so conveniently happens that the MikroORM CLI does load it if it's installed... So if users do the same in their app's startup process, they'd also have ts files considered... but if they don't, they don't get ts config files?

docs/docs/quick-start.md
```diff
@@ -264,13 +264,19 @@ $ npx mikro-orm
 $ yarn mikro-orm
 ```
 
-For CLI to be able to access your database, you will need to create `mikro-orm.config.js` file that exports your ORM configuration.
+For CLI to be able to access your database, you will need to create a configuration file that exports your ORM configuration(s).
```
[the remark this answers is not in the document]
boenrobot: > maybe its finally time to create a dedicated CLI guide?

Agree in general... For this PR, I just wanted to first split what was already on this page, but is CLI only vs app only vs CLI+app, as the docs are very much conflating the two.

docs/docs/quick-start.md
boenrobot: I've adjusted the wording there to say "unless loaded via `ts-node`" which I guess is the least confusing way to put it.

B4nan: Well, it depends. For ESM projects, the loader is needed, and that is what the `mikro-orm-esm` CLI endpoint (or how to call it) is doing. For CJS projects its fine to call `require('ts-node').register()` and that is what the ORM will do for you unless you explicitly disable `useTsNode` in the package.json. This is happening only for the CLI, not for apps. For apps you are responsible to call it yourself, e.g. by replacing `node main.js` with `ts-node main.ts` or similar.

B4nan: This is the only place where we try to register ts-node:

https://github.com/mikro-orm/mikro-orm/blob/master/packages/cli/src/CLIConfigurator.ts#L51

boenrobot: Hm... I think things might be made less confusing if `ts-node` is a required dependency of `@mikro-orm/cli`. `@mikro-orm/cli` itself is (or at least should be) usually installed as a dev dependency anyway.

That would let the docs to simply say "MikroORM CLI always considers .ts files, unless explicitly disabled by setting `useTsNode` to `false`", along with later saying "By default, TS config files are not considered when running your app, unless you run it via `ts-node`. You can set `useTsNode` to `false` to disable TS config files, even if the app is ran with `ts-node`."

For now, I think the latest wording:

> By default, TS config files are not considered when running your app, unless you run it via `ts-node`. You can use the `alwaysAllowTs` option in your `package.json` file, which will enable checking the TS files even if `ts-node` is not used, as well as in the MikroORM CLI. This can be handy if you run your app via [Bun](https://bun.sh).

is informative enough, while not being too confusing. I'm not sure how to put it better without going into unnecessary technicalities that the vast majority of users won't need to care about.

B4nan: > Hm... I think things might be made less confusing if ts-node is a required dependency of @mikro-orm/cli

You can use the CLI with compiled files, e.g. in production environments. I would rather not make the CLI dependent on TS.

I know this is rather an edge case, but I believe there are people doing this, e.g. to run compiled migrations in production.

docs/docs/quick-start.md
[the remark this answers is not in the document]
B4nan: yeah we can leave that for a separate PR

docs/docs/quick-start.md
boenrobot: Hm... fair... Not an untypical scenario indeed. Not sure if "optional dependency" would do it then...

I mean, it would work for the 99% case, since npm and other package managers install optional dependencies by default, but in the scenario above, it also means we'd force users to use `--omit=optional` (or the equivalent for other package managers) when deploying to production + explicitly add the optional dependencies they do need. That's a good optimization to have regardless, but do we want to force it on users is the real question.

B4nan: I'd rather leave it like it is now, especially because there are alternatives to ts-node, and in modern runtimes like `bun` you don't even need this since it's supported out of box.

docs/docs/quick-start.md
```diff
@@ -341,57 +424,104 @@ const config: Options = {
...
-To verify your setup, you can use `mikro-orm debug` command.
+then you will need to start your app with the `MIKRO_ORM_CONTEXT_NAME` environment variable set to `example1` to load the database `tenant_example1`, and similarly, when running CLI, you can use
+
+```sh
+$ npx mikro-orm --contextName=example1
+```
+
+to operate on that particular tenant's database instance. Not specifying either option will point you to the "tenant_default" database.
+
+You can also combine arrays and factory functions. Array members will be preferred, and any functions in the array will be executed from top to bottom. The first function to return an object will be what ends up being used.
```
B4nan, line 478: this is nice, quite powerful 👍

docs/docs/quick-start.md
boenrobot: Ok... And any suggestions on how to reword this from the latest phrasing?

packages/cli/src/CLIConfigurator.ts
```diff
@@ -36,6 +36,11 @@ export class CLIConfigurator {
         array: true,
         desc: `Set path to the ORM configuration file`,
       })
+      .option('contextName', {
```
B4nan, line 39: I wonder if this should be named just `context` instead of `contextName` (keeping the ORM option and env var intact). Thoughts?

packages/core/src/utils/ConfigurationLoader.ts
```diff
@@ -15,34 +15,85 @@ import { Utils } from './Utils';
...
-        if (tmp instanceof Function) {
-          tmp = tmp();
-        }
+        let tmp: unknown = config.default ?? config;
 
         if (tmp instanceof Promise) {
           tmp = await tmp;
         }
 
+        if (Array.isArray(tmp)) {
```
B4nan, line 42: would be great to introduce some private helpers to simply the code, its both quite long and very nested now. e.g. the function detection + invocation + possible promise handling seems to be repeated

packages/core/src/utils/ConfigurationLoader.ts
```diff
@@ -15,34 +15,85 @@ import { Utils } from './Utils';
...
+            tmp = tmp(contextName);
+
+            if (tmp instanceof Promise) {
+              tmp = await tmp;
+            }
+            if (!isValidConfigFactoryResult(tmp)) {
+              throw new Error(`MikroORM config '${contextName}' was not what the function exported from '${path}' provided. Ensure it returns a config object with no name, or name matching the requested one.`);
+            }
+          } else {
+            if (!(configFinder(tmp))) {
```
B4nan, line 83:
```suggestion
            if (!configFinder(tmp)) {
```

docs/docs/quick-start.md
B4nan: I think the current version is good

packages/cli/src/CLIConfigurator.ts
boenrobot: I think keeping it to "contextName" would be better, as it makes it more searchable. The word "context" alone is too... context dependent (pun intended).

B4nan: But what else could `--context` refer to than a context name? :]

boenrobot: Right, but I mean, trying to search more information about context names, and why they're present in the CLI would give me too much noise if I just search "context" (and many search engines would kind of treat "--context" as being just "context"), whereas "contextName" would maybe include some mention about RequestContext that would be a bit of a noise in this case... but the majority of results would be referring to the env and config option, which in turn explain the multi config process.

packages/core/src/utils/ConfigurationLoader.ts
boenrobot: The promise thing in particular can maybe be simplified to just unconditionally await...

i.e.

```
          if (tmp instanceof Function) {
            tmp = await tmp(contextName);
```

and analogously in the other spots.

Not quite sure about the function detection, as there's a difference between it being an array and not being an array.

packages/cli/src/CLIConfigurator.ts
B4nan: Maybe it could be an alias? I guess it's not a huge deal, but it feels more visually pleasing to have just `npx mikro-orm --context foo`. But I would only do this if we want to change the docs to use this short name, which I believe you would also rather not do, for discoverability reasons?

boenrobot: > But I would only do this if we want to change the docs to use this short name, which I believe you would also rather not do, for discoverability reasons?

yes, exactly

A non-default alias or other shorthand for readability's sake is fine, but the longer one being searchable is also important IMO.

B4nan: I would still maybe add the alias, even if not used in the docs.

B4nan (approved):
all right, lets roll! 

i guess i will ship stable later today or tomorrow morning, just need to finish the blog post. could you maybe prepare a few sentences about this feature i can add there? nothing huge, i'd just link the docs from there about the advanced things

B4nan:
Shall we close #6067 with this or do you envision more work in that direction?

B4nan:
It's a blog post, it shouldn't add anything that is not in the docs, it should promote the new feature, linking to the docs for more context. I'd mention what it is about, that it opens new ways for multi-tenant setups, that there will be integrations building on top of this in the near future, and one small from the docs. I often just copy (part of) the docs in there and reword a bit.

PR #6213, opened 2024-11-06
refactor: make `ConfigurationLoader.getConfiguration()` backwards compatible

boenrobot:
If true is explicitly given at the first argument, it will trigger the legacy behavior, where `--config` is also searched for in `process.argv`.

If left without arguments, the new default is used, which is still mostly compatible, except `--config` is not searched in `process.argv`.

In either situation, the paths to search for default to the result of `ConfigurationLoader.getConfigPaths()`.

### DIFF
--- a/packages/core/src/utils/ConfigurationLoader.ts
+++ b/packages/core/src/utils/ConfigurationLoader.ts
@@ -15,7 +15,42 @@ import { Utils } from './Utils';
  */
 export class ConfigurationLoader {
 
-  static async getConfiguration<D extends IDatabaseDriver = IDatabaseDriver, EM extends D[typeof EntityManagerType] & EntityManager = EntityManager>(contextName = 'default', paths: string[], options: Partial<Options> = {}): Promise<Configuration<D, EM>> {
+  /**
+   * Gets a named configuration from a set of paths
+   *
+   * @param contextName Name of context to load from the config. Used when config file exports array or factory function.
+   * @param paths Array of possible paths for a configuration file. Files will be checked in order, and the first existing one will be used.
+   * @param options Additional options to augment the final configuration with.
+   */
+  static async getConfiguration<D extends IDatabaseDriver = IDatabaseDriver, EM extends D[typeof EntityManagerType] & EntityManager = EntityManager>(contextName: string, paths: string[], options?: Partial<Options>): Promise<Configuration<D, EM>>;
+  /**
+   * Gets a named configuration from a set of paths
+   *
+   * @param contextName Name of context to load from the config. Used when config file exports array or factory function. Defaults to "default".
+   * @param paths Array of possible paths for a configuration file. Files will be checked in order, and the first existing one will be used. Defaults to the output of {@link ConfigurationLoader.getConfigPaths}
+   * @param options Additional options to augment the final configuration with.
+   *
+   * @deprecated Prefer to explicitly set the parameter values. The defaults are available for backwards compatibility.
+   */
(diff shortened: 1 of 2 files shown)


packages/core/src/utils/ConfigurationLoader.ts
```diff
@@ -15,7 +15,42 @@ import { Utils } from './Utils';
...
+  /**
+   * Gets a named configuration from a set of paths
+   *
+   * @param contextName Name of context to load from the config. Used when config file exports array or factory function. Defaults to "default".
+   * @param paths Array of possible paths for a configuration file. Files will be checked in order, and the first existing one will be used. Defaults to the output of {@link ConfigurationLoader.getConfigPaths}
+   * @param options Additional options to augment the final configuration with.
+   *
+   * @deprecated Prefer to explicitly set the parameter values. The defaults are available for backwards compatibility.
+   */
+  static async getConfiguration<D extends IDatabaseDriver = IDatabaseDriver, EM extends D[typeof EntityManagerType] & EntityManager = EntityManager>(contextName?: string, paths?: string[], options?: Partial<Options>): Promise<Configuration<D, EM>>;
```
B4nan, line 35: do we really need two signatures here? they are the same except the optionality of the `paths`, right?

I would only care about the old signature from stable versions, for the same reason I am marking this as just refactor, since it is not fixing anything that would be released to stable channel

tests/features/cli/CLIHelper.test.ts
```diff
@@ -54,7 +55,11 @@ describe('CLIHelper', () => {
         case 'mikro-orm-factory.config.js': return (contextName: string) => (contextName === 'boom' ? undefined : Object.assign(
           {},
           config,
-          { dbName: `tenant_${contextName}` } satisfies Options<MongoDriver>,
+          { dbName: `tenant_${contextName}`,
+            logger: message => {
```
B4nan, line 59:
```suggestion
          {
            dbName: `tenant_${contextName}`,
            logger: message => {
```

packages/core/src/utils/ConfigurationLoader.ts
boenrobot: I was hoping to have the argument-less version also be deprecated, while still available for BC.

I suppose that overload could instead be just that - a zero arguments one, rather than the "true" one but with defaults.

B4nan: The old signature was `(validate?: boolean, options?: Partial<Options>)`, and I am fine deprecating that, but why should we have two overloads for deprecating it?

Maybe I am missing something, and in the end I am fine with what you have in here, it's just a bit confusing to have 3 signatures now.

boenrobot: The old signature allowed the first arg to be false... these overloads, as written, cause a type error on false, while still allowing true and not adding the arg.

B4nan: But why? The point of doing this was to not cause any failures, even type errors are a BC.

boenrobot: Validate = false woudn't work anyway... might as well alert integrations that use it.

If any framework integration out there needs it, they can stilldo that with `new Configuration({}, false)`, which is cross version compatible.

The nestjs one doesn't use it, and I'd bet other integrations don't either.

B4nan: Ok, let's leave it like this, worst case we can fix it if someone reports any problems.

PR #6215, opened 2024-11-07
refactor(core): add missing backward compatibility code

lsh4711:


lsh4711:
I made a slight modification to the signature, and I think it is necessary to check whether this change does not go against boenrobot's original intent.

PR #6216, opened 2024-11-07
feat(core): validate non-persistent relations are not composite

B4nan:
Adding `persist: false` to relations is supported, but rather discouraged, it was designed to be used on scalars instead. Using it on a relation targeting composite PK will cause issues and is now disallowed. You can opt out of this validation via `checkNonPersistentCompositeProps` discovery option.

### DIFF
--- a/packages/core/src/errors.ts
+++ b/packages/core/src/errors.ts
@@ -258,6 +258,10 @@ export class MetadataError<T extends AnyEntity = AnyEntity> extends ValidationEr
     return this.fromMessage(meta, prop, `targets abstract entity ${prop.type}. Maybe you forgot to put @Entity() decorator on the ${prop.type} class?`);
   }
 
+  static nonPersistentCompositeProp(meta: EntityMetadata, prop: EntityProperty) {
+    return this.fromMessage(meta, prop, `is non-persistent relation which targets composite primary key. This is not supported and will cause issues, 'persist: false' should be added to the properties representing single columns instead.`);
+  }
+
   static propertyTargetsEntityType(meta: EntityMetadata, prop: EntityProperty, target: EntityMetadata) {
     /* istanbul ignore next */
     const suggestion = target.embeddable ? 'Embedded' : 'ManyToOne';
--- a/packages/core/src/metadata/MetadataValidator.ts
+++ b/packages/core/src/metadata/MetadataValidator.ts
@@ -49,7 +49,7 @@ export class MetadataValidator {
 
     for (const prop of Utils.values(meta.properties)) {
       if (prop.kind !== ReferenceKind.SCALAR) {
-        this.validateReference(meta, prop, metadata);
+        this.validateReference(meta, prop, metadata, options);
         this.validateBidirectional(meta, prop, metadata);
       } else if (metadata.has(prop.type)) {
         throw MetadataError.propertyTargetsEntityType(meta, prop, metadata.get(prop.type));
@@ -119,7 +119,7 @@ export class MetadataValidator {
     });
   }
 
-  private validateReference(meta: EntityMetadata, prop: EntityProperty, metadata: MetadataStorage): void {
+  private validateReference(meta: EntityMetadata, prop: EntityProperty, metadata: MetadataStorage, options: MetadataDiscoveryOptions): void {
     // references do have types
     if (!prop.type) {
       throw MetadataError.fromWrongTypeDefinition(meta, prop);
@@ -132,6 +132,10 @@ export class MetadataValidator {
 
     if (metadata.find(prop.type)!.abstract && !metadata.find(prop.type)!.discriminatorColumn) {
       throw MetadataError.targetIsAbstract(meta, prop);
+    }
+
+    if (prop.persist === false && metadata.find(prop.type)!.compositePK && options.checkNonPersistentCompositeProps) {
+      throw MetadataError.nonPersistentCompositeProp(meta, prop);
     }
   }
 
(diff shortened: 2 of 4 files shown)


B4nan:
Why would you use persist: false on M:N? 

This validation is meant for m:1 and 1:1, those will break since the ORM only represents such properties as `prop.fieldNames[0]`.

B4nan:
This is one problem: 

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/metadata/MetadataDiscovery.ts#L1440-L1442

Here is a test failing case that motivated this new validation:

```ts
import {
  BaseEntity,
  Collection,
  Entity,
  ManyToOne,
  MikroORM,
  OneToMany,
  PrimaryKey,
  PrimaryKeyProp,
  Property,
  Ref,
} from "@mikro-orm/sqlite";
import { v4 } from "uuid";

@Entity()
class OrganizationMikroModel extends BaseEntity {
  @PrimaryKey({ columnType: "uuid" })
  id!: string;
}

@Entity()
class TaskReviewerMikroModel extends BaseEntity {
  [PrimaryKeyProp]?: ["id", "organization"];

  @PrimaryKey({ columnType: "uuid" })
  id!: string;

  @ManyToOne({ entity: () => OrganizationMikroModel, ref: true, primary: true })
  organization!: Ref<OrganizationMikroModel>;

  @ManyToOne({
    entity: () => TaskMikroModel,
    ref: true,
    deleteRule: "cascade",
    joinColumns: ["task_id", "organization_id"],
  })
  task!: Ref<TaskMikroModel>;

  @ManyToOne({
    entity: () => TaskReviewerMikroModel,
    deleteRule: "set null",
    ref: true,
    nullable: true,
    joinColumns: ["previous_reviewer_id", "organization_id"],
    persist: false,
  })
  previousReviewer?: Ref<TaskReviewerMikroModel> | null;

  @Property({ type: "uuid", nullable: true })
  previousReviewerId: string | null = null;
}

@Entity()
class TaskMikroModel extends BaseEntity {
  [PrimaryKeyProp]?: ["id", "organization"];

  @PrimaryKey({ columnType: "uuid" })
  id!: string;

  @ManyToOne({ entity: () => OrganizationMikroModel, ref: true, primary: true })
  organization!: Ref<OrganizationMikroModel>;

  @OneToMany({
    entity: () => TaskReviewerMikroModel,
    mappedBy: "task",
    orphanRemoval: true,
  })
  reviewers = new Collection<TaskReviewerMikroModel>(this);
}

let orm: MikroORM;

beforeAll(async () => {
  orm = await MikroORM.init({
    dbName: ":memory:",
    entities: [OrganizationMikroModel, TaskReviewerMikroModel, TaskMikroModel],
    debug: ["query", "query-params"],
    allowGlobalContext: true, // only for testing
  });
});

afterAll(async () => {
  await orm.close(true);
});

test("fun times", async () => {
  const orgId = v4();
  orm.em.create(OrganizationMikroModel, { id: orgId });
  await orm.em.flush();
  orm.em.clear();

  const taskId = v4();
  const reviewerId1 = v4();
  const reviewerId2 = v4();

  const newTask = orm.em.create(TaskMikroModel, {
    id: taskId,
    organization: orgId,
  });
  orm.em.assign(newTask, {
    reviewers: [
      orm.em.create(TaskReviewerMikroModel, {
        id: reviewerId1,
        task: [taskId, orgId],
        organization: orgId,
      }),
      orm.em.create(TaskReviewerMikroModel, {
        id: reviewerId2,
        previousReviewerId: reviewerId1,
        task: [taskId, orgId],
        organization: orgId,
      }),
    ],
  });
  await orm.em.flush();
  orm.em.clear();

  const task = await orm.em.findOneOrFail(
    TaskMikroModel,
    { id: taskId },
    { populate: ["reviewers"] }
  );

  // Delete reviewer 1
  orm.em.assign(task, {
    reviewers: [
      orm.em.getReference(TaskReviewerMikroModel, [reviewerId2, orgId]),
    ],
  });

  await orm.em.flush();

  // mimic flush to able to log/assert the changesets
  orm.em.getUnitOfWork().computeChangeSets();
  console.log(orm.em.getUnitOfWork().getChangeSets());
  expect(orm.em.getUnitOfWork().getChangeSets()).toHaveLength(0);
});
```

B4nan:
> A pivot table with auxiliary data...

But you would add `persist: false` on the pivot entity on the to-one properties, not on the M:N, right?

If you have an actual use case, please provide a complete example for that. I'll be happy to target this only to to-one relations, but I want to see what I am dealing with first.

B4nan:
Changed the validation to only trigger for to-one properties in a436ca67f07a9bb492718de595ae2facc0f1fbb0. 

I see you have some `persist: false` on M:N in the entity generator snapshots, and as said above, I am not sure what is that for exactly, but maybe there is some use case I am missing. IIRC such relations are ignored in schema diffing, the option is not just about persistence, maybe we could split it into two options for better control, if the motivation was having readonly relations. The problem I was trying to solve with the validation is definitely not about them. I haven't found any to-one relations with `persist: false` in there, only the scalar props are marked that way, which is the correct way to do things (and exactly what this validation is trying to enforce).

B4nan:
Yes, it should be working on all relations. But cascade persist (which I guess you have on your mind) is always enforced, same for all relation kinds. 99% people misused this by setting `cascade: [remove]` without realizing they also disable persistence by reachability, and that is just wrong, making that configurable was a mistake, at least with this kind of API. I would rather introduce something like `readonly: true`, I think someone was already asking for something like that in the past.

PR #6224, opened 2024-11-09
docs: add v6.4 blogpost

B4nan:


docs/blog/2024-11-06-mikro-orm-6-4-released.md
```diff
@@ -0,0 +1,164 @@
...
+});
+```
+
+## Support for multiple ORM configurations
+
+[//]: # (TODO)
+
+## App-level `--config` argument deprecated
+
+[//]: # (TODO)
```
B4nan, line 156: @boenrobot if you could write a few sentences here, it would be great. My current plan is to ship stable on Monday morning.

docs/blog/2024-11-11-mikro-orm-6-4-released.md
```diff
@@ -0,0 +1,217 @@
...
+});
+```
+
+## Support for multiple ORM configurations
+
+[//]: # (TODO)
+
+## App-level `--config` argument deprecated
+
+[//]: # (TODO)
```
boenrobot, line 156:
```suggestion
When you use `MikroORM.init()` without arguments, MikroORM tries to figure out the configuration in a few different ways. As part of this step, the command line of the process (`process.argv`) is checked for an option called "--config" with the value being a path to a MikroORM config file.

While convinient for those who want this fearure, there is no way to opt out of it for those who do not, short of explicitly setting the config. Even worse that that though, it can cause conflicts if you intend to use the same argument name for something else, and when you inevitably hit an error due to this conflict, it can be difficult to diagnose if you are not keenly aware MikroORM is doing this.

So, in v7, MikroORM will no longer be checking the command line for the path to the config file. You can still specify a custom path with the `MIKRO_ORM_CONFIG` environment variable, or you can parse the command line yourself and pass the resulting config to the `MikroORM.init()` call.

Because this change is breaking, and yet subtle enough for users not realizing they're relying on it, in v6.4, MikroORM will log a deprecation warning on the console output if your config is loaded based on the commane line arguments. This will let you see if you would be affected by this change in v7.

This change also marks the first time MikroORM is having formal deprecation warnings system. There may be more such warnings added before v7, or in later releases.
```

docs/blog/2024-11-06-mikro-orm-6-4-released.md
B4nan: thanks, modified it a bit, I'd suggest installing grammarly extension btw, it helps with the typos tremendously :]

could you please also write a bit on the multi ORM configs?

docs/blog/2024-11-11-mikro-orm-6-4-released.md
```diff
@@ -0,0 +1,229 @@
...
+    const hint = createQueryHint(store);
+
+    return sql + hint;
+  },
+});
+```
+
+## Support for multiple ORM configurations
+
+[//]: # (TODO)
```
boenrobot, line 152:
```suggestion
From the very beginning, you could have multiple instances of MikroORM running in your application, each with a different config. Examples of when you may want to do this are multi-tenant applications or applications where certain features are delegated to a separate database for whatever reason.

When it comes to actually using those multiple instances however, you were always on own. Any framework intergration that wants to let you run multiple instances must also provide you its own means of passing either the config object itself, or a path to a config file to be imported for each instance.

With this release, this is starting to change. A single MikroORM config file can now export a function that dynamically produces a config object based on a name as an argument, or an array with a mix of config objects and functions. Based on a "contextName" option, MikroORM's CLI will pick one of the objects.

Framework integrations will be able to use a similar approach, where they can take your one config file, and create each instance you use based on your available config objects and functions.

There is more work to be done to actually make such integrations, but at least there is now a standard way they can rely on, and thus have the framework integration itself "just work".
```

I do not have access to my computer (will do tomorrow...). I'm writing this from a phone, and adding a code example would be a huge pain. Feel free to copy from the docs if you want to add examples too. Being from a phone is also the reason for the typos 😅.  No red squiggly here. Just suggestions if the cursor is at a wring word.

B4nan: Thanks! You could have just said you are AFK till monday :] I already started writing this down as well, so will try to merge my few sentences with yours :] I'll release it in the morning, but we can polish the article a bit during the day if needed.

boenrobot: Yes, but I didn't want to delay you (again), so I did the next best thing😅

PR #6253, opened 2024-11-22
feat(core): allow defining entities with inferred interfaces

### LINKED ISSUE #6242: Enhance type of EntitySchema
At this stage, in order to define Entity without using decorator magic, we use EntitySchema. like below:

```ts
import { EntitySchema, t, Collection, Reference } from '@mikro-orm/core';

export interface IBook {
  title: string;
  price?: number;
  author?: Reference<IAuthor>;
  tags: Collection<IBookTag>;
}

export const Book = new EntitySchema<IBook>({
  name: 'Book',
  properties: {
    title: { type: t.string },
    price: { type: t.number, nullable: true, default: 0 },
    author: { entity: () => Author, ref: true, nullable: true },
    tags: { entity: () => BookTag, ref: true, array: true },
  }
})
```

It works well, based on this, we can make types more secure, as well as infer interfaces from EntitySchema to enhance the development experience.

We can infer interfaces from EntitySchema to make it easier to use.
In this way, we write Entity only once, and TypeScript can infer the interface automatically.
And I think providing a Zod-Like API would improve the development experience significantly:

```ts
export const Book = EntitySchema.define({
  name: 'Book',
  properties: (t) => ({
    title: t.string(),
    price: t.number().nullable().default(0),
    author: t.manyToOne(() => Author).nullable(),
    tags: t.manyToMany(() => BookTag),
  })
})

export interface IBook extends InferEntity<typeof Book> {}
```

Currently, EntitySchema only checks the keys of an interface when it is defined and don't check the nullable, value type or relationship. We can use TypeScript to do more strict checking to avoid accidental errors:

```ts
import { EntitySchema, t, Collection, Reference } from '@mikro-orm/core';

export interface IBook {
  title: string;
  price?: number;
  author?: Reference<IAuthor>;
  tags: Collection<IBookTag>;
}

export const Book = new EntitySchema<IBook>({
  name: 'Book',
  properties: {
    title: { type: t.string },
    price: { entity: () => Author, ref: true, nullable: true }, // TypeScript should tell developer that this is wrong
    author: { entity: () => Author, ref: true, nullable: true },
    tags: { entity: () => BookTag, ref: true, array: true },
  }
})
```

To achieve this, we need to refactor the `EntitySchemaMetadata` type.
In this way, we still write interfaces first, which may be more verbose than Infer interfaces.

**Additional context**
The following ORMs define Table using Zod-like APIs instead of decorators:
* [Drizzle ORM](https://orm.drizzle.team/docs/sql-schema-declaration#example)
* [Orchid ORM](https://orchid-orm.netlify.app/guide/orm-and-query-builder.html#define-a-base-table)

xcfox:
This PR implements the `Zod-Like API` discussed in the [previous discussion](https://github.com/mikro-orm/mikro-orm/issues/6242), which allows defining schemas using JavaScript code and automatically infers interfaces using TypeScript's inference capabilities.

This brings two significant improvements:
- Reducing the amount of code;
- Enhances type safety;

#### Directly using EntitySchema
```ts
export interface IPublisher4 extends Omit<IBaseEntity5, typeof OptionalProps> {
  [OptionalProps]?: 'name' | 'type' | IBaseEntity5[typeof OptionalProps];
  name: string;
  type: PublisherType;
  books: Collection<IBook4>;
  tests: Collection<ITest4>;
  enum3?: number;
}

export enum PublisherType {
  LOCAL = 'local',
  GLOBAL = 'global',
}

export const Publisher4 = new EntitySchema<IPublisher4, IBaseEntity5>({
  name: 'Publisher4',
  extends: BaseEntity5,
  properties: {
    name: { type: 'string', default: 'asd' },
    type: { enum: true, items: () => PublisherType, default: PublisherType.LOCAL },
    enum3: { enum: true, items: [1, 2, 3], nullable: true },
    books: { kind: '1:m', entity: 'Book4', mappedBy: 'publisher' },
    tests: { kind: 'm:n', entity:  'Test4', fixedOrder: true },
  },
});
```

#### Using defineEntity

```ts
import { defineEntity, InferEntity } from '@mikro-orm/core';

export enum PublisherType {
  LOCAL = 'local',
  GLOBAL = 'global',
}

export const Publisher5 = defineEntity({
  name: 'Publisher5',
  extends: BaseEntity5,
  properties: p => ({
    name: p.string().default('asd'),
    type: p.enum(() => PublisherType).default(PublisherType.LOCAL),
    enum3: p.enum([1, 2, 3]).nullable(),
    books: () => p.oneToMany(Book4).mappedBy('publisher'),
    tests: () => p.manyToMany(Test4).fixedOrder(),
  }),
});

export type IPublisher5 = InferEntity<typeof Publisher5>;
```

Closes #6242


xcfox:
Additionally, for complete type safety, we need a hook that can populate default values for entities when `em.create()` is called.
Currently, default values declared via `onCreate` or `default` only take effect during `em.flush()`, which means that between `em.create()` and `em.flush()`, the default values do not exist. Perhaps we need to add an `onInit()` method to properties, so that when `em.create` is called, `onInit` is invoked to populate the default values for the entity.

B4nan:
I wonder if it's a good idea to implement this in the `EntitySchema` class. Seeing that you mostly added new stuff and didn't really reuse much of that class, I would rather use something separate to reduce the complexity of it, as well as isolate this functionality. Could be a separate `defineEntity` export maybe?

xcfox:
`defineEntity` is a very appropriate name. I will refactor a version as soon as possible.

packages/core/src/metadata/EntitySchema.ts
```diff
@@ -403,4 +552,10 @@ export class EntitySchema<Entity = any, Base = never> {
     } as EntityProperty<T>;
   }
 
+  '~types'?: {
```
B4nan, line 555: I'd use a `declare symbol` instead of hardcoded name, like in here

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/typings.ts#L76-L77

B4nan:
> as soon as possible.

Btw no need to rush, I'd keep this for the next feature release so we can promote it properly. Maybe during next month or even in January.

tests/entities-schema/FooBar4.ts
```diff
@@ -31,3 +31,14 @@ export const FooBar4 = new EntitySchema<IFooBar4, IBaseEntity5>({
     virtual: { type: String, persist: false },
   },
 });
+
+export const FooBar5 = EntitySchema.define({
```
B4nan, line 35: I'd remove all those changes in the tests since you are not using those entities anywhere, they are not really tests in the first place. Let's create a new suite for this, and actually use the entities to verify they work both runtime and build time/type-wise. Let's use sqlite for that so it doesn't add a lot of overhead.

Also, this will need quite a lot of new documentation, preferably all the examples with code tabs would have a new tab with this approach, as well as text description of this new approach somewhere.

packages/core/src/types/index.ts
```diff
@@ -19,15 +19,15 @@ import { StringType } from './StringType';
...
+import { IType, TransformContext, Type, InferJSType } from './Type';
 import { Uint8ArrayType } from './Uint8ArrayType';
 import { UnknownType } from './UnknownType';
 import { UuidType } from './UuidType';
 
 export {
   Type, DateType, TimeType, DateTimeType, BigIntType, BlobType, Uint8ArrayType, ArrayType, EnumArrayType, EnumType,
   JsonType, IntegerType, SmallIntType, TinyIntType, MediumIntType, FloatType, DoubleType, BooleanType, DecimalType,
-  StringType, UuidType, TextType, UnknownType, TransformContext, IntervalType, IType, CharacterType,
+  StringType, UuidType, TextType, UnknownType, TransformContext, IntervalType, IType, CharacterType, InferJSType,
```
B4nan, line 30: why is this one publicly exported? feels internal

packages/core/src/decorators/OneToMany.ts
```diff
@@ -37,7 +38,7 @@ export function OneToMany<Target, Owner>(
   return createOneToDecorator(entity, mappedBy, options, ReferenceKind.ONE_TO_MANY);
 }
 
-export interface OneToManyOptions<Owner, Target> extends ReferenceOptions<Owner, Target> {
+export interface OneToManyOptions<Owner, Target, ValueType = Target extends object ? Collection<Target> : never> extends ReferenceOptions<Owner, Target, ValueType> {
```
B4nan, line 41: this feels wrong, target of 1:m is not a collection.

packages/core/src/decorators/OneToOne.ts
```diff
@@ -1,18 +1,19 @@
...
-  options: Partial<OneToOneOptions<Owner, Target>> = {},
+  options: Partial<OneToOneOptions<Owner, Target, any>> = {},
 ) {
   const mappedBy = typeof mappedByOrOptions === 'object' ? mappedByOrOptions.mappedBy : mappedByOrOptions;
   options = typeof mappedByOrOptions === 'object' ? { ...mappedByOrOptions, ...options } : options;
   return createOneToDecorator<Target, Owner>(entity as string, mappedBy, options, ReferenceKind.ONE_TO_ONE);
 }
 
-export interface OneToOneOptions<Owner, Target> extends Partial<Omit<OneToManyOptions<Owner, Target>, 'orderBy'>> {
+export interface OneToOneOptions<Owner, Target, ValueType = Target extends object ? Reference<Target> : never> extends Partial<Omit<OneToManyOptions<Owner, Target, ValueType>, 'orderBy'>> {
```
B4nan, line 16: target of 1:1 doesnt have to be a `Reference` wrapper

packages/core/src/typings.ts
```diff
@@ -73,6 +73,9 @@ export const EagerProps = Symbol('EagerProps');
 export const HiddenProps = Symbol('HiddenProps');
 export const Config = Symbol('Config');
 
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export const __types = Symbol('__types');
```
B4nan, line 77: does it have to be a real property? i was hoping we can use the exact same approach as with the other internal type symbols, so `declare const __types: unique symbol;`

also lets try to find a better name than `__types`

xcfox: What name do you think is more appropriate? As references:
- [zod](https://github.com/colinhacks/zod/blob/main/src/types.ts#L173) use `_type`, `_output`, `_input`
- [valibot](https://github.com/fabian-hiller/valibot/blob/main/library/src/types/schema.ts#L62) use `~types`
- [yup](https://github.com/jquense/yup/blob/master/src/schema.ts#L154) use `__outputType`
- [drizzle](https://github.com/drizzle-team/drizzle-orm/blob/main/drizzle-orm/src/table.ts#L53) just use `_`
- [orchid-orm](https://github.com/romeerez/orchid-orm/blob/main/packages/orm/src/baseTable.ts#L136) use `columns.shape`

packages/core/src/decorators/OneToOne.ts
xcfox: Yes, we can use function overloads to infer whether the type of the user's JS code is `Reference`.

packages/core/src/decorators/OneToMany.ts
xcfox: I think I made a mistake. In a 1:m relationship, the value type of the property should be `Collection<Owner>`.

packages/core/src/decorators/OneToOne.ts
xcfox: Additionally, as a more type-safe approach, do you agree with me enabling `{ ref: true }` by default in `oneToOneFactory`?

B4nan: > Additionally, as a more type-safe approach, do you agree with me enabling { ref: true } by default in oneToOneFactory?

Having different defaults among the various ways to define an entity sounds quite wrong to me.

package.json
```diff
@@ -133,6 +133,7 @@
     "rimraf": "6.0.1",
     "ts-jest": "29.2.0",
     "ts-node": "10.9.2",
+    "tsd": "^0.31.2",
```
B4nan, line 136: not sure what is this doing here?

packages/core/src/decorators/Property.ts
```diff
@@ -105,6 +105,12 @@ export interface PropertyOptions<Owner, Value = any> {
    * Add the property to the `returning` statement.
    */
   returning?: boolean;
+
+  /**
+   * Automatically set the property value when new instance of entity is created,
+   * either manually em.create(), or automatically when new entities are loaded from database
```
B4nan, line 111: so this is an equivalent of the `@OnInit()` hook on property level? adding default values to managed entities is often wrong idea, why not use the existing `onCreate`?

xcfox: `onInit` is used to populate default values for entities defined using `EntitySchema`, while `onCreate` is not triggered when calling `em.create`.
Or is there another way to automatically populate default values for entities defined by `EntitySchema` when created with `em.create`?

```ts
const Foo = EntitySchema.define({
  name:'Foo',
  properties: t => ({
    id: t.integer({ primary: true }),
    createdAt: t.datetime({ onCreate: () => new Date() }),
    initAt: t.datetime({ onInit: () => new Date() }),
    byDefault: t.text({ default: 'foo' }),
  }),
});

// same in class and decorators
@Entity()
class Foo {
  @PrimaryKey()
  id: number;

  @Property({ onCreate: () => new Date() })
  createdAt: Date;

  @Property()
  initAt: = new Date();

  @Property({ default: 'foo' })
  byDefault: string;
}

it('should create entity with default values', async () => {
  const foo = orm.em.create(Foo, {} as any);
  expect(foo.initAt).toBeInstanceOf(Date);
  expect(foo.createdAt).toBeUndefined();
  expect(foo.byDefault).toBeUndefined();
  await orm.em.flush();
  expect(foo.initAt).toBeInstanceOf(Date);
  expect(foo.createdAt).toBeInstanceOf(Date);
  expect(foo.byDefault).toEqual('foo');
});
```

B4nan: My point is that you usually don't want to do this for managed entities, you want values from database there, instead of overriding them with some runtime default.

We could enable using the `default` values directly in `em.create`, although I am not so sure about the value of this. DB defaults are handled by the database on flush.

xcfox: > We could enable using the `default` values directly in `em.create`, although I am not so sure about the value of this. DB defaults are handled by the database on flush.

This sounds like a better solution than `onInit`, because `onInit` executes when loading model instances from the database, whereas default values only apply during `em.create` and first flush.

B4nan: Yes, that's exactly what I am trying to say here. its fine to add this, but its not the right tool for what you want to use it.

B4nan: Also adding this should be handled separately, it doesn't have much common with this PR, it needs to work with all variants for defining entities. Same applies to dealing with the `default` in `em.create`, let's not do those things in this PR, they need to have a separate entry in the changelog.

packages/core/src/decorators/Property.ts
```diff
@@ -125,7 +120,7 @@ export interface PropertyOptions<Owner, Value = any> {
    * Specify default column value for {@link https://mikro-orm.io/docs/schema-generator Schema Generator}.
    * This is a runtime value, assignable to the entity property. (SQL only)
    */
-  default?: string | string[] | number | number[] | boolean | null;
+  default?: string | Value | (() => Value) | null;
```
B4nan, line 123: why the callback? this option represents database defaults, it needs to be a value. you are not supposed to be using this as `() => new Date()`. we have `onCreate` for that, which only adds the runtime value and has no effect on the schema.

xcfox: EntitySchema needs a way to fill in default values when `em.create` is called. Consider we have an entity with a uuid as the primary key, and we need to assign a different uuid to each entity:

```ts
const Foo = EntitySchema.define({
  name: 'Foo',
  properties: {
    id: t.string({ primary: true, default: () => randomUUID() })
  }
})
```

In this commit, I strictly distinguish whether the value in `default` is a callback function:
- For callback functions: They will be called when creating the entity, and the returned value will be used as the default value. In this case, you can still use `defaultRaw` to specify the default value in the database.
- For primitive values: They will not be assigned to the entity when it is created, but will wait for `em.flush()`, and then set as the default value in the database.

Do you think it is more suitable to call `onCreate` instead of `default` when `em.create` is called?

B4nan: Yes, `onCreate` is there exactly for this purpose, unlike `default` which is primarily for the schema generator. Let's not bend the `default` option for what the `onCreate` is there.

B4nan:
Maybe you could extract some smaller part of this PR which we could merge right ahead, so the PR checks would start for you without me allowing them (you need at least one contribution, otherwise I need to push the button on every commit you add). Could be some typo fix or a small update to the types, or some of the features we discussed above that should have their own changelog entry (e.g. processing `onCreate` directly inside `em.create`).

xcfox:
@B4nan  The new `defineEntity` feature has been completed and includes comprehensive tests.

Next, we can consider updating the documentation.

B4nan (commented):
few more notes

B4nan:
also the PR title should be changed, this is a new feature, a new way to define entities, it shouldn't be connected to `EntitySchema` anyhow

packages/core/src/metadata/EntitySchema.ts
```diff
@@ -403,4 +409,14 @@ export class EntitySchema<Entity = any, Base = never> {
     } as EntityProperty<T>;
   }
 
+  declare readonly [__types]?: {
+    entity: Entity;
+  };
```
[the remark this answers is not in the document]
xcfox: This is a common practice for implementing type inference for schemas, and almost all schema libraries (such as [zod](https://github.com/colinhacks/zod/blob/main/src/types.ts#L173), [valibot](https://github.com/fabian-hiller/valibot/blob/main/library/src/types/schema.ts#L62), [drizzle](https://github.com/drizzle-team/drizzle-orm/blob/main/drizzle-orm/src/table.ts#L53), [orchid-orm](https://github.com/romeerez/orchid-orm/blob/main/packages/orm/src/baseTable.ts#L136)) include similar properties.

With the help of the `__types` property, third-party libraries (such as [tRPC](https://github.com/trpc/trpc/blob/next/packages/server/src/unstable-core-do-not-import/parser.ts#L55)) can infer entity types without directly depending on the schema library.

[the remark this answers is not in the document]
B4nan: I am asking why it's attached to `EntitySchema` class, so saying it's a common practice feels more than weird, this is our class, and doesn't really have much in common with the libraries you listed. Internally, we use `EntityMetadata` class to actually store things, `EntitySchema` lives only as a way for users to define entities (and a mechanism for normalizing the defaults during discovery) - here, we are adding a new way.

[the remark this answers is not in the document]
xcfox: More specifically, `__types` is prepared for `InferEntity`.

Without `__types`, we can only use the following `InferEntity`:

```ts
export type InferEntity<T extends EntitySchema<any, any>> = T extends EntitySchema<infer E> ? E : never;
```

With the help of `__types`, we can use the following `InferEntity`:

```ts
export type InferEntity<T> = T extends {[__types]?: {entity?: any}} ? NonNullable<T[__types]>['entity'] : never;
```

Compared to the first method, the second method is usually more precise and does not depend on `EntitySchema`. This means that for third-party libraries, there is no need to add `@mikro/core` as a dependency.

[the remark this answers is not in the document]
xcfox: I think you might believe that `__types` is better suited to be placed in `EntityMetadata`. This is feasible, and I will refactor this part of the code.

docs/docs/embeddables.md
```diff
@@ -363,6 +427,13 @@ address!: Address;
 ```ts title="./entities/User.ts"
 @Embedded({ prefix: false })
 address!: Address;
+```
+
+  </TabItem>
+  <TabItem value="define-entity">
+
+```ts title="./entities/User.ts"@Entity()
+address: p.embedded(()=> Address, { prefix: false }),
```
B4nan, line 436: mind the spacing, especially in the docs

```suggestion
address: p.embedded(() => Address, { prefix: false }),
```

docs/docs/embeddables.md
```diff
@@ -363,6 +427,13 @@ address!: Address;
 ```ts title="./entities/User.ts"
 @Embedded({ prefix: false })
 address!: Address;
+```
+
+  </TabItem>
+  <TabItem value="define-entity">
+
+```ts title="./entities/User.ts"@Entity()
```
B4nan, line 435: this doesnt look right

```suggestion
```ts title="./entities/User.ts"
@Entity()
```

docs/docs/embeddables.md
```diff
@@ -445,6 +523,13 @@ addresses: Address[] = [];
 ```ts title="./entities/User.ts"
 @Embedded()
 addresses: Address[] = [];
+```
+
+  </TabItem>
+  <TabItem value="define-entity">
```
B4nan, line 529: here you use `define-entity` while on line 510 you use `defineEntity`. this needs to be the same, that's the whole point of the tab ids

B4nan (commented):
few random things I spotted when peeking at the docs changes

docs/docs/embeddables.md
xcfox: Yes, I'm keeping the same value `define-entity` everywhere and using `defineEntity` as the label, just like `entity-schema` - `EntitySchema` .

xcfox:
@B4nan I've added fully documents for `defineEntity`.
If there are any areas that could benefit from improvement or further discussion, please feel free to let me know.

rickylabs:
is it planned to merge this anytime soon ?

B4nan:
It contains some small hidden breaking changes, notably the early execution of `onCreate` property callbacks (#6263). If we want to keep it that way, it would have to wait till v7.

I guess we could add some opt-in for that and keep the current behavior to unblock this. But still, I wouldn't call it _anytime soon_, I still didn't find the time to properly test and review this.

xcfox:
@B4nan Thank you for the suggestions you provided regarding this PR.  
This PR has completed a well-refined `Zod-like` API and provided relevant documentation.

However, there is still some room for improvement, including the Relational API and the Chained API.

## Relational API
The API currently used in this PR looks like this:

```ts
const userProperties = defineEntityProperties({
  ...baseProperties,
  name: p.string(),
  email: p.string({ unique: true }),
});

export interface IUser extends InferEntityFromProperties<typeof userProperties> {
  friend: IUser | null | undefined;
}

export const User: EntitySchema<IUser> = defineEntity({
  name: 'User',
  properties: {
    ...userProperties,
    friend: p.manyToOne(() => User, { nullable: true }),
  },
});
```

Due to the limitations of TypeScript, we need to manually annotate the types of relational fields.

Inspired by the [Drizzle Relational API v2](https://rqbv2.drizzle-orm-fe.pages.dev/docs/relations-v2), we can improve this API to make it look like this:

```ts
// We need to export this entity to define relationships for other entities
export const UserEntity = defineEntity((p) => ({
  ...baseProperties,
  name: p.string(),
  email: p.string().unique(),
}));

export const { User } = defineRelations({ User: UserEntity }, (r) => ({
  friend: r.manyToOne.User().nullable(),
}));
```

## Chained API
The API currently used in this PR looks like this:

```ts
export const Book = defineEntity((p) => ({
  name: 'Book',
  properties: {
    id: p.string({ primaryKey: true }),
    createdAt: p.date({ onCreate: () => new Date() }),
    updatedAt: p.date({ onCreate: () => new Date(), onUpdate: () => new Date() }),
    title: p.string(),
    description: p.string({ nullable: true }),
  },
}));
```

The advantage of this design is that it is more in line with the writing style of the `EntitySchema` constructor.

We can use the chained API to improve this API to provide the best development experience, making it look like this:

```ts
export const Book = defineEntity((p) => ({
  name: 'Book',
  properties: {
    id: p.string().primaryKey(),
    createdAt: p.date().onCreate(() => new Date()),
    updatedAt: p.date()
      .onCreate(() => new Date())
      .onUpdate(() => new Date()),
    title: p.string(),
    description: p.string().nullable(),
  },
}));
```

---

What do you think of those API designs? Let me know your thoughts, and I'll improve this PR according to your suggestions.

jadejr:
I didn't see https://standardschema.dev/ mentioned in this or the mentioned discussion. Would be nice to be inspired by what they are doing where it fits.

xcfox:
@jadejr Thank you for your interest in StandardSchema!

In fact, when the standard-schema was proposed, I had the intention of integrating it into MikroORM. We had thorough discussions, and you can review the previous discussions in [this PR](https://github.com/mikro-orm/mikro-orm/pull/6147).

Ultimately, I decided not to directly incorporate standard-schema into the core of MikroORM for the following reasons: Standard-schema is designed for validation libraries (such as Zod, Valibot, and ArkType), and it only includes the Input and Output types of the schema. However, for MikroORM and other ORM libraries, we need to have Create, Update, and various Select types. Standard-schema cannot effectively describe the entities in the database.

Moreover, we now have [mikro-orm-standard-schema](https://github.com/boenrobot/mikro-orm-standard-schema), which enables MikroORM and standard-schema to work well together.

jadejr:
Ah, i read the discussion, but i didn't click through the PR so i missed it :(

Thanks for the link mikro-orm-standard-schema!

xcfox:
## April 2025 Update

I have reimplemented the `defineEntity`. Compared with the previous version, the current `defineEntity` has witnessed a significant improvement in the development experience.

Inspired by the recursive types in [zod v4](https://v4.zod.dev/v4#true-recursive-types), we have broken through the limitations of circular application in TypeScript. We no longer need to manually annotate the types of any relationships.
I believe this is the most elegant way to define relationships in the TypeScript ecosystem. (If there is a better approach, please let me know.) We only need to write the type definitions, and TypeScript will automatically infer the correct types and relationships.

Here are some examples:

#### Chained Calls

```ts
export const User = defineEntity({
  name: 'User',
  properties: t => ({
     ...BaseEntityProperties,
     name: t.string().default('unknown'),
     email: t.string().nullable().default('unknown@example.com'),
     role: t.enum(['user', 'admin']).default('user'),
   }),
});

export type IUser = InferEntity<typeof User>;
```

#### Relationships

```ts
export const Post = defineEntity({
  name: 'Post',
  properties: t => ({
    ...BaseEntityProperties,
    title: t.string().nullable(false),
    content: t.string().nullable(false),
    author: () => t.manyToOne(User).nullable(), // Use the entire field as a callback function to break through the circular reference limitation of TypeScript
    replies: () => t.oneToMany(Post).mappedBy('parent'), // Self-reference, and TypeScript can infer the correct type for this
    parent: () => t.manyToOne(Post).nullable(),
  }),
});

export type IPost = InferEntity<typeof Post>;
```

### This PR Also Includes

- 📝 **Complete Documentation**: Updated all parts of the documentation related to entity definitions and provided a wealth of examples of `defineEntity`.
- 🧪 **100% Testing**: Added a large number of test cases, achieving 100% code coverage to ensure the correctness of the type inference and relationship definition of `defineEntity`.
- 🚥 **No Breaking Changes**: Does not contain any breaking changes and can be safely merged into V6. (Waiting for [trigger onCreate when em.create is called](https://github.com/mikro-orm/mikro-orm/pull/6263) to provide full type safety.)

tests/defineEntity.test.ts
```diff
@@ -0,0 +1,1106 @@
...
+        files: () => p.oneToMany(File).mappedBy('folder'),
+      }),
+    });
+
+    const File = defineEntity({
+      name: 'File',
+      properties: p => ({
+        id: p.integer().primary().autoincrement(),
+        name: p.string(),
+        folder: () => p.manyToOne(Folder),
```
B4nan, line 300: this sounds more in line with how things work
```suggestion
        folder: p.manyToOne(() => Folder),
```

tests/defineEntity.test.ts
```diff
@@ -0,0 +1,1106 @@
+import { Cascade, Collection, defineEntity, EntityDTO, EntityMetadata, EntitySchema, Hidden, InferEntity, Opt, Ref, Reference, ScalarReference, types } from '@mikro-orm/core';
+import { IsExact, assert } from 'conditional-type-checks';
+
+describe('defineEntity', () => {
+  it('should define entity', () => {
+    const Foo = defineEntity({
+      name: 'Foo',
+      properties: p => ({
```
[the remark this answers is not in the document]
xcfox: There is no need to use a `callback` here.
I have considered three ways to get the properties builder `p`:

### Through defineEntity.properties
```ts
import { defineEntity } from '@mikro/core'

const p = defineEntity.properties

const Foo = defineEntity({
  name: 'Foo',
  properties: {
    name: p.string(),
  },
})
```
- ✅ The `@mikro/core` only needs to export the `defineEntity` method externally.
- ⚠️ An additional line is required to get `p`.


### Through the properties callback function
```ts
import { defineEntity } from '@mikro/core'

const Foo = defineEntity({
  name: 'Foo',
  properties: p => ({
    name: p.string(),
  }),
})
```
- ✅ The `@mikro/core` only needs to export the `defineEntity` method externally.
- ✅ There is no need to use an additional statement to get `p`.

### Export `p`
```ts
import { defineEntity, p } from '@mikro/core'

const Foo = defineEntity({
  name: 'Foo',
  properties: {
    name: p.string(),
  },
})
```
- ✅ There is no need to use an additional statement to get `p`.
- ⚠️ The `@mikro/core` needs to export both the `defineEntity` method and `p` externally. Maybe it seems a bit confusing to export a single letter? I'm not really sure either.

tests/defineEntity.test.ts
```diff
@@ -0,0 +1,1106 @@
...
+    expect(Foo.meta).toEqual(asSnapshot(FooSchema.meta));
+  });
+
+  it('should define entity with reference scalar property', () => {
+    const Foo = defineEntity({
+      name: 'Foo',
+      properties: p => ({
+        id: p.integer().primary().autoincrement(),
+        name: p.string().ref(),
+        email: p.string().ref().ref(false),
```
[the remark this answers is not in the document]
xcfox: Here we are actually testing the type of `.ref(false)`. At the JavaScript level, `{ ref: false }` should override `{ ref: true }`. In terms of types, `.ref(false)` should rewrite `email: Ref<string>` as `email: string`. I will replace this with a more complete use case later.

tests/defineEntity.test.ts
xcfox: Yes, `folder: p.manyToOne(() => Folder)` definitely looks better, but it will cause an error related to circular references in TypeScript, ultimately resulting in the relationship type not being correctly inferred. This issue also plagues [Valibot](https://valibot.dev/guides/other/) and [zod](https://zod.dev/?id=recursive-types).

`folder: () => p.manyToOne(Folder)` is a reluctant choice. Inspired by [zod v4](https://v4.zod.dev/v4#true-recursive-types), as long as we use the entire property as a callback function, we can bypass TypeScript's circular reference check.

Another approach is to define the property using a getter, but I think this syntax is not as concise as the callback function:
```ts
const File = defineEntity({
  name: 'File',
  properties: p => ({
    name: p.string(),
    get folder() {
      return p.manyToOne(() => Folder),
    },
  }),
});
```

tests/defineEntity.test.ts
```diff
@@ -0,0 +1,1106 @@
...
+    });
+
+    expect(Foo.meta).toEqual(asSnapshot(FooSchema.meta));
+  });
+
+  it('should define entity with formula', () => {
+    const Box = defineEntity({
+      name: 'Box',
+      properties: p => ({
+        objectVolume: p.formula('obj_length * obj_height * obj_width').$type<number>(),
```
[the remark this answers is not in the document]
xcfox: Good idea. I'll add a generic parameter for `.formula`.

docs/docs/entity-schema.md
```diff
@@ -154,9 +157,82 @@ export const schema = new EntitySchema<FooBar>({
 ```
 
 > As a value for `type` you can also use one of `String`/`Number`/`Boolean`/`Date`.
+## `defineEntity`
```
B4nan, line 161:
```suggestion

## `defineEntity`
```

docs/docs/entity-schema.md
```diff
@@ -154,9 +157,82 @@ export const schema = new EntitySchema<FooBar>({
...
+});
+
+// We can use `InferEntity` to infer the type of an entity
+export interface IBook extends InferEntity<typeof Book> {}
+```
+
+`defineEntity.properties` provides all [MikroORM built-in types](./custom-types#types-provided-by-mikroorm). To use [custom types](./custom-types), we can also use `p.type()`.
+
+```ts
+const properties = defineEntityProperties(p => ({
```
B4nan: i guess this should be `defineEntity.properties`? i dont see `defineEntityProperties` import or definition anywhere

docs/docs/entity-schema.md
```diff
@@ -154,9 +157,82 @@ export const schema = new EntitySchema<FooBar>({
...
+    publisher: () => p.oneToOne(Publisher).inversedBy('books'),
+    tags: () => p.manyToMany(BookTag).inversedBy('books').fixedOrder(),
+  }),
+});
+
+// We can use `InferEntity` to infer the type of an entity
+export interface IBook extends InferEntity<typeof Book> {}
+```
+
+`defineEntity.properties` provides all [MikroORM built-in types](./custom-types#types-provided-by-mikroorm). To use [custom types](./custom-types), we can also use `p.type()`.
```
B4nan, line 195: speaking of custom types, do we have a way to work with the `IType` too? i guess the `$type` callback would work there?

https://mikro-orm.io/docs/next/custom-types#mapping-to-objects-and-type-safety

ideally that page about custom types would also use tabs (or a separate section) and show examples of this with the new `defineEntity` helper.

packages/core/src/entity/defineEntity.ts
```diff
@@ -0,0 +1,925 @@
...
+  /** Set the constraint type. Immediate constraints are checked for each statement, while deferred ones are only checked at the end of the transaction. Only for postgres unique constraints. */
+  deferMode(deferMode: DeferMode | `${DeferMode}`): OneToOneOptionsBuilder<TargetValue> {
+    return new OneToOneOptionsBuilder({ ...this['~options'], deferMode });
+  }
+
+}
+
+
+function createPropertyBuilders<Types extends Record<string, any>>(
+	options: Types,
```
B4nan: tabs on several places in this file

B4nan (commented):
i got few final nits

B4nan:
also please update the PR description to reflect current state

packages/core/src/entity/defineEntity.ts
```diff
@@ -0,0 +1,925 @@
+import type { EntityManager, CheckCallback, SerializeOptions, EntityMetadata, Cascade, LoadStrategy, DeferMode, ScalarReference, Reference, Opt, Hidden, EnumOptions, Dictionary, OneToManyOptions, Collection, EmbeddedOptions, EmbeddedPrefixMode, ManyToManyOptions, FilterQuery, QueryOrderMap, EntityName, OneToOneOptions } from '..';
+import type { ColumnType, PropertyOptions, ManyToOneOptions, ReferenceOptions } from '../decorators';
+import type { AnyString, GeneratedColumnCallback, Constructor } from '../typings';
+import type { Type } from '../types';
+import { EntitySchema } from '../metadata/EntitySchema';
+import { types } from '../types/index';
```
B4nan: dont import from the index files, this will create import cycles (and is not supported in ESM at all), import directly from the files that define those things

docs/docs/entity-schema.md
xcfox: This was a misstake, and I've updated the document, using `defineEntity.properties`.

docs/docs/entity-schema.md
xcfox: I enhanced the `$type` method, and it will use `IType` as the type when more than two type parameters are used.
And I have added examples in the document.

packages/core/src/entity/defineEntity.ts
xcfox: I reorganized the import statements to avoid circular imports.

packages/core/src/entity/defineEntity.ts
```diff
@@ -0,0 +1,943 @@
+import type { EntityManager } from '../EntityManager';
+import type { ColumnType, PropertyOptions, ManyToOneOptions, ReferenceOptions, EnumOptions, EmbeddedOptions, EmbeddedPrefixMode, ManyToManyOptions, OneToManyOptions, OneToOneOptions } from '../decorators';
```
B4nan: this is still an index import

packages/core/src/entity/defineEntity.ts
```diff
@@ -0,0 +1,943 @@
...
+import type { AnyString, GeneratedColumnCallback, Constructor, Opt, Hidden, CheckCallback, FilterQuery, EntityName, Dictionary, EntityMetadata } from '../typings';
+import type { Reference, ScalarReference } from './Reference';
+import type { SerializeOptions } from '../serialization';
```
B4nan: same here

B4nan (changes requested):

packages/core/src/entity/defineEntity.ts
xcfox: I updated the import statement to ensure that nothing is imported from the barrel file except for 'types'.

PR #6263, opened 2024-11-27
refactor(core): trigger `onCreate` when `em.create` is called.

xcfox:


packages/core/src/entity/EntityFactory.ts
```diff
@@ -287,6 +287,10 @@ export class EntityFactory {
           .forEach(prop => delete entity[prop.name]);
       }
 
+      if (options.newEntity) {
+        this.assignDefaultValues(entity, meta);
+      }
```
B4nan, line 292: i feel like the place where you run this actually is a breaking change, as right now the callback gets an entity on the time of flush, with all the values present. people can (and I believe they use it this way) infer the value based on other things, which were possibly set explicitly, not just via property initializers. now you run this right after the constructor, so no values outside of property initializers are present.

we can consider running this after the `hydrate` call, but even that feels breaking to me to be honest. i would be fine doing that opt-in right now, and we can run a poll to see how the community feels about this change before we'd switch the default in v7.

xcfox: It's indeed better to call after 'hydrate'.

Yes, this is a breaking change. Although I believe most projects won't be affected by this update, and the PR didn't add many new test cases, we should still consider the community's feedback to avoid any potential issues.

B4nan:
I'd be up for merging this into v6 if we make it opt-in. Next release will be a feature one, I want to include the `defineEntity` helper too. I got some vacations ahead, so it will take a few weeks (maybe a month) to prepare things, so no rush, I still have one feature I want to look into for this release myself ("hybrid" loading strategy).

PR #6292, opened 2024-12-11
docs: add jsdoc to make it clear that ArrayCollection.add() is idempotent

alcalyn:
Just starting with MikroOrm.

I found some places where there is no jsdoc not documentation, e.g for those two methods, I had to check source code to know if I need to do `if (collection.contains())` before adding an item or not.

### DIFF
--- a/packages/core/src/entity/Collection.ts
+++ b/packages/core/src/entity/Collection.ts
@@ -159,6 +159,9 @@ export class Collection<T extends object, O extends object = object> extends Arr
     return super.toJSON() as unknown as EntityDTO<TT>[];
   }
 
+  /**
+   * @inheritDoc
+   */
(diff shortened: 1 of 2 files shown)


packages/core/src/entity/Collection.ts
```diff
@@ -159,6 +159,9 @@ export class Collection<T extends object, O extends object = object> extends Arr
     return super.toJSON() as unknown as EntityDTO<TT>[];
   }
 
+  /**
```
B4nan, line 162: This needs to be over the `set` method too. Ideally on all the collection methods that are overriding the arraycollection ones.

alcalyn: `set` seems not to be overriden in Collection.

Sorry I take time to respond because I was confused, and I'm currently replacing typeorm by mikro-orm, and I want to make sure I understand mikro-orm before adding jsdoc

PR #6309, opened 2024-12-17
fix(core): correctly handle populate of `mapToPk` relations

Ruby184:
### The Problem

Related issue: [#6265](https://github.com/mikro-orm/mikro-orm/pull/6265). Before the changes, populating a `mapToPk` relation did not extract just the key; instead, the entire entity object was assigned to the property.

### The Solution

I have added tests for both simple and composite primary keys. While adding tests for the `select-in` loading strategy, I noticed that the second query (for loading the primary key entity) is not executed. This makes sense because the keys are already selected.  

This led me to reconsider whether we should use a join for the `joined` strategy, as the keys are already present on the owning entity. 

**Select-in strategy:**

```sql
select `t0`.* 
from `team` as `t0` 
where `t0`.`id` = 'team1' 
limit 1
```
**Joined strategy:**

```sql
select `t0`.*, `c1`.`id` as `c1__id`
from `team` as `t0`
-- this join is redundant
left join `order` as `c1` on `t0`.`current_order_id` = `c1`.`id`
where `t0`.`id` = 'team1'
limit 1
```

EDIT: I have also added a test for the inverse side of a 1:1 relation using `mapToPk` to verify that it works as expected. Another redundant populate query was being made to the `Team` entity, which I addressed as well.

```sql
select `o0`.*, `o1`.`id` as `o1__id`
from `order` as `o0`
left join `team` as `o1` on `o0`.`id` = `o1`.`current_order_id`
where `o0`.`id` = 'order1'
limit 1
-- this populate query is redundant because we do not want the full entity
select `t0`.*
from `team` as `t0`
where `t0`.`current_order_id` in ('order1')
```

### DIFF
--- a/packages/core/src/entity/EntityLoader.ts
+++ b/packages/core/src/entity/EntityLoader.ts
@@ -223,7 +223,7 @@ export class EntityLoader {
     }
 
     const where = await this.extractChildCondition(options, prop);
-    const data = await this.findChildren<Entity>(entities, prop, populate, { ...options, where, orderBy: innerOrderBy! }, !!ref);
+    const data = await this.findChildren<Entity>(entities, prop, populate, { ...options, where, orderBy: innerOrderBy! }, !!(ref || prop.mapToPk));
     this.initializeCollections<Entity>(filtered, prop, field, data, innerOrderBy.length > 0);
 
     return data;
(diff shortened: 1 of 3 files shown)


B4nan:
The original idea was that you should not populate a relation with `mapToPk`, since there is nothing to populate. Only exception for that is inverse side of 1:1 relation.

But we should indeed handle the case where people try to do so, even if they don't need to, and I can imagine in some cases (like joined filters on relations) it might be needed or done internally.

B4nan (approved):
thanks!

PR #6315, opened 2024-12-24
docs: use native uuid example

twlite:


README.md
```diff
@@ -299,13 +299,11 @@ Or if you want to use UUID primary keys:
...
 
 ```typescript
-import { v4 } from 'uuid';
-
 @Entity()
 export class UuidBook {
 
   @PrimaryKey()
-  uuid = v4();
+  uuid = crypto.randomUUID();
```
B4nan, line 306: don't you need to import crypto first?

twlite: It is a global object that works even in browsers

B4nan: maybe nowadays, we support node 18, and from what i see this was made global in node 19, so lets add the explicit import please.

README.md
```diff
@@ -299,13 +299,13 @@ Or if you want to use UUID primary keys:
 **`./entities/UuidBook.ts`**
 
 ```typescript
-import { v4 } from 'uuid';
+import { randomUUID } from 'crypto';
```
B4nan, line 302:
```suggestion
import { randomUUID } from 'node:crypto';
```

B4nan (approved):
all right, thanks! 

if you care, there will be many places like this in the actual docs, so PR welcome for those

PR #6326, opened 2025-01-06
feat(mongo): include MikroORM version in MongoDB handshake

alexbevi:
The PR incorporates MongoDB's [wrapping client library](https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.md#supporting-wrapping-libraries) specification for the connection handshake to allow library details to be included in the metadata written to `mongos` or `mongod` logs.

For example, this change would allow server-side logs such as the following:

> {"t":{"$date":"2024-01-27T23:10:40.108Z"},"s":"I","c":"NETWORK","id":51800,"ctx":"conn16235","msg":"client metadata","attr":{"remote":"127.0.0.1:1094","client":"conn16235","doc":{"driver":{`"name":"nodejs|MikroORM"`,`"version":"6.12.0|6.4.2"`},"platform":"Node.js v18.18.2, LE","os":{"name":"linux","architecture":"x64","version":"5.15.133+","type":"Linux"}}}}

For anyone hosting clusters with connections coming from different applications this can help differentiate connections and facilitate log analysis.

### DIFF
--- a/packages/mongodb/src/MongoConnection.ts
+++ b/packages/mongodb/src/MongoConnection.ts
@@ -147,6 +147,12 @@ export class MongoConnection extends Connection {
       ret.maxPoolSize = pool.max;
     }
 
+    const version = Utils.getORMVersion();
+    ret.driverInfo = {
+      name: 'MikroORM',
+      version: version
+    }
+    
     return Utils.mergeConfig(ret, this.config.get('driverOptions'));
   }
 
--- a/packages/mongodb/src/MongoConnection.ts
+++ b/packages/mongodb/src/MongoConnection.ts
@@ -147,10 +147,9 @@ export class MongoConnection extends Connection {
       ret.maxPoolSize = pool.max;
     }
 
-    const version = Utils.getORMVersion();
     ret.driverInfo = {
       name: 'MikroORM',
-      version: version
+      version: Utils.getORMVersion(),
     }
     
     return Utils.mergeConfig(ret, this.config.get('driverOptions'));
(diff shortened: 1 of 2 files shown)


alexbevi:
For context this is the same change proposed/merged at https://github.com/meteor/meteor/pull/13222

packages/mongodb/src/MongoConnection.ts
```diff
@@ -147,6 +147,12 @@ export class MongoConnection extends Connection {
       ret.maxPoolSize = pool.max;
     }
 
+    const version = Utils.getORMVersion();
+    ret.driverInfo = {
+      name: 'MikroORM',
+      version: version
+    }
```
B4nan, line 154:
```suggestion
    ret.driverInfo = {
      name: 'MikroORM',
      version: Utils.getORMVersion(),
    }
```

B4nan (approved):
thanks!

packages/mongodb/src/MongoConnection.ts
```diff
@@ -147,6 +147,11 @@ export class MongoConnection extends Connection {
       ret.maxPoolSize = pool.max;
     }
 
+    ret.driverInfo = {
+      name: 'MikroORM',
+      version: Utils.getORMVersion(),
+    }
+
```
B4nan, line 154:
```suggestion
    };

```

B4nan (changes requested):
You'll need to update the assertion in the failing test.

PR #6328, opened 2025-01-08
fix(entity-generator): skip bidirectional relation if target metadata doesn't exist

patrykbojczuk:
This fixes a problem with generating entities when some referenced tables are skipped/not taken and `bidirectionalRelations` option is set to `true`.


tests/features/entity-generator/EntityGenerator.mysql.test.ts
```diff
@@ -63,6 +63,18 @@ describe('EntityGenerator', () => {
     await orm.close(true);
   });
 
+  test('generate entities with bidirectional relations and skipped referenced table [mysql]', async () => {
```
B4nan, line 66: for tests that are not mysql specific i would prefer any other driver. MySQL is the slowest one when it comes to working with schema (both create table queries as well as info schema queries). ideally create an isolated test that is not using those huge "initORMxxx` methods.

packages/entity-generator/src/EntityGenerator.ts
```diff
@@ -279,7 +279,10 @@ export class EntityGenerator {
       : metadata.filter(m => !m.pivotTable || this.referencedEntities.has(m));
     for (const meta of filteredMetadata) {
       for (const prop of meta.relations) {
-        const targetMeta = metadata.find(m => m.className === prop.type)!;
+        const targetMeta = metadata.find(m => m.className === prop.type);
+        if (!targetMeta) {
+          continue;
+        }
```
B4nan, line 285:
```suggestion

        if (!targetMeta) {
          continue;
        }

```

tests/features/entity-generator/EntityGenerator.mysql.test.ts
patrykbojczuk: Sure, I created an isolated test according to your suggestion.

B4nan:
thanks! looking good, any reason you kept this as a draft?

boenrobot (approved):

PR #6333, opened 2025-01-12
feat(core): improved configuration loaders

octet-stream:
### Details

This PR implements improves configuration loading with support for new transpilers, like [jiti](https://www.npmjs.com/package/jiti) and [tsx](https://www.npmjs.com/package/tsx). It also includes documentation updates, reflecting these changes. This PR also deprecates some of the old CLI settings: `useTsNode`, `preferTs`, and `alwaysAllowTs` in favor of a single new option called `loader`, and this new option allows to opt-in to a specific transpiler, or use auto-detection (the default), or opt-out to runtime's native `import()` (when set to either false or `native`).

### Notable changes

- [x] Implement new loaders: `ts-node`, `jiti`, `tsx`, and `native`;
- [x] Add new type-safe ESM utilities:
  - [x] `tryModule` - it resolves a promise, returned by `import()` and catches the errors with the `ERR_MODULE_NOT_FOUND` code, then wraps it into a `TryModuleError`, so it can be handled later. Unlike existent utilities (Utils.dynamicImport and Utils.tryRequire), the resulting value has correct types (though it only supports the `import()` - it can be easily modified to support `require()` function as well);
  - [x] `requireDefault` - resolves the `default` member from a module's exports while also providing correct types;
- [x] Add in-code documentation for CLI Settings interface;
- [x] Deprecate `useTsNode`, `preferTs`, and `alwaysAllowTs` in favor of `loaders`;
- [ ] Integrate loaders into `ConfigurationLoader` class;
- [ ] Add tests for loaders and `ConfigurationLoader` integration;

I also updated git hooks to use local binaries of `lint-staged` and `commitlint`, because there's no point to have those installed globally (yarn can pick up both local and global installations), nor to assume everyone has these by default.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,6 +5,9 @@ module.exports = {
     'es2020': true,
     'node': true,
   },
+  globals: {
+    NodeJS: true // So ESLint will no complain about NodeJS namespace
+  },
   extends: [
     'eslint:recommended',
     'plugin:@typescript-eslint/eslint-recommended',
--- a/packages/core/src/utils/ConfigurationLoader.ts
+++ b/packages/core/src/utils/ConfigurationLoader.ts
@@ -9,6 +9,7 @@ import { colors } from '../logging/colors';
 import type { Dictionary } from '../typings';
 import { Configuration, type Options } from './Configuration';
 import { Utils } from './Utils';
+import type { Loaders } from './loader';
 
 /**
  * @internal
@@ -167,12 +168,23 @@ export class ConfigurationLoader {
 
   static getSettings(): Settings {
     const config = ConfigurationLoader.getPackageConfig();
-    const settings = { ...config['mikro-orm'] };
+    const settings: Settings = { ...config['mikro-orm'] };
     const bool = (v: string) => ['true', 't', '1'].includes(v.toLowerCase());
-    settings.useTsNode = process.env.MIKRO_ORM_CLI_USE_TS_NODE != null ? bool(process.env.MIKRO_ORM_CLI_USE_TS_NODE) : settings.useTsNode;
+
+    function loaderNameFromEnv(value?: string): Loaders {
+      if (!value || ['false', 'f', '0', 'no', 'n', '', 'true', 't', '1', 'yes', 'y'].includes(value.toLowerCase())) {
+        return undefined;
+      }
+
+      return value as Loaders;
+    }
+
     settings.tsConfigPath = process.env.MIKRO_ORM_CLI_TS_CONFIG_PATH ?? settings.tsConfigPath;
-    settings.alwaysAllowTs = process.env.MIKRO_ORM_CLI_ALWAYS_ALLOW_TS != null ? bool(process.env.MIKRO_ORM_CLI_ALWAYS_ALLOW_TS) : settings.alwaysAllowTs;
     settings.verbose = process.env.MIKRO_ORM_CLI_VERBOSE != null ? bool(process.env.MIKRO_ORM_CLI_VERBOSE) : settings.verbose;
+
+    settings.useTsNode = process.env.MIKRO_ORM_CLI_USE_TS_NODE != null ? bool(process.env.MIKRO_ORM_CLI_USE_TS_NODE) : settings.useTsNode;
+    settings.alwaysAllowTs = process.env.MIKRO_ORM_CLI_ALWAYS_ALLOW_TS != null ? bool(process.env.MIKRO_ORM_CLI_ALWAYS_ALLOW_TS) : settings.alwaysAllowTs;
+    settings.loader = process.env.MIKRO_ORM_CLI_LOADER != null ? loaderNameFromEnv(process.env.MIKRO_ORM_CLI_LOADER) : settings.loader;
 
     if (process.env.MIKRO_ORM_CLI_CONFIG?.endsWith('.ts')) {
       settings.useTsNode = true;
@@ -444,10 +456,72 @@ export class ConfigurationLoader {
 
 }
 
+/**
+ * Command line settings
+ */
 export interface Settings {
-  alwaysAllowTs?: boolean;
+  /**
+   * Enable verbose logging (e.g. print queries used in seeder or schema diffing)
+   */
   verbose?: boolean;
-  useTsNode?: boolean;
+
+  /**
+   * A custom path to your `tsconfig.json` file
+   */
   tsConfigPath?: string;
+
+  /**
+   * Custom paths for Mikro ORM config lookup
+   */
   configPaths?: string[];
+
+  /**
+   * A loader to import Mikro ORM config with.
+   * This option enables TypeScript support if the runtime of your choice can't do that for you.
+   *
+   * You can use `MIKRO_ORM_CLI_LOADER` to set this option via environment variables.
+   *
+   * The value can be either of these: [`'ts-node'`](https://www.npmjs.com/package/ts-node), [`'jiti'`](https://www.npmjs.com/package/jiti), [`'tsx'`](https://www.npmjs.com/package/tsx), `'auto'`, `'native'`, `false`, `null`, or `undefined`.
+   *
+   * When set to `'native'`, Mikro ORM will try and use runtime's native [`import()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import) to load config, whether or not the runtime can handle TypeScript files.
+   *
+   * When set to `'auto'`, Mikro ORM will try and use loaders in following order:
+   *
+   * 1. `ts-node`
+   * 2. `jiti`
+   * 3. `tsx`
+   * 4. `import()`
+   *
+   * If none of these can read the config, you will be asked to install either of the packages.
+   *
+   * The use of `ts-node` as config loader is discouraged and it might be removed in a future releases.
+   *
+   * @default 'auto'
+   */
+  loader?: Loaders;
+
+  /**
+   * Whether or not to bypass TypeScript `loader` and let the runtime to hanlde it
+   *
+   * @default false
+   *
+   * @deprecated use `loader` option instead
+   */
+  alwaysAllowTs?: boolean;
+
+  /**
+   * Whether or not use `ts-node` package to import the config.
+   *
+   * **The package must be installed!**
+   *
+   * @deprecated use `loader` option instead
+   */
+  useTsNode?: boolean;
+
+  /**
+   * An alias for `useTsNode`
+   *
+   * @deprecated use `loader` option instead
+   */
+  preferTs?: boolean;
 }
--- a/packages/core/src/utils/Utils.ts
+++ b/packages/core/src/utils/Utils.ts
@@ -1360,3 +1360,76 @@ export class Utils {
   }
 
 }
+
+/**
+ * @internal
+ */
+export interface TryModuleErrorOptions extends ErrorOptions {
+  cause: NodeJS.ErrnoException;
+}
+
+/**
+ * This error is a noop flavor for `ERR_MODULE_NOT_FOUND` error with required `cause` option.
+ *
+ * @internal
+ */
+
+export class TryModuleError extends Error {
+
+  override readonly cause: NodeJS.ErrnoException;
+
+  /**
+   * Module specifier
+   */
+  readonly specifier: string;
+
+  constructor(specifier: string, options: TryModuleErrorOptions) {
+    const { cause, ...rest } = options;
+
+    super(`Unable to import module "${specifier}"`, rest);
+
+    this.specifier = specifier;
+    this.cause = cause;
+  }
+
+}
+
+/**
+ * @internal
+ */
+export interface TryModuleOptions {
+  specifier: string;
+}
+
+/**
+ * Takes a `promise` returned from an `import()` call, then resolves it to have correct typings, and catches any `ERR_MODULE_NOT_FOUND` error.
+ * If such error occurs, then it throws a `TryModuleError` with the original error as its `cause`.
+ *
+ * @param promise - A promise that resolves the module
+ * @param options - Extra options
+ *
+ * @internal
+ */
+export const tryModule = <TModuleResult>(
+  promise: Promise<TModuleResult>,
+  options: TryModuleOptions,
+): Promise<TModuleResult> => promise.catch((cause): never => {
+  if (!(cause instanceof Error) || (cause as NodeJS.ErrnoException).code !== 'ERR_MODULE_NOT_FOUND') {
+    throw cause;
+  }
+
+  throw new TryModuleError(options.specifier, { cause });
+});
+
+/**
+ * Returns a value from `default` property of given ES module `value` object.
+ *
+ * You can use this function to get a value returned from a `import()` call.
+ *
+ * @param value - An ES module object
+ *
+ * @internal
+ */
+export const requireDefault = <T>(
+  value: T | {default: T},
+): T => Utils.isObject(value) && 'default' in value ? value.default : value;
--- a/packages/core/src/utils/loader.ts
+++ b/packages/core/src/utils/loader.ts
@@ -1,0 +1,157 @@
+import { join, isAbsolute } from 'node:path';
+
+import { tryModule, TryModuleError, requireDefault, Utils } from './Utils';
+import type { Settings } from './ConfigurationLoader';
+import type { Options } from './Configuration';
+
+/**
+ * A set of supported TypeScript loaders
+ */
+export type Loaders = 'ts-node' | 'jiti' | 'tsx' | 'auto' | 'native' | false | null | undefined;
+
+interface Loader {
+  name: string;
+  import(specifier: string): Promise<Options>;
+}
+
+type LoaderFactory = (projectRoot: string, cliSettings: Settings) => Promise<Loader>;
+
+const createLoaderFactory = (fn: LoaderFactory): LoaderFactory => fn;
+
+const createNativeLoader = createLoaderFactory(async () => ({
+  name: 'import',
+  import: async specifier => requireDefault(await import(specifier)), // TODO: Handle module not found errors and report if the specifier is .ts file
+}));
+
+/**
+ * Creates loader factory for [`ts-node`](https://www.npmjs.com/package/ts-node).
+ *
+ * Note: Unlike the other loaders, this will *register* ts-node as `require` hook.
+ */
+const createTsNodeLoader = createLoaderFactory(async (root, settings) => {
+  const name = 'ts-node';
+  const loader: Loader = {
+    name,
+    import: async specifier => requireDefault(await import(specifier)),
+  };
+
+  // If ts-node is already registered, we can just return the loader
+  if (Utils.detectTsNode()) {
+    return loader;
+  }
+
+  const tsNode = requireDefault(await tryModule(import('ts-node'), {
+    specifier: name,
+  }));
+
+  const configPath = settings.tsConfigPath || 'tsconfig.json';
+  const tsConfigPath = isAbsolute(configPath) ? configPath : join(process.cwd(), configPath);
+
+  const { options } = tsNode.register({
+    cwd: root,
+    project: tsConfigPath,
+    transpileOnly: true,
+    compilerOptions: {
+      module: 'nodenext',
+      moduleResolution: 'nodenext',
+    },
+  }).config;
+
+  // Register tsconfig-paths to support custom module resolution
+  if (Utils.isObject(options.paths) && Object.keys(options.paths).length > 0) {
+    const paths = requireDefault(await tryModule(import('tsconfig-paths'), {
+      specifier: 'tsconfig-paths',
+    }));
+
+    paths.register({
+      baseUrl: options.baseUrl ?? '.',
+      paths: options.paths,
+    });
+  }
+
+  return loader;
+});
+
+/**
+ * Creates loader factory for [`jiti`](https://www.npmjs.com/package/jiti) transpiler
+ */
+const createJitiLoader = createLoaderFactory(async root => {
+  const name = 'jiti';
+
+  const { createJiti } = await tryModule(import('jiti'), {
+    specifier: name,
+  });
+
+  const jiti = createJiti(root);
+
+  return {
+    name,
+    import: specifier => jiti.import(specifier, {
+      default: true,
+    }),
+  };
+});
+
+/**
+ * Creates loader factory for [`tsx`](https://www.npmjs.com/package/tsx) transpiler
+ */
+const createTsxLoader = createLoaderFactory(async root => {
+  const name = 'tsx';
+
+  const { tsImport } = await tryModule(import('tsx/esm/api'), {
+    specifier: name,
+  });
+
+  return {
+    name,
+    import: async specifier => requireDefault(await tsImport(specifier, root)),
+  };
+});
+
+const factories = [createTsNodeLoader, createJitiLoader, createTsxLoader];
+
+/**
+ * Auto detects available transpiler by iterating over the internal `factories` array (see above) and creating each loader from the list.
+ *
+ * If a loader factory throws `TryModuleError` that means there's no transpiler for this package installed.
+ *
+ * If no loader has been successfully created, it will return native loader and let the runtime to deal with config loading.
+ */
+export const createAutoLoader = createLoaderFactory(async (root, settings) => {
+  for (const createLoader of factories) {
+    try {
+      return await createLoader(root, settings);
+    } catch (error) {
+      if (!(error instanceof TryModuleError)) {
+        throw error;
+      }
+    }
+  }
+
+  return createNativeLoader(root, settings);
+});
+
+/**
+ * Creates a loader depending on given `settings.loader` value.
+ *
+ * If no loader specified, it will auto-detect whatever transpiler is installed within the project's `node_modules` by importing it via `import()`.
+ */
+export const createLoader = createLoaderFactory((root, settings) => {
+  if (settings.alwaysAllowTs || settings.preferTs === false || settings.useTsNode === false) {
+    return createNativeLoader(root, settings);
+  }
+
+  switch (settings.loader) {
+    case 'ts-node':
+      return createTsNodeLoader(root, settings);
+    case 'jiti':
+      return createJitiLoader(root, settings);
+    case 'tsx':
+      return createTsxLoader(root, settings);
+    case 'native':
+    case false:
+      return createNativeLoader(root, settings);
+    default:
+      return createAutoLoader(root, settings);
+  }
+});
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,1 +1,1 @@
-commitlint --edit $1
+yarn commitlint --edit $1
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,1 @@
-lint-staged
+yarn lint-staged
--- a/package.json
+++ b/package.json
@@ -127,6 +127,7 @@
     "guid-typescript": "1.0.9",
     "husky": "9.1.7",
     "jest": "29.7.0",
+    "jiti": "2.4.2",
     "lerna": "8.1.9",
     "lint-staged": "15.3.0",
     "mongodb": "6.12.0",
@@ -134,6 +135,7 @@
     "rimraf": "6.0.1",
     "ts-jest": "29.2.0",
     "ts-node": "10.9.2",
+    "tsx": "4.19.2",
     "typescript": "5.7.3",
     "uuid": "11.0.4"
   },
--- a/tests/Utils.test.ts
+++ b/tests/Utils.test.ts
@@ -1,6 +1,6 @@
 import path from 'node:path';
 import { ObjectId } from 'bson';
-import { EntityMetadata, MikroORM, sql, compareObjects, Utils } from '@mikro-orm/core';
+import { EntityMetadata, MikroORM, sql, compareObjects, Utils, requireDefault, tryModule, TryModuleError } from '@mikro-orm/core';
 import { Author } from './entities';
 import { initORMMongo, BASE_DIR } from './bootstrap';
 import FooBar from './entities/FooBar';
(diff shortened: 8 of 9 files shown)


B4nan:
> I also updated git hooks to use local binaries of lint-staged and commitlint, because there's no point to have those installed globally (yarn can pick up both local and global installations), nor to assume everyone has these by default.

I believe those were already resolved to the locally installed versions. I don't have them installed globally and it works just fine.

> Deprecate useTsNode, preferTs, and alwaysAllowTs in favor of loaders;

This feels weird to me, at `preferTs` should stay as a way to disallow using TS even if supported (= production builds). But I haven't checked the changes yet, so maybe I am missing something. Note that `useTsNode` is no longer needed since v6.4, and serves only as a way to disallow TS in the CLI (for the very same reason, production builds).

octet-stream:
> I believe those were already resolved to the locally installed versions. I don't have them installed globally and it works just fine.

I got an error when I tried to commit for the first time saying it can't find `lint-staged` (which I don't have globally installed), so I don't think this is the case. The only change I've made was adding `yarn` before `lint-staged` and `commitlint` in git hooks. This fixed the problem.

image

> This feels weird to me, at least preferTs should stay as a way to disallow using TS even if supported

This is still supported if the `loader` option or `MIKRO_ORM_CLI_LOADER` variable is set to either `"native"` or `false`. The option itself defaults to `"auto"`, meaning that Mikro ORM will try to detect whatever of supported transpilers installed. The `MIKRO_ORM_CLI_LOADER` will override whatever is set in `package.json`, just like it currently works.

For backward compatibility reasons I'll keep `ts-node` as the fist option (one the `createAutoLoader` will try fist), before you can decide whether to deprecate it or not. I think `jiti` is a good replacement, because it seem to me the most compatible (unlike tsx, which is using esbuild under the hood).

Also, I found that `preferTs` option is not documented, nor normalized (I don't see a CLI counterpart for this option anywhere). Should I include docs and add CLI option for it in this PR?

octet-stream:
>  Note that useTsNode is no longer needed since v6.4, and serves only as a way to disallow TS in the CLI (for the very same reason, production builds).

The new `loader` option can both disable TS detection, or choose a specific transpiler to handle TS files.

For example, when set to `tsx` - it will try and import it. If it's not installed - it will fail and ask user to add it to project's dependencies :)

B4nan:
Ok good. So the PR as is should be good for 6.x release, no BCs, right? 

> Also, I found that preferTs option is not documented,

It is mentioned [here](https://mikro-orm.io/docs/guide/project-setup#testing-the-endpoint) and [here](https://mikro-orm.io/docs/metadata-providers#tsmorphmetadataprovider), but feel free to mention it on other places. The thing is, you are not supposed to use this outside of test, and the value is inferred automatically, but having some explicit docs around it won't hurt for sure.

> nor normalized (I don't see a CLI counterpart for this option anywhere). Should I include docs and add CLI option for it in this PR?

You mean the settings in package.json? For that we have `useTsNode`, the value is used as a default for `preferTs` internally. If you mean env var, yeah, sure, let's add that.

Also, does this work with ESM projects? Will the regular `mikro-orm` CLI endpoint work? I've added the `mikro-orm-esm` because I haven't found any way to register the support programmatically with ts-node without using `node --loader`, I would love to get rid of it in v7.

packages/core/src/utils/loader.ts
```diff
@@ -0,0 +1,157 @@
+import { join, isAbsolute } from 'node:path';
+
+import { tryModule, TryModuleError, requireDefault, Utils } from './Utils';
+import type { Settings } from './ConfigurationLoader';
+import type { Options } from './Configuration';
+
+/**
+ * A set of supported TypeScript loaders
+ */
+export type Loaders = 'ts-node' | 'jiti' | 'tsx' | 'auto' | 'native' | false | null | undefined;
```
B4nan, line 10: feels weird to call this in plural, IMO it should be just `Loader` (and the current private `Loader` interface would be renamed, since its private, it can be anything really). or we can keep `Loader` interface and rename this to something like `LoaderType`. having it in plural is inconsistent with the rest of the codebase, I don't think we have anything called in plural right now.

also, since you deprecated `alwaysAllowTs`, how is this supposed to work with bun, will `native` support TS with bun while `false` force using JS files?

.eslintrc.js
```diff
@@ -5,6 +5,9 @@ module.exports = {
     'es2020': true,
     'node': true,
   },
+  globals: {
+    NodeJS: true // So ESLint will no complain about NodeJS namespace
+  },
```
B4nan, line 10: this is also weird, we already use the `node:` prefix on many places, never saw a problem with eslint in that regard. or is it for something else?

octet-stream:
> Ok good. So the PR as is should be good for 6.x release, no BCs, right?

Yes, and I will cover my code with test to make sure everything works as expected + since you have integration test for CLI, I should be able to catch potential bugs and regression in the process. But the code itself should behave the same: ts-node is the default, and old flags supported and will take precedence if `loader` option is not set.

Also, I'm planning to use fixtures for my tests, just like you do for tests against databases, and as I [already did](https://github.com/octet-stream/mikro-orm-load-config/tree/342c9fe197c6b7860a24c27fe9ce68ba3d0fb305/src/fixtures) in my proposal, so everything will be tested against real configs.

> Also, does this work with ESM projects? Will the regular mikro-orm CLI endpoint work?

Everything will be handled by either jiti, or tsx. They offer programmatic API that does not affect the whole runtime, unlike with ts-node, which behaves like loader or require hook. So, yes, a single entry point should be enough, and I already tested it against both CJS and ESM in my proposal. I will port these tests to Mikro ORM too. :)

See:
* tsx: https://tsx.is/dev-api/ts-import
* jiti: https://www.npmjs.com/package/jiti

packages/core/src/utils/ConfigurationLoader.ts
```diff
@@ -9,6 +9,28 @@ import { colors } from '../logging/colors';
...
+/**
+ * Returns valid value for `loader` option from given environment `MIKRO_ORM_CLI_LOADER` variable.
+ *
+ * @param value - A raw value of the environment variable
+ *
+ * @internal
+ */
+function loaderNameFromEnv(value?: string): Loaders {
+  const maybeBoolean = value?.toLowerCase();
+  if (!maybeBoolean || ['false', 'f', '0', 'no', 'n', ''].includes(maybeBoolean)) {
```
B4nan, line 23: we use different logic for reading bools from env vars, I would like to stay consistent with that. its used in this very file inside `loadEnvironmentVars`:

https://github.com/mikro-orm/mikro-orm/blob/59f05935e498735ca33f82fdd61f10d23634f0cd/packages/core/src/utils/ConfigurationLoader.ts#L296

i don't think we need to care about `y/n` or `yes/no` (or we should respect them universally, not just here). the code should be reused ideally (we can move the current helpers out of there to the top level functions).

packages/core/src/utils/ConfigurationLoader.ts
```diff
@@ -9,6 +9,28 @@ import { colors } from '../logging/colors';
...
+    return false;
+  }
+
+  // Return `undefined` if the `value` can be converted to `true`, so behaviour will be the same as in createLoader
+  if (['true', 't', '1', 'yes', 'y'].includes(maybeBoolean.toLowerCase())) {
```
B4nan, line 28: same here

packages/core/src/utils/Utils.ts
```diff
@@ -1360,3 +1360,76 @@ export class Utils {
...
+  cause: NodeJS.ErrnoException;
+}
+
+/**
+ * This error is a noop flavor for `ERR_MODULE_NOT_FOUND` error with required `cause` option.
+ *
+ * @internal
+ */
+
+export class TryModuleError extends Error {
```
B4nan, line 1377:
```suggestion
 */
export class TryModuleError extends Error {
```

also maybe think of a better name? but i guess we can keep this too, it just feels weird to have the `Try` prefix in an error name.

and move this to the https://github.com/mikro-orm/mikro-orm/blob/678119475cee860dd4684c9f4d903a3d4028b8a3/packages/core/src/errors.ts file

---

btw why is it internal, do we always catch it internally or can it bubble to the user? if it can, i wouldn't mark it as internal

packages/core/src/utils/Utils.ts
```diff
@@ -1360,3 +1360,76 @@ export class Utils {
   }
 
 }
+
+/**
+ * @internal
+ */
+export interface TryModuleErrorOptions extends ErrorOptions {
+  cause: NodeJS.ErrnoException;
```
B4nan, line 1368: is this `NodeJS` namespace why you altered the eslint config?

packages/core/src/utils/Utils.ts
```diff
@@ -1360,3 +1360,76 @@ export class Utils {
...
+export interface TryModuleOptions {
+  specifier: string;
+}
+
+/**
+ * Takes a `promise` returned from an `import()` call, then resolves it to have correct typings, and catches any `ERR_MODULE_NOT_FOUND` error.
+ * If such error occurs, then it throws a `TryModuleError` with the original error as its `cause`.
+ *
+ * @param promise - A promise that resolves the module
+ * @param options - Extra options
```
B4nan, line 1409:
```suggestion
 * @param promise A promise that resolves the module
 * @param options Extra options
```

packages/core/src/utils/Utils.ts
```diff
@@ -1360,3 +1360,76 @@ export class Utils {
...
+
+  throw new TryModuleError(options.specifier, { cause });
+});
+
+/**
+ * Returns a value from `default` property of given ES module `value` object.
+ *
+ * You can use this function to get a value returned from a `import()` call.
+ *
+ * @param value - An ES module object
```
B4nan, line 1429:
```suggestion
 * @param value An ES module object
```

packages/core/src/utils/Utils.ts
```diff
@@ -1360,3 +1360,76 @@ export class Utils {
...
+    throw cause;
+  }
+
+  throw new TryModuleError(options.specifier, { cause });
+});
+
+/**
+ * Returns a value from `default` property of given ES module `value` object.
+ *
+ * You can use this function to get a value returned from a `import()` call.
```
B4nan, line 1427:
```suggestion
 * You can use this function to get a value returned from an `import()` call.
```

packages/core/src/utils/Utils.ts
```diff
@@ -1360,3 +1360,76 @@ export class Utils {
...
+ *
+ * @param value - An ES module object
+ *
+ * @internal
+ */
+export const requireDefault = <T>(
+  value: T | {default: T},
```
B4nan, line 1434: weird that the linter doesnt complain about it
```suggestion
  value: T | { default: T },
```

packages/core/src/utils/loader.ts
```diff
@@ -0,0 +1,157 @@
...
+ */
+const createTsNodeLoader = createLoaderFactory(async (root, settings) => {
+  const name = 'ts-node';
+  const loader: Loader = {
+    name,
+    import: async specifier => requireDefault(await import(specifier)),
+  };
+
+  // If ts-node is already registered, we can just return the loader
+  if (Utils.detectTsNode()) {
```
B4nan, line 39: note that the name here is a bit wrong, the `detectTsNode` is used to detect TS support, not just via ts-node

packages/core/src/utils/loader.ts
```diff
@@ -0,0 +1,157 @@
...
+    return loader;
+  }
+
+  const tsNode = requireDefault(await tryModule(import('ts-node'), {
```
B4nan, line 43: I wonder if this won't be a problem for bundlers, `ts-node` is not a peer dependency and I believe bundlers will try to bundle it this way. That's why we use the `tryRequire` which won't suffer from it, since bundlers won't see a literal import of a package, they see just an import of some unknown dependency.

packages/core/src/utils/loader.ts
```diff
@@ -0,0 +1,157 @@
...
+    transpileOnly: true,
+    compilerOptions: {
+      module: 'nodenext',
+      moduleResolution: 'nodenext',
+    },
+  }).config;
+
+  // Register tsconfig-paths to support custom module resolution
+  if (Utils.isObject(options.paths) && Object.keys(options.paths).length > 0) {
+    const paths = requireDefault(await tryModule(import('tsconfig-paths'), {
```
B4nan, line 62: same here

packages/core/src/utils/loader.ts
```diff
@@ -0,0 +1,157 @@
...
+  };
+});
+
+/**
+ * Creates loader factory for [`tsx`](https://www.npmjs.com/package/tsx) transpiler
+ */
+const createTsxLoader = createLoaderFactory(async root => {
+  const name = 'tsx';
+
+  const { tsImport } = await tryModule(import('tsx/esm/api'), {
```
B4nan, line 101: same here

packages/core/src/utils/loader.ts
```diff
@@ -0,0 +1,157 @@
...
+const factories = [createTsNodeLoader, createJitiLoader, createTsxLoader];
+
+/**
+ * Auto detects available transpiler by iterating over the internal `factories` array (see above) and creating each loader from the list.
+ *
+ * If a loader factory throws `TryModuleError` that means there's no transpiler for this package installed.
+ *
+ * If no loader has been successfully created, it will return native loader and let the runtime to deal with config loading.
+ */
+export const createAutoLoader = createLoaderFactory(async (root, settings) => {
```
B4nan, line 120: I would prefer regular functions over arrow functions, especially for things that are publicly documented. arrow functions are rendered on the index page in API docs, while regular functions have their own page. Not to mention it's more readable that way.

packages/core/src/utils/Utils.ts
octet-stream: > btw why is it internal, do we always catch it internally or can it bubble to the user? if it can, i wouldn't mark it as internal

It is for internal use only. We will catch it and use its data (that's what is the `specifier` field for) to display a message asking for ts-node/jiti/tsx installation. Unlike Node.js' errors it can be tested with `instanceof` and type casting is not needed anymore.

I think `tryModule` and `requireDefault` should also be inaccessible from `@mikro-orm/core` package, since they supposed to use internally.

packages/core/src/utils/loader.ts
octet-stream: Yeah, I noticed. But then again, if it returns true, then `createTsNodeLoader` will just return loader (which in case with ts-node is just `import()`). So I think it should work, at least for Bun.

packages/core/src/utils/loader.ts
octet-stream: Hm. I never though about that. I will see if that is gonna be the problem. Maybe there's a way (universal, hopefully) to ask bundlers to skip bundling some of the dependencies for package authors.

packages/core/src/utils/Utils.ts
B4nan: > I think tryModule and requireDefault should also be inaccessible from @mikro-orm/core package, since they supposed to use internally.

Makes sense, as long as it won't be part of the barrel file, it shouldn't be exported. I guess we should rename them to `exports.ts` or similar, I don't want to use them internally (it can cause issues with cycles), they should be there currently only for this use case.

packages/core/src/utils/loader.ts
octet-stream: What do you mean?

In this case I just use createLoaderFactory to apply types, so I don't have to add them myself for each loader. I think it's neat.

octet-stream:
> I would love to get rid of it in v7.

I forgot to mention one important thing: As long as we have to support ts-node - this might not work in a single CLI entry point, so in v6 we sure have to keep `mikro-orm-esm` around. For new loaders, the `mikro-orm` should work.

packages/core/src/utils/ConfigurationLoader.ts
octet-stream: > we use different logic for reading bools from env vars, I would like to stay consistent with that. its used in this very file inside loadEnvironmentVars

Yes, I saw these. In this case I don't need to *convert* the value to a boolean. I just need to see if it *can* be converted. But I agree, that makes sense. So I'll just move these helpers outside of Utils class and and use in my comparison like so:

```ts
if (!value || bool(value) === false) {
  // ...
}
```

.eslintrc.js
octet-stream: ESLint thinks NodeJS is a variable that wasn't declared, while in reality it's just TS namespace. This is just configuration problem and sure needs better fix. But I'll better keep this one for you :)

Ideally this should be handled by typescript-eslint.

packages/core/src/utils/loader.ts
octet-stream: > feels weird to call this in plural, IMO it should be just Loader (and the current private Loader interface would be renamed, since its private, it can be anything really). or we can keep Loader interface and rename this to something like LoaderType. having it in plural is inconsistent with the rest of the codebase, I don't think we have anything called in plural right now.

Makes sense. I can rename it to loader. But I was thinking about another name, but for this type and cli option - `configLoader`, because it less generic than just `loader`. Though, I like the fact that the current name is just short :) But I can do how you say.

> also, since you deprecated alwaysAllowTs, how is this supposed to work with bun, will native support TS with bun while false force using JS files?

When `loader` is set to `false` or `"native"` - it will use native `import()`, so the runtime should handle it for use, but you have to specify this value explicitly, no matter which runtime you choose. Isn't this is how it works now?

packages/core/src/utils/loader.ts
B4nan: AFAIK wrapping the import inside a helper function and passing the module name dynamically is enough.

packages/core/src/utils/Utils.ts
octet-stream: > but you'll have to adjust the barrel file, I think it now reexports this whole file.

Which one? The Utils? Yup, You can see how I access my utils in the tests.

By the way, is the `errors.ts` also a barrel file? Some of the errors must be part of the public API surface, so you probably re-export them too, and I want `TryModuleError` class (or whatever the name will be at the end) to be inaccessible outside of `@mikro-orm/core`, since it's for internal use only.

packages/core/src/utils/Utils.ts
octet-stream: Huh, somehow ESLint was not complaining about that.

packages/core/src/utils/loader.ts
octet-stream: I see, but tryModule is not only catches ERR_MODULE_NOT_FOUND error, but also returns correct types, unlike `Utils.tryRequire`, which returns any. If there's no way to keep bundlers from bundling these calls - I will figure out another solution how to have both types and keep dependencies from bundling.

packages/core/src/utils/loader.ts
B4nan: `createAutoLoader` is a variable holding an arrow function, not a function itself. 

I wanted to use an actual function signature:

```ts
export function createAutoLoader(...) {
  return createLoaderFactory(...);
}
```

Arrow functions used to be rendered on the reference page in typedoc, and it was far away from usable. But I just checked and apparently, this has changed over time, and arrow functions are now rendered on their separate pages as part of the functions category with the same layout. I still prefer function signatures for readability, but let's keep it here since I understand and makes things less verbose.

B4nan:
>  so in v6 we sure have to keep mikro-orm-esm around

Removing it as part of v6 would be a BC, so that's out of question anyway.

packages/core/src/utils/loader.ts
B4nan: I don't think you can have both. Either it will break bundlers (and will require config changes to mark those as external in users' projects), or it will be typed to `any`.

packages/core/src/utils/Utils.ts
B4nan: > Which one? 

https://github.com/mikro-orm/mikro-orm/blob/8f4790f01b0704cb18d31ff89686ec2eb80b6537/packages/core/src/utils/index.ts#L5

> By the way, is the errors.ts also a barrel file?

Barrel files reexport things, so in that sense, nope, its not. But we reexported all its contents as part of the upper barrel file.

https://github.com/mikro-orm/mikro-orm/blob/18a425c091af0535d6b43dfb797edade58822fd2/packages/core/src/index.ts#L16

Maybe move all of this into the `loaders.ts` file and don't export it at all?

packages/core/src/utils/loader.ts
octet-stream: If there's no way to make it work with the bundlers the way I wan't to, then I can always add types manually :D

Something like

```ts
import type * as tsxApi from "tsx/esm/api"

const tsx: tsxApi = tryModule("tsx/esm/api")
```

This will be worse, but still better than having `any`. Because if the API changes, and I will use something incorrectly - TypeScript will make sure I notice it.

packages/core/src/utils/Utils.ts
octet-stream: Sounds good, I'll move all these new utilities into loaders.ts, and if you'll ever need tryModule or requireDefault, you'll just change the code.

packages/core/src/utils/loader.ts
B4nan: You can't do that either, such type imports would end up in the `.d.ts` files, failing the users' project build unless they use `skipLibsCheck`.

octet-stream: Are they? If my code does not have these types exposed somewhere, like in parameters or return type. I use these only inside of specific loader, I think TS should just strip this import out of the resulting .d.ts file.

For example, this is how .d.ts file looks in my proposal's code, when you build it:
image

As you can see, the 'lilconfig' module is there, because I use it in public-facing type alias. Everything else is removed. Though, I use `tsup` to build this project instead of tsc, because it does not support `.ts` extension in module specifier (it cannot transform) *yet* (in stable release), which is using Microsoft's `api-extractor` to produce .d.ts files, I think tsx itself should work the same. As long as these types used in internally this should not be a problem.

octet-stream: I could not find anything of use, for now, if there's any way to exclude these from bundling. Maybe I should just ask the developers of Vite, Webpack and Rollup. Or I'll just drop this solution, because I don't mind as long as it will *not* break Mikro ORM. It's just nice to have types inferred by TypeScript, rather than writing them by hand.

octet-stream: Found something for Vite: https://vite.dev/guide/ssr.html#ssr-externals

It will not include `@mikro-orm/*` packages (as well as any other package from node_modules) to SSR bundle, but that's just Vite.

octet-stream: I'm gonna rewrite this function as you suggested for backward compatibility, but I'll keep types for better DX. I don't think TS will keep them in `.d.ts`, so this should not be a problem.

But I think this is something to consider for v7:
1. Move configuration loader to its own package. I can see that the current ConfigurationLoader class is used in quite a [few places](https://github.com/search?q=repo%3Amikro-orm%2Fmikro-orm%20configurationloader&type=code). We can also reuse these utilities to implement directory-based discovery and make this package official way do do it (if user's codebase is not using any bundler, see below).
2. Deprecate and remove `MikroORM.init()` without arguments - this way the new package will be part of CLI, or user's code if they want to use directory-based discovery.
3. Write official recipe for directory-based discovery when using MikroORM with bundlers, at least Vite (https://vite.dev/guide/features.html#glob-import) and Webpack (https://webpack.js.org/api/module-methods/#dynamic-expressions-in-import) have their own way to bundle code with dynamic imports. Afaik, it's not even possible to do it with Mikro ORM's own APIs, because bundlers will just ignore utilities, like `Utils.tryRequire` and `Utils.dynamicImport`. If you want to support bundlers, then let's make it official :)

packages/core/src/utils/Utils.ts
octet-stream: Moved this error to loader.ts. I don't think the name matters quite a lot, but I can rename it when I figure out how to test this error's appearance (I think it can be done by mocks if I throw NodeJS.ErrnoError instead of returning the exports?), sure.

packages/core/src/utils/loader.ts
octet-stream: Updated tryModule implementation so it will accept specifier as the first argument, instead of promise returned by `import()`, now it shouldn't be a problem.

B4nan:
> Can I disable or change the rule for this file?

Which one? We use dynamic imports in the codebase, e.g. [here](https://github.com/mikro-orm/mikro-orm/blob/1aa940bbd2b1eeddfb3d662c740b346e72e4d74b/packages/core/src/utils/Utils.ts#L191).

octet-stream:
> Which one?

This one: https://typescript-eslint.io/rules/consistent-type-imports/

The error says following:

```
Error: node_modules/jiti/lib/jiti.d.cts(1,24): error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("./types.js")' call instead.
```

So, from what I understand, I need to replace

```ts
import type * tsxApi from "tsx/esm/api"
```

with

```ts
Type TsxApi = typeof import("tsx/esm/api") // Except, I will use it right in `tryModule` type parameter
```

packages/core/src/utils/loader.ts
octet-stream: > Note that folder-based discovery is one thing, but we still do similar stuff for seeders and migrations. I don't think we can just drop that part, so we won't get rid of those issues completely just by moving the discovery part.

Yes, but aren't they separated packages? Do *they* load responsible for loading migrations and seeders?

> But I was never a huge fan of bundling on backend

I couldn't agree more, but I usually use Mikro ORM in full-stack projects, where both backend and frontend are the same app and it's built with one of the so called "meta frameworks", and all of them use one of the bundlers. Like, Qwik, SvelteKit and React Router v7 (ex Remix) use Vite.

octet-stream:
> So, from what I understand, I need to replace

Nope, that didn't help :)

octet-stream:
`skipLibCheck` helps (I added it to tests/tsconfig.json). Why don't you have this option enabled? At least for tests. It hurts performance when tsc checks everything in `node_modules`, and I don't see you test any `.d.ts` files in tests.

B4nan:
> Why don't you have this option enabled? 

Because I want to see if we do something wrong when integrating 3rd party libraries. AFAIK you can end up with issues without this, that you won't see but people can see if they don't use `skipLibsCheck` in their projects. 

I am fine with having that disabled for tests.

octet-stream:
> Because I want to see if we do something wrong when integrating 3rd party libraries.

I mean, it's not like you can do about 3rd parties.

> AFAIK you can end up with issues without this, that you won't see but people can see if they don't use skipLibsCheck in their projects.

The documentation says this: https://www.typescriptlang.org/tsconfig/#skipLibCheck

I'm not sure what they mean by

> TypeScript will type check the code you specifically refer to in your app’s source code.

Will it verify *public*-facing types from 3rd party libraries, or just the source code?

Anyway. Turns out tsc will also fail when I try and build the packages, with the same error. But at the same time the `loader.ts` is compiled, I can see the result in `dist` directory and as I said, there there's no `jiti` and `tsx` type declarations in the resulting `loader.d.`ts file. Too sad TS still complains about these anyway during compilation. I should probably just use `any` and be done with that (maybe even reuse Utils.dynamicImport). `@ts-ignore` doesn't help, of course, because it tries to check 3rd party .d.ts file.

octet-stream:
Also, I think tsc will apply same tsconfig.json for 3rd parties as in mikro orm which might cause the problem in the first place - 3rd parties may have different configs, but their code is will valid. tsc looks at their .d.cts file, which imports `./types.js` file as type-only import and thinks this is ES module. Weird. Ok, I can just use `any`, if you want to keep `skipLibCheck` disabled. But then we will not be able to catch errors if jiti or tsx change their public API - afaik the types I import will still be validated, tsc will just skip what's inside the imported module. At least I will see the errors if I try to import or use something that doesn't exist in the package I'm referencing in my code (no matter whether skipLibCheck disabled or not).

octet-stream:
I have this error when I'm using imports in my tests:

image

So, Jest does not support esm by default?

Also, this error with mts files:

image

tsc can't pick up mikro orm typings in esm for some reason (even if I add package.json with {type: "module"} and configure tsc, just for fixtures). I don't see such problem when I install it from npm and use in my esm projects (for example - I don't have this problem in my proposal's repo). Weird.

I'll probably skip tests for esm for now, since Mikro ORM seem to be configured as CJS package, but they should work.

octet-stream:
I believe this is also Jest-related issue. I get it only with tsx, though.

image

This probably related: https://github.com/jestjs/jest/issues/9580

I will skip tsx tests for now, and see if I can do anything about it. The loader should still work though. I am not much familiar with Jest, so any help from you will be appreciated.

packages/core/src/utils/ConfigurationLoader.ts
octet-stream: I updated this code, but I can't just use `bool` function for type casting for `loader` option, because it can be a boolean, or a string.

packages/core/src/utils/ConfigurationLoader.ts
```diff
@@ -39,6 +67,9 @@ export class ConfigurationLoader {
    */
   static async getConfiguration<D extends IDatabaseDriver = IDatabaseDriver, EM extends D[typeof EntityManagerType] & EntityManager = EntityManager>(validate: boolean, options?: Partial<Options>): Promise<Configuration<D, EM>>;
   static async getConfiguration<D extends IDatabaseDriver = IDatabaseDriver, EM extends D[typeof EntityManagerType] & EntityManager = EntityManager>(contextName: boolean | string = 'default', paths: string[] | Partial<Options> = ConfigurationLoader.getConfigPaths(), options: Partial<Options> = {}): Promise<Configuration<D, EM>> {
+    const settings = ConfigurationLoader.getSettings();
+    const basePath = options.baseDir ?? process.cwd();
```
octet-stream, line 71: I'm not sure what is the baseDir option for, but the `projectRoot` argument is meant to be configurable (for tests and for use), so I had to get it from somewhere and not just re-use `process.cwd()`.

octet-stream:
@B4nan I might need your help with `ConfigurationLoader.getConfiguration()` testing. I just have no idea how to test it against real config file, because my test case always fails validation at Configuration class saying I should provide `driver` option. I even added test for loaders + Configuration class to make sure that returned file, and it works, but not for getConfiguration (even with the same config file).

Also, I've tried to pass base path by overriding `process.cwd` for my test case (as you can see in the code) and by setting `basePath` option - the result is the same. My initial though was that merge helper doesn't work properly and wipes out driver for some reason if I set `basePath`, but it doesn't seem to be the case.

I know you test against mocks, but I want to make sure my loaders integrated properly so I won't break anything in v6, so I run my tests against real config files.

B4nan:
I wonder if it would be better to target v7 here, I've already set up the branch and I am almost there with ESM rewrite using vitest instead of jest.

B4nan:
Why useless? It adds support for other TS support providers, or did I miss something?

octet-stream:
> Why useless? It adds support for other TS support providers, or did I miss something?

I meant, it would probably need to be rewritten, because it might not fit in its current state in a future, and as I said - it would depend on whether or not it will be put into a separate package. The loaders implementation is completed thought, no matter how it will be integrated - it tested and it should work.

> I am not completely opposed to doing that but so far I haven't seen a good enough reason.

Well, as I already said, I think it might help to improve support for bundlers, if we clear the core package from dynamic import and things like `node:fs` (or fs-extra) usage, plus we can avoid tricks with like you mentioned in your previous comments and have dynamic imports properly typed, like I did in my [proposal](https://github.com/octet-stream/mikro-orm-load-config) repo, because we move all Node.js specific stuff to CLI, and to a separate package for those who don't care about bundlers. Have to say bundlers not a huge concern for me, I'm fine eight way as long as I can help to replace `ts-node` with more modern solutions. :)

octet-stream:
Btw, do we still need ts-node support if we move this to v7 instead?

octet-stream:
I just found out that jiti having problems when it encounters class properties without initializer: https://github.com/unjs/jiti/issues/57

So, I guess this pr is stuck anyway until and if they fix the problem. tsx does not have this problem, but then again - they don't support decorators metadata.

B4nan:
I started testing out the ESM rewrite (#6403) and the ts-node solution we have now is not working at all, so there is your answer - ts-node is going out from v7 right now :] I just swapped it with `@swc-node/register` and it looks good, works fine when registered dynamically, so no need for the ESM bin script. I will only make a simple replacement for now so I can unblock the ESM PR, I would still like to see support for the other solutions you mentioned, as long as it can all stay optional, I am all in for making things less opinionated.

octet-stream:
Glad to hear that. I'll remove ts-node support then, when we will move this PR forward. Hope jiti lans PR with [fix](https://github.com/unjs/jiti/pull/361) for the issue I was talking about in my previous comment, otherwise the only two loaders left will be tsx and native. Or maybe I can replace ts-node loader with swc (as I initially proposed, in addition to other options), as far as I can tell, the implementation will be the same
https://github.com/mikro-orm/mikro-orm/blob/a1469599ddc9465cd9423fb7ee78f546e3d5d1dc/packages/core/src/utils/ConfigurationLoader.ts#L236-L239

octet-stream:
Think I better close this and start over, but for v7 instead, as you proposed, once I have time for it. Of course that if this feature is still needed at all, since you've moved to SWC and Node.js now also supports TS type stripping and transformation.

B4nan:
I surely think it makes sense to let people choose what they want for TS support, SWC sounds like a better default since it works with the legacy decorators and reflection too, but for those who want to use tsmorph or don't want to use decorators at all, using esbuild/tsx/native type stripping or others is a valid choice. Let me know if you'd like to start working on this, I do rebase the v7 branch once in a while but I'll be happy to hold it when there is a PR from other contributors to reduce conflicts.

PR #6352, opened 2025-01-22
fix(schema): fixed handling of check constraints on MariaDB before 10.5.10

### LINKED ISSUE #6350: Database version when using Entity Generator
Using Entity generator on MariaDB throws an error on [this query](https://github.com/labzy88/mikro-orm/blob/85832c5bd064551512572bf04817ed79851fd715/packages/mariadb/src/MariaDbSchemaHelper.ts#L147-L151)

The error "Unknown column 'tc.level' in 'field list'"

I'm running MariaDB v 10.4, 'level' was introduced in 10.5.10 [here](https://mariadb.com/kb/en/information-schema-check_constraints-table/)

At the moment, is it possible to specify a version in the entity generator to accommodate this?

boenrobot:
New minimum version is 5.3.1, when MariaDB specific execution comments were introduced.

Fixes #6350

### DIFF
--- a/packages/mariadb/src/MariaDbSchemaHelper.ts
+++ b/packages/mariadb/src/MariaDbSchemaHelper.ts
@@ -144,11 +144,15 @@ export class MariaDbSchemaHelper extends MySqlSchemaHelper {
   }
 
   protected override getChecksSQL(tables: Table[]): string {
-    return `select tc.constraint_schema as table_schema, tc.table_name as table_name, tc.constraint_name as name, tc.check_clause as expression,
-      case when tc.level = 'Column' then tc.constraint_name else null end as column_name
-      from information_schema.check_constraints tc
-      where tc.table_name in (${tables.map(t => this.platform.quoteValue(t.table_name))}) and tc.constraint_schema = database()
-      order by tc.constraint_name`;
+    return `select
+              tc.constraint_schema as table_schema,
+              tc.table_name as table_name,
+              tc.constraint_name as name,
+              tc.check_clause as expression,
+              /*M!100510 case when tc.level = 'Column' then tc.constraint_name else */ null /*M!100510 end */ as column_name
+            from information_schema.check_constraints tc
+            where tc.table_name in (${tables.map(t => this.platform.quoteValue(t.table_name))}) and tc.constraint_schema = database()
+            order by tc.constraint_name`;
   }
 
   /* istanbul ignore next */


packages/mariadb/src/MariaDbSchemaHelper.ts
```diff
@@ -144,11 +144,15 @@ export class MariaDbSchemaHelper extends MySqlSchemaHelper {
...
-      case when tc.level = 'Column' then tc.constraint_name else null end as column_name
-      from information_schema.check_constraints tc
-      where tc.table_name in (${tables.map(t => this.platform.quoteValue(t.table_name))}) and tc.constraint_schema = database()
-      order by tc.constraint_name`;
+    return `select
+              tc.constraint_schema as table_schema,
+              tc.table_name as table_name,
+              tc.constraint_name as name,
+              tc.check_clause as expression,
+              /*M!100510 case when tc.level = 'Column' then tc.constraint_name else */ null /*M!100510 end */ as column_name
```
B4nan, line 152: for older versions the query wont be valid since you left the `,` in there

B4nan: hmm or it will put `null` in there, ok that can work too

B4nan:
i dont think this is entity generator issue, same will happen if you try to create migrations or use schema generator, so lets change the scope to just `schema`.

packages/mariadb/src/MariaDbSchemaHelper.ts
boenrobot: But there's two comments, not one...

The final query for older versions should be ending up as
```
tc.check_clause as expression,
null as column_name
```

B4nan (approved):

packages/mariadb/src/MariaDbSchemaHelper.ts
```diff
@@ -144,11 +144,15 @@ export class MariaDbSchemaHelper extends MySqlSchemaHelper {
   }
 
   protected override getChecksSQL(tables: Table[]): string {
-    return `select tc.constraint_schema as table_schema, tc.table_name as table_name, tc.constraint_name as name, tc.check_clause as expression,
-      case when tc.level = 'Column' then tc.constraint_name else null end as column_name
-      from information_schema.check_constraints tc
-      where tc.table_name in (${tables.map(t => this.platform.quoteValue(t.table_name))}) and tc.constraint_schema = database()
-      order by tc.constraint_name`;
+    return `select
```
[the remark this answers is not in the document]
boenrobot: It should, yes... As long as there's no delimiter in the query, which is certainly the case here.

PR #6393, opened 2025-02-06
feat(seeder): make `.each()` function receive `index: number` argument

jtomaszewski:
Sometimes I want to dynamically assign different params to my seeded entities, depending on their index.

Example:

```ts
const participants = await new TournamentParticipantFactory(em)
      .each((p, i) => {
        p.seed = i + 1;
      })
      .create(16, {
        tournament,
      });
```

Without this, I have to write something like

```ts
const participants = await new TournamentParticipantFactory(em)
      .create(16, {
        tournament,
      });
participants.forEach((p, i) => { p.seed = i + 1; });
await em.persistAndFlush(participants);
```

This in part resolves https://github.com/mikro-orm/mikro-orm/discussions/3002 .


B4nan (approved):
Thanks!

PR #6396, opened 2025-02-06
feat(seeder): pass params to `.definition()` function as argument

jtomaszewski:
Some of my entities have a lot of attributes, that can have a lot of combinations. It would be nice if I could pass a property to the `.definition()` method (when making or creating an entity), so that I can dynamically define how a generated entity should look like.

Example usage:

```ts
export class TournamentFactory extends Factory<
  TournamentEntity,
  EntityData<TournamentEntity> & { groupsCount?: number }
> {
  model = TournamentEntity;

  async definition(
    params?: EntityData<TournamentEntity> & { groupsCount?: number }
  ): EntityData<TournamentEntity> {
    return {
      name: "XXX Olympic Games",
      ...params,
      groups: params.groups ?? [...Array(params?.groupsCount ?? 0)].map((v, i) =>
        new TournamentGroupFactory(this.em).makeEntity({
          number: i + 1,
        })
      ),
    };
  }
}

// Final usage

new TournamentFactory(em).createOne({ groupsCount: 4 }) // creates a tournament with four groups
```

Without this, I currently have :

- to create dozens of classes (e.g. `FourGroupsTournamentFactory` `EightGroupsTournamentFactory`), or
- use `.each()`, but that doesn't scale, as I'd prefer to keep the above logic written only once, in my factory class, or
- do some wrappers around the factory, like `createTournament(em, { groupsCount: 4 })`, which becomes a factory-wrapper for another factory.

### DIFF
--- a/docs/docs/seeding.md
+++ b/docs/docs/seeding.md
@@ -295,7 +295,7 @@ export class AuthorFactory extends Factory<
   }
 }
 
-# Finally
+// Finally
 new AuthorFactory(em).createOne({ booksCount: 4 })
 ```
 
--- a/docs/docs/seeding.md
+++ b/docs/docs/seeding.md
@@ -243,7 +243,11 @@ const authors = await new AuthorFactory(orm.em).create(5, {
 
 ### Factory relationships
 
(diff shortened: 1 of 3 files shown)


B4nan (approved):
lets mention this in the docs too

jtomaszewski:
Good point. Updated the docs.

PR #6400, opened 2025-02-08
feat: replace knex with kysely

B4nan:
Closes #5646


packages/knex/src/AbstractSqlConnection.ts
```diff
@@ -72,15 +73,23 @@ export abstract class AbstractSqlConnection extends Connection {
    * @inheritDoc
    */
   async checkConnection(): Promise<{ ok: true } | { ok: false; reason: string; error?: Error }> {
+    if (!this.connected) {
+      return { ok: false, reason: 'Connection not established' };
+    }
+
     try {
-      await this.getKnex().raw('select 1');
+      await this.client.executeQuery(CompiledQuery.raw('select 1'));
```
lookfirst, line 81: Stupid minor, but you could save calling a whole bunch of Object.freeze()'s by caching the .raw() call.

B4nan: Does it really matter? This is used once during startup when the database connection is established (plus when user invokes `orm.checkConnection()`).

lookfirst: That's why I said `stupid minor`...

PR #6403, opened 2025-02-10
refactor: migrate to native ESM

B4nan:
MikroORM v7 is a native ESM package now. It can be still consumed from a CJS project, as long as you use TypeScript and Node.js version that supports `require(esm)`.

Also, `jest` has been replaced with `vitest`.

BREAKING CHANGES:
- TypeScript 5.8+ required. Older versions might work too, but only if your project is also ESM.
- TypeScript support was previously provided by `ts-node`, which is now replaced with `@swc-node`.
- `mikro-orm-esm` CLI script is removed, use the standard `mikro-orm` script instead.
- `tsNode` config option removed in favor of `preferTs`
- `useTsNode` is renamed to `preferTs`


packages/cli/src/CLIConfigurator.ts
```diff
@@ -1,15 +1,15 @@
...
+import { ClearCacheCommand } from './commands/ClearCacheCommand.js';
+import { CreateDatabaseCommand } from './commands/CreateDatabaseCommand.js';
+import { CreateSeederCommand } from './commands/CreateSeederCommand.js';
+import { DatabaseSeedCommand } from './commands/DatabaseSeedCommand.js';
+import { DebugCommand } from './commands/DebugCommand.js';
+import { GenerateCacheCommand } from './commands/GenerateCacheCommand.js';
+import { GenerateEntitiesCommand } from './commands/GenerateEntitiesCommand.js';
+import { ImportCommand } from './commands/ImportCommand.js';
+import { MigrationCommandFactory } from './commands/MigrationCommandFactory.js';
+import { SchemaCommandFactory } from './commands/SchemaCommandFactory.js';
```
octet-stream, line 12: Perhaps not of a much importance: TS supports rewriting `.ts` extensions to `.js` for relative paths during compilation since 5.7, and since Node.js has TS support (they just strip types, so not all syntax is supported, but this is not important since you use tsc anyway) out of the box, maybe it would be better to use `.ts` extension in specifiers when importing TypeScript modules?

See:
https://www.typescriptlang.org/tsconfig/#rewriteRelativeImportExtensions
https://nodejs.org/api/typescript.html

packages/cli/src/cli.ts
```diff
@@ -1,17 +1,10 @@
...
-  .style('yellow', 'required')
-  .helpStyle('green')
-  .errorsStyle('red');
-
-import { CLIHelper } from './CLIHelper';
-import { CLIConfigurator } from './CLIConfigurator';
+import { CLIHelper } from './CLIHelper.js';
+import { CLIConfigurator } from './CLIConfigurator.js';
 
 void (async () => {
```
octet-stream, line 6: No need for async IIFE in ESM btw :)
For CLI it would be fine.

packages/cli/src/CLIConfigurator.ts
B4nan: This won't have any effect, right? I am happy to make the switch for a reason, so lets leave that for when we have one, changing it wont be breaking anyhow, so it can happen anytime.

octet-stream: Sure, do as you think is best. And, yeah, I don't think it's gonna have any effect, it will just output .ts extension as .js.

PR #6416, opened 2025-02-13
feat(core): improve typings for `@Entity`, `@Index` and `@Unique` decorators

cAttte:
hi! this PR improves the typings for these decorator options:
- `@Entity` → `discriminatorColumn`
- `@Index`, `@Unique` → `properties`

so that they only accept property keys from the decorated class, instead of just any string.

this particular example is pretty cool, as it shows how this supersedes the runtime [`unknownIndexProperty`](https://github.com/mikro-orm/mikro-orm/blob/a6c48fd1b436edcd2bcead74ebfec42bb65da043/packages/core/src/errors.ts#L214) error (at least in simple cases where every property is a @​Property):

```ts
@Unique({ properties: ["platform_id" }) // Type '"platform_id"' is not assignable to type 'keyof Thing'. Did you mean '"platformID"'?
@Unique({ properties: ["platform", "platformID"] }) // good
class Thing { id: number, platform: string, platformID: number }
```

there are minor side effects to this change (which i'm happy to explain), but i don't think they should be a problem

### DIFF
--- a/packages/core/src/decorators/Entity.ts
+++ b/packages/core/src/decorators/Entity.ts
@@ -1,6 +1,6 @@
 import { MetadataStorage } from '../metadata';
 import { Utils } from '../utils';
-import type { Constructor, Dictionary, EntityClass, FilterQuery } from '../typings';
+import type { AnyString, Constructor, Dictionary, EntityClass, FilterQuery } from '../typings';
 import type { FindOptions } from '../drivers/IDatabaseDriver';
 
 export function Entity<T extends EntityClass<unknown>>(options: EntityOptions<T> = {}) {
@@ -19,7 +19,7 @@ export type EntityOptions<T> = {
   tableName?: string;
   schema?: string;
   collection?: string;
-  discriminatorColumn?: T extends EntityClass<infer P> ? keyof P : string;
+  discriminatorColumn?: (T extends EntityClass<infer P> ? keyof P : string) | AnyString;
   discriminatorMap?: Dictionary<string>;
   discriminatorValue?: number | string;
   forceConstructor?: boolean;
--- a/packages/core/src/decorators/Entity.ts
+++ b/packages/core/src/decorators/Entity.ts
@@ -1,19 +1,17 @@
 import { MetadataStorage } from '../metadata';
 import { Utils } from '../utils';
-import type { Constructor, Dictionary, FilterQuery } from '../typings';
+import type { Constructor, Dictionary, EntityClass, FilterQuery } from '../typings';
 import type { FindOptions } from '../drivers/IDatabaseDriver';
 
-export function Entity(options: EntityOptions<any> = {}) {
-  return function <T>(target: T & Dictionary) {
+export function Entity<T extends EntityClass<unknown>>(options: EntityOptions<T> = {}) {
+  return function (target: T) {
     const meta = MetadataStorage.getMetadataFromDecorator(target);
     Utils.mergeConfig(meta, options);
-    meta.class = target as unknown as Constructor<T>;
+    meta.class = target;
 
     if (!options.abstract || meta.discriminatorColumn) {
       meta.name = target.name;
     }
-
-    return target;
   };
 }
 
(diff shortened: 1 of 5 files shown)


cAttte:
oops sorry, completely forgot to check the tests/tsc. did the Entity part there, so now i'm working on the dot path support. the type itself is gonna be a little monstruous but i think it's gonna be worth it to have exhaustive auto-complete (where possible, otherwise just default to `string` ofc)

cAttte:
here is my draft for this utility type. the `Keyof` is there to exclude array indices, and you could add other specific types with internal properties. i excluded properties of certain (non-JSON-in-DB) types altogether (functions, collections, refs) to reduce false positives, but it can never be 100% strict because of e.g. `class Book { author: Author }`

```ts
type Keyof<T> = T extends keyof any[] ? never : keyof T;
// `Visited` is there to avoid infinite recursion, but i'm not sure how effective it is in real circular structures
type Path<T, Prev extends string = '', Visited extends unknown[] = []> = T extends object
  ? T extends Visited[number]
    ? never
    : {
        [K in Keyof<T> & string]: T[K] extends Function | Collection<any> | Ref<any>
          ? never
          : Prev extends ''
            ? `${K}` | Path<T[K], `${K}`, [T, ...Visited]>
            : `${Prev}.${K}` | Path<T[K], `${Prev}.${K}`, [T, ...Visited]>;
      }[Keyof<T> & string]
  : never;

class TestClass { myProp: { some: number; and: { a: { path: Record<string, boolean> } } } }
type Test = Path<TestClass> // "myProp" | "myProp.some" | "myProp.and" | "myProp.and.a" | "myProp.and.a.path" | `myProp.and.a.path.${string}`
```

if you think this type is too much, we could of course just do `Prop | ˋ${Prop}.${string}ˋ` to still keep some amount autocompletion and strictness. we can also add `| AnyString` to (a) fall back in case of untyped properties (which i wouldn't recommend but oh well) and (b) simplify the display type (in errors, etc) to `Path<MyEntity>` instead of `"a" | "b" | "c" | ...`. let me know

B4nan:
We don't need a new type for this, we have `AutoPath`.

cAttte:
great, i added AutoPath there. but this breaks every other usage of `IndexOptions` and `UniqueOptions` in the codebase ([e.g.](https://github.com/mikro-orm/mikro-orm/blob/996503cdd856f830583629ed16c17d539c796fe6/packages/knex/src/schema/DatabaseTable.ts#L242)), because of that `P` / `Hint` type parameter (is that there to avoid generating the union of every single possible path? also am i even using it correctly?). other than lifting this type param up, do you think there's a better way e.g. adding a type param that just skips checking `properties`?

B4nan:
> . but this breaks every other usage of IndexOptions and UniqueOptions in the codebase 

You'll need to address this in the codebase, those are internal places that did not expect the option to be type safe. No way around that. I don't mind some casting or `any` inside the code that is not affecting users.

Also, you say _every other_, but I can see only ~10 places, that is really trivial to fix by hand.

> is that there to avoid generating every single possible path

The type does not do that, that's why it has the type argument, so it doesn't need to. Doing so is a bad idea perf-wise, there are cycles in most of the entity definitions.

packages/core/src/decorators/Entity.ts
```diff
@@ -1,27 +1,25 @@
 import { MetadataStorage } from '../metadata';
 import { Utils } from '../utils';
-import type { Constructor, Dictionary, FilterQuery } from '../typings';
+import type { AnyString, Constructor, Dictionary, EntityClass, FilterQuery } from '../typings';
 import type { FindOptions } from '../drivers/IDatabaseDriver';
 
-export function Entity(options: EntityOptions<any> = {}) {
-  return function <T>(target: T & Dictionary) {
```
B4nan, line 7: this is actually tiny bit breaking, you are removing the type argument people are likely already using, we should leave it there and drop it in v7

B4nan: also, the `@Filter`, `@Check` and `@Formula` decorators should be updated to do the same

B4nan (approved):
I've played a bit with the other decorators but I wasn't able to find a way to make things work without BCs, so will leave that for v7.

Anyway, thanks for this, I feel dumb for not knowing this was possible, always thought it requires ES decorators for the added typesafety 🙃

cAttte:
sorry i've been busy, yeah luckily the decorator types are expressive enough to allow for this sort of thing as well as e.g. only allowing class targets, good luck with the other decorators. and thanks a lot for fixing the errors & merging!

PR #6424, opened 2025-02-16
feat: add entityname as argument for filters

signalize:
This PR adds an extra argument to the filters condition function.

When you execute a relational query, a filter function is executed multiple times. By using the entity name, you can deduce which layer the filter is applied to.


B4nan (approved):
thanks

PR #6466, opened 2025-02-28
feat(entity-generator): pass through `orderBy` option to the output

jtomaszewski:
I am using entity generator to maintain source code of my entities, and also I'd like to provide `orderBy` option for some of my collections.

Currently one can alter a lot of entities' code through the `onProcessedMetadata` callback - but just not the `orderBy` option, as it is never saved to the generated code output.

This PR fixes it - allowing to specify `orderBy` option for some of the relations in the metadata callbacks, so that it's later placed in the generated entities code.

### DIFF
--- a/packages/entity-generator/src/SourceFile.ts
+++ b/packages/entity-generator/src/SourceFile.ts
@@ -770,6 +770,10 @@ export class SourceFile {
     this.entityImports.add(prop.type);
     options.entity = `() => ${prop.type}`;
 
+    if (prop.orderBy) {
+      options.orderBy = JSON.stringify(prop.orderBy);
+    }
+
     if (prop.mappedBy) {
       options.mappedBy = this.quote(prop.mappedBy);
       return;
@@ -808,6 +812,9 @@ export class SourceFile {
     this.entityImports.add(prop.type);
     options.entity = `() => ${prop.type}`;
     options.mappedBy = this.quote(prop.mappedBy);
+    if (prop.orderBy) {
+      options.orderBy = JSON.stringify(prop.orderBy);
+    }
   }
 
   protected getEmbeddedPropertyDeclarationOptions(options: Dictionary, prop: EntityProperty) {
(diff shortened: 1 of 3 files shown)


B4nan (changes requested):
thanks, looking good, but let's make sure the output doesn't look messy as with `JSON.stringify`

packages/entity-generator/src/SourceFile.ts
```diff
@@ -770,6 +770,10 @@ export class SourceFile {
     this.entityImports.add(prop.type);
     options.entity = `() => ${prop.type}`;
 
+    if (prop.orderBy) {
+      options.orderBy = JSON.stringify(prop.orderBy);
```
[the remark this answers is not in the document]
jtomaszewski: Sure. Didn't know what function to use here, thanks for the recommendation. Fix pushed

packages/entity-generator/src/SourceFile.ts
```diff
@@ -808,6 +813,9 @@ export class SourceFile {
     this.entityImports.add(prop.type);
     options.entity = `() => ${prop.type}`;
     options.mappedBy = this.quote(prop.mappedBy);
+    if (prop.orderBy) {
+      options.orderBy = JSON.stringify(prop.orderBy);
```
B4nan, line 817: same here

packages/entity-generator/src/SourceFile.ts
```diff
@@ -22,6 +22,7 @@ import {
 } from '@mikro-orm/core';
 import { parse, relative } from 'node:path';
 import { POSSIBLE_TYPE_IMPORTS } from './CoreImportsHelper';
+import { inspect } from 'node:util';
```
B4nan, line 25: please put this one above the relative import (so just one line above)

B4nan (changes requested):

B4nan (approved):

PR #6467, opened 2025-02-28
fix(core): respect Collection property `orderBy` when dataloader is enabled

jtomaszewski:
After adding default `orderBy` to my collection in the entity class `@OneToMany(...)` decorator, I noticed that my collections are sometimes sorted correctly and sometimes not at all.

Upon debugging it appeared that this happens only if `dataloader` is enabled (and it is, in my case).

This PR fixes it, making the `Collection#loadItems` load items in the default order (rather than none) also when dataloader is enabled. (Currently, it happens only when it is disabled.)

### DIFF
--- a/packages/core/src/entity/Collection.ts
+++ b/packages/core/src/entity/Collection.ts
@@ -300,7 +300,7 @@ export class Collection<T extends object, O extends object = object> extends Arr
     if (options.dataloader ?? [DataloaderType.ALL, DataloaderType.COLLECTION].includes(DataloaderUtils.getDataloaderType(em.config.get('dataloader')))) {
       const order = [...this.items]; // copy order of references
       const orderBy = this.createOrderBy(options.orderBy);
-      const customOrder = !!orderBy;
+      const customOrder = orderBy.length > 0;
       // eslint-disable-next-line dot-notation
       const items: TT[] = await em['colLoader'].load([
         this,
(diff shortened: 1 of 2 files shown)


tests/features/collection/collection-order.test.ts
```diff
@@ -0,0 +1,79 @@
+import { Collection, Entity, ManyToOne, MikroORM, OneToMany, PrimaryKey, Property, SimpleLogger } from '@mikro-orm/better-sqlite';
```
B4nan, line 1: please use `sqlite` driver so it's easier to rebase this to v7 (which removed the `better-sqlite` driver and is using better-sqlite3 in the `sqlite` one)

tests/features/collection/collection-order.test.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+
+let orm: MikroORM;
+
+beforeAll(async () => {
+  orm = await MikroORM.init({
+    entities: [Author],
+    dbName: ':memory:',
+    loggerFactory: SimpleLogger.create,
```
B4nan, line 48: this is not used anywhere, so can be omitted

B4nan (changes requested):
thanks! 

just two nits

B4nan (approved):

packages/core/src/entity/Collection.ts
```diff
@@ -299,9 +299,13 @@ export class Collection<T extends object, O extends object = object> extends Arr
 
     if (options.dataloader ?? [DataloaderType.ALL, DataloaderType.COLLECTION].includes(DataloaderUtils.getDataloaderType(em.config.get('dataloader')))) {
       const order = [...this.items]; // copy order of references
-      const customOrder = !!options.orderBy;
+      const orderBy = this.createOrderBy(options.orderBy);
+      const customOrder = !!orderBy;
```
B4nan, line 303: this method always returns an array, so you cant just do `!!orderBy`, it will always be true

```suggestion
      const orderBy = this.createOrderBy(options.orderBy);
      const customOrder = orderBy.length > 0;
```

(uncovered by the coverage report)

B4nan (approved):

PR #6489, opened 2025-03-07
feat(mssql): add support for raw query fragments with TVPs

patrykbojczuk:
This adds support for passing table-valued parameters (TVPs) to raw query fragments when using MsSql driver.
Currently, objects used to construct TVP by tedious are stringified to JSON as any other object.

Table-valued parameters allow multi-column comparisons without the need for dynamic generation of conditions with ORs between, and makes the query shorter, in some cases improving the performance.

I know that knex dependency removal is planned for the next major version, but the changes here are very simple, so I think it can be easily rewritten for the new knex-less solution.

### DIFF
--- a/packages/core/src/platforms/Platform.ts
+++ b/packages/core/src/platforms/Platform.ts
@@ -494,6 +494,18 @@ export abstract class Platform {
     return value;
   }
 
+  shouldBindParamWithKnex(value: any): boolean {
+    return false;
+  }
+
+  formatValue(value: any): string {
+    if (this.shouldBindParamWithKnex(value)) {
+      return '?';
+    }
+
+    return this.quoteValue(value);
+  }
+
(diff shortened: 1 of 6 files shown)


B4nan:
Thanks for the PR, but I am not really sure if this is something to be merged, at least not into v6. The changes are quite hacky, e.g. the `patchConfig`, and depend on knex (`shouldBindParamWithKnex`), which won't work at all in v7 most likely. Also, I really don't like the idea of introducing `formatValue` on such a low level because of such a niche feature of a single driver.

Feel free to try to PR this into v7, once we have a working solution there, we can talk about getting it into v6. But the solution cannot depend on what kysely/knex is doing on lower levels. 

I am honestly not sold on the need for this, is this a common practice in SQL Server? It feels quite unnecessary.

> I know that knex dependency removal is planned for the next major version

Not just planned, it's already gone in v7 branch.

PR #6492, opened 2025-03-07
fix(core): improve TS detection logic

### LINKED ISSUE #6491: Regression: v6.4.9 breaks tsx support
### Describe the bug

Up until now, using `tsx` (instead of `ts-node`) was working great. The latest version just broke this though, so using `tsx` now no longer enables TS support.

The regression commit is this one:
https://github.com/mikro-orm/mikro-orm/commit/3efdcd0a00d038b2eb24a668329f4b1cea46b2a2

### Reproduction

Try to run a project with `tsx` instead of `ts-node`.

None

### MikroORM version

v6.4.9

### Node.js version

v22.14.0

### Operating system

_No response_

### Validations

B4nan:
Closes #6491

### DIFF
--- a/packages/core/src/utils/Utils.ts
+++ b/packages/core/src/utils/Utils.ts
@@ -800,8 +800,12 @@ export class Utils {
       || !!process.env.TS_JEST // check if ts-jest is used (works only with v27.0.4+)
       || !!process.env.VITEST // check if vitest is used
       || !!process.versions.bun // check if bun is used
-      || process.argv.slice(1).some(arg => arg.includes('ts-node')) // registering ts-node runner
-      || process.execArgv.some(arg => arg === 'ts-node/esm'); // check for ts-node/esm module loader
+      || process.argv.slice(1).some(arg => arg.match(/\.([mc]?ts|tsx)$/)) // executing `.ts` file
+      || process.execArgv.some(arg => {
+        return arg.includes('ts-node') // check for ts-node loader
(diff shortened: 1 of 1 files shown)


packages/core/src/utils/Utils.ts
```diff
@@ -800,8 +800,12 @@ export class Utils {
       || !!process.env.TS_JEST // check if ts-jest is used (works only with v27.0.4+)
       || !!process.env.VITEST // check if vitest is used
       || !!process.versions.bun // check if bun is used
-      || process.argv.slice(1).some(arg => arg.includes('ts-node')) // registering ts-node runner
-      || process.execArgv.some(arg => arg === 'ts-node/esm'); // check for ts-node/esm module loader
+      || process.argv.slice(1).some(arg => arg.endsWith('.ts')) // executing `.ts` file
```
B4nan, line 803: @DASPRiD what do you think about this bit? not sure if it can be dangerous, but it seems quite good, we would also detect if people try to use native node stripping on a TS file this way, and it would be up to them to decide if its working for them (it wont parse decorators, that was the reason to change this logic in the first place, but it can work with `EntitySchema`)

DASPRiD: Hmmm, this looks a little fragile in my opinion. There are a lot more typescript extensions to consider here I think. From the top of my head I can think of "tsx", "mts" and "cts". So maybe `/\.(tsx|[mc]?ts)$/.test(arg)`?

PR #6500, opened 2025-03-11
docs: fix typo in Foreign keys and `forceObject` title

thomasgoldstein:
Very minor fix.
I believe the title "Foreign keys are `forceObject`" must be a typo, because it does not make sense to me.
I think the intended title is "Foreign keys _and_ `forceObject`".


B4nan (approved):
thanks!

PR #6501, opened 2025-03-11
fix(postgres): cockroachdb may not require unmarshallArray

CAJ2:
This PR fixes an issue I had when using the migration tool using CockroachDB.

When running migration:create or similar, any existing enums are fetched from the database. For some reason with CockroachDB, Knex or the driver automatically deserializes the array, so `unmarshallArray` will fail.


B4nan (approved):
thanks!

PR #6517, opened 2025-03-17
fix(mongo): do not create migrations collection by default

amadeus-torwell:


B4nan (changes requested):
you will also need to remove this line https://github.com/mikro-orm/mikro-orm/blob/918f05ac1c721ef7cd315c32167c6c24a938ce5d/tests/features/schema-generator/SchemaGenerator.mongo.test.ts#L24

B4nan (approved):
thanks!

PR #6536, opened 2025-03-20
feat(core): add `includeCount` option to `em.findByCursor`

Nexfader:
This PR restores the changes from the original PR #4626 (which were possibly lost due to a bad rebase) and resolves merge conflicts with the current master branch.

As B4nan suggested in the original PR, I've made the changes to the Cursor constructor parameters non-breaking, so the feature can be released under the current major version.

P.S. Thanks for the awesome ORM! :)


packages/core/src/utils/Cursor.ts
```diff
@@ -64,12 +64,13 @@ export class Cursor<
...
   readonly hasPrevPage: boolean;
   readonly hasNextPage: boolean;
+  readonly totalCount?: number;
 
   private readonly definition: (readonly [EntityKey<Entity>, QueryOrder])[];
 
   constructor(
     readonly items: Loaded<Entity, Hint, Fields, Excludes>[],
-    readonly totalCount: number,
+    totalCount: number | undefined,
```
B4nan, line 73: maybe we could make it work conditionally in `em.findByCursor`, with method overloads i guess, so without the new `includeCount` option being set to `false` explicitly, we would return `Cursor & { totalCount: number }`. i still feel like this is a tiny bit breaking, not functionally, but on type level

B4nan: alternatively, i would prefer keeping the type in the property as mandatory, since that is the default behaviour

Nexfader: Resolved the issue by making totalCount conditional through a generic parameter, so the changes became non-breaking at all

B4nan (approved):
thanks!

packages/core/src/EntityManager.ts
```diff
@@ -652,9 +657,15 @@ export class EntityManager<Driver extends IDatabaseDriver = IDatabaseDriver> {
...
-
-    return new Cursor(entities, count, options, this.metadata.get(entityName));
+    const [entities, count] = options.includeCount !== false
+      ? await em.findAndCount(entityName, where, options)
+      : [await em.find(entityName, where, options)];
+    return new Cursor(
+      entities,
+      count as IncludeCount extends true ? number : undefined,
+      options,
+      this.metadata.get(entityName)
```
B4nan, line 667:
```suggestion
      this.metadata.get(entityName),
```

PR #6564, opened 2025-04-08
Fix entity path discovery for bun 1.2.7

tuaki:
In bun 1.2.7 the error stack trace (used in `lookupPathFromDecorator`) was changed (see the image below, the line starting with `--` was dropped ). Therefore, mikro-orm doesn't work on this bun version (and the following versions). This PR should fix it. Works for both 1.2.6 and 1.2.7.

image

### DIFF
--- a/packages/core/src/utils/Utils.ts
+++ b/packages/core/src/utils/Utils.ts
@@ -825,7 +825,8 @@ export class Utils {
     // but those are also present in node, so we need to check this only if they weren't found.
     if (line === -1) {
       // here we handle bun which stack is different from nodejs so we search for reflect-metadata
-      const reflectLine = stack.findIndex(line => Utils.normalizePath(line).includes('node_modules/reflect-metadata/Reflect.js'));
+      // Different bun versions might have different stack traces. The "last index" works for both 1.2.6 and 1.2.7.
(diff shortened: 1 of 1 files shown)


mariansam:
To add more context, users were complaining [on Discord](https://discord.com/channels/1214904142443839538/1355884747477352538/1358897803090333737) that in Bun versions 1.2.7 and 1.2.8, MikroORM throws with error "Source file './relative/path/to/my/entity.ts' not found" during entities discovery. The reason is the discovery is done using error stack traces which changed since Bun 1.2.7 (see picture above). We've created a minimal reproduction with steps to easily test the bug and this fix: https://github.com/tuaki/mikro-orm-bug-bun-1.2.7. I can confirm the fix works in Bun versions 1.2.6 to 1.2.8. Haven't tested it in Node or Deno.

B4nan (approved):
would be great to have some tests for this in https://github.com/mikro-orm/mikro-orm/blob/master/tests/Utils.test.ts#L359

I'll merge it right ahead since this doesn't introduce any regression in the code coverage and everything is still passing, but a followup PR with more tests for what did change in bun would be more than welcome

Anyway, thanks!

tuaki:
Thanks for the quick merge! The followup PR is here: #6569.

PR #6569, opened 2025-04-09
test: lookupPathFromDecorator for bun 1.2.7

tuaki:
Adds test for the entity path discovery fix for bun 1.2.7 (merged in #6564).

I have added the current stack trace for `bun run`. I am not sure about the other use cases (e.g., `bun build`) in bun 1.2.7, but they work quite differently so they shouldn't be affected by this change.


B4nan (approved):
thanks!

PR #6572, opened 2025-04-09
fix(postgres): fix schema diffing for `default` and `nullable` of check constraint enums

### LINKED ISSUE #6571: When setting a default value for an enum field or making the enum field nullable after it was added to a table, migrations cannot pick it up
### Describe the bug

Originally reported in https://github.com/mikro-orm/mikro-orm/discussions/6570

- When a new entity with an enum field is created and migrations are generated, the migration is correct for the enum field.
- If later down the road we want to add a default value to the enum field, I can modify the entity but trying to generate a second migration doesn't work, I get "schema is up-to-date" message.
- Another observation is that trying to make the enum field nullable down the road also doesn't generate a second migration.

### Reproduction

https://github.com/mahieyin-rahmun/reproduction

@mikro-orm/postgresql

### MikroORM version

@mikro-orm/postgresql@6.4.12 -> ./node_modules/.pnpm/@mikro-orm+postgresql@6.4.12_@mikro-orm+core@6.4.12/node_modules/@mikro-orm/postgresql

### Node.js version

Node v20.18.1, typescript@5.4.5

### Operating system

Ubuntu 24.04.1 LTS

### Validations

mahieyin-rahmun:
- Add functionality to handle default values and nullable properties in the PostgreSqlSchemaHelper.
- Update tests to verify changes, including scenarios for adding/removing default values and nullable constraints in enum fields.

Closes #6571

### DIFF
--- a/packages/postgresql/src/PostgreSqlSchemaHelper.ts
+++ b/packages/postgresql/src/PostgreSqlSchemaHelper.ts
@@ -450,6 +450,14 @@ export class PostgreSqlSchemaHelper extends SchemaHelper {
         return table.specificType(column.name, column.type);
       }
 
+      if (changedProperties.has('default')) {
+        return column.default ? table.specificType(column.name, column.type).defaultTo(column.default) : table.specificType(column.name, column.type);
+      }
+
+      if (changedProperties.has('nullable')) {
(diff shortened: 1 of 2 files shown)


B4nan (approved):
thanks!

looking good, as long as the CI is green

PR #6580, opened 2025-04-11
fix(core): make `populate` parameter of `em.populate()` readonly

### LINKED ISSUE #6579: Add `readonly` to the type for the `populate` parameter in `EntityManager.populate`
I was trying to use the `populate` value of a `FindOptions` object with `em.populate`, but TypeScript threw an error since the types mismatched.

In the implementation, `EntityManager.populate` only uses the `populate` parameter as a value for `FindOptions`, so I can't see a reason why it couldn't be `readonly`.

Example:

```ts
public async specificFindQuery(options: FindOptions<User>): Promise<User[]> {
  const users = await this.em.createQueryBuilder(User).select('*').getResult();
  if (options.populate) {
    await this.em.populate(users, options.populate);
  }
  return users;
}
```

This gives me the following error on the `options.populate` block:

```
Argument of type 'readonly never[]' is not assignable to parameter of type 'false | never[]'.
  The type 'readonly never[]' is 'readonly' and cannot be assigned to the mutable type 'never[]'.ts(2345)
```

As described in the issue title, my proposal is to make the type `readonly`, which would (should) not impact the functionalty if I'm not mistaken

Tlepel:
Closes #6579

### DIFF
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -1875,7 +1875,7 @@ export class EntityManager<Driver extends IDatabaseDriver = IDatabaseDriver> {
     Hint extends string = never,
     Fields extends string = '*',
     Excludes extends string = never,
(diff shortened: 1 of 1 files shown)


B4nan (approved):

Tlepel:
Fair enough, I wasn't sure whether it'd be a feat or fix

Issue #6584, opened 2025-04-13
[Easy fix] Bug in tryCache resultCache when the cached result is `null`


belmeopmenieuwesim:
Currently the `tryCache` in the EntityManager does not handle the case were the result of `em.resultCache.get` is `null`, which is not a cache miss, but can be a cache hit, a valid cache result. Only `undefined` should be treated as a cache miss.  This is btw only an issue for the em.findOne variants, empty arrays work fine (e.g. em.find etc.).

https://github.com/mikro-orm/mikro-orm/blob/86910191c9829a6a8552ce0d92d8603be9d402a2/packages/core/src/EntityManager.ts#L2266-L2304

This causes a massive performance degradation for `null` cached results in very common use-cases.

Luckily this is very easy to fix.

@B4nan What do you think? Was this one on purpose or should we fix this?

B4nan:
I guess you are right. Are you up for a PR?

PR #6585, opened 2025-04-13
fix(result-cache) properly handle valid `null` cache results

belmeopmenieuwesim:
Initial proposal to fix https://github.com/mikro-orm/mikro-orm/issues/6584


B4nan (approved):
thanks!

PR #6591, opened 2025-04-17
docs:  remove mention for manual typescript config in `package.json`

kei-mp:
Change was made in this PR: https://github.com/mikro-orm/mikro-orm/pull/5650 - the code sample was updated but not this sentence. Great package by the way, I'm new coming from EF Core that is also based on Data Mapper and really enjoying it so far.

I've gone through the entire guide now and don't see any other mentions of `mikro-orm` in `package.json`


B4nan (approved):
thanks!

PR #6600, opened 2025-04-21
fix(core): add support for zero count result cache

belmeopmenieuwesim:
Result cache for count queries that resulted in `0` should not make a DB call. This PR fixes this.


B4nan (commented):
what about empty string?

B4nan (approved):
thanks!

PR #6618, opened 2025-04-28
fix(mssql): handle mapping for order by expressions with nulls first/last

SteadEXE:
This PR fixes an issue with order by with nulls first and nulls last

### DIFF
--- a/packages/mssql/src/MsSqlPlatform.ts
+++ b/packages/mssql/src/MsSqlPlatform.ts
@@ -250,30 +250,18 @@ export class MsSqlPlatform extends AbstractSqlPlatform {
   }
 
   override getOrderByExpression(column: string, direction: QueryOrder): string[] {
-    const ret: string[] = [];
-
     switch (direction.toUpperCase()) {
-      case QueryOrder.ASC:
-        ret.push(`${column} asc`);
-        break;
-      case QueryOrder.DESC:
-        ret.push(`${column} desc`);
-        break;
       case QueryOrder.ASC_NULLS_FIRST:
-        ret.push(`case when ${column} is null then 0 else 1 end, ${column} asc`);
-        break;
+        return `case when ${column} is null then 0 else 1 end, ${column} asc`;
       case QueryOrder.ASC_NULLS_LAST:
-        ret.push(`case when ${column} is null then 1 else 0 end, ${column} asc`);
-        break;
+        return `case when ${column} is null then 1 else 0 end, ${column} asc`;
       case QueryOrder.DESC_NULLS_FIRST:
-        ret.push(`case when ${column} is null then 0 else 1 end, ${column} desc`);
-        break;
+        return `case when ${column} is null then 0 else 1 end, ${column} desc`;
       case QueryOrder.DESC_NULLS_LAST:
-        ret.push(`case when ${column} is null then 1 else 0 end, ${column} desc`);
-        break;
+        return `case when ${column} is null then 1 else 0 end, ${column} desc`;
+      default:
+        return `${column} ${direction.toLowerCase()}`;
     }
-
-    return ret;
   }
 
 }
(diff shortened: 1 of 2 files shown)


packages/mssql/src/MsSqlPlatform.ts
```diff
@@ -248,4 +249,31 @@ export class MsSqlPlatform extends AbstractSqlPlatform {
...
+        break;
+      case QueryOrder.DESC_NULLS_FIRST:
+        ret.push(`case when ${column} is null then 0 else 1 end, ${column} desc`);
+        break;
+      case QueryOrder.DESC_NULLS_LAST:
+        ret.push(`case when ${column} is null then 1 else 0 end, ${column} desc`);
+        break;
+    }
+
+    return ret;
```
B4nan, line 276: since you never use more than the one item in the array, you don't need an array, you can just return from the cases directly. and for asc/desc you can use the default clause together.

```suggestion
switch (direction.toUpperCase()) {
  case QueryOrder.ASC:
    ret.push(`${column} asc`);
    break;
  case QueryOrder.DESC:
    ret.push(`${column} desc`);
    break;
  case QueryOrder.ASC_NULLS_FIRST:
    return `case when ${column} is null then 0 else 1 end, ${column} asc`;
  case QueryOrder.ASC_NULLS_LAST:
    return `case when ${column} is null then 1 else 0 end, ${column} asc`;
  case QueryOrder.DESC_NULLS_FIRST:
    return `case when ${column} is null then 0 else 1 end, ${column} desc`;
  case QueryOrder.DESC_NULLS_LAST:
    return `case when ${column} is null then 1 else 0 end, ${column} desc`;
  default:
    return `${column} ${direction.toLowerCase()}`;
}
```

tests/platforms/order-by.test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  @Property({ length: 255, nullable: true })
+  value?: string | null;
+
+}
+
+const options = {
+  'sqlite': { dbName: ':memory:' },
+  'libsql': { dbName: ':memory:' },
+  'better-sqlite': { dbName: ':memory:' },
```
B4nan, line 20: i'd remove this case, since it will be removed in v7, but not a huge deal to leave it

B4nan (changes requested):
few nits about the test

packages/mssql/src/MsSqlPlatform.ts
SteadEXE: I had to return an array to respect Abstract class, but I kept the logic

tests/platforms/order-by.test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    // Should be escaped twice.
+    const test3 = new Test();
+    test3.value = 'bar';
+
+    orm.em.persist(test);
+    orm.em.persist(test2);
+    orm.em.persist(test3);
+
+    await orm.em.flush();
```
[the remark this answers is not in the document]
SteadEXE: I could in fact just remove it, we never actually use the result

tests/platforms/order-by.test.ts
```diff
@@ -0,0 +1,121 @@
...
+  @PrimaryKey()
+  id!: number;
+
+  @Property({ length: 255, nullable: true })
+  value?: string | null;
+
+}
+
+const options = {
+  mysql: { port: 3308 },
```
B4nan, line 18: please leave the sqlite and libsql drivers, i meant just the better-sqlite one

tests/platforms/order-by.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  test(`order-by`, async () => {
+    const mock = mockLogger(orm, ['query', 'query-params']);
+
+    await orm.em.findAll(Test,
+      {
+        orderBy: {
+          value: QueryOrder.ASC,
+        },
+    });
```
B4nan, line 51: all of those should be written like that, i can imagine this wont pass the linter
```suggestion
    await orm.em.findAll(Test, {
      orderBy: {
        value: QueryOrder.ASC,
      },
    });
```

B4nan, line 51: all of those should be written like that, i can imagine this wont pass the linter
```suggestion
    await orm.em.findAll(Test, {
      orderBy: {
        value: QueryOrder.ASC,
      },
    });
```

B4nan (approved):
thanks!

PR #6623, opened 2025-05-01
feat(schema): allow setting a custom FK constraint name

### LINKED ISSUE #6624: Allow defining a custom foreign contraint name when creating a relationship
I'm dealing with an existing database with custom foreign constraint names which do not follow the general naming convention. Anytime I'm auto-generating schema migrations with Mikro-ORM, the migration wants to rename these foreign constraints. This is problematic because the project I'm working on relies on these constraint names for `ON CONFLICT` style queries, so I cannot change them easily.

I would like Mikro-ORM to provide a way to name the foreign contraint on a relation basis.

I already have a working PR: https://github.com/mikro-orm/mikro-orm/pull/6623

Let me know if I can help on this.

kozlek:
Closes #6624


B4nan (changes requested):
thanks!

i guess i would rather go with `foreignKeyConstraintName` or just `foreignKeyName`, also a test case would be nice (nothing huge, you could just modify some of the existing [ones that handle FK diffing](https://github.com/mikro-orm/mikro-orm/blob/101c0a85cacfd20e7d4857646d7c7242e4ec1cd1/tests/features/schema-generator/fk-diffing.postgres.test.ts))

kozlek:
> thanks!
> 
> i guess i would rather go with `foreignKeyConstraintName` or just `foreignKeyName`, also a test case would be nice (nothing huge, you could just modify some of the existing [ones that handle FK diffing](https://github.com/mikro-orm/mikro-orm/blob/101c0a85cacfd20e7d4857646d7c7242e4ec1cd1/tests/features/schema-generator/fk-diffing.postgres.test.ts))

Thanks for your feedback, I applied your recommendations regarding the name of the parameter (using `foreignKeyName`).
I also modified the test file you mentioned by adding a custom foreign key name for both `ManyToOne` and `OneToOne`.

B4nan (approved):
thanks!

PR #6662, opened 2025-05-19
feat(mssql): Allow inserting when triggers are present in mssql tables

SteadEXE:
When inserting in SQL Server, you cannot use output statement if some triggers are set.
You need to output in a temporary table and then select from this temporary table.

### DIFF
--- a/tests/platforms/output-statements.test.ts
+++ b/tests/platforms/output-statements.test.ts
@@ -1,6 +1,6 @@
 import { Entity, PrimaryKey, Property, SimpleLogger } from '@mikro-orm/core';
 import { MikroORM, MsSqlDriver } from '@mikro-orm/mssql';
-import { mockLogger } from '../helpers.js';
+import { mockLogger } from '../helpers';
 
 @Entity({ hasTriggers: true })
 class WithTriggers {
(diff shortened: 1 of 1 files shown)


packages/mssql/src/MsSqlDriver.ts
```diff
@@ -58,13 +58,38 @@
...
+        const selections = returningFields
+          .map((field: string) => `[t].${this.platform.quoteIdentifier(field)}`)
+          .join(',');
+
+        const returns = returningFields
+          .map((field: string) => `inserted.${this.platform.quoteIdentifier(field)}`)
+          .join(',');
+
+        let outputSql = `select top(0) ${selections} into #out from ${tableName} as t left join ${tableName} on 0=1; `;
+        outputSql += sql.replace(/\soutput\s(.+)\svalues\s/, ` output ${returns} into #out values `) + '; ';
```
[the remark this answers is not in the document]
SteadEXE: Replaced a regex

packages/mssql/src/MsSqlDriver.ts
```diff
@@ -64,7 +64,9 @@ export class MsSqlDriver extends AbstractSqlDriver<MsSqlConnection> {
       });
     }
 
-    return super.nativeInsertMany(entityName, data, options);
+    return super.nativeInsertMany(entityName, data, {
+      ...options,
```
B4nan, line 68: why are you making a copy of the options here?

SteadEXE: This was a left over, from a previous attempt

PR #6671, opened 2025-05-22
fix(mssql): only add temporary table when insert query actually returns fields

SteadEXE:


B4nan (approved):
Looking good. Lets not call this a fix, since the last commit wasnt released yet.

Issue #2548, opened 2021-12-17
Add option to disable generation of FKs


ramiel:
To support myISAM. Not to be confused with `migrations.disableForeignKeys` config option.

B4nan:
`disableForeignKeys` is not an option to "disable generation of FKs". it never was. it only controls if the schema commands will be wrapped in `set foreign_key_checks = 0` and `set foreign_key_checks = 1`.

ramiel:
Oh, in version 4 it disabled foreign key constraint creation. What changed here? How do I disable foreign key creation? Not all the mysql DB use the InnoDB engine

B4nan:
It did not change, it never worked that way.

B4nan:
What changed is that FKs are not properly diffed, in general schema diffing now works as it should.

ramiel:
Ok. Question remains. Is there a way to avoid creating foreign key constraints?

B4nan:
No it is not, there never was one. Feel free to send a PR, but it would have to be a top level configuration (that will be then feeded to the `SchemaGenerator` methods as part of their options), nothing about migrations, migrations just work with `SchemaGenerator`. 

Note that you still can use v5, you just need to remove the FK queries from your migrations manually if you don't want them there. That is the whole point of migrations, as you have absolute control over that.

ramiel:
Thanks. I have no time to provide a PR and the codebase is too big, sorry.

>Note that you still can use v5, you just need to remove the FK queries from your migrations manually if you don't want them there

Can you explain more about this?

B4nan:
You have migration file with SQL queries, so you can just remove those you dont want there. Not sure what more you want me to explain about that :]

Note that the codebase might be big, but this problem is about a single file. 

https://github.com/mikro-orm/mikro-orm/blob/master/packages/knex/src/schema/SchemaGenerator.ts#L69
https://github.com/mikro-orm/mikro-orm/blob/master/packages/knex/src/schema/SchemaGenerator.ts#L157
https://github.com/mikro-orm/mikro-orm/blob/master/packages/knex/src/schema/SchemaGenerator.ts#L166
https://github.com/mikro-orm/mikro-orm/blob/master/packages/knex/src/schema/SchemaGenerator.ts#L261
https://github.com/mikro-orm/mikro-orm/blob/master/packages/knex/src/schema/SchemaGenerator.ts#L303

(and maybe few more places in that file)

B4nan:
[...]

PR #6702, opened 2025-06-06
feat(schema): allow disabling foreign key constraint creation, per relation

nicomouss:
This PR follows my comment [here](https://github.com/mikro-orm/mikro-orm/issues/2548#issuecomment-2765607556) regarding the possibility to disable foreign key constraint, per relation.

This work complements what had been previously done in fcdb236.

Being able to disable foreign key creation on specific relations is useful when using DDD (Domain Driven Design) philosophy to design your app. Inside a DDD aggregate, entities are linked to each other by object references, and at the DB level we want to have strong integrity using foreign key constraint. Now for cross-aggregate links, we usually want to use _soft references_, meaning at the db level we just want an id referencing another entity but without the foreign key constraint.

Down the road it allows to implement eventual consistency between aggregates when we delete data from the db. Having foreign keys everywhere is sometimes too strict of a design. There is good literature on the subject [here](https://stackoverflow.com/a/1077696), for those interested to dive into the subject.

Hence the proposed PR here.

Please find below some explanations about how to use the feature implemented in the PR.

If you need to disable the creation of the underlying SQL foreign key constraint for a specific relation, you can now set `createForeignKeyConstraint` to `false` on the relation on the owning side.

```ts
@Entity()
export class Book {

  @ManyToOne(() => Author, { createForeignKeyConstraint: false })
  author1: Author;

}
```

This option is valid to be set on `ManyToOne`, `OneToOne`, and `ManyToMany` relations.

Note that if you globally disable the creation of all foreign key contraints by setting `createForeignKeyConstraints` to `false`, then no foreign key constraint is created whatsoever on any relation.

```ts
  ...
  schemaGenerator: {
    createForeignKeyConstraints: false,
  },
});
```

### DIFF
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -51,8 +51,8 @@ export class MetadataDiscovery {
   private readonly discovered: EntityMetadata[] = [];
 
   constructor(private readonly metadata: MetadataStorage,
-    private readonly platform: Platform,
-    private readonly config: Configuration) {
+              private readonly platform: Platform,
+              private readonly config: Configuration) {
     this.namingStrategy = this.config.getNamingStrategy();
     this.metadataProvider = this.config.getMetadataProvider();
     this.cache = this.config.getMetadataCacheAdapter();
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -51,8 +51,8 @@ export class MetadataDiscovery {
   private readonly discovered: EntityMetadata[] = [];
 
   constructor(private readonly metadata: MetadataStorage,
-              private readonly platform: Platform,
-              private readonly config: Configuration) {
+    private readonly platform: Platform,
+    private readonly config: Configuration) {
     this.namingStrategy = this.config.getNamingStrategy();
     this.metadataProvider = this.config.getMetadataProvider();
     this.cache = this.config.getMetadataCacheAdapter();
@@ -826,6 +826,7 @@ export class MetadataDiscovery {
       autoincrement: false,
       updateRule: prop.updateRule,
       deleteRule: prop.deleteRule,
+      createForeignKeyConstraint: prop.createForeignKeyConstraint,
     } as EntityProperty;
 
     if (selfReferencing && !this.platform.supportsMultipleCascadePaths()) {
--- a/packages/knex/src/schema/DatabaseTable.ts
+++ b/packages/knex/src/schema/DatabaseTable.ts
@@ -32,8 +32,8 @@ export class DatabaseTable {
   public comment?: string;
 
   constructor(private readonly platform: AbstractSqlPlatform,
-    readonly name: string,
-    readonly schema?: string) {
+              readonly name: string,
+              readonly schema?: string) {
     Object.defineProperties(this, {
       platform: { enumerable: false, writable: true },
     });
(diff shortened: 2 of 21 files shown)


packages/core/src/metadata/MetadataDiscovery.ts
```diff
@@ -51,8 +51,8 @@ export class MetadataDiscovery {
   private readonly discovered: EntityMetadata[] = [];
 
   constructor(private readonly metadata: MetadataStorage,
-              private readonly platform: Platform,
-              private readonly config: Configuration) {
+    private readonly platform: Platform,
+    private readonly config: Configuration) {
```
B4nan: please remove formatting changes

packages/knex/src/schema/DatabaseTable.ts
```diff
@@ -32,8 +32,8 @@ export class DatabaseTable {
   public comment?: string;
 
   constructor(private readonly platform: AbstractSqlPlatform,
-              readonly name: string,
-              readonly schema?: string) {
+    readonly name: string,
+    readonly schema?: string) {
```
B4nan: same here

tests/features/createForeignKeyConstraint/createForeignKeyConstraint.mysql.test.ts
```diff
@@ -0,0 +1,64 @@
...
+describe('createForeignKeyConstraint [mysql]', () => {
+
+  test('create SQL schema', async () => {
+
+    /**
+    * In this test, some foreign key constraints creations are disabled, on
+    * some specific OneToOne, ManyToOne, ManyToMany relations (aka when
+    * "createForeignKeyConstraint" is set to false on the owning side of
+    * the relation).
+    */
```
B4nan: missing leading space, same for all the commend blocks you've added
```suggestion
    /**
     * In this test, some foreign key constraints creations are disabled, on
     * some specific OneToOne, ManyToOne, ManyToMany relations (aka when
     * "createForeignKeyConstraint" is set to false on the owning side of
     * the relation).
     */
```

tests/features/createForeignKeyConstraint/createForeignKeyConstraint.mysql.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    * "createForeignKeyConstraint" is set to true on a given relation.
+    */
+
+    const orm = await MikroORM.init({
+      entities: [Author, AuthorAddress, BaseEntity, Book, BookTag, Publisher, PublisherAddress],
+      dbName: `db-${v4()}`, // random db name
+      port: 3308,
+      driver: MySqlDriver,
+      schemaGenerator: {
+        createForeignKeyConstraints: false,
```
B4nan, line 53: we already have tests for the global flag, i'd just remove those

nicomouss: Actually those tests were to make sure that even if we set `createForeignKeyConstraint` to `true` for a given relation, then the global `createForeignKeyConstraints` parameter takes precedence over everything.

I wait for your confirmation to remove those tests or not, then I will push a refactor commit to finalize everything.

B4nan: ok, let's leave it

nicomouss:
> looking good, thanks!
> 
> i got just a few code style nits

Thanks for the review,
0d80fcc should do

packages/core/src/metadata/MetadataDiscovery.ts
nicomouss: Sorry, those are unexpected code change triggered under the hood on my dev env when I hit `option+shift+F` to format (vscode on mac).

As `yarn lint` still passed, I did not see this.

Yet I can't find a way to force my vscode to respect this style rule automatically, is there something I can parameterize to prevent this specific code style nits to happen in the future ?

B4nan (approved):

PR #6703, opened 2025-06-07
fix(schema): properly handle array defaults

### LINKED ISSUE #6688: Migrator: Array of embeddables not properly assigned default value
### Describe the bug

The title already describes it quite good. I am trying to set an empty array as default for an array of embeddables but the migrator does not create a proper SQL statement for that.

So the following entities:
```ts
@Embeddable()
export class Account {
  @Property()
  accountName: string;

  @Property()
  accountPassword: string;
}

@Entity()
export class User {
  @PrimaryKey()
  id: number;

  @Embedded({ array: true, default: [] })
  accounts: Account[] = [];
}
```

result in this migration:

```ts
override async up(): Promise<void> {
    this.addSql(`create table "user" ("id" serial primary key, "accounts" jsonb not null default );`);
}
```

Our current workaround is:

```ts
@Entity()
export class User {
  @PrimaryKey()
  id: number;

  @Embedded({ array: true, defaultRaw: "'[]'" })
  accounts: Account[] = [];
}
```

### Reproduction

https://github.com/NicklasKnell/reproduction/tree/array-default

@mikro-orm/postgresql

### MikroORM version

6.4.15

### Node.js version

22.9.0

### Operating system

Ubuntu

### Validations

Ruby184:
Fixes #6688. While testing across all databases, MySQL was failing with the following error:

```sql
DriverException: create table `a` (
  `id` int unsigned not null auto_increment primary key,
  `languages` text not null default 'en',
  `accounts` json not null default '[]',
  `items` text not null default ''
) default character set utf8mb4 engine = InnoDB;
- BLOB, TEXT, GEOMETRY or JSON column 'languages' can't have a default value
```
See: https://dev.mysql.com/doc/refman/9.0/en/data-type-defaults.html

> The [BLOB](https://dev.mysql.com/doc/refman/9.0/en/blob.html), [TEXT](https://dev.mysql.com/doc/refman/9.0/en/blob.html), GEOMETRY, and [JSON](https://dev.mysql.com/doc/refman/9.0/en/json.html) data types can be assigned a default value only if the value is written as an expression, even if the expression value is a literal

Fixed by wrapping the default value in an expression when the column type is one of the types mentioned above.

### DIFF
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -1307,12 +1307,8 @@ export class MetadataDiscovery {
       return;
     }
 
-    if (Array.isArray(prop.default)) {
-      if (prop.customType instanceof ArrayType) {
-        val = prop.customType.convertToDatabaseValue(prop.default, this.platform)!;
-      } else if (prop.kind === ReferenceKind.EMBEDDED && prop.array) {
-        val = this.platform.convertJsonToDatabaseValue(prop.default) as string;
-      }
+    if (Array.isArray(prop.default) && prop.customType) {
+      val = prop.customType.convertToDatabaseValue(prop.default, this.platform)!;
     }
 
     prop.defaultRaw = typeof val === 'string' ? `'${val}'` : '' + val;
@@ -1368,6 +1364,10 @@ export class MetadataDiscovery {
     }
 
     if (prop.kind === ReferenceKind.SCALAR && !prop.customType && prop.columnTypes && ['json', 'jsonb'].includes(prop.columnTypes[0])) {
+      prop.customType = new JsonType();
+    }
+
+    if (prop.kind === ReferenceKind.EMBEDDED && !prop.customType && (prop.object || prop.array)) {
       prop.customType = new JsonType();
     }
 
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -1307,8 +1307,12 @@ export class MetadataDiscovery {
       return;
     }
 
-    if (prop.customType instanceof ArrayType && Array.isArray(prop.default)) {
-      val = prop.customType.convertToDatabaseValue(prop.default, this.platform)!;
+    if (Array.isArray(prop.default)) {
+      if (prop.customType instanceof ArrayType) {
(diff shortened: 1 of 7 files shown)


packages/core/src/metadata/MetadataDiscovery.ts
```diff
@@ -1307,8 +1307,12 @@ export class MetadataDiscovery {
...
     }
 
-    if (prop.customType instanceof ArrayType && Array.isArray(prop.default)) {
-      val = prop.customType.convertToDatabaseValue(prop.default, this.platform)!;
+    if (Array.isArray(prop.default)) {
+      if (prop.customType instanceof ArrayType) {
+        val = prop.customType.convertToDatabaseValue(prop.default, this.platform)!;
+      } else if (prop.kind === ReferenceKind.EMBEDDED && prop.array) {
+        val = this.platform.convertJsonToDatabaseValue(prop.default) as string;
+      }
```
B4nan: so `prop.customType` is not set for the embedded array properties? maybe it would be a better fix to set it during the discovery (to the JSON type), but maybe it will break more things, in such case i would be fine with what you did

tests/issues/GH6688.test.ts
```diff
@@ -0,0 +1,65 @@
...
+  items: Opt<EnumItems[]> = [];
+
+}
+
+const options = {
+  'sqlite': { dbName: ':memory:' },
+  'better-sqlite': { dbName: ':memory:' },
+  'mysql': { dbName: 'default_array_values', port: 3308 },
+  'mariadb': { dbName: 'default_array_values', port: 3309 },
+  'postgresql': { dbName: 'default_array_values' },
```
B4nan, line 40: can you test mssql too please?

B4nan (changes requested):

packages/core/src/metadata/MetadataDiscovery.ts
Ruby184: Yes, it was not set. I looked into it and added a condition to apply the JsonType custom type in the case of embedded properties, but a few tests broke. 😕 I’ll try to fix that.

Ruby184: I’ve fixed the tests - searched for `customType instanceof JsonType` and updated the conditions where necessary. So it shouldn’t break anything.

B4nan (approved):
thanks!

PR #6706, opened 2025-06-09
feat(schema): allow wilcard-schema friendly index expression

nicomouss:
This feature allows to use index expressions on wilcard-schema entities.

Now `expression` accepts either a SQL expression as a raw `string` (aka same behavior as til now); or a callback function that returns the SQL expression to be used, the dynamic schema being passed as a parameter.

```ts
@Entity({ schema: '*' })
@Index({ expression: (schema?: string) => `create index "custom_idx_on_name" on "${schema}"."author" ("name")` })
export class Author {
  ...
  @Property()
  name: string;
  ...
}
```

Logic is as follow:
- If the entity is setup with schema: "my_schema", "my_schema" is passed as a parameter.
- If the entity is setup with schema: "*", the dynamic schema is passed as a parameter (aka the targeted schema while calling `schemaGenerator.getCreateSchemaSQL({ schema: "..." })`).

Closes #6446


packages/entity-generator/src/SourceFile.ts
```diff
@@ -137,7 +137,7 @@ export class SourceFile {
       indexOpt.name = this.quote(index.name);
     }
 
-    if (index.expression) {
+    if (typeof index.expression === 'string') {
```
B4nan, line 140: i guess we should also support working with the callback here, since people can use callbacks dynamically via metadata hooks in case of a schema first approach. now a callback would be ignored completely.

B4nan (commented):
thanks!

Ruby184:
It would be great if the full table name were passed to the expression function, and also if the function received a second argument with the column mapping, similar to how it's currently done with the [@Check](https://mikro-orm.io/docs/defining-entities#check-constraints) decorator.

So your example would be:

```ts
@Entity({ schema: '*' })
@Index({ expression: (table, columns) => `create index "custom_idx_on_name" on "${table.split('.').join('"."')}" ("${columns.name}")` })
export class Author {
  ...
  @Property()
  name: string;
  ...
}
```

B4nan:
Good point. Maybe we could even have an object in the first parameter, it would have both table name and schema name as properties, and a `quoted` (or `quotedName`?) property, with already quoted and fully qualified name, so you could do ``@Index({ expression: (table, columns) => `create index "custom_idx_on_name" on ${table.quoted} ("${columns.name}")` })``.

packages/entity-generator/src/SourceFile.ts
nicomouss: Right, fixed in c853aa6

nicomouss:
> Good point. Maybe we could even have an object in the first parameter, it would have both table name and schema name as properties, and a `quoted` (or `quotedName`?) property, with already quoted and fully qualified name, so you could do `` @Index({ expression: (table, columns) => `create index "custom_idx_on_name" on ${table.quoted} ("${columns.name}")` }) ``.

I second the idea of two separated properties `table.name` and `table.schema` in the first param (plus the columns dictionary as the other param). imo much cleaner compared to have to split/join stuff inside the callback.

Regarding the `quotedName` param, does this add much value ?

imo, right out of the box the intent of this code:
`create index "custom_idx_on_name" on "${table.schema}"."${table.name}" ("${columns.name}")`
is clearer than this:
`create index "custom_idx_on_name" on ${table.quotedName} ("${columns.name}")`
or this:
`create index "custom_idx_on_name" on "${table.split('.').join('"."')}" ("${columns.name}")`

nicomouss:
> Good point. Maybe we could even have an object in the first parameter, it would have both table name and schema name as properties, and a `quoted` (or `quotedName`?) property, with already quoted and fully qualified name, so you could do `` @Index({ expression: (table, columns) => `create index "custom_idx_on_name" on ${table.quoted} ("${columns.name}")` }) ``.

I second the idea of two separate properties `table.name` and `table.schema` in the first param (plus the columns dictionary as the other param). imo much cleaner compared to have to split/join stuff inside the callback.

As for the other property, should we choose `table.fullyQualifiedQuotedName` instead of `table.quotedName` ?

B4nan:
> As for the other property, should we choose table.fullyQualifiedQuotedName instead of table.quotedName ?

That's way too huge, I would be fine with `table.quoted`, clear and concise, `quotedName` is fine as well, but nothing longer please.

packages/core/src/metadata/MetadataValidator.ts
```diff
@@ -207,7 +207,7 @@ export class MetadataValidator {
     }
   }
 
-  private validateIndexes(meta: EntityMetadata, indexes: { properties: string | string[] }[], type: 'index' | 'unique'): void {
+  private validateIndexes(meta: EntityMetadata, indexes: { properties?: string | string[] }[], type: 'index' | 'unique'): void {
```
B4nan, line 210: so the index properties are optional when its an expression on entity level, right?

packages/entity-generator/src/SourceFile.ts
nicomouss: As for now, just added test coverage for this, in c83b244

As a side note, there was a little bit of tiny side effect, as I had to change `properties` into `properties?` in a couple of places.

Otherwise, if `properties` stayed mandatory, then within metadata hook it was not possible to dynamically add a custom index with an expression callback like so:

```ts
entity.indexes.push({
  name: 'author2_custom_idx_on_email',
  expression: (schema?: string) => `create index "author2_custom_idx_on_email" on "${schema}"."author2" ("email")`,
});
```

packages/knex/src/schema/DatabaseTable.ts
```diff
@@ -878,10 +878,10 @@ export class DatabaseTable {
   }
 
   addIndex(meta: EntityMetadata, index: {
-    properties: string | string[];
+    properties?: string | string[];
```
B4nan, line 897: maybe we should validate that either properties or expression is provided? maybe in the validation method in metadata validator

tests/features/entity-generator/MetadataHooks.mysql.test.ts
```diff
@@ -97,6 +97,14 @@ const initialMetadataProcessor: MetadataProcessor = (metadata, platform) => {
           propOptions.type = 'IdentitiesContainer';
         }
       });
+
+      // Adding a custom index with an expression's callback.
+      // We need to make sure the expression's callback is generated on the entity.
+      // From https://github.com/mikro-orm/mikro-orm/pull/6706
```
B4nan: i dont think we need to link PRs that add code, we have git blame for that.

packages/entity-generator/src/SourceFile.ts
nicomouss: Oops, forgot to add test for uniques as well. Done in acd0d2f.

packages/core/src/metadata/MetadataValidator.ts
nicomouss: See my comment underneath your first review.

If not optional, then it's not possible to do do this in a metadata hook:

```ts
entity.indexes.push({
  name: 'author2_custom_idx_on_email',
  expression: (schema?: string) => `create index "author2_custom_idx_on_email" on "${schema}"."author2" ("email")`,
});
```

Actually `properties` was already optional for decorators and in entity schemas (so that we could indeed define and index with only an expression and no properties).
Only when we wanted to dynamically add an index (aka an index with only an expression, and no properties) in a metadata hook, then it was not possible.

packages/knex/src/schema/DatabaseTable.ts
nicomouss: Sure, can do.

Just to be clear on that point: I only made `properties` optional out of necessity to be able to add an expression-only index from a metadata hook, which was not possible otherwise.

But previous to that, it was already possible to declare an index like below:

```ts
@Index({})
export class Author {
  ...
}
```

No validation was done, nor any exception was raised, nor any test was in place.

Now at least, the behavior is unified over the codebase.

In the end, do you want me to incorporate the new validation behavior as part of this PR ?

B4nan: That's invalid for sure, on entity level you need to specify properties or expression. If you want to do it separately, I don't mind, but we should validate this.

nicomouss:
f3a5702 brings the implementation of the `expression` callback now accepting two parameters:

- first parameter `table` is an object containing 3 properties:
  - `name` is the name of the table
  - `schema` is the name of the schema
  - `quoted` is the fully qualified-quoted name (aka "mychema"."mytable")

- second parameter `columns` is a js-property-to-db-column dictionnary of the given entity

This is working for both `@Index` and `@Unique`.

So it is now possible to do stuff like below, similar to how `@Check` decorator is working:

```ts
@Entity({ schema: '*' })
@Index<Author>({ expression: (table, columns) => `create index "custom_idx_on_name" on "${table.schema}"."${table.name}" ("${columns.name}")` })
export class Author {
  ...
  @Property()
  name: string;
  ...
}
```

Other examples:

```ts
// Using table.quoted
@Index<Author>({ expression: (table, columns) => `create index "custom_idx_on_name" on ${table.quoted} ("${columns.name}")` })

// By omitting the type (@Index instead of @Index<Author>), columns is not type-checked
@Index({ expression: (table, columns) => `create index "custom_idx_on_name" on ${table.quoted} ("${columns.my_prop}")` })
```

The side effect of this implementation, is that the `properties` decorator option on `@Index` and `@Unique` decorators is not type-checked anymore:

```ts
// This will not raise any type-checking errors anymore
@Index({ name: 'custom_idx_1', properties: ['name', 'email','not_existing_prop'] })
```

Indeed, this does not seem to be possible to bring both type-checkings (aka for `properties` option, and for the `columns` parameter in the expression callback) working at the same time on a given decorator implementation.

Tried lots and lots of ways to figure this out, always hitting a dead end constantly.

That is actually why probably the `@Check` decorator does not type-check its `property` option either. But instead only type-checks the `columns` parameter within the `expression` callback.

B4nan:
Hmm, that's weird, I would very much like to keep the type safety we have now, it would be a downgrade to lose that. Did you try the [same condition](https://github.com/mikro-orm/mikro-orm/blob/833a6403ca393c341fca8e6207652d063fdb0677/packages/core/src/decorators/Indexed.ts#L33) as we have for the `properties` option now? That's what deals with the `typeof User` vs `User` difference. I'll have a look, I am sure there is a way to make this work.

B4nan:
It seems to work fine for me with that approach:

```ts
expression?: string | (T extends EntityClass<infer P> ? IndexCallback<P> : IndexCallback<T>);
```

It won't allow to pass the type explicitly, but that's something I don't want from users at all, those should be inferred (and it is not allowed in the current version either).

nicomouss:
All right, all good now ! Thanks for having brought fresh eyes on this.

The reason of my initial confusion stems from `@Check` decorator, which is explicitly requiring to pass the type in order to activate type-checking (`@Check<Author>({ ... }`). By trying to replicate the same behavior in `@Index` (aka allowing to do `@Index<Author>({ ... })`, I reached those incompatibilities with the type-checkings.

Indeed by inferring the type under the hood, it is way better.
Should be `@Check` implemented the same way then ?

So now we want to use the `@Index` and `@Unique` that way:

```ts
@Entity({ schema: '*' })
@Index({ expression: (table, columns) => `create index "custom_idx_on_name" on "${table.schema}"."${table.name}" ("${columns.name}")` })
export class Author {
  ...
  @Property()
  name: string;
  ...
}
```

```ts
// Using table.quoted
@Index({ expression: (table, columns) => `create index "custom_idx_on_name" on ${table.quoted} ("${columns.name}")` })
```

With all the types inferred automatically under the hood.

B4nan:
The inference in those decorators is rather new addition, and its incompatible with the explicitly provided type argument, IIRC I left the check decorator as is since it felt breaking (it would require people to remove the explicit type argument), but we should surely do this in v7.

nicomouss:
> The inference in those decorators is rather new addition, and its incompatible with the explicitly provided type argument, IIRC I left the check decorator as is since it felt breaking (it would require people to remove the explicit type argument), but we should surely do this in v7.

Yes, it will be a breaking change for those already using `@Check`.
v7 target sounds legit then.

packages/core/src/typings.ts
```diff
@@ -428,7 +428,15 @@ export type EntityDTO<T, C extends TypeConfig = never> = {
...
 };
 
-type CheckKey<T> = IsUnknown<T> extends false ? keyof T : string;
+type IndexKey<T> = IsUnknown<T> extends false ? keyof T : string;
+type IndexTable = { name: string; schema?: string; quoted: string };
+export type IndexCallback<T> = (table: IndexTable, columns: Record<IndexKey<T>, string>) => string;
+
+type UniqueKey<T> = IndexKey<T>;
+type UniqueTable = IndexTable;
+export type UniqueCallback<T> = (table: UniqueTable, columns: Record<UniqueKey<T>, string>) => string;
```
B4nan: do we really need this duplicity? those types are exactly the same

packages/knex/src/schema/DatabaseTable.ts
```diff
@@ -634,12 +635,12 @@ export class DatabaseTable {
...
    */
-  getShortestName(): string {
+  getShortestName(quoted = false): string {
     if (!this.schema || this.name.startsWith(this.schema + '.')) {
-      return this.name;
+      return quoted ? `"${this.name}"` : this.name;
     }
 
-    return `${this.schema}.${this.name}`;
+    return quoted ? `"${this.schema}"."${this.name}"` : `${this.schema}.${this.name}`;
```
B4nan: quoting needs to be platform specific, you need to use `platform.quoteIdentifier` method for that.

B4nan (changes requested):
two nits, also we need to update the docs

packages/core/src/typings.ts
nicomouss: Under the hood they are the same, semantically they are different (index vs unique). That was the reason, to bring clarity in the naming of things.

If the namings `IndexTable`, `IndexKey` and `IndexCallback` make sense to be used for uniques as well, then I can remove the specific type names introduced for uniques.

B4nan: I'd prefer to have a single type for this, and use it for the check constraints too in v7, so better use some generic names. The second parameter would be useful for formula properties too, and I guess there will be other places like that.

nicomouss: Ok, what generic prefix name for both indexes, uniques, checks would you suggest ? 

`Constraint` ? That would give `ConstraintTable`, `ConstraintKey`, `ConstraintCallback` types.

Or still stay with `Index` ? (does not feel right knowing that an unique is an index, but a check is not really an index ...).

B4nan: I wouldnt be too concerned about where those types are used, think more about what they represent. None of them is actually tied to index/unique/check behaviour, one describes a table name helper, the other a column name mapping. Maybe something like this

```ts
type PropertyName<T> = IsUnknown<T> extends false ? keyof T : string;
type TableName = { name: string; schema?: string; quoted: string };
type ColumnNameMapping<T> = Record<PropertyName<T>, string>; // or maybe just Columns<T> would work fine
```

This can be reused in the check constraint types as well as for the formula properties.

nicomouss: Adding thoughts about this, we should also have quoted column names available, alongside the raw column names.

Regarding the types, we could then have something like so:

```ts
type PropertyName<T> = IsUnknown<T> extends false ? keyof T : string;
type TableName = { name: string; schema?: string; quoted: string };
type ColumnNames<T> = { nameMapping: Record<PropertyName<T>, string>; quotedMapping: Record<PropertyName<T>, string> };
```

That way down the road it will allow to do this:

```ts
@Index({ expression: (table, columns) => `create index custom_idx_on_name on ${table.quoted} (${columns.quotedMapping.name})` })
```

B4nan: I wonder if it actually makes sense to have only quoted names in those properties. Would be breaking, so something for v7, but I cant think of a reason why would you want to use an unquoted name in those places. So I'd skip this now, but in v7 we would quote the column names automatically.

Same could apply to the table name, I actually think it could be just a string property in v7, including the fully qualified and quoted name, since why would you want to use just the table name without schema prefix? 

```ts
@Index({ expression: (table, columns) => `create index custom_idx_on_name on ${table} (${columns.name})` })
```

I was also thinking of defining a `toString()` on the table name object, if we do that, you could just use `${table}` to get the quoted fqn.

Ruby184: I think a better solution would be to allow returning a `raw` fragment as the expression, so quoting can be handled using `??` parameters.

B4nan: I am not so sure about that, the readability would be worse. But we can surely allow it too.

nicomouss:
The platform-specific quoting is now implemented.
Added tests for various platforms to make sure the platform-specific quoting is indeed respected by the `@Index` or `@Unique` expressions.

Also types have been refactored following the conversation above, with the toString method added to the `TableName` object:

```ts
type PropertyName<T> = IsUnknown<T> extends false ? keyof T : string;
type TableName = { name: string; schema?: string; quoted: string; toString: () => string };
type ColumnNameMapping<T> = Record<PropertyName<T>, string>;
```

I left only one callback type, for both indexes and uniques:

```ts
export type IndexCallback<T> = (table: TableName, columns: Record<PropertyName<T>, string>) => string;
```

That means the expression for `@unique` is accepting an `IndexCallback` now (instead of an `UniqueCallback` which was basically an alias for it).

A unique being a special case of an index, that makes some sense.
If that causes trouble in the clarity of things, I can add the `UniqueCallback` type alias back into play.

**Note: for whatever reason, I hit some weird stuff regarding type checking, in case I do this for the callback**:

```ts
export type IndexCallback<T> = (table: TableName, columns: ColumnNameMapping<T>) => string;
```

instead of this:

```ts
export type IndexCallback<T> = (table: TableName, columns: Record<PropertyName<T>, string>) => string;
```

The former breaks the type-checking in `@Index` and `@Unique` decorators.
Probably missing something trivial along the lines, so fresh eyes are welcome on this ...

packages/core/src/typings.ts
nicomouss: Ok so I implemented the `toString()` method and refactored the different types, in 119c24a and 24c0018.

I'd suggest we leave it as it is for now, and indeed make everything quoted for all properties in v7 (either for `TableName` and for `ColumnNameMapping`). That is what makes the more sense in my eyes as well.

Ruby184: Sure, it's less readable, but when passing already escaped columns, there's no way to escape the index name `custom_idx_on_name`, for example, in a database-agnostic way—you have to escape it manually, even when the other columns and the table are already escaped.

nicomouss: > there's no way to escape the index name `custom_idx_on_name`, for example, in a database-agnostic way—you have to escape it manually, even when the other columns and the table are already escaped.

For this, another idea would be to pass the escaping character(s) as extra parameter(s) somewhere ('`' in case of MySql, '"' in case of postgres, '[' and ']' in case of mssql, and so on ...).
So that it is usable if needed.

And/or an utility method that can be called within the expression, which would call the platform's specific escaping logic under the hood.

For example, something along those lines:

```ts
@Index({ expression: (table, columns, utils) => `create index ${utils.quoteStart}custom_idx_on_name${utils.quoteEnd} on ${table} (${columns.name})` })
```

```ts
@Index({ expression: (table, columns, utils) => `create index ${utils.quote(custom_idx_on_name)} on ${table} (${columns.name})` })
```

nicomouss:
(Sorry, unexpected request for review triggered on my side).
That was not intended.

packages/knex/src/schema/DatabaseTable.ts
nicomouss: > That's invalid for sure, on entity level you need to specify properties or expression. If you want to do it separately, I don't mind, but we should validate this.

Yes, can do a specific PR later on when work on this one is done.

Actually, `@Index({})` works in current master, because of this check which silently ignores such things:

https://github.com/mikro-orm/mikro-orm/blob/c83b244d74318683c54822c880418324f0431cee/packages/knex/src/schema/DatabaseTable.ts#L924-L926

So two options there I guess:
- leave it as it is (Mikroorm silently ignoring those empty indexes)
- upgrade metadata validation logic to turn Mikroorm stricter on that point (Mikroorm will raise an exception in case of a empty index) - that might be a breaking change for some having empty indexes in their codebases, so shall this target v7 ?

B4nan: I'd leave things as they are now and make validation stricter in v7.

packages/core/src/typings.ts
B4nan: Maybe the third param could be a tagged template literal function for quoting identifiers:

```ts
@Index({ expression: (table, columns, quote) => quote`create index ${'custom_idx_on_name'} on ${table} (${columns.name})` })
```

I guess this looks better than the idea with `toString` on the `table` object. Maybe we could have both (tag the object so you can distinguish when processing the quote function params). I don't like the idea with `quoteStart` and `quoteEnd`.

Ruby184: I think using `raw` is consistent with how it works in other places, like object queries. Yes, it’s not very readable, but it’s simple and consistent.

```ts
@Index({
  expression: (table, columns) => raw('create index ?? on ?? (??)', ['custom_idx_on_name', table, columns.name])
})
```

B4nan: We can support that too, but what I proposed is much nicer (and trivial to implement). I'd like to have both.

nicomouss: > Maybe the third param could be a tagged template literal function for quoting identifiers:
> 
> ```ts
> @Index({ expression: (table, columns, quote) => quote`create index ${'custom_idx_on_name'} on ${table} (${columns.name})` })
> ```
> 
> I guess this looks better than the idea with `toString` on the `table` object. Maybe we could have both

I'd definitely vote to keep both. Indeed if I know that my index name is ok without quoting (as it is 99.999% of the time, as far as my experience is concerned), then as a mikroorm user I'd prefer to write this (aka in v7 when `columns.name` returns the already-quoted string):

```ts
@Index({ expression: (table, columns) => `create index custom_idx_on_name on ${table} (${columns.name})` })
```

instead of this:

```ts
@Index({ expression: (table, columns, quote) => quote`create index ${'custom_idx_on_name'} on ${table} (${columns.name})` })
```

The later introducing unneeded noise in that case.

> Maybe we could have both (tag the object so you can distinguish when processing the quote function params)

Just to be sure to be on the same page, what do you mean by _tag the object_ ?
If I haven't miss something, in order to differentiate the quote function params, we could use a bit of stronger typings using classes:

```ts
type PropertyName<T> = IsUnknown<T> extends false ? keyof T : string;
class TableName { // here instead of type TableName { ... }
  name: string; 
  schema?: string; 
  quoted: string;
  ...
};
class ColumnName {
  name: string;
  ...
}
type ColumnNameMapping<T> = Record<PropertyName<T>, ColumnName>; // here instead of type Record<PropertyName<T>, string>
```

So that within the quote function when processing the params, we can distinguish them based on their runtime types:

```ts
if (myparam instanceof TableName) { 
  /* myparam.quoted or myparam.toString() contains the table quoted name */
} else { 
  /* we know that we have to quote myparam */
}
```

Also this will be helpful for v7 implementation (see note(1) below).

**note(1)** in v7 (aka when `columns.name` will return the already-quoted string), while processing the quote function params, we will need to be able to differentiate between `${columns.name}` and `${'custom_idx_on_name'}`. The former will be already quoted, while the later will need to be quoted. Having a class for `ColumnName` instead of a type will be of help here as well.

**note(2)** In mikroorm codebase, at the moment, there is the `DatabaseTable.getShortestName(true)` which returns the fully qualified quoted name of the table (example: I have a table named `user` within the `admin` schema. The method returns `"admin"."user"` when using postgres).

Now in order to not reinvent the wheel, is there a method available somewhere in the codebase to quote a column name, per-platform ? (example: I have a column named `facebook.id`. I want the quoted version of it, which would be `"facebook.id"` when using postgres).

Waiting for your feedback on that, then I'll introduce the changes.

B4nan: I think the `quoted` helper would be enough, I wouldn't quote the column names if we have that. By tagging I meant adding some symbol property to the object so you can check on runtime its that object, you can as well use a class instance and instanceof check. I didnt mean anything on type level.

> Now in order to not reinvent the wheel, is there a method available somewhere in the codebase to quote a column name, per-platform ? (example: I have a column named facebook.id. I want the quoted version of it, which would be "facebook.id" when using postgres).

It should just use the `platform.quoteIdentifier` method directly. We can add a parameter to it to disable what it does with the dots now explicitly when processing column names.

nicomouss: All good, so regarding the automatic quoting in v7, we forget about that below, right ?:

> I wonder if it actually makes sense to have only quoted names in those properties. Would be breaking, so something for v7, but I cant think of a reason why would you want to use an unquoted name in those places. So I'd skip this now, but in v7 we would quote the column names automatically.

> By tagging I meant adding some symbol property to the object so you can check on runtime its that object, you can as well use a class instance and instanceof check. I didnt mean anything on type level.

All right, coming from Java/C#/OOP, using the object's type/class to check those kinds of things, is what comes naturally to my mind.

> It should just use the platform.quoteIdentifier method directly. We can add a parameter to it to disable what it does with the dots now explicitly when processing column names.

Sure

nicomouss: > Maybe the third param could be a tagged template literal function for quoting identifiers:
> 
> ```ts
> @Index({ expression: (table, columns, quote) => quote`create index ${'custom_idx_on_name'} on ${table} (${columns.name})` })
> ```
> 
> I guess this looks better than the idea with `toString` on the `table` object. Maybe we could have both (tag the object so you can distinguish when processing the quote function params).

Done in f397d9b

nicomouss:
> two nits, also we need to update the docs

Docs now up to date with current implementation.
See ec649de

packages/core/src/platforms/Platform.ts
```diff
@@ -481,8 +481,8 @@ export abstract class Platform {
     return value as string;
   }
 
-  quoteIdentifier(id: string, quote = '`'): string {
-    return `${quote}${id.toString().replace('.', `${quote}.${quote}`)}${quote}`;
+  quoteIdentifier(id: string, splitDots = true, quote = '`'): string {
```
B4nan: changing the meaning of the second parameter is technically breaking change

packages/core/src/typings.ts
```diff
@@ -429,10 +429,19 @@ export type EntityDTO<T, C extends TypeConfig = never> = {
...
-type TableName = { name: string; schema?: string; quoted: string; toString: () => string };
+export class TableName {
+
+  constructor(readonly name: string, readonly schema: string | undefined, readonly quoted: string) {}
+
+  toString = () => {
+    return this.quoted;
+  };
+
+}
```
B4nan: if its a class, use class methods instead of properties with arrow functions. also please add blank lines around the class declaration

packages/knex/src/schema/DatabaseTable.ts
```diff
@@ -32,6 +33,15 @@ export class DatabaseTable {
   public nativeEnums: Dictionary<{ name: string; schema?: string; items: string[] }> = {}; // for postgres
   public comment?: string;
 
+  private quoteFunction = (expParts: readonly string[], ...values: unknown[]) => expParts.reduce<string>((exp, expPart, i) => {
```
B4nan: this function should be available as a global export, just like `raw` and `sql`, it shouldnt be an internal of `DatabaseTable` class. i would move the `TableName` class to its own file and declare this `quoted` function in that file instead.

edit: it should work the same as the `raw` helper, so it wont work directly with the platform, it will map to a raw fragment using the `??` under the hood. that way we will have out of box support for the raw fragments too. this will be handy for other places, e.g. recently i needed exactly this for a formula property (in v7, where I already added support for the raw helper there - and if this new `quoted` helper will also produce a raw fragment, it will work there right ahead)

packages/core/src/platforms/Platform.ts
nicomouss: hmmmm would you mind elaborating where is this a breaking change ?

All around the codebase, there is no call explicitly passing the second parameter.
Plus the overrides in MsSqlPlatform and PostgreSqlPlatform ignore this param completely. Making it kinda useless if for whatever reason a mikroorm user would call `platform.quoteIdentifier(myId, '!')` on those platforms.

Also, for platforms where quoteStart and quoteEnd are not the same (aka MsSqlPlatform with '[' and ']'), having to override a base method ```quoteIdentifier(id: string, quote = '`')``` has no meaning. Should be ```quoteIdentifier(id: string, quoteStart = '`', quoteEnd = '`')``` in that case in Platform.ts then.

So things are a bit shaky around these things.

Anyway, just let me know the method signature you want in `Platform.ts`, that would go straight to the point.

B4nan: It's a public method, in a publicly exported class. So even if its not used inside this code base, some people might have a custom driver, using it directly.

> Also, for platforms where quoteStart and quoteEnd are not the same (aka MsSqlPlatform with '[' and ']'), having to override a base method quoteIdentifier(id: string, quote = '`') has no meaning. Should be quoteIdentifier(id: string, quoteStart = '`', quoteEnd = '`') in that case in Platform.ts then.

Sounds a bit overengineered, the current solution for MSSQL doesn't use the parent implementation, I would like to keep that, just implement handling of this new parameter there too (so a super simple if statement).

packages/core/src/typings.ts
nicomouss: Yes, my bad, some code refactoring that went under my radar.

packages/knex/src/schema/DatabaseTable.ts
nicomouss: > this function should be available as a global export, just like raw and sql, it shouldnt be an internal of DatabaseTable class.

I just followed what you proposed above, aka passing the quote function as a third parameter of the expression callback:

> Maybe the third param could be a tagged template literal function for quoting identifiers:
> 
> ```ts
> @Index({ expression: (table, columns, quote) => quote`create index ${'custom_idx_on_name'} on ${table} (${columns.name})` })
> ```
> 
> I guess this looks better than the idea with `toString` on the `table` object. Maybe we could have both

Sorry if i am missing something here: if you want the quote method available globally, what is the point to pass it at the third parameter ?

Never used raw fragments whatsoever in mikroorm, so please give me little bit more context about the ```it will map to a raw fragment using the ?? under the hood```.

packages/core/src/platforms/Platform.ts
nicomouss: > Sounds a bit overengineered, the current solution for MSSQL doesn't use the parent implementation, I would like to keep that, just implement handling of this new parameter there too (so a super simple if statement).

It doesn't use, but still needs to `override` the base method in terms of OOP and typescript compilation to work. So if we have this in Platform.ts:

```ts
quoteIdentifier(id: string, quote = '`', splitDots = true): string {
  return `${quote}${splitDots ? id.toString().replace('.', `${quote}.${quote}`) : id}${quote}`;
}
```

Then we'll have this in MsSqlPlatform:
```ts
// That's the ugly part here with this "quote" parameter defined but never used.
// Indeed it has no sense for Mssql to have only one quoting param, so the meaning 
// of this parameter alone is inherently wrong for this platform.
override quoteIdentifier(id: string, quote = '[', splitDots = true): string {
  return `[${splitDots ? id.replace('.', `].[`) : id}]`;
}
```

Plus now the call to the method needs now to explicitly pass `undefined` for the `quote` param, in order to be able to pass the third param as `false`, without altering the implemented platform-specific quoting behavior:
```ts
// The other 'ugly' part
this.platform.quoteIdentifier((id as any).toString(), undefined, false)
```

If you are happy with those tweaks, then done in 8d1a956

B4nan: hmm, i see, that's indeed ugly. all right, i guess i am fine with the tiny BC instead of all that.

packages/knex/src/schema/DatabaseTable.ts
Ruby184: > edit: it should work the same as the raw helper, so it wont work directly with the platform, it will map to a raw fragment using the ?? under the hood. that way we will have out of box support for the raw fragments too.

If you want to make that global, maybe just add support for raw fragments in expressions, as I proposed earlier, and remove the quote argument altogether. Then, a universal quote function could be implemented in a separate PR. I can look into that.

B4nan: Yes, that would be a requirement for this to work, since the return value of this method would be a raw fragment instance.

B4nan: @nicomouss yes, if we make it global, no need to have it in the parameter

> Never used raw fragments whatsoever in mikroorm, so please give me little bit more context about the it will map to a raw fragment using the ?? under the hood.

If you do `raw('... ?? ...', ['id'])` you get a raw fragment instance which can be evaluated via `platform.formatQuery()` conditionally, to get a platform specific interpretation of quoting of both values (`?`) and identifiers (`??`). So you would basically replace all parameters of `quote` helper with `??` and pass this string down to `raw` with the actual param names in the second argument.

```ts
// calling this
quote`create index ${'custom_idx_on_name'} on ${table} (${columns.name})` 
// would resolve internally to this
raw('create index ?? on ?? (??)', ['custom_idx_on_name', table.toString(false), columns.name])
// `table.toString(false)` would return the schema.name unquoted, or maybe we could just drop the idea of quoting things in there?
```

There are many examples over the codebase, e.g. this is how its supported for `prop.default`: https://github.com/mikro-orm/mikro-orm/blob/534193c327b2721ffae1703ddd259eb0a3a2cbe4/packages/core/src/metadata/MetadataDiscovery.ts#L430-L435

In v7, this is supported on all places that were previously accepting a `knex.raw` instance (and many more). Generally speaking, I'd like to support this everywhere where it makes sense.

nicomouss: > If you do raw('... ?? ...', ['id']) you get a raw fragment instance which can be evaluated via platform.formatQuery() conditionally

I see ... Is this documented somewhere, as far as current available v6.4 docs available online ?
Only thing I can find about `raw` is very succinct in https://mikro-orm.io/docs/raw-queries#raw-helper.

nicomouss: > If you want to make that global, maybe just add support for raw fragments in expressions, as I proposed earlier, and remove the quote argument altogether. Then, a universal quote function could be implemented in a separate PR. I can look into that.

The quote function is already initiated in this PR, it is just waiting to be turned into a global one using the raw helper under the hood.

Once the expression callback is made compatible with raw fragments, it will be easy to do it here also, no problem.

packages/core/src/platforms/Platform.ts
nicomouss: Well that's a tiny BC, but the real underlying problem is else: the whole codebase is bugged down anyway regarding dotted column names. Just looking at the codebase, up to what is done in this PR regarding ```this.platform.quoteIdentifier(...)```, no distinction is made between quoting a column name and quoting a table name.

Which means that in current mikroorm version
- quoting `myschema.myTable` returns `"myschema"."myTable"` (which is right).
- quoting column name `facebook.id` returns `"facebook"."id"` (which is wrong. Should return `"facebook.id"`).

Haven't written proper tests, but just looking at the codebase while working on this PR makes me think that if you try using dotted columns names in your db, then mikroorm will crumble in various places.

That is why I initially asked if there was another method to specifically quote a column:

> Now in order to not reinvent the wheel, is there a method available somewhere in the codebase to quote a column name, per-platform ? (example: I have a column named facebook.id. I want the quoted version of it, which would be "facebook.id" when using postgres).

Anyway, the new version of the method added in this PR provides the way to fix that, thanks to the `splitDots` param (which should be passed `false` for quoting column names):
```
quoteIdentifier(id: string, splitDots = true, quote = '`'): string { ... }
```

This means that once this PR is merged, there will be calls to that method to be updated all over the codebase if you want to fix the dotted column problem.

That is something you might want to keep in mind.

B4nan: Have you actually seen anyone using dots in column names? I haven't, nobody ever filed a report about it (and tbh it feels quite insane). Not saying we shouldn't support it, but the whole discussion here is getting really derailed from the original problem, and I also see this change as a potential BC, so not interested in making some wider fixes in that regard in v6.

Let's not try to solve every issue you come across in this PR. I would be fine if we remove the whole `splitDots` argument completely from the PR and leave that for later (v7).

nicomouss: > Have you actually seen anyone using dots in column names. I haven't, nobody ever filed a report about it (and tbh it feels quite insane)

Yes, I have seen that in various projects with stuff like `facebook.id`, otherwise I would not take the time to write a detailed feedback, dont' you think ?

That's not because you never encountered something in life that it is to be considered insane.
For example if I prefix my embbeddables with `address.`, is that considered so ? That's very subjective, right ?

> the whole discussion here is getting really derailed from the original problem

Bro, it has already diverted a lot from the start, not because of this last feedback ... 
But because of the various ideas an moving requirements popping up from everybody, that came into play over the course of the PR, which split from the initial target (aka just a `schema` param).

> Let's not try to solve every issue you come across in this PR

I am not trying to fix anything at all, I am giving a feedback on something that came clear to my eyes while working on the PR.

B4nan: Can we please focus on finishing the PR instead? I see you got triggered by this, that wasn't my intention. Note that I specifically said we should support this too, however insane it _feels_ to me (just like column names with spaces, that's a very similar category of weird naming ideas). But it's not something we should solve as part of this PR, there is no value in fixing this only for the index expressions, because the rest of the codebase would still fail to work with such columns (both runtime querying and schema diffing most likely).

---

I would drop the handling of `splitDots` from the PR, add the global `quote` helper and merge it. If you consider the support for dots in column names important, we can take a closer look at fixing that in v6, but I would personally leave that for v7, where we can afford small BCs like changing the order of parameters.

Ruby184: > Bro, it has already diverted a lot from the start, not because of this last feedback ...
But because of the various ideas an moving requirements popping up from everybody, that came into play over the course of the PR, which split from the initial target (aka just a schema param).

I didn’t want to clutter the PR, I just wanted to generalize the idea. Yes, adding schema as a parameter is what you needed, but it’s a good idea to consider broader use cases before merging something, as it becomes harder to change later without breaking existing user code.

For example, we needed to create indexes with WHERE conditions without writing the expression as a raw string everywhere, so we had to work around that. That’s why I thought it would be great to support passing the table and column into the function directly and also just support `raw` for quoting columns if needed.

nicomouss:
Implemented index expression using `raw` and global `quote` helpers.

Use them as follows:

```ts
@Entity({ tableName: 'author', schema: '*' })
// Using raw SQL string
@Index({ name: 'custom_idx_on_name', expression: (table, columns) => `create index custom_idx_on_name on "${table.schema}"."${table.name}" ("${columns.name}")` })
// Using quote helper
@Index({ name: 'custom_idx_on_country', expression: (table, columns) => quote`create index ${'custom_idx_on_country'} on ${table} (${columns.country})` })
// Using raw function
@Unique({ name: 'custom_unique_on_email', expression: (table, columns) => raw(`alter table ?? add constraint ?? unique (??)`, [table, 'custom_unique_on_email', columns.email]) })
export class Author {
  ...
}
```

I Removed the `splitDots` stuff from `platform.quoteIdentifier(...)`, so it is left untouched from how it is in current master. You can change things as you wish in v7.

I left `TableName` class within `typings.ts` as for now. There is already EntityMetadata class in there as well, so ... move it as you wish.

Global `quote` helper is in its own file in the same folder alongside the other helpers.

packages/knex/src/schema/DatabaseTable.ts
nicomouss: > If you want to make that global, maybe just add support for raw fragments in expressions, as I proposed earlier, and remove the quote argument altogether. Then, a universal quote function could be implemented in a separate PR. I can look into that.

90dfcf5

That was a one-line function ...

```ts
export function quote(expParts: readonly string[], ...values: unknown[]) {
  return raw(expParts.join('??'), values);
}
```

packages/core/src/typings.ts
```diff
@@ -428,8 +428,21 @@ export type EntityDTO<T, C extends TypeConfig = never> = {
...
 
-type CheckKey<T> = IsUnknown<T> extends false ? keyof T : string;
-export type CheckCallback<T> = (columns: Record<CheckKey<T>, string>) => string;
+type PropertyName<T> = IsUnknown<T> extends false ? keyof T : string;
+export class TableName {
+
+  constructor(readonly name: string, readonly schema: string | undefined) {}
+
+  toString() {
+    return `${this.schema}.${this.name}`;
```
B4nan:
```suggestion
    if (this.schema) {
      return `${this.schema}.${this.name}`;
    }

    return this.name;
```

packages/core/src/platforms/Platform.ts
```diff
@@ -506,7 +506,7 @@ export abstract class Platform {
 
     if (sql[0] === '?') {
       if (sql[1] === '?') {
-        ret += this.quoteIdentifier(params[j++]);
+        ret += this.quoteIdentifier(params[j++].toString());
```
B4nan: i guess you need this because the params can be also the new TableName instances? i am not sure if this will work in v7, since we natively support raw fragments there too. it would be better to check for the table instance in the `quoteIdentifier` method instead of stringifying the values we pass into it. .

packages/core/src/utils/Quote.ts
```diff
@@ -0,0 +1,5 @@
+import { raw } from './RawQueryFragment';
+
+export function quote(expParts: readonly string[], ...values: unknown[]) {
+  return raw(expParts.join('??'), values);
+}
```
B4nan: i'd move this to the `RawQueryFragment` file since its just a thin layer on top of what it does

packages/core/src/platforms/Platform.ts
```diff
@@ -526,7 +526,7 @@ export abstract class Platform {
         ret += sql.substring(pos, idx - 1) + '?';
         pos = idx + 1;
       } else if (sql.substring(idx, idx + 2) === '??') {
-        ret += sql.substring(pos, idx) + this.quoteIdentifier(params[j++]);
+        ret += sql.substring(pos, idx) + this.quoteIdentifier(params[j++].toString());
```
B4nan: same here

packages/knex/src/schema/DatabaseTable.ts
```diff
@@ -634,12 +637,12 @@ export class DatabaseTable {
   /**
    * The shortest name is stripped of the default namespace. All other namespaced elements are returned as full-qualified names.
    */
-  getShortestName(): string {
+  getShortestName(quoted = false): string {
```
B4nan: is this `quoted` parameter used anywhere now? if not, i'd revert those changes too.

packages/core/src/platforms/Platform.ts
nicomouss: Well my initial intention was indeed to do the type check, but it appeared cleaner to use the `toString()` methods while I went through it.
Will revert back to the explicit type check then.

packages/knex/src/schema/DatabaseTable.ts
nicomouss: Yeah definitely some code left behind from the various changes made. This was used for the `table.quoted` property, which is now gone.

packages/core/src/typings.ts
nicomouss: Good one, thanks

packages/core/src/platforms/Platform.ts
```diff
@@ -505,11 +505,12 @@ export abstract class Platform {
     let ret = '';
 
     if (sql[0] === '?') {
+      const param = params[j++];
       if (sql[1] === '?') {
-        ret += this.quoteIdentifier(params[j++]);
+        ret += this.quoteIdentifier(param instanceof TableName ? param.toString() : param);
```
B4nan: what i meant here was moving the check inside the `quoteIdentifier` method, that should be the only place you need to alter in the platform code, the `formatQuery` method should stay as is.

nicomouss: a227246

Actually within ```quoteIdentifier``` base method in ```Platform.ts```, in master it is already calling ```id.toString()```. Not sure if this was intended, so I left it like so.

Only in the overrides for the different platform (MsSqlPlatform and PostgreSqlPlatform) that the TableName type check has been added.

B4nan: Hmm, now I wonder if we even need this handling and the `TableName` class, it was added because of the quoting issue which is now resolved in another way, with the `quote` helper. Maybe we should just call `id.toString` in the existing overrides too. It would indeed be faster than having instanceof checks on such a hot path.

I don't mind keeping the class itself, but the `instanceof` check is what we could get rid of.

nicomouss: Yes, I just also realized the TableName type check is indeed useless ...

[f5990aa](https://github.com/mikro-orm/mikro-orm/pull/6706/commits/f5990aa8425a8d96f1a42760b2a301ccf1784cae) - I just changed the signature to clearly indicate that the method now accepts `any` objects (not only `string`) as the `id` parameter. Under the hood, it's calling `id.toString()` to stringify the id.

nicomouss: I left the TableName class, but it has also become useless indeed. We could come back to using an object literal as the `table` parameter in the expression callback, with a basic `type` to define its structure.

B4nan: I am not sure if typing it to `any` is correct here, things like `null` or `undefined` would be allowed that way, and those do not have `toString` method. `string | { toString(): string }` would be more precise. 

> I left the TableName class, but it has also become useless indeed. We could come back to using an object literal as the table parameter in the expression callback, with a basic type to define its structure.

Will leave that up to you.

nicomouss: Yes, you are right, using `any` type might potentially cause some issues here (as often with `any` ...). The inline type you propose is definitely the way to go, way stronger.

And I guess it is better to leave things as clean as possible, so `TableName` is now back to its initial form, a `type` instead of a `class`.

We should have something pretty much close to the finish line now I guess.

[24ee815](https://github.com/mikro-orm/mikro-orm/pull/6706/commits/24ee815fbd1856494fc6ee4f74906da869479b9e)

packages/core/src/utils/RawQueryFragment.ts
```diff
@@ -244,3 +244,7 @@ sql.ref = <T extends object>(...keys: string[]) => raw<T, RawQueryFragment>('??'
 sql.now = (length?: number) => raw<Date, string>('current_timestamp' + (length == null ? '' : `(${length})`));
 sql.lower = <T extends object>(key: string | ((alias: string) => string)) => createSqlFunction('lower', key);
 sql.upper = <T extends object>(key: string | ((alias: string) => string)) => createSqlFunction('upper', key);
+
+export function quote(expParts: readonly string[], ...values: unknown[]) {
```
B4nan: one very last thing - lets add a jsdoc block similar to the `sql` helper above, and also brifly mention this method in the [raw fragments section](https://mikro-orm.io/docs/next/raw-queries) 

btw the `values` type can be `(string | { toString(): string })[]` here too i think, no?

nicomouss: [a9eb865](https://github.com/mikro-orm/mikro-orm/pull/6706/commits/a9eb865208b9f7fe5138c944ddc9ebcd5cd991bc)

> one very last thing - lets add a jsdoc block similar to the sql helper above, and also brifly mention this method in the [raw fragments section](https://mikro-orm.io/docs/next/raw-queries)

I also enriched documentation regarding the use of `'??'` within `raw` helper. Apart if I missed it, this did not seem to be documented yet; as users can also take that approach now to write their expression, I felt the documentation was needed on that as well.

> btw the values type can be (string | { toString(): string })[] here too i think, no?

Yes, thanks

nicomouss:
> perfect, thank you!

You're welcome

PR #6715, opened 2025-06-17
fix(sql): correctly handle lazy scalar properties with custom types

simonbrunel:
This PR fixes lazy support for custom type properties, which were always selected even if not explicitly populated.

### DIFF
--- a/packages/knex/src/AbstractSqlDriver.ts
+++ b/packages/knex/src/AbstractSqlDriver.ts
@@ -1673,17 +1673,20 @@ export abstract class AbstractSqlDriver<Connection extends AbstractSqlConnection
     }
 
     if (ret.length > 0 && !hasExplicitFields && addFormulas) {
-      meta.props
-        .filter(prop => prop.formula && !lazyProps.includes(prop))
-        .forEach(prop => {
+      for (const prop of meta.props) {
+        if (lazyProps.includes(prop)) {
+          continue;
+        }
+
+        if (prop.formula) {
           const a = this.platform.quoteIdentifier(alias);
           const aliased = this.platform.quoteIdentifier(prop.fieldNames[0]);
           ret.push(raw(`${prop.formula!(a)} as ${aliased}`));
-        });
-
-      meta.props
-        .filter(prop => !prop.object && (prop.hasConvertToDatabaseValueSQL || prop.hasConvertToJSValueSQL))
-        .forEach(prop => ret.push(prop.name));
+        }
+        if (!prop.object && (prop.hasConvertToDatabaseValueSQL || prop.hasConvertToJSValueSQL)) {
+          ret.push(prop.name);
+        }
+      }
     }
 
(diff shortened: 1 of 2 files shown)


packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -1673,17 +1673,20 @@ export abstract class AbstractSqlDriver<Connection extends AbstractSqlConnection
...
           const a = this.platform.quoteIdentifier(alias);
           const aliased = this.platform.quoteIdentifier(prop.fieldNames[0]);
           ret.push(raw(`${prop.formula!(a)} as ${aliased}`));
-        });
-
-      meta.props
-        .filter(prop => !prop.object && (prop.hasConvertToDatabaseValueSQL || prop.hasConvertToJSValueSQL))
-        .forEach(prop => ret.push(prop.name));
+        }
+        if (!prop.object && (prop.hasConvertToDatabaseValueSQL || prop.hasConvertToJSValueSQL)) {
```
simonbrunel, line 1686: I kept the logic from the previous code, which was that if `prop.formula` is true, it was still checking against the `hasConvertTo*SQL` methods. But I think this line could be a `} else if {` so we save a few checks in case the property is a formula.

packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -1673,17 +1673,20 @@ export abstract class AbstractSqlDriver<Connection extends AbstractSqlConnection
     }
 
     if (ret.length > 0 && !hasExplicitFields && addFormulas) {
-      meta.props
-        .filter(prop => prop.formula && !lazyProps.includes(prop))
-        .forEach(prop => {
+      for (const prop of meta.props) {
```
simonbrunel, line 1676: I merged the 2 `forEach` into a single regular `for ... of` loop so it's a bit more efficient since we don't check 2 times if `lazyProps` includes `prop`.

tests/issues/GH6715.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  location!: GeoPoint;
+
+}
+
+let orm: MikroORM;
+
+beforeAll(async () => {
+  orm = await MikroORM.init({
+    entities: [Place],
+    dbName: 'GHxxxx',
```
B4nan:
```suggestion
    dbName: '6715',
```

tests/issues/GH6715.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    port: 5433,
+  });
+
+  await orm.schema.execute('create extension if not exists postgis');
+  await orm.schema.refreshDatabase();
+});
+beforeEach(async () => {
+  await orm.schema.clearDatabase();
+});
```
B4nan: we have just one test, no need for this

```suggestion

```

B4nan (approved):
looking good, thanks!

PR #6716, opened 2025-06-17
feat(core): add `loadOrFail` to `ScalarReference` wrapper

Webbly25:
Add `loadOrFail()` to ScalarRef

### DIFF
--- a/packages/core/src/entity/Reference.ts
+++ b/packages/core/src/entity/Reference.ts
@@ -1,6 +1,7 @@
 import { inspect } from 'node:util';
 import type { FindOneOptions, FindOneOrFailOptions } from '../drivers/IDatabaseDriver';
 import { DataloaderType } from '../enums';
+import { NotFoundError } from '../errors';
 import type {
   AddEager,
   AddOptional,
@@ -234,7 +235,7 @@ export class ScalarReference<Value> {
       options.failHandler ??= wrapped.__em!.config.get('findOneOrFailHandler');
       const entityName = this.entity!.constructor.name;
       const where = wrapped.getPrimaryKey();
-      throw options.failHandler!(entityName, where);
+      throw new NotFoundError(`${entityName} (${where}) failed to load property '${this.property}'`);
     }
 
     return ret;
--- a/packages/core/src/entity/Reference.ts
+++ b/packages/core/src/entity/Reference.ts
@@ -1,4 +1,6 @@
 import { inspect } from 'node:util';
+import type { FindOneOptions, FindOneOrFailOptions } from '../drivers/IDatabaseDriver';
+import { DataloaderType } from '../enums';
 import type {
   AddEager,
   AddOptional,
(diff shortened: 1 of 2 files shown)


Webbly25:
I am unsure as to the errors to actually throw.

Also, where should I write the tests?
I saw there are tests for `Ref.loadOrFail()` in `types.test.ts`. Should I write them there?

packages/core/src/entity/Reference.ts
```diff
@@ -132,11 +132,11 @@ export class Reference<T extends object> {
...
-      const wrapped = helper(this.entity);
-      options.failHandler ??= wrapped.__em!.config.get('findOneOrFailHandler');
-      const entityName = this.entity.constructor.name;
-      const where = wrapped.getPrimaryKey() as any;
-      throw options.failHandler!(entityName, where);
+        const wrapped = helper(this.entity);
+        options.failHandler ??= wrapped.__em!.config.get('findOneOrFailHandler');
+        const entityName = this.entity.constructor.name;
+        const where = wrapped.getPrimaryKey() as any;
+        throw options.failHandler!(entityName, where);
```
B4nan: we use 2 spaces for indent

packages/core/src/entity/Reference.ts
```diff
@@ -222,6 +222,32 @@ export class ScalarReference<Value> {
     return this.value;
   }
 
+   /**
+   * Ensures the underlying entity is loaded first (without reloading it if it already is loaded).
+   * Returns the entity or throws an error just like `em.findOneOrFail()` (and respects the same config options).
+   */
```
B4nan: this is wrongly indented

```suggestion
  /**
   * Ensures the underlying entity is loaded first (without reloading it if it already is loaded).
   * Returns the entity or throws an error just like `em.findOneOrFail()` (and respects the same config options).
   */
```

B4nan (changes requested):
thanks! 

re testing, either add a specific test case for this feature (into `tests/features`) or just modify the existing tests for scalar refs, the big entity manager suites include entities with scalar refs, e.g. [Book2.perex](https://github.com/mikro-orm/mikro-orm/blob/9fa5fad5e3955cdcdee89aa12c8b3dd4841b2045/tests/entities-sql/Book2.ts#L51-L52) is one

packages/core/src/entity/Reference.ts
```diff
@@ -222,6 +222,32 @@ export class ScalarReference<Value> {
...
+  async loadOrFail(options: Omit<LoadReferenceOrFailOptions<any, any>, 'populate' | 'fields' | 'exclude'> = {}): Promise<Value> {
+    const ret = await this.load(options);
+
+    if (!ret) {
+      if (this.entity) {
```
B4nan: `this.entity` will always be available, no need to check this

packages/core/src/entity/Reference.ts
```diff
@@ -222,6 +222,32 @@ export class ScalarReference<Value> {
...
+        const where = wrapped.getPrimaryKey() as any;
+        throw options.failHandler!(entityName, where);
+      }
+
+      if (this.property) {
+        throw new Error(`${this.property} is not initialized`);
+      }
+
+      throw new Error(`ScalarRef not initialized`);
+    }
```
B4nan: i am not sure what you are trying to do here, is this code even reachable? i'd remove it completely

B4nan (changes requested):

B4nan (approved):
looking good, thanks!

PR #6725, opened 2025-06-21
feat(logger): add slow query logging support

### LINKED ISSUE #4695: Feature: Log long running queries
I'd like to be able to log slow queries to console/file and keep track of average query times

Add an option in config with `maxExecutionTime ` in ms  and log queries that exceed this time limit.

I have been using console.time and console.timeEnd but that is just the hacky way and is not precise

**Additional context**
https://thorben-janssen.com/hibernate-slow-query-log/#activating-hibernates-slow-query-log
https://orkhan.gitbook.io/typeorm/docs/logging#log-long-running-queries

sridhar-5:
closes #4695 

- Introduced a new logging namespace - `slowQuery`
- Logging all the Queries that takes beyond the configurable threshold of **>=200** ms. 
- Modified the log level to 'warning` for the slow queries. 
- Highlighting the meta data color of slow queries to yellow 

This is my first contribution to Mikro ORM, i do not have a lot of experience with typescript either, i tried to do this while learning the same. So please feel free to let me know if i missed some nuances, i'm happy to address the feedback.

### DIFF
--- a/packages/core/src/connections/Connection.ts
+++ b/packages/core/src/connections/Connection.ts
@@ -158,8 +158,10 @@ export abstract class Connection {
   }
 
   protected logQuery(query: string, context: LogContext = {}): void {
+    const isSlowQuery = typeof context.took === 'number' && context.took >= 200;
+
     this.logger.logQuery({
-      level: 'info',
+      level: isSlowQuery ? 'warning' : 'info',
       connection: {
         type: this.type,
         name: this.options.name || this.config.get('name') || this.options.host,
--- a/packages/core/src/logging/DefaultLogger.ts
+++ b/packages/core/src/logging/DefaultLogger.ts
@@ -8,6 +8,7 @@ export class DefaultLogger implements Logger {
   readonly writer: (message: string) => void;
   private readonly usesReplicas?: boolean;
   private readonly highlighter?: Highlighter;
+  private readonly slowQueryThreshold: number = 200;
 
   constructor(private readonly options: LoggerOptions) {
     this.debugMode = this.options.debugMode ?? false;
@@ -83,7 +84,10 @@ export class DefaultLogger implements Logger {
    * @inheritDoc
    */
   logQuery(context: { query: string } & LogContext): void {
-    if (!this.isEnabled('query', context)) {
+    const isQueryLoggingEnabled = this.isEnabled('query', context);
+    const shouldLogSlowQueries= this.isSlowQueryLoggingEnabled('slowQuery', context);
+
+    if (!isQueryLoggingEnabled && !shouldLogSlowQueries) {
       return;
     }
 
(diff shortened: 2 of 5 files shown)


packages/core/src/connections/Connection.ts
```diff
@@ -158,8 +158,10 @@ export abstract class Connection {
   }
 
   protected logQuery(query: string, context: LogContext = {}): void {
+    const isSlowQuery = typeof context.took === 'number' && context.took >= 200;
```
sridhar-5: was a bit unhappy about leaving the constant here, please do suggest if there is a better way to do this.

sridhar-5: Oh! May be i could change this slowQueryTheshold into a configurable option.

sridhar-5:
@B4nan - could i please get a review on this ?

B4nan:
thanks! so how is this supposed to work in production when logs are disabled? we need a way to log stuff to the file system specifically for the slow query logs (regardless of how the default logging is set up), your PR only adds the new logger namespace. you can already set up a file logger but that means all logs would go there, maybe we could allow having multiple loggers, and set them up with different namespaces. slow query logging could be enabled automatically this way (logging to a txt file only the slow queries, regardless of the `debug` ORM option)

sridhar-5:
Oh! Okay, so we want the `slowQuery` logs to appear even with the debug logs are disabled in the production. Got it. I can make those changes. Thanks for the feedback @B4nan .

B4nan:
I am not sure if this should be enabled by default, but to me this is a production feature, and in production you don't usually enable query logging, since it would slow down the app and pollute the app logs like crazy. I'd say the best way to support this would be allowing multiple loggers, they can log to different places (console, file, some 3rd party service), based on different settings (log levels). Alternatively, only the slow query logs could be treated as a special case, but the former approach sounds cleaner. I haven't been thinking much about this, also not sure how it works in other ORMs, let's first agree on how this could look like before spending more time with implementation.

dunklesToast:
Is this still being worked on? If not, could I take this over because we'd need such a feature.

B4nan:
This is one of those PRs where you provide a review and you never hear back from the author 🙃 

Feel free to work on this, but please target v7, I won't be merging more features to v6.

sridhar-5:
My sincerely apologies. I was managing applying to graduate schools and a full-time job and i lost track of this PR. I'm just finishing up my last application [will be done in a day]. I intend to continue contributing here from Sunday. But if that doesn't work, please feel free to continue. Please do let me know wha you think. Again my apologies, ghosting was not the intention.

sridhar-5:
> My sincerely apologies. I was managing applying to graduate schools and a full-time job and i lost track of this PR. I'm just finishing up my last application [will be done in a day]. I intend to continue contributing here from Sunday. But if that doesn't work, please feel free to continue. Please do let me know wha you think. Again my apologies, ghosting was not the intention.

Re-checking, should i continue on this or is it being taken up by @dunklesToast ? Let me know if i should close this.

dunklesToast:
I haven't begun working on this yet - wanted to give you some more time to check in. If you want to, you can still work on this from my side, but I guess @B4nan's comment regarding the v6 feature freeze would mean that you should re-create this feature from the v7 branch or merge v7 into this branch.

Also, on a slightly related note I wondered if we could also use this to clean up the logging params inside the MikroORM config object? Maybe a nested object like this:

```
logging: {
  ignoreDeprecations: boolean | string[];
  verbose: boolean;
  debug: boolean;
  logger: MikroORMLogger
  highlight: boolean;
  hightlightTheme: SQLHighlighter;
  maxExecutionTime: undefined | number; (
}
```

The variable name `maxExecutionTime` was adapted from [TypeORM](https://orkhan.gitbook.io/typeorm/docs/docs/advanced-topics/5-logging#log-long-running-queries) - not sure if y'all like it 🤷🏻

sridhar-5:
It does sound like a good refactor to me. I will see if I can do that too within the scope of this work. I will put up details about the approach I'm planning to take for slow query logging here tomorrow; we can work towards getting that aligned first and then i can make modifications to the code.

B4nan:
Note that v7 is already merged to master, but your branch is 517 commits behind, so better recreate the PR instead of trying to rebase this.

sridhar-5:
Makes sense. I will close this PR and recreate a new one. I will use the issue to discuss the approach first.

Issue #6739, opened 2025-07-06
Unexpected updates on composite FK relationships during select queries


jaichinger:
### Describe the bug

I think I’ve found an bug where change detection in the EM is causing unnecessary updates for certain select queries. The updates are just setting the same values.

We make heavy use of composite primary keys in our database schema, and the related functionality of sharing columns in foreign keys via fieldNames/ownColumns. So I’ve included that configuration where relevant for completeness.

I did some debugging and it looks like in the change set, the original pet entity had a object for the owner property (owner: { org: 1, id: 1 }) but the update payload’s value is (I’m assuming) a PK array (owner: [1, 1]).

I'm not sure if it's a hydration issue or not, as PK cond objects are being passed to the EntityComparator, when the comparator always expects a PK array.

**UPDATE:** I was able to provide a much simpler test case in `example-simple.test.ts`.

The new case has just a single m:1 relationship `Book` -> `Author`.

### Reproduction

https://github.com/jaichinger/mikro-orm-reproduction

### What driver are you using?

@mikro-orm/postgresql

### MikroORM version

6.4.16

### Node.js version

22.16.0

### Operating system

Linux 6.10.14

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/mikro-orm/mikro-orm/blob/master/CONTRIBUTING.md).
- [x] Read the [docs](https://mikro-orm.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/mikro-orm/mikro-orm/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mikro-orm/mikro-orm/discussions) or join our [Discord](https://discord.gg/w8bjxFHS7X).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

jaichinger:
Hey @B4nan I pushed a _much_ simpler test case to my reproduction repo, in a file called `example-simple.test.ts`.

I've kept the original complicated one around as well.

jaichinger:
Closing this issue for now. After investigating further, the issue we're experiencing looks to be caused by something else. I'll make another issue to keep things clean.

PR #6740, opened 2025-07-08
fix(core): normalize values when comparing array values in EntityComp…

jaichinger:
This PR modifies `EntityComparator.getPropertyComparator` to normalize properties of type array before comparing them. 

Sometimes a PK cond object is being passed to the comparator but the comparator expects last and current to both be arrays.

See #6739

### DIFF
--- a/packages/core/src/utils/EntityComparator.ts
+++ b/packages/core/src/utils/EntityComparator.ts
@@ -715,7 +715,17 @@ export class EntityComparator {
     }
 
     if (['array'].includes(type) || type.endsWith('[]')) {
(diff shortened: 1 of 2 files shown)


B4nan:
This feels like a workaround rather than a proper fix to me, I'll have a closer look at the repro to see if I can think of a more correct way to deal with that.

jaichinger:
> This feels like a workaround rather than a proper fix to me, I'll have a closer look at the repro to see if I can think of a more correct way to deal with that.

Yeah, I agree. I'm admittedly a little out of my depth with the codebase. Anyway, today I found other scenarios on our application where this workaround does not work, and unexpected updates are still being triggered.

I'll do some more investigating and update the repro/test cases

PR #6756, opened 2025-07-15
feat(core): add `RequiredNullable` helper type

zachkirsch:
Followup from https://github.com/mikro-orm/mikro-orm/discussions/6744#discussioncomment-13766303

This PR adds a type `RequiredNullable ` that can be used to make an entity field nullable, but still require the user to supply a value (`null` or otherwise) when instantiating 

```ts
export class Article extends BaseEntity {

  @Property({ length: 1000 })
  description: RequiredNullable<string>;

  // ...

}

em.create(Article, { description: "Description!" }); // ok
em.create(Article, { description: null }); // ok
em.create(Article, {}); // compile error: missing description
```

### DIFF
--- a/docs/docs/defining-entities.md
+++ b/docs/docs/defining-entities.md
@@ -394,20 +394,64 @@ properties: {
 
 To make a nullable field required in methods like `em.create()` (i.e. you cannot omit the property), use `RequiredNullable` type. Such property needs to be provided explicitly in the `em.create()` method, but will accept a `null` value.
 
-```ts title='book.entity.ts'
-export class Book extends BaseEntity {
 
-  @Property()
-  title: RequiredNullable<string>;
+<Tabs
+  groupId="entity-def"
+  defaultValue="reflect-metadata"
+  values={[
+    {label: 'reflect-metadata', value: 'reflect-metadata'},
+    {label: 'ts-morph', value: 'ts-morph'},
+    {label: 'defineEntity', value: 'define-entity'},
+    {label: 'EntitySchema', value: 'entity-schema'},
+  ]
+  }>
+  <TabItem value="reflect-metadata">
 
-  // ...
-
-}
+```ts title='./entities/Book.ts'
+@Property()
+title: RequiredNullable<string>;
 
 em.create(Book, { title: "Alice in Wonderland" }); // ok
 em.create(Book, { title: null }); // ok
 em.create(Book, {}); // compile error: missing title
 ```
+
+  </TabItem>
+  <TabItem value="ts-morph">
+
+```ts title='./entities/Book.ts'
+@Property()
+title: RequiredNullable<string>;
+
+em.create(Book, { title: "Alice in Wonderland" }); // ok
+em.create(Book, { title: null }); // ok
+em.create(Book, {}); // compile error: missing title
+```
+
+  </TabItem>
+  <TabItem value="define-entity">
+
+```ts title='./entities/Book.ts'
+// TODO
+
+em.create(Book, { title: "Alice in Wonderland" }); // ok
+em.create(Book, { title: null }); // ok
+em.create(Book, {}); // compile error: missing title
+```
+
+  </TabItem>
+  <TabItem value="entity-schema">
+
+```ts title="./entities/Author.ts"
+// TODO
+
+em.create(Book, { title: "Alice in Wonderland" }); // ok
+em.create(Book, { title: null }); // ok
+em.create(Book, {}); // compile error: missing title
+```
+
+  </TabItem>
+</Tabs>
 
 ## Default values
 
--- a/docs/docs/defining-entities.md
+++ b/docs/docs/defining-entities.md
@@ -392,6 +392,23 @@ properties: {
   </TabItem>
 </Tabs>
 
+To make a nullable field required (i.e. you cannot omit the property), use `RequiredNull`:
+
+```ts title='book.entity.ts'
+export class Book extends BaseEntity {
+
+  @Property()
+  title: string | RequiredNull;
+
+  // ...
+
+}
+
+em.create(Book, { title: "Alice in Wonderland" }); // ok
+em.create(Book, { title: null }); // ok
+em.create(Book, {}); // compile error: missing description
+```
+
 ## Default values
 
 We can set default value of a property in 2 ways:
--- a/docs/docs/defining-entities.md
+++ b/docs/docs/defining-entities.md
@@ -392,13 +392,13 @@ properties: {
   </TabItem>
 </Tabs>
 
-To make a nullable field required (i.e. you cannot omit the property), use `RequiredNull`:
+To make a nullable field required (i.e. you cannot omit the property), use `RequiredNullable`:
 
 ```ts title='book.entity.ts'
 export class Book extends BaseEntity {
 
   @Property()
-  title: string | RequiredNull;
+  title: RequiredNullable<string>;
 
   // ...
 
--- a/docs/docs/defining-entities.md
+++ b/docs/docs/defining-entities.md
@@ -392,7 +392,7 @@ properties: {
   </TabItem>
 </Tabs>
 
-To make a nullable field required (i.e. you cannot omit the property), use `RequiredNullable`:
+To make a nullable field required in methods like `em.create()` (i.e. you cannot omit the property), use `RequiredNullable` type. Such property needs to be provided explicitly in the `em.create()` method, but will accept a `null` value.
 
 ```ts title='book.entity.ts'
 export class Book extends BaseEntity {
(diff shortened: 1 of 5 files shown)


zachkirsch:
@B4nan where's the best place to add tests for this?

docs/versioned_docs/version-6.1/guide/02-relationships.md
```diff
@@ -371,6 +371,23 @@ export class Article extends BaseEntity {
 }
 ```
 
+By default, any nullable properties are optional when creating an instance of an entity. To make a nullable field required (i.e. you cannot omit the property), use `RequiredNull`:
```
B4nan: you've modified v6.1 docs snapshot, this should go [here](https://github.com/mikro-orm/mikro-orm/tree/master/docs/docs) instead

also should be documented [here](https://mikro-orm.io/docs/defining-entities#optional-properties)

docs/versioned_docs/version-6.1/guide/02-relationships.md
```diff
@@ -371,6 +371,23 @@ export class Article extends BaseEntity {
...
+
+```ts title='article.entity.ts'
+export class Article extends BaseEntity {
+
+  @Property({ length: 1000 })
+  description: string | RequiredNull;
```
B4nan: this way the type wont be considered nullable for the TS compiler, right? 

I was suggesting the approach with generics (`RequiredNullable<T>`) as that could work fine in that regard, while you'd still be able to catch that type when [deciding if a property is required or optional](https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/typings.ts#L336-L344) (thanks to the brand). I think that is the only place that needs to be touched, the rest of your changes wouldn't be necessary that way.

B4nan (changes requested):
tests could go here

https://github.com/mikro-orm/mikro-orm/blob/master/tests/types.test.ts

docs/versioned_docs/version-6.1/guide/02-relationships.md
B4nan: Another approach would be with the symbol (like `OptionalProps`), it's less ergonomic, but you don't have to adjust the property type at all.

zachkirsch:
@B4nan just pushed up with changes - take a look!

packages/core/src/typings.ts
```diff
@@ -300,6 +304,8 @@ export type EntityDataProp<T, C extends boolean> = T extends Date
 
 export type RequiredEntityDataProp<T, O, C extends boolean> = T extends Date
   ? string | Date
+  : { [__requiredNullable]?: 1 } extends T
+  ? T | null
```
B4nan: please adjust the indents so it was like before - indent the next branch (more spacing for the `?` and the following `:` line), same for the other changes in this file.

packages/core/src/typings.ts
```diff
@@ -337,6 +343,8 @@ type IsOptional<T, K extends keyof T, I> = T[K] extends Collection<any, any>
   ? true
   : ExtractType<T[K]> extends I
     ? true
+    : { [__requiredNullable]?: 1 } extends T[K]
+    ? false
     : K extends ProbablyOptionalProps<T>
```
B4nan: maybe it would be better to handle this in the `ProbablyOptionalProps` type?

packages/core/src/typings.ts
```diff
@@ -1084,6 +1092,8 @@ export type ExpandProperty<T> = T extends Reference<infer U>
 type LoadedLoadable<T, E extends object> =
   T extends Collection<any, any>
   ? LoadedCollection<E>
+  : { [__requiredNullable]?: 1 } extends T
+  ? T | null
```
B4nan: is this needed? if so, can you add a test case? the property will already include `| null`.

tests/types.test.ts
```diff
@@ -1031,4 +1032,23 @@ describe('check typings', () => {
     }
   });
 
+  test('RequiredNull', () => {
```
B4nan:
```suggestion
  test('RequiredNullable', () => {
```

tests/types.test.ts
```diff
@@ -1031,4 +1032,23 @@ describe('check typings', () => {
...
+    }
+
+    const em = { create: jest.fn() as any } as EntityManager;
+
+    // @ts-expect-error
+    em.create(User, { });
+
+    em.create(User, { requiredNullableString: 'some string' });
+    em.create(User, { requiredNullableString: null });
+
```
B4nan, line 1052: I'd also like to see test with the usage of the property, e.g. can you still do `user.requiredNullableString?.toLowerCase()`?

packages/core/src/typings.ts
zachkirsch: for sure! btw is there a formatter like prettier i can use here? wasn't able to get it working

zachkirsch:
@B4nan it seems that `null` is treated specially by typescript and keep disappearing the types


Screenshot 2025-07-16 at 9 46 52 AM

B4nan:
What about `export type RequiredNullable<T = never> = T & { [__requiredNullable]?: 1 } | null`?

zachkirsch:
that worked great!! just pushed - take a look at your convenience

docs/docs/defining-entities.md
```diff
@@ -392,6 +392,23 @@ properties: {
   </TabItem>
 </Tabs>
 
+To make a nullable field required (i.e. you cannot omit the property), use `RequiredNullable`:
+
+```ts title='book.entity.ts'
+export class Book extends BaseEntity {
```
B4nan: the example should use tabs as all the other examples on this page, showing different metadata providers

docs/docs/defining-entities.md
```diff
@@ -392,6 +392,23 @@ properties: {
   </TabItem>
 </Tabs>
 
+To make a nullable field required (i.e. you cannot omit the property), use `RequiredNullable`:
```
B4nan:
```suggestion
To make a nullable field required in methods like `em.create()` (i.e. you cannot omit the property), use `RequiredNullable` type. Such property needs to be provided explicitly in the `em.create()` method, but will accept a `null` value.
```

docs/docs/guide/02-relationships.md
```diff
@@ -373,6 +373,23 @@ export class Article extends BaseEntity {
 }
 ```
 
+By default, any nullable properties are optional when creating an instance of an entity. To make a nullable field required (i.e. you cannot omit the property), use `RequiredNullable`:
```
B4nan: let's adjust the text in the same way here

docs/docs/defining-entities.md
```diff
@@ -392,6 +392,67 @@ properties: {
...
+em.create(Book, { title: "Alice in Wonderland" }); // ok
+em.create(Book, { title: null }); // ok
+em.create(Book, {}); // compile error: missing title
+```
+
+  </TabItem>
+  <TabItem value="define-entity">
+
+```ts title='./entities/Book.ts'
+// TODO
```
zachkirsch: @B4nan can you help me with the example here?

docs/docs/defining-entities.md
```diff
@@ -392,6 +392,67 @@ properties: {
...
+em.create(Book, { title: "Alice in Wonderland" }); // ok
+em.create(Book, { title: null }); // ok
+em.create(Book, {}); // compile error: missing title
+```
+
+  </TabItem>
+  <TabItem value="entity-schema">
+
+```ts title="./entities/Author.ts"
+// TODO
```
zachkirsch: and here

zachkirsch:
@B4nan this is ready for another review when you have a chance!

zachkirsch:
@B4nan bump on this :)

B4nan:
Instead of bumping, you could try to resolve those two TODOs 🙃 I don't have the answer myself, so I will need to try - just like you can try yourself.

I'll get to this eventually (before next release).

docs/docs/defining-entities.md
B4nan: this will likely need to use the `$type` method, to override the resulting property type

https://mikro-orm.io/docs/next/defining-entities#mapping-directly-to-primary-keys
https://mikro-orm.io/docs/next/custom-types#mapping-to-objects-and-type-safety

docs/docs/defining-entities.md
B4nan: this should look the same as the examples for https://mikro-orm.io/docs/next/defining-entities#default-values i guess, so maybe better to expand all the variants with a more verbose version like we have for the default values

B4nan:
Sorry for the wait, and thank you for your contribution!

zachkirsch:
Thank you!! Was hoping to try to finish it up this week but you beat me to it :)

Issue #470, opened 2020-04-08
Some entities are not persisted when inserting in parallel


Yaty:
Hello

I'm writing some tests for my application. For those tests I need to create multiple users and persist them in the database.

I'm using MikroORM with Nest.js.

If I create them this way it works, every user is persisted and every user entity I get in return have an id :
```ts
const users = [
  await utils.givenUser(),
  await utils.givenUser(),
  await utils.givenUser(),
  await utils.givenUser(),
];
```

But if I use `Promise.all` multiple weird things are happening.

First case, I try to create two users : 

```ts
const users = await Promise.all([
  utils.givenUser(),
  utils.givenUser(),
]);
```

`users[0].id` is undefined and `users[1].id` is defined

With three users : `users[0].id` and `users[2].id` are defined, `users[1].id` is undefined

With four users : `users[0].id` and `users[2].id` are defined, `users[1].id` and `users[3].id` are undefined

With five users : `users[0].id`, `users[2].id` and `users[3].id` are defined, `users[1].id` and `users[4].id`  are undefined

With six users and more an error is thrown : 

```
TypeError: Cannot read property 'setValue' of undefined

    at ChangeSetPersister.mapPrimaryKey (/home/hugo/projets/paffme/node_modules/mikro-orm/dist/unit-of-work/ChangeSetPersister.js:49:53)
    at ChangeSetPersister.persistEntity (/home/hugo/projets/paffme/node_modules/mikro-orm/dist/unit-of-work/ChangeSetPersister.js:39:18)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at ChangeSetPersister.persistToDatabase (/home/hugo/projets/paffme/node_modules/mikro-orm/dist/unit-of-work/ChangeSetPersister.js:20:9)
    at UnitOfWork.commitChangeSet (/home/hugo/projets/paffme/node_modules/mikro-orm/dist/unit-of-work/UnitOfWork.js:208:9)
    at UnitOfWork.persistToDatabase (/home/hugo/projets/paffme/node_modules/mikro-orm/dist/unit-of-work/UnitOfWork.js:342:13)
```

Sometimes it works with 6 users but it seems the more I increase the number of entity to create the more chance I have to get the error.

I had some cases where the error was thrown for 4 or 5 users creation. The ordering between defined and undefined id seems to be the same every time.

I cleaned the table, tried to create 6 users, got the error and only 5 users were created : 1, 2, 3, 5, 6

[...]

PR #6762, opened 2025-07-17
test: add test for entity removal in parallel forks

jtomaszewski:
After battling for a few hours, I think I've found a bug in Mikro ORM.

Basically we have a complex web app that has separate event consumers running in parallel, each doing its' own thing.

Think of it as running a `Promise.all(...)` on a few promises, where each promise is running code in its own EM fork, via the `EntityManager#transactional(...)` callback.

Now, we've noticed that if we remove an entity in one of such forks, the entity gets unintentionally repersisted whenever `.flush()` is called in any other forks (e.g. when their `.transactional()` callback comes to an end).

I've wrote a test case that reproduces this.

@B4nan , any idea what can we do about this?

It seems to be a serious blocker to us, because:

- it may cause our app to randomly not to work as we wished. we've met this issue accidentally, because some validations in the entity have started failing when it had been repersisted after being deleted; and it took us a good half of day to find the root cause. i'm afraid we'll meet this issue many more times in the future.
- I can't find of a workaround? we definitely want transaction wrapping and forking for each of our event consumer's code.

### DIFF
--- a/tests/issues/entity-removal-parallel-forks.test.ts
+++ b/tests/issues/entity-removal-parallel-forks.test.ts
@@ -39,3 +39,28 @@ test('if multiple forks are created in parallel, an entity can be removed in one
   const countAfterSecondFlush = await orm.em.count(Author3.entity, { id: entity.id });
   expect(countAfterSecondFlush).toBe(0);
 });
+
+
+test('if multiple transactional calls are being made in parallel, an entity can be removed in one callback without being re-persisted in another', async () => {
+  const entity = new Author3.entity('test author', 'test@test.com');
+  await orm.em.persistAndFlush(entity);
+
+  const tx1 = orm.em.transactional(async em => {
+    em.remove(entity);
+    await em.flush();
+
+    const countAfterDelete = await orm.em.count(Author3.entity, {
+      id: entity.id,
+    });
+    expect(countAfterDelete).toBe(0);
+  });
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  const tx2 = orm.em.transactional(() => {});
+
+  await Promise.all([tx1, tx2]);
+
+  const countAfterAll = await orm.em.count(Author3.entity, {
+    id: entity.id,
+  });
+  expect(countAfterAll).toBe(0);
+});
--- a/tests/issues/entity-removal-parallel-forks.test.ts
+++ b/tests/issues/entity-removal-parallel-forks.test.ts
@@ -1,0 +1,41 @@
+import { MikroORM } from '@mikro-orm/core';
+import { SqliteDriver } from '@mikro-orm/sqlite';
+import { initORMSqlite } from '../bootstrap';
+
+const { Author3, Book3, BookTag3, Publisher3, Test3, BaseEntity4 } = require('../entities-js/index');
+
+let orm: MikroORM<SqliteDriver>;
+
+beforeAll(async () => {
+  orm = await initORMSqlite();
+  await orm.schema.ensureDatabase();
+  await orm.schema.refreshDatabase();
+});
+
+beforeEach(async () => {
+  await orm.schema.clearDatabase();
+});
+
+afterAll(async () => {
+  await orm.close(true);
+});
+
+test('if multiple forks are created in parallel, an entity can be removed in one fork without being re-persisted in another', async () => {
+  const entity = new Author3.entity('test author', 'test@test.com');
+  await orm.em.persistAndFlush(entity);
+
+  const fork1 = orm.em.fork({ clear: false });
+  const fork2 = orm.em.fork({ clear: false });
+
+  fork1.remove(entity);
+  await fork1.flush();
+
+  const countAfterDelete = await orm.em.count(Author3.entity, { id: entity.id });
+  expect(countAfterDelete).toBe(0);
+
+  const anotherEntity = new Publisher3.entity('test publisher');
+  await fork2.persistAndFlush(anotherEntity);
+
+  const countAfterSecondFlush = await orm.em.count(Author3.entity, { id: entity.id });
+  expect(countAfterSecondFlush).toBe(0);
+});


tests/issues/entity-removal-parallel-forks.test.ts
```diff
@@ -0,0 +1,41 @@
...
+  await fork1.flush();
+
+  const countAfterDelete = await orm.em.count(Author3.entity, { id: entity.id });
+  expect(countAfterDelete).toBe(0);
+
+  const anotherEntity = new Publisher3.entity('test publisher');
+  await fork2.persistAndFlush(anotherEntity);
+
+  const countAfterSecondFlush = await orm.em.count(Author3.entity, { id: entity.id });
+  expect(countAfterSecondFlush).toBe(0);
```
jtomaszewski, line 40: This currently fails:

image

If one comments out these lines, then it works:

```
  const anotherEntity = new Publisher3.entity('test publisher');
  await fork2.persistAndFlush(anotherEntity);
```

But whether publisher3 is created or not, shouldn't have any effect on author3 entities, so it should work anyways, no matter if these lines are commented out, or not.

B4nan:
To me this is a bug in your code, not in the ORM. You have two copies of the identity map, since you create two forks without clearing the context, and you later flush the other one that still holds the entity reference. This is a misusage to me, I don't see anything to fix really - you should not fork without clearing context if you want to do stuff in parallel like this.

jtomaszewski:
> you should not fork without clearing context if you want to do stuff in parallel like this.

In my actual app, all I actually do is just use `@Transactional` decorator in some methods of my classes, e.g. event consumers.

Perhaps the default options of the `.transactional` function should be to fork EM with a clean identity map? [Currently it is the opposite.](https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/EntityManager.ts#L1289) Which leads to very unintuitive behaviour. 

In https://github.com/mikro-orm/mikro-orm/issues/470#issuecomment-2865523326, you've mentioned

> Nowadays you can flush in parallel

Perhaps, but when it comes to `.transactional` function and how it works currently, **one simply cannot use `transactional` callbacks** in parallel. The current behaviour is unpredictable. Entities maybe will get removed, maybe not.

To elaborate my point, I've just pushed a second example, this time with usage of `.transactional()`. Please take a look at that.

Could we now agree there's something to be improved in here?

tests/issues/entity-removal-parallel-forks.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  });
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  const tx2 = orm.em.transactional(() => {});
+
+  await Promise.all([tx1, tx2]);
+
+  const countAfterAll = await orm.em.count(Author3.entity, {
+    id: entity.id,
+  });
+  expect(countAfterAll).toBe(0);
```
jtomaszewski, line 65: this currently also fails - the author doesn't get removed in the end. I guess it is being removed in `tx1`, and then re-persisted by the flush of the `tx2`.

B4nan:
> one simply cannot use transactional callbacks in parallel

That's a bit too exaggerated, you just need to call it on a fresh fork, or use `clear: true` in the options. Flushing multiple forks in parallel that share context (which is what your test is doing) was always a bad idea.

> Perhaps the default options of the .transactional function should be to fork EM with a clean identity map?

That would be a breaking change. We don't clear in that place because it was never meant to be a "way to run stuff in parallel", it is a way to "run things in explicit transaction". If you want to run things in parallel, use fresh forks, it's simple as that. Maybe we could change how the `@Transactional` decorator works, that's something I'd consider, but even that change would be breaking - for now you'll need to be explicit and use `clear: true` in the decorator options.

jtomaszewski:
Shouldn't we add at least a warning somewhere in docs, that the `transactional` decorator/method can't be used in parallel code paths (that relies on e.g. `Promise.all`), unless one passes `clear: true` option for all the transactions?

E.g. in here https://mikro-orm.io/docs/transactions . What do you think? I can submit a PR.

Looking back, I think I've met this issue several times in my last few months, in total losing quite a bit of time on this, until I found out about the root cause and this problem reproduction here.

I had no idea that MikroORM transactions _by default_ don't work with parallelization. Making operations in parallel on the back-end has always been a standard practice to me.

B4nan:
We could try to find a way to resolve the re-insert, that might be relatively simple, but my point is you are doing something wrong, and it might have more side effects than just this (e.g. unwanted updates, loss of state, unrelated changes in entity state in general caused by _the other fork_). You have a shared state in your forks (forking with `clear: false` uses the exact same entity references, just the identity maps are duplicated), which is the root cause of your problem.

B4nan:
Yeah, docs can always be improved, no problems with that, you are right that this is not mentioned there at all, neither the fact that the changes from the inner context will be propagated up.

And I'll open an RFC about this, I am not completely opposed to changing the defaults of `em.transactional` in v7.

jtomaszewski:
You're saying I'm doing something _wrong_ — yet all I'm doing, is just adding/removing data in a `.transactional` callbacks in parallel way. I wouldn't call that _wrong_. I think it's something I'd expect from all back-end devs to be doing in general. (Wrapping their operations in transactions; and parallelizing as much of their operations as possible.)

> We could try to find a way to resolve the re-insert, that might be relatively simple, but my point is you are doing something wrong, and it might have more side effects than just this (e.g. unwanted updates, loss of state, unrelated changes in entity state in general caused by the other fork).

I think I understand. We can have parallel code paths doing changes to the very same objects. How would MikroORM know whether a particular change `x` was caused by the transaction (context / async stack trace) `t1` or `t2`, to know whether that change should be flushed inside of the `t1` or `t2` ? :s 

Perhaps there's no easy fix to this.

> Yeah, docs can always be improved,  (...) And I'll open an RFC about this

Thank you a lot.

B4nan:
> You're saying I'm doing something wrong — yet all I'm doing, is just adding/removing data in a .transactional callbacks in parallel way. I wouldn't call that wrong. I think it's something I'd expect from all back-end devs to be doing in general. (Wrapping their operations in transactions; and parallelizing as much of their operations as possible.)

Well, you missed the important bit here, the _wrong_ part is that you first work with the upper EM and keep stuff in its identity map before you do this.

B4nan:
#6763

B4nan:
I'll close this one too in favor of the RFC

PR #6764, opened 2025-07-18
Removed items still appear in the reloaded collection (if prior to that they've been removed in a clean transaction)

jtomaszewski:
After the discussion in https://github.com/mikro-orm/mikro-orm/pull/6762 , I'm now replacing all `@Transactional()` calls in my codebase to be `@Transactional({ clear: true })` .

However, now I've met another strange behaviour (perhaps unrelated completely to that other ticket). 

After removing an entity in the transaction, I am still getting that entity in another entity's collection when calling `.loadItems({ refresh: true })` , even though the entity does not exist in the database anymore.

Am I doing something wrong, or is it MikroORM incorrectly not reloading the collection when asked to?

Note: If I remove `clear: true` transaction option, then the test passes.

### DIFF
--- a/tests/issues/item-removal-in-clean-transaction.test.ts
+++ b/tests/issues/item-removal-in-clean-transaction.test.ts
@@ -1,4 +1,4 @@
-import { MikroORM } from '@mikro-orm/core';
+import { DataloaderType, MikroORM } from '@mikro-orm/core';
 import { SqliteDriver } from '@mikro-orm/sqlite';
 import { initORMSqlite } from '../bootstrap';
 
@@ -7,7 +7,9 @@ const { Author3, Book3, BookTag3, Publisher3, Test3, BaseEntity4 } = require('..
 let orm: MikroORM<SqliteDriver>;
 
 beforeAll(async () => {
-  orm = await initORMSqlite();
+  orm = await initORMSqlite({
+    dataloader: DataloaderType.ALL,
+  });
   await orm.schema.ensureDatabase();
   await orm.schema.refreshDatabase();
 });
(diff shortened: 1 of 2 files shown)


tests/issues/item-removal-in-clean-transaction.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    expect(books.map(b => b.title)).toEqual(['book 2', 'book 3']);
+  }, { clear: true });
+
+  expect(await orm.em.findOne(Book3.entity, { title: 'book 1' })).toBeNull();
+
+  const booksBeforeReload: { title: string }[] = author.books.getItems();
+  expect(booksBeforeReload.map(b => b.title)).toEqual(['book 1', 'book 2', 'book 3']);
+
+  const booksAfterReload: { title: string }[] = await author.books.loadItems({ reload: true });
+  expect(booksAfterReload.map(b => b.title)).toEqual([
```
jtomaszewski, line 49: incorrectly, this currently returns "book 1" as well.

image

B4nan: there is no `reload` option, its called `refresh`.

B4nan: would be better if you pick any other base for your test and not the JS entities, since those are not type safe. these tests are actually removed from v7 already.

jtomaszewski: I've renamed `reload` to `refresh`. The test started passing.

**But, after adding `dataloader: DataloaderType.ALL` option to the ORM config, the test fails again.**

So there's still some issue here (but it may also be connected to dataloader).

(That's how I've met this issue in my app - I am using dataloader too.)

> would be better if you pick any other base for your test and not the JS entities, since those are not type safe. these tests are actually removed from v7 already.

Will do next time, thanks.

B4nan: I see, that indeed looks like a bug, the queries are fired correctly and return the two books, but it gets hydrated to an empty array for some reason.

B4nan: Looking closer at this, the problem originates in the propagation of changes from `em.transactional`, which is triggered regardless of the `clear` option (that only controls if the inner for is clear, not the propagation to the upper context). it currently works by registering the entities from the fork to the upper context - but if the entity was already there, we are not trying to merge the state with it, so you end up with two copies of the same entity. Dataloader depends on the identity when it tries to match the entities with the relations, so that's why it fails.

B4nan:
2b608fc fixes the issue with dataloader, and adds some docs around the context propagation in `em.transactional`

jtomaszewski:
Thank you, the docs seem okay, and the warning there should especially be useful.

PR #6782, opened 2025-07-31
typescript type support in where clause with subquery

dbwhddn10:
typescript type support subquery in where clause like below code.
```
  const users = await em.find(User, {
    id: {
      $in: subQuery,
    },
  });
```

I change only ObjectQuery to QBFilterQuery

QBFilterQuery definition is 
```
type QBFilterQuery<T = any> = ObjectQuery<T> | Dictionary;
```

### DIFF
--- a/packages/core/src/typings.ts
+++ b/packages/core/src/typings.ts
@@ -183,7 +183,7 @@ export type ExpandQuery<T> = T extends object
 export type EntityProps<T> = { -readonly [K in EntityKey<T>]?: T[K] };
 export type ObjectQuery<T> = OperatorMap<T> & FilterObject<T>;
 export type FilterQuery<T> =
-  | ObjectQuery<T>
+  | QBFilterQuery<T>
   | NonNullable<ExpandScalar<Primary<T>>>
   | NonNullable<EntityProps<T> & OperatorMap<T>>
   | FilterQuery<T>[];


B4nan:
> I change only ObjectQuery to QBFilterQuery

This literally means removing type safety, that's not an option.

I'd like to support this through the `raw` helper instead.

dbwhddn10:
@B4nan thanks for reply.

I suggested it because it seems simpler and is already supported in QueryBuilder, as shown in the code below.

```
const subQuery = em.createQueryBuilder(User).select('id').where(...).getKnexQuery();
const posts = em.createQueryBuilder(Post)
  .where({author_id: subQuery}) // <- QBFilterQuery
  .getResultList();
```


Is it safe from SQL injection attack when using raw helper?
Is this your recommended way? 

```
const subQuery = em.createQueryBuilder(User).select('id').where(...);
const result = await em.findAll(Post, {
    where: {
      author: {
        id: {
          $in: raw(`(${subQ.getQuery()})`, subQ.getParams()),
        },
      },
    },
  });
```

B4nan:
QB is not type safe, you can put anything in the `qb.where()`. For now I would just cast the subquery to `any` and call it a day, the `raw` helper also works thanks to the return type being `any`. What I meant with the `raw` helper is that I want to have a version exported from the SQL drivers that will work directly with a QB instance (both the ORM QB and knex QB), e.g. `raw(qb)`.

PR #6786, opened 2025-08-02
docs: clarify why `JsonType` is needed

adriancuadrado:
Explained how the json type is persisted differently depending on the database used.


B4nan (approved):
thanks

Issue #1603, opened 2021-03-25
Allow passing collation (or other extended options) to mongoDB query


mrkswrnr:
**Is your feature request related to a problem? Please describe.**
To retrieve users by their email address, I would like to use a (previously created) collation index. Unfortunately I can't pass the [necessary options](https://docs.mongodb.com/manual/reference/method/db.collection.find/#specify-collation) to the find method of the mongo client. In the `MongoDriver` unknown `FindOneOptions`/`FindOptions`-properties are not passed to the `MongoConnection` or `mongodb.collection.find`.

**Describe the solution you'd like**
I would like to pass a property `collation` in the FindOptions for `findOne` or `find`, which is passed directly to the mongodb.collection.find function. 

**Describe alternatives you've considered**
Instead of using a collation index, I could make the search case-insensitive by using a regular expression. However, this does not seem to me to be the recommended way. 

Alternatively, I could set the collation for the entire collection, which would result in this collation being used by default - without having to include the collation in the `FindOptions`. But then all other queries would also use this collation, which is not desired e.g. for first/last names.

**Additional context**
I could imagine different approaches to solve this problem: 
1. the find methods in `MongoDriver` and `MongoConnection` are extended by `options.collation`.
2. the find methods are changed so that excess options are passed directly to the `mongo.collection.find` method.
3. another hook `beforeFind` allows to manipulate the query (this seems to me a disproportionate amount of effort to fulfill this requirement, but might allow future use cases) 

Is this feature something you would consider? In any case, thanks for the work on MikroORM!

luffy1-1:
Any update on this or workaround? I need to use collation for case-insensitive sorting.

PR #6792, opened 2025-08-05
feat: MongoDB CommandOperationOptions

peter-lapin-powerus:
### 🛠 Changes
This PR adds support for passing native MongoDB `CommandOperationOptions` (such as `comment`, `maxTimeMS`, etc.) from the MikroORM driver layer all the way down to the native MongoDB client.

### Motivation
In our use case, we wanted to leverage MongoDB's native [comment](https://mongodb.github.io/node-mongodb-native/5.0/interfaces/CommandOperationOptions.html#comment) option to associate contextual metadata (such as audit information) with each write operation. However, MikroORM currently does not expose a way to pass these options through its driver abstraction.

### What's changed?
A new optional field `commandOperationOptions?: Record<string, any>` has been added to `NativeInsertUpdateOptions<T>` and related option interfaces.

This parameter is propagated through:
- `MongoDriver.nativeInsert`, `nativeInsertMany`, `nativeUpdate`, `nativeUpdateMany`, `nativeDelete`
- `MongoConnection.deleteMany`
- Internally in `MongoConnection.runQuery()`, where it is spread into the final native client operation options

This makes it possible to pass custom options like:

```
em.nativeUpdate(EntityName, where, data, {
  commandOperationOptions: { comment: { lastUpdatedBy: { user: 12345 } }
});
```

### 🚧 Limitations
No new tests have been added yet, but the existing test suite is passing (yarn test).
This feature currently assumes that commandOperationOptions is a flat object — no validation or type safety for deeper command semantics has been implemented.

### ✅ Next Steps
Add dedicated test coverage for the propagation logic.
Optionally consider stricter typing by using `Partial<CommandOperationOptions>` instead of `Record<string, any>`.

### DIFF
--- a/packages/mongodb/src/MongoConnection.ts
+++ b/packages/mongodb/src/MongoConnection.ts
@@ -38,7 +38,7 @@ import {
   type TransactionEventBroadcaster,
   type UpsertOptions,
   type UpsertManyOptions,
-  type LoggingOptions, DeleteOptions,
+  type LoggingOptions,
 } from '@mikro-orm/core';
 
 export class MongoConnection extends Connection {


packages/mongodb/src/MongoConnection.ts
```diff
@@ -232,8 +232,11 @@ export class MongoConnection extends Connection {
     return this.runQuery<T>('bulkUpdateMany', collection, data, where, ctx, upsert, upsertOptions);
   }
 
-  async deleteMany<T extends object>(collection: string, where: FilterQuery<T>, ctx?: Transaction<ClientSession>): Promise<QueryResult<T>> {
-    return this.runQuery<T>('deleteMany', collection, undefined, where, ctx);
+  async deleteMany<T extends object>(collection: string, where: FilterQuery<T>, deleteOptions: {
```
B4nan, line 235: this is a breaking change on API level, we cant just change shape of existing parameters

packages/core/src/drivers/IDatabaseDriver.ts
```diff
@@ -222,6 +222,8 @@ export interface NativeInsertUpdateOptions<T> {
   /** `nativeUpdate()` only option */
   upsert?: boolean;
   loggerContext?: LogContext;
+  /** MongoDb only @CommandOperationOptions */
+  commandOperationOptions?: Record<string, any>;
```
B4nan, line 226: i would rather not pollute the shared interfacing with mongo specific options

B4nan (changes requested):
i would very much like to see this done differently, isolated inside the `@mikro-orm/mongodb` package instead of leaking on the core level (`IDatabaseDriver.ts`)

we can reexport necessary interfacing from the driver package to alter them if there isnt a better way

(also some tests are needed)

packages/mongodb/src/MongoConnection.ts
peter-lapin-powerus: Thanks a lot for reviewing the request and for the quick feedback 🙌

You're absolutely right — changing the method signature of deleteMany is a breaking change at the API level.

When exploring the codebase, I noticed that deleteMany on the connection level is currently only used internally by MongoDriver, so I went with updating the method signature to support passing CommandOperationOptions. But you're correct — it’s better to avoid such changes unless strictly necessary.

packages/core/src/drivers/IDatabaseDriver.ts
peter-lapin-powerus: Thanks again for the thoughtful review and valuable feedback 

You're absolutely right — avoiding Mongo-specific options in shared interfaces makes perfect sense, and I appreciate you pointing that out. I’ll revisit the problem and explore an alternative that respects the current abstractions.

If you have a preferred way of handling this kind of driver-specific option (like passing comment), I’d be grateful for your guidance.

Alternatively, would it make more sense to raise a feature request first so we can discuss the use case and potential solutions there?

Thanks again.

packages/mongodb/src/MongoConnection.ts
B4nan: i dont mind changing the parameter now, but it needs to support the old signature too, which could be tricky in this case (it already was an object). better just add a new parameter and clean this up in v7

packages/core/src/drivers/IDatabaseDriver.ts
B4nan: i think we already have an issue for this, its basically the same as #1603, right? 

IMO the best way would be to extend the interface from the core package with the new options, on the mongo package level, and reexport it with the same name. we do the same with the driver specific `MikroORM` and `EntityManager` classes

B4nan:
Closing as stale

PR #6798, opened 2025-08-12
docs: clarify configuration file export options

radicaled:
I discovered by exploratory programming that the configuration file accepts a function that exports a promise, rather than just synchronous methods.

This behavior doesn't seem to be documented anywhere but is important for people dynamically pulling configuration values from remote sources, so this PR adds that information.


B4nan (approved):
thanks!

it is noted somewhere (i remember mentioning it), but its also a very small note, so small i couldnt find it right now 🙃

PR #6802, opened 2025-08-18
feat(core): add comprehensive transaction propagation support

### LINKED ISSUE #6788: Feature Request: Support `propagation` in `MikroORM`'s `@Transactional()` Decorator
There is not enough control using `@Transactional()` decorator for transaction propagation.

`nestjs-cls` has a plugin for `@Transactional()` decorator and they have a spring-style concept of transaction propagation like in java's `@Transactional()` annotation. 
The propagation option is controlled by the Propagation enum, which has the following values:
- `Required`: (default) Reuse the existing transaction or create a new one if none exists.
- `RequiresNew`: Create a new transaction even if one already exists. The new transaction is committed independently of the existing one.
- `NotSupported`: Run without a transaction even if one exists. The existing transaction is resumed once the callback completes.
- `Mandatory`: Reuse an existing transaction, throw an exception otherwise.
- `Never`: Run without a transaction, throw an exception if one already exists.
- `Supports`: Reuse the existing transaction or continue without a transaction if none exists.
- `Nested`: Create a subtransaction if supported; otherwise, reuse the existing transaction or create a new one if none exists. 

It would be really nice to add the propagation options to `@Transactional({ propagation: 'Required' })` decorator

`nestjs-cls` does not yet have an adapter for `MicroORM`, they only have it for `TypeORM` and other orm's and query builders ([full list](https://papooch.github.io/nestjs-cls/plugins/available-plugins/transactional)). Maybe another solution would be to add the adapter for `MicroORM` to `nestjs-cls` library.

junhyeongkim2:
## Summary

Implements comprehensive transaction propagation support for MikroORM, providing fine-grained control over transaction boundaries

  ## What's Changed

  - Added `TransactionManager` class to handle transaction lifecycle and propagation
  - Implemented all 7 standard propagation types:
    - `REQUIRED`: Join existing transaction or create new one
    - `REQUIRES_NEW`: Always create independent transaction
    - `NESTED`: Create savepoint if transaction exists, otherwise new transaction
    - `SUPPORTS`: Use transaction if available, otherwise execute non-transactionally
    - `MANDATORY`: Require existing transaction, throw error if none exists
    - `NEVER`: Must execute without transaction, throw error if one exists
    - `NOT_SUPPORTED`: Suspend existing transaction and execute without transaction
  - Added comprehensive documentation with examples for each propagation type
  - Full test coverage for all propagation scenarios
  - TypeScript-safe implementation with proper error handling

  ## Breaking Changes

  None - fully backward compatible. The default behavior remains unchanged.

  Fixes #6788


packages/core/src/transaction/TransactionManager.ts
```diff
@@ -0,0 +1,330 @@
...
+      return this.executeTransactionFlow(fork, cb, propagateToUpperContext, parentEm);
+    }, { ...options, eventBroadcaster });
+  }
+
+  /**
+   * Suspends the current transaction and returns the suspended resources.
+   */
+  private suspendTransaction(em: EntityManager): unknown {
+    const suspended = em.getTransactionContext();
+    em.setTransactionContext(null!);
```
B4nan:
```suggestion
    em.resetTransactionContext();
```

packages/core/src/EntityManager.ts
```diff
@@ -2378,6 +2338,14 @@ export class EntityManager<Driver extends IDatabaseDriver = IDatabaseDriver> {
     return this.getContext(false)._id;
   }
 
+  /**
+   * Checks whether transactions are disabled for this EntityManager instance.
+   * @internal
+   */
+  get isTransactionsDisabled(): boolean {
```
B4nan: the naming here is pretty bad, ideally we wouldnt have to have this method at all, is there any reason why you just dont pass the `disableTransactions` option directly to the `TransactionManager`?

or even better, why should we create the tx manager if transactions are disabled? feels like we can just skip it completely

packages/core/src/index.ts
```diff
@@ -32,3 +32,4 @@ export * from './naming-strategy';
 export * from './metadata';
 export * from './cache';
 export * from './decorators';
+export * from './transaction';
```
B4nan: i dont think we need a folder with a single file, `TransactionManager` can go to `utils`

packages/core/src/transaction/TransactionManager.ts
```diff
@@ -0,0 +1,330 @@
...
+        }
+        return this.createNewTransaction(em, cb, options);
+
+      default:
+        throw new Error(`Unsupported transaction propagation type: ${propagation}`);
+    }
+  }
+
+  /**
+   * Executes a transaction with default behavior (backward compatibility).
```
B4nan: wdym by backwards compatibility here? also what is "default behaviour"? feels like this comment is only adding more confusion

B4nan:
@citron0137 can you take a look at this, since you were proposing the feature?

packages/core/src/transaction/TransactionManager.ts
junhyeongkim2: Fixed, using resetTransactionContext() now.

packages/core/src/EntityManager.ts
junhyeongkim2: You're right - checking disableTransactions before creating TransactionManager is more efficient.
  Removed the getter and now skip manager creation when transactions are disabled.

packages/core/src/index.ts
junhyeongkim2: Moved to utils/ as suggested.

packages/core/src/transaction/TransactionManager.ts
junhyeongkim2: You're right, the comment was misleading. It just handles standard nested transactions when no propagation is specified. Updated to be clearer.

junhyeongkim2:
@B4nan Thanks for the review! All suggested changes have been implemented. Please let me know if there's anything else that needs adjustment!

citron0137:
I love this update!!

Because it effectively solved the constraints that existed in the existing propagation. 

And I think this will be useful for putting propagation options in annotation.

B4nan:
> 📝 Step 2: MongoDB Support
> 📝 Step 3: SQLite Support

What exactly is needed for those? The changes are already implemented on the core level, so they should just work with all drivers, no? Would be good to test it with those drivers too.

> 📝 Step 4: Platform Adapter Integration

I don't understand what this should be about.

> 📝 Step 5: Documentation & Examples

This is a requirement for merging any PR. Ideally, I'd like to see all the steps implemented here. If you want to split into multiple PRs, you can do so, but I won't be merging incomplete PRs to master - ideally this would end up as a single commit in there.

junhyeongkim2:
@B4nan Thank you for your review!

  1. MongoDB/SQLite Support

 You're right. No additional implementation is needed.

  When I ran the complete test suite across all six databases (MariaDB, MongoDB, MSSQL, MySQL, PostgreSQL, SQLite),       I encountered test failures in MongoDB and SQLite:
  - MongoDB: Failed on NESTED propagation tests with "Transaction already in progress" error
  - SQLite: Failed on REQUIRES_NEW propagation tests due to concurrent transaction limitations

 I initially thought these failures might need fallback handling, but I now understand these are expected behaviors due to the
  physical limitations of each database:
  - MongoDB doesn't support nested transactions
  - SQLite only allows single write connections
  - The core implementation is already working correctly

  If you'd like, I can add comprehensive test suites for all databases following the same structure as the [TransactionPropagation.postgre.test.ts](https://github.com/mikro-orm/mikro-orm/pull/6802/files#diff-d58d1cf46bc79c907d1b24239e57c8dbb3a173427e617b3d44dcdc1b2987bdf9),
with specific test cases to verify that MongoDB and SQLite properly throw errors for their unsupported scenarios.

  2. Platform Adapter Integration

  I understand now that this is unnecessary. I had initially conceived this to automatically fallback to supported propagation types, but I realize maintaining database-specific constraints is the correct approach.

  3. Documentation & Examples

 Thank you for the clarification. I now understand my proposed support implementations are unnecessary.
 Please let me know what's needed for this PR to be complete. Happy to add docs if required, or if the current state is sufficient, that works too.

B4nan:
Let's test the other drivers and add some docs about this - I'd go with this page: https://mikro-orm.io/docs/transactions but feel free to mention it elsewhere too.

B4nan:
Let's put those tests to `tests/features/transactions`, once we merge this, I'll move some other tx related tests there.

junhyeongkim2:
@B4nan  Added tests and documentation as requested

  - **Tests**: All database drivers covered. Note: SQLite excluded (60s timeout due to single-writer limitation), MongoDB validates "Transaction already in progress" error for NESTED.

  - **Docs**: Minimal propagation section with table and example added to avoid duplication.

  Please take a look and let me know if any adjustments needed. Thanks!

tests/TransactionPropagation.mariadb.test.ts
```diff
@@ -0,0 +1,1270 @@
...
+  @Property({ onCreate: () => new Date(), nullable: true })
+  createdAt?: Date;
+
+}
+
+describe('Transaction Propagation - MariaDB', () => {
+  let orm: MikroORM<MariaDbDriver>;
+
+  beforeAll(async () => {
+    orm = await initORMMySql<MariaDbDriver>('mariadb', {
```
B4nan: please dont use those init methods from the bootstrap file, they have a lot of advanced options enabled, better stay to the defaults and just use `MikroORM.init` directly, with an explicit db name to ensure there is no collision (tests run in parallel, each should have its own db)

also if the tests are exactly the same, it would be a better approach to have one test using `describe.each`, similarly to e.g. this test:

https://github.com/mikro-orm/mikro-orm/blob/master/tests/features/upsert/upsert.test.ts#L154-L184

you can make the tests conditional if the behaviour is different

junhyeongkim2:
> Let's put those tests to `tests/features/transactions`, once we merge this, I'll move some other tx related tests there.

Sure, I'll move them to tests/features/transactions/

junhyeongkim2:
1. Migrated transaction propagation tests to /tests/features/transactions/ with direct MikroORM.init() calls
  2. Unified PostgreSQL/MySQL/MariaDB/MSSQL tests using describe.each pattern (47 tests × 4 DBs)
  3. Separated SQLite and MongoDB due to unique transaction limitations
  4. Skipped 7 MSSQL tests due to database engine limitations

  @B4nan Would appreciate your review when convenient. Thanks!

docs/docs/transactions.md
```diff
@@ -83,6 +83,32 @@ Explicit transaction demarcation is required when you want to include custom DBA
...
+| `REQUIRED` (default) | Join existing transaction or create new one |
+| `REQUIRES_NEW` | Always create independent transaction |
+| `NESTED` | Create savepoint within existing transaction |
+| `NOT_SUPPORTED` | Execute without transaction |
+
+```ts
+// REQUIRES_NEW: creates independent transaction
+await em.transactional(async (em1) => {
+  const user = em1.create(User, { name: 'John' });
+  await em1.persist(user);
```
B4nan: this is not necessary when using `em.create`, also its a sync method

```suggestion
```

docs/docs/transactions.md
```diff
@@ -83,6 +83,32 @@ Explicit transaction demarcation is required when you want to include custom DBA
...
+  
+  await em1.transactional(async (em2) => {
+    const article = em2.create(Article, { title: 'Test' });
+    await em2.persist(article);
```
B4nan: dtto
```suggestion
```

docs/docs/transactions.md
```diff
@@ -83,6 +83,32 @@ Explicit transaction demarcation is required when you want to include custom DBA
 
 `em.transactional(cb)` and the `@Transactional()` will flush the inner `EntityManager` before transaction commit.
 
+#### Transaction Propagation
+
+Control how nested transactions behave using the `propagation` option:
```
B4nan: this is pretty sparse, could you try to expand the docs a bit, also briefly explaining what this is good for.

docs/docs/transactions.md
```diff
@@ -83,6 +83,32 @@ Explicit transaction demarcation is required when you want to include custom DBA
...
+
+| Propagation | Description |
+| ----------- | ----------- |
+| `REQUIRED` (default) | Join existing transaction or create new one |
```
B4nan: well, but this is not exactly the default behaviour now, nested `em.transactional` call always creates a nested transaction. if there is no tx context, it creates one, if there is, it creates a savepoint.

i guess this behaviour is still maintained, given the existing tests are passing, just the docs are confusing. please try to expand on those descriptions a bit more

docs/docs/transactions.md
```diff
@@ -83,6 +83,32 @@ Explicit transaction demarcation is required when you want to include custom DBA
...
+| `REQUIRES_NEW` | Always create independent transaction |
+| `NESTED` | Create savepoint within existing transaction |
+| `NOT_SUPPORTED` | Execute without transaction |
+
+```ts
+// REQUIRES_NEW: creates independent transaction
```
B4nan: the example should be described in text above it, its not very clear what it tries to show. i'd rather see a sentence or two about what it does instead of those tiny comments

B4nan:
I am looking at the other propagation options described in #6788, and it feels like they should be pretty trivial to implement. If that's the case, can you give it a try?

docs/docs/transactions.md
B4nan: the sprint docs could be used as an inspiration, we dont need to be that verbose, but what we have now is way too little

https://www.baeldung.com/spring-transactional-propagation-isolation

junhyeongkim2:
@B4nan Thanks for the opportunity! I'll implement the remaining propagation options and update the docs with clearer descriptions as suggested.

junhyeongkim2:
Hi @B4nan, thanks for the review! Here's what I've addressed:

  1. Documentation expanded ✅

  - Added comprehensive explanations following Spring docs style and existing MikroORM patterns
  - Each type now includes description, use case, and examples

  2. Additional propagations implemented ✅

  - SUPPORTS, MANDATORY, NEVER added as suggested

  3. SUPPORTS, MANDATORY, NEVER Tests added ✅

  - 36 test cases covering all scenarios
  - Works across all databases

  Let me know if you'd like any adjustments!

packages/core/src/utils/TransactionManager.ts
```diff
@@ -0,0 +1,343 @@
+import type { EntityManager } from '../EntityManager';
+import { type TransactionOptions, TransactionPropagation } from '../enums';
+import { type FlushEventArgs, TransactionEventBroadcaster } from '../events';
+import { TransactionContext } from '../utils/TransactionContext';
+import { ChangeSetType } from '../unit-of-work';
+
+/**
+ * Manages transaction lifecycle and propagation for EntityManager.
+ * Simplified implementation supporting 4 core propagation types.
```
B4nan: this is no longer true, also update the PR description to reflect latest changes

docs/docs/transactions.md
```diff
@@ -83,6 +83,351 @@ Explicit transaction demarcation is required when you want to include custom DBA
...
+    author.name = 'Eric Evans';
+    em2.persist(author);
+    throw new Error('Nested transaction failed');
+  });
+  
+  // Book is still saved despite inner transaction failure
+});
+
+// Using @Transactional() decorator
+class BookService {
```
B4nan, line 124: lets add `constructor(private readonly em: EntityManager) { }` to all those examples with `@Transactional` decorator

docs/docs/transactions.md
```diff
@@ -83,6 +83,351 @@ Explicit transaction demarcation is required when you want to include custom DBA
...
+
+##### MANDATORY Propagation
+
+Enforces that a method must be called within an existing transaction:
+
+```ts
+// Using em.transactional()
+async function updateBookStock(bookId: number, quantity: number) {
+  await em.transactional(async (em) => {
+    const book = await em.findOne(Book, bookId);
```
B4nan: use `findOneOrFail` in those examples, otherwise they wouldnt type check

junhyeongkim2:
Hi @B4nan,

  I've addressed all the requested changes:

  - ✅ Added `constructor(private readonly em: EntityManager) { }` to all `@Transactional()` decorator examples (8 classes)
  - ✅ Updated MANDATORY examples to use `findOneOrFail` as requested

  Additionally, I fixed similar TypeScript issues in other examples:
  - SUPPORTS: `findBookWithAuthor` now uses `findOneOrFail`
  - NESTED: `publishBook` now uses `findOneOrFail`
  - NOT_SUPPORTED: `updateAndReport` now uses `findOneOrFail`

  I've also updated the PR title and description to reflect the current implementation with all 7 propagation types.

  All documentation examples should now compile without TypeScript errors. Let me know if you need any additional changes!

packages/core/src/utils/TransactionManager.ts
```diff
@@ -0,0 +1,342 @@
...
+  async handle<T>(
+    cb: (em: EntityManager) => T | Promise<T>,
+    options: TransactionOptions = {},
+  ): Promise<T> {
+    const em = this.em.getContext(false) as EntityManager;
+
+    // If no explicit propagation is set, use default behavior
+    if (!options.propagation) {
+      return this.executeDefaultTransaction(em, cb, options);
+    }
```
B4nan: isn't the default behaviour exactly the same as `NESTED`? how do they differ if not?

packages/core/src/utils/TransactionManager.ts
```diff
@@ -0,0 +1,342 @@
...
+
+  /**
+   * Joins existing transaction context without creating savepoint.
+   */
+  private async joinExistingTransaction<T>(
+    em: EntityManager,
+    cb: (em: EntityManager) => T | Promise<T>,
+    options: TransactionOptions,
+  ): Promise<T> {
+    const fork = this.createFork(em, options);
```
B4nan: if we are joining an existing transaction, should we really create the fork and new `TransactionContext`?

packages/core/src/utils/TransactionManager.ts
junhyeongkim2: You're right - the default behavior and NESTED are functionally identical. 

  We could eliminate this duplication by setting NESTED as the explicit default:
  options.propagation ??= TransactionPropagation.NESTED;

  Then remove executeDefaultTransaction entirely. What do you think?

packages/core/src/utils/TransactionManager.ts
junhyeongkim2: You're right Simply returning cb(em) correctly represents the joining semantics without unnecessary overhead. Thanks for catching this!

packages/core/src/utils/TransactionManager.ts
B4nan: Yes please, also update the docs about this. It feels really weird to have some default mode that is not one of the existing options, and clearly this is a code duplication too.

B4nan: I wonder if we should have different defaults for the `@Transactional` decorator, it kinda makes sense to use `REQUIRED` mode for it. For `em.transactional`, the `NESTED` mode feels correct.

Changing the semantics now feels a bit breaking, but we could do that in v7.

junhyeongkim2: I agree, that would be a better approach!

junhyeongkim2:
@B4nan Thanks for the review! I've made the following changes:

  1. **Removed `executeDefaultTransaction` and `processTransactionCallback`** - Set NESTED as explicit default instead
  2. **Simplified `joinExistingTransaction`** - Now just returns `cb(em)` without creating unnecessary fork/TransactionContext
  3. **Updated documentation** - Clarified that NESTED is the default propagation type

  All tests are passing. Let me know if you'd like any adjustments!

docs/docs/transactions.md
```diff
@@ -83,6 +83,375 @@ Explicit transaction demarcation is required when you want to include custom DBA
...
+
+// Throws error - no transaction exists
+await inventoryService.updateStock(1, 10);
+```
+
+Use for critical operations that should never run outside a transaction context.
+
+##### NESTED Propagation
+
+Creates a savepoint when a transaction exists, otherwise creates a new transaction (default behavior for nested calls):
```
B4nan:
```suggestion
Creates a savepoint when a transaction exists, otherwise creates a new transaction. This is the default behavior.
```

docs/docs/transactions.md
```diff
@@ -83,6 +83,375 @@ Explicit transaction demarcation is required when you want to include custom DBA
...
+  await inventoryService.updateStock(1, 10);
+});
+
+// Throws error - no transaction exists
+await inventoryService.updateStock(1, 10);
+```
+
+Use for critical operations that should never run outside a transaction context.
+
+##### NESTED Propagation
```
B4nan: lets move this to the first position and remove the current "Default Behavior" section, they are about the same thing (up to you which example you want to keep)

packages/core/src/utils/TransactionManager.ts
```diff
@@ -0,0 +1,292 @@
...
+  constructor(private readonly em: EntityManager) {}
+
+  /**
+   * Main entry point for handling transactional operations with propagation support.
+   */
+  async handle<T>(
+    cb: (em: EntityManager) => T | Promise<T>,
+    options: TransactionOptions = {},
+  ): Promise<T> {
+    const em = this.em.getContext(false) as EntityManager;
```
B4nan: why are you casting this? feels unnecessary

junhyeongkim2: Agreed, removing the cast.

junhyeongkim2:
@B4nan  Thanks for pointing out the docs ordering - fixed! Also removed the unnecessary type cast. Appreciate the detailed review.

packages/core/src/utils/TransactionManager.ts
```diff
@@ -0,0 +1,292 @@
...
+  /**
+   * Joins existing transaction context without creating savepoint.
+   */
+  private async joinExistingTransaction<T>(
+    em: EntityManager,
+    cb: (em: EntityManager) => T | Promise<T>,
+    options: TransactionOptions,
+  ): Promise<T> {
+    return cb(em);
+  }
```
B4nan: let's inline this

packages/core/src/utils/TransactionManager.ts
```diff
@@ -0,0 +1,292 @@
...
+  private createDeletionHandler(parent: EntityManager): (args: FlushEventArgs) => void {
+    return (args: FlushEventArgs) => {
+      const deletionChangeSets = args.uow.getChangeSets()
+        .filter(cs => cs.type === ChangeSetType.DELETE || cs.type === ChangeSetType.DELETE_EARLY);
+
+      deletionChangeSets.forEach(cs =>
+        parent.getUnitOfWork(false).unsetIdentity(cs.entity),
+      );
+    };
+  }
```
B4nan: i'd inline this one too, also prefer `for of` instead of `forEach`

junhyeongkim2:
@B4nan Thanks for the review! All suggestions applied.

packages/core/src/utils/TransactionManager.ts
```diff
@@ -0,0 +1,275 @@
...
+
+      case TransactionPropagation.SUPPORTS:
+        if (hasExistingTransaction) {
+          return cb(em);
+        }
+        return this.executeWithoutTransaction(em, cb, options);
+
+      case TransactionPropagation.MANDATORY:
+        if (!hasExistingTransaction) {
+          throw new Error(`No existing transaction found for transaction marked with propagation "${propagation}"`);
```
B4nan: one final nit, lets create a custom error class for those thrown by the tx manager

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/errors.ts

junhyeongkim2: @B4nan Custom error class added. Happy to adjust the naming if needed.

packages/core/src/errors.ts
```diff
@@ -301,3 +301,19 @@ export class NotFoundError<T extends AnyEntity = AnyEntity> extends ValidationEr
   }
 
 }
+
+export class TransactionStateError<T extends AnyEntity = AnyEntity> extends ValidationError<T> {
```
B4nan: I'd remove the type argument, such errors are not really bound to entity type

```suggestion
export class TransactionStateError extends ValidationError {
```

junhyeongkim2: @B4nan Thanks! Updated as suggested.

B4nan (approved):
all right, we are good to go, thanks a lot for the PR!

junhyeongkim2:
@B4nan Thank you for your patience and guidance throughout this PR. It's been an honor to contribute to MikroORM!

injae-kim:
Aweeeeeesome work @junhyeongkim2 👍
Thanks for the kind review to reviewer.

PR #6807, opened 2025-08-21
fix(core): ensure correct runtime value of `decimal` and `float` properties

### LINKED ISSUE #6806: Decimals are returned as numbers instead of strings
### Describe the bug

Columns with `DECIMAL` type are returned as `number` when they should be returned as `string`.

I've originally thought that this might be entity generator error, as it returns `type: 'decimal'` instead of `type: DecimalType`, but after manually changing it to use `DecimalType` (even explicitly specifying to use the `'string'` mode) it still returns `number`s instead of `string`s.

Note that the reproduction is based on SQLite, but I've first encountered the issue while using mssql driver.

### Reproduction

Repro: https://github.com/patrykbojczuk/decimal-as-number

@mikro-orm/mssql

### MikroORM version

6.4.17-dev.88

### Node.js version

v20.15.0, typescript@5.9.2

### Operating system

macOS

### Validations

patrykbojczuk:
This simple PR makes sure the value returned from the `convertToJSValue` is always `string` if not explicitly set to return `number`.

This is a workaround for #6806, because some drivers, e.g. [tediousjs/tedious](https://github.com/tediousjs/tedious) serialize `DECIMAL`s and `FLOAT`s to `number` instead of `string`.

Closes #6806

### DIFF
--- a/tests/issues/GH6806.test.ts
+++ b/tests/issues/GH6806.test.ts
@@ -27,7 +27,6 @@ beforeAll(async () => {
   orm = await MikroORM.init({
     dbName: ':memory:',
     entities: [Amount],
-    debug: ['query', 'query-params'],
   });
   await orm.schema.refreshDatabase();
 });


tests/issues/GH6806.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  Property,
+} from '@mikro-orm/sqlite';
+
+@Entity()
+class Amount {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property({ type: new DecimalType('string'), precision: 5, scale: 2 })
```
B4nan: lets add a test for the `DoubleType` too

(i'd just add another property to this test, no need for a separate one)

B4nan:
> (i'd just add another property to this test, no need for a separate one)

i was too late with my edit 🙃 lets just add another property, no need for two tests

PR #6822, opened 2025-08-30
fix(core): onlyOwnProperties allow owned many_to_many

acrosett:
allow update of many_to_many IDs on owning side when onlyOwnProperties: true. Closes discussion #6813

### DIFF
--- a/docs/docs/entity-helper.md
+++ b/docs/docs/entity-helper.md
@@ -168,7 +168,7 @@ When you have some properties in the payload that are not represented by an enti
 
 #### `onlyOwnProperties`
 
-With `onlyOwnProperties` enabled, to-many relations are skipped, and payloads of to-one relations are converted to foreign keys. Defaults to `false`.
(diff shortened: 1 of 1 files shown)


packages/core/src/entity/EntityAssigner.ts
```diff
@@ -325,7 +333,7 @@ export interface AssignOptions<Convert extends boolean> {
   onlyProperties?: boolean;
 
   /**
-   * With `onlyOwnProperties` enabled, to-many relations are skipped, and payloads of to-one relations are converted
+   * With `onlyOwnProperties` enabled, (not owned) to-many relations are skipped, and payloads of other relations are converted
```
B4nan:
```suggestion
   * With `onlyOwnProperties` enabled, inverse sides of to-many relations are skipped, and payloads of other relations are converted
```

also let's also update the docs here:

https://mikro-orm.io/docs/entity-helper#assign-options

B4nan (changes requested):

B4nan (approved):

B4nan:
thanks!

PR #6826, opened 2025-09-03
fix(query-builder): skip branching on to-one joins

### LINKED ISSUE #6824: Condition on inverse one-to-one join applying to incorrect alias
### Describe the bug

Hi @B4nan, I think this might be related to #6819 but during my testing today with v6.5.2, there seems to be a specific situation where the alias still isn't being mapped correctly.

The issue arises when adding a condition on the inverse side of a OneToOne relationship, and the field is one of the composite primary key fields.

The following orm query:
```typescript
  const user = await orm.em.findOneOrFail(User, {
    request: {
      id: 13,
    },
  });
```
Generates sql:
```sql
select `u0`.*,
    `r1`.`org_id` as `r1__org_id`,
    `r1`.`id` as `r1__id`
from `user` as `u0`
    left join `request` as `r1` on `u0`.`org_id` = `r1`.`org_id`
    and `u0`.`id` = `r1`.`user_id`
where `u0`.`request_id` = 13
limit 1 -- SQLITE_ERROR: no such column: u0.request_id
```
The `u0.request_id` doesn't exist, rather it should be filtering on `r1.id`

Weirdly, if I nest such relationships and add a condition on a field 2+ deep, it works.

I've updated my repro with a test for this case.

Thanks again, for you help on #6819. 

### Reproduction

https://github.com/jaichinger/mikro-orm-reproduction-3

@mikro-orm/sqlite

### MikroORM version

6.5.2

### Node.js version

22

### Operating system

macOS

### Validations

jaichinger:
- Set preferNoBranch to true if relationship is :1. This allows the child alias to be correctly matched.

Closes #6824

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,6 +53,7 @@ services:
   mssql:
     image: mcr.microsoft.com/mssql/server:2025-CTP2.0-ubuntu-22.04
     restart: always
+    platform: linux/amd64
     ports:
       - '1433:1433'
     environment:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,7 +53,6 @@ services:
   mssql:
     image: mcr.microsoft.com/mssql/server:2025-CTP2.0-ubuntu-22.04
     restart: always
-    platform: linux/amd64
     ports:
       - '1433:1433'
     environment:
--- a/tests/issues/GH6824.test.ts
+++ b/tests/issues/GH6824.test.ts
@@ -1,0 +1,228 @@
+import {
+  Collection,
+  Entity,
+  ManyToOne,
+  MikroORM,
+  OneToMany,
+  OneToOne,
+  PrimaryKey,
+  Property,
+  Ref,
+} from '@mikro-orm/sqlite';
+
+@Entity()
+class Organisation {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property()
+  name!: string;
+
+}
+
+@Entity()
+class User {
+
+  @ManyToOne({
+    entity: () => Organisation,
+    fieldName: 'org_id',
+    primary: true,
+    ref: true,
+  })
+  org!: Ref<Organisation>;
+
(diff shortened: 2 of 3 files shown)


docker-compose.yml
```diff
@@ -53,6 +53,7 @@ services:
   mssql:
     image: mcr.microsoft.com/mssql/server:2025-CTP2.0-ubuntu-22.04
     restart: always
+    platform: linux/amd64
```
B4nan: is this really necessary?

tests/issues/GH6824.test.ts
```diff
@@ -0,0 +1,228 @@
...
+}
+
+let orm: MikroORM;
+
+beforeAll(async () => {
+  orm = await MikroORM.init({
+    dbName: ':memory:',
+    entities: [Organisation, User, Profile, UserRequest],
+    debug: ['query', 'query-params'],
+    allowGlobalContext: true, // only for testing
```
B4nan:
```suggestion
```

B4nan (changes requested):

B4nan:
> so this prevents errors on arm based machines (e.g. apple silicon)

What errors specifically? I am on M4 and I don't think I get any errors.

docker-compose.yml
jaichinger: Ah yeah, my bad. It creates a warning, not an error. Will remove this.

B4nan: Maybe it should be here, I am just wondering if it won't make things worse for e.g. windows contributors somehow.

B4nan (approved):
thanks!

docker-compose.yml
jaichinger: Ah right, yeah not sure. It might be okay as docker desktop on windows still uses linux under the hood I think, and the image doesn't come in anything other than `linux/amd64`.

Anyway, I've reverted my change just to be safe.

PR #6859, opened 2025-09-18
test(core): test on entities with circular reference, and auto-incremented ids

leonardbinet:
https://github.com/mikro-orm/mikro-orm/discussions/6266#discussioncomment-14442615


B4nan:
I am not sure if this can be done without making the column nullable (or having predefined PK values), how would you do this in raw SQL? You can't create the `Unit` row without the value, the idea behind deferred constraint is you have a value for it, its just not yet in the database - it can't be omitted and updated later (even in the same tx).

We could improve [this condition](https://github.com/mikro-orm/mikro-orm/blob/83ca9c14cc4fd443b45ea40f1945c5a249336747/packages/core/src/unit-of-work/UnitOfWork.ts#L1008) so your test works with both deferred constraint and a nullable column, but the column would still have to be nullable for this to work. If you don't make it deferred, the test is passing already (that's the only part I can fix).

leonardbinet:
Ohhh indeed, I didn't think it through, I thought it would work with CTEs or deferred constraint, but indeed the nullability is necessary.

Thanks for pointing our that removing the deferred constraint already works, it fixed the issue on my side.

I've updated the PR to simply check this usecase without any deferred constraint, do you want me to close this PR? or is this test worth it (I've no idea if this "circular reference" with auto-increment ids use-case is already tested, or worth it)

PR #6860, opened 2025-09-18
fix: correctly map raw result fields to entity properties with @Property({ name: 'x_name' })

### LINKED ISSUE #6861: Mapped entity property is overwritten by raw field with the same name
### Describe the bug

When mapping raw database results to entity fields in MikroORM, a property decorated to map to a differently named column (e.g., @Property({ fieldName: 'a_tmp' }) a: string;) will be populated from a_tmp.
However, if the raw result from the database also includes a column named a, currently the automatic mapping logic in MikroORM copies all remaining unmapped raw fields into the entity object, potentially overwriting the already mapped value of a with the raw value from a.

### Reproduction

Define an entity property with a custom column mapping:
```typescript
@Property({ fieldName: 'a_tmp' })
a: string;
```
Use database result that includes both a and a_tmp.
After MikroORM maps the result, the value of a in the entity may come from the raw a column instead of the intended a_tmp.

https://github.com/isaryy/mapped-field-overwrite-reproduction/tree/master

@mikro-orm/postgresql

### MikroORM version

6.5.4

### Node.js version

22

### Operating system

macOs

### Validations

isaryy:
reproduction repo – https://github.com/isaryy/mapped-field-overwrite-reproduction

closes #6861

### DIFF
--- a/tests/issues/GH6861.test.ts
+++ b/tests/issues/GH6861.test.ts
@@ -25,8 +25,6 @@ beforeAll(async () => {
   orm = await MikroORM.init({
     dbName: ':memory:',
     entities: [User],
-    debug: ['query', 'query-params'],
-    allowGlobalContext: true, // only for testing
   });
   await orm.schema.refreshDatabase();
   await orm.em.execute('alter table "user" add column name text');
--- a/tests/issues/GH6861.test.ts
+++ b/tests/issues/GH6861.test.ts
@@ -1,0 +1,46 @@
+import { Entity, MikroORM, PrimaryKey, Property } from '@mikro-orm/sqlite';
+
+@Entity()
+class User {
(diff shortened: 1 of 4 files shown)


tests/issues/GH6861.test.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+let orm: MikroORM;
+
+beforeAll(async () => {
+  orm = await MikroORM.init({
+    dbName: ':memory:',
+    entities: [User],
+    debug: ['query', 'query-params'],
+    allowGlobalContext: true, // only for testing
```
B4nan:
```suggestion
```

B4nan:
Please update the failing snapshots.

Issue #6863, opened 2025-09-20
Partial load + nullable property hydration results in `undefined` instead of `null`


chloeelim:
### Describe the bug

I understand that there will only ever be a single entity instance per ID in the identity map; hence the suggestion to either fork a separate context (or use disableIdentityMap / refresh: true) as referenced here: [#3507](https://github.com/mikro-orm/mikro-orm/issues/3507).

That said, according to the v5 docs, this should already be handled automatically without needing a separate context: https://mikro-orm.io/blog/mikro-orm-5-released#automatic-refreshing-of-loaded-entities

I’m running into a hydration issue whereby if an entity has a nullable property that wasn’t included in the first partial load (q1), it ends up being hydrated as undefined instead of null in a subsequent full load (q2) (see example below).

This issue disappears if I either use a separate context or refresh the entity, but that feels like it shouldn’t be necessary in v5.

---

### Reproduction

🐛  **Reproduction repo:** https://github.com/chloeelim/mikro-partial-load-nullable-reproduction

Consider an Author entity with this schema:
```ts
@Entity()
export class Author {
  @PrimaryKey({ type: "number" })
  id: number;

  @Property({ type: "string" })
  name: string;

  @Property({ type: "string", nullable: true })
  email: string | null = null;
}
```

1. Suppose the DB contains this row: `{ id: 1, name: "John", email: null }`
2. Now suppose we run these 2 queries one after the other
    ```ts
    // q1: partial load (select only 'id')
    const q1 = await orm.em.find(Author, { id: 1 }, { fields: ['id'] });
    
    // q2: full load (select all fields)
    const q2 = await orm.em.find(Author, { id: 1 });
    ```

**Expected identity map behaviour:**
- After q1, the entity should look like: `{ id: 1 }` in the entity map
- After q2, the author entity should be properly merged and hydrated: `{ id: 1, name: "John", email: null }`

**Actual behaviour:**
- q1 behaves as expected, the author entity in the identity map looks like `{ id: 1 }` ✅ 
- However, after q2, the nullable property is not properly hydrated, and the author entity looks like `{ id: 1, name: "John", email: undefined }` instead
 
- This triggers schema validation issues (as email is `undefined`, when the schema expected `null | string`)

---

🤔 **Hunch:**

This seems to be a result of MikroORM’s identity map and merge logic treating undefined and null as equivalent when computing diffs during mergeData/diffEntities. Concretely:

[...]

PR #6864, opened 2025-09-20
fix(core): correctly diff nullable properties between undefined and null

chloeelim:
Previously, merging entities after a partial load could result in nullable properties being hydrated as `undefined` instead of `null` because the entity comparator treated `undefined` and `null` as equivalent, skipping updates in `mergeData`.

This updates `getGenericComparator` to explicitly handle cases where `last` is `undefined` and `current` is `null`, ensuring nullable fields are properly merged and hydrated in the identity map.

Fixes issues with schema validation when nullable columns are not included in partial loads.

Closes https://github.com/mikro-orm/mikro-orm/issues/6863.

Reproduction repo: https://github.com/chloeelim/mikro-reproduction/tree/partial-load-nullable

### DIFF
--- a/tests/issues/GH6863.test.ts
+++ b/tests/issues/GH6863.test.ts
@@ -26,7 +26,6 @@ beforeAll(async () => {
   orm = await MikroORM.init({
     dbName: ':memory:',
     entities: [User],
-    allowGlobalContext: true, // only for testing
   });
   await orm.schema.refreshDatabase();
 });


tests/issues/GH6863.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+}
+
+let orm: MikroORM;
+
+beforeAll(async () => {
+  orm = await MikroORM.init({
+    dbName: ':memory:',
+    entities: [User],
+    debug: ['query', 'query-params'],
```
B4nan:
```suggestion
```

tests/issues/GH6863.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    allowGlobalContext: true, // only for testing
```
B4nan:
```suggestion
```

B4nan (approved):
Thanks!

PR #6866, opened 2025-09-21
Allow sha256 hashing for fips

ganeshharihara:
For FIPS deployments, MD5 is globally disabled, so I added an option for the user to select sha256 as the hashing algorithm

### DIFF
--- a/packages/core/src/utils/Configuration.ts
+++ b/packages/core/src/utils/Configuration.ts
@@ -105,6 +105,7 @@ export class Configuration<D extends IDatabaseDriver = IDatabaseDriver, EM exten
     ensureDatabase: true,
     ensureIndexes: false,
     batchSize: 300,
+    hashAlgorithm: 'md5',
     debug: false,
     ignoreDeprecations: false,
     verbose: false,
@@ -324,7 +325,7 @@ export class Configuration<D extends IDatabaseDriver = IDatabaseDriver, EM exten
    * Gets instance of metadata CacheAdapter. (cached)
    */
   getMetadataCacheAdapter(): SyncCacheAdapter {
-    return this.getCachedService(this.options.metadataCache.adapter!, this.options.metadataCache.options, this.options.baseDir, this.options.metadataCache.pretty);
+    return this.getCachedService(this.options.metadataCache.adapter!, this.options.metadataCache.options, this.options.baseDir, this.options.metadataCache.pretty, this.options.hashAlgorithm);
   }
 
   /**
@@ -675,6 +676,7 @@ export interface MikroORMOptions<D extends IDatabaseDriver = IDatabaseDriver, EM
   metadataProvider: { new(config: Configuration): MetadataProvider };
   seeder: SeederOptions;
   preferReadReplicas: boolean;
+  hashAlgorithm: 'md5' | 'sha256';
   dynamicImportProvider: (id: string) => Promise<unknown>;
 }
 
--- a/packages/core/src/cache/FileCacheAdapter.ts
+++ b/packages/core/src/cache/FileCacheAdapter.ts
@@ -12,7 +12,8 @@ export class FileCacheAdapter implements SyncCacheAdapter {
 
   constructor(private readonly options: { cacheDir: string; combined?: boolean | string },
               private readonly baseDir: string,
-              private readonly pretty = false) { }
+              private readonly pretty = false,
+              private readonly hashAlgorithm: 'md5' | 'sha256' = 'md5') { }
 
   /**
    * @inheritDoc
@@ -96,7 +97,7 @@ export class FileCacheAdapter implements SyncCacheAdapter {
 
     const contents = readFileSync(origin);
 
-    return Utils.hash(contents.toString() + this.VERSION);
+    return Utils.hash(contents.toString() + this.VERSION, undefined, this.hashAlgorithm);
   }
 
 }
--- a/packages/core/src/utils/Utils.ts
+++ b/packages/core/src/utils/Utils.ts
@@ -998,8 +998,8 @@ export class Utils {
     return Utils.normalizePath(path);
   }
 
-  static hash(data: string, length?: number): string {
-    const hash = createHash('md5').update(data).digest('hex');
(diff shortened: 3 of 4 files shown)


B4nan (changes requested):
thanks!

looks good, lets mention it in the docs too, e.g. here:

https://github.com/mikro-orm/mikro-orm/blob/master/docs/docs/configuration.md

B4nan:
there are other places that use `Utils.hash`, maybe we could respect the new config option there too?

B4nan:
I guess I wouldn't bother with this place, to support is correctly, we'd have to use env vars most likely (set during `MikroORM.init`), and it might be error prone, since this code path is used from the decorators and those can be executed before the `init` method.

https://github.com/mikro-orm/mikro-orm/blob/9af0e3815c9e8200ace3e6a09e5ebd23c6f768f8/packages/core/src/metadata/MetadataStorage.ts#L21

We could just change the algo in v7 to sha256 there, sounds safe. 

But the other places that use the `hash` method all have the config object at hand, so let's respect it there too.

Maybe we could change the default for the other methods too, it might produce some schema diffs for index names that are too long, but people can opt in to md5 if they want to (or just rename their indexes, the ORM will produce the queries correctly, so its just about running a single migration).

ganeshharihara:
> there are other places that use `Utils.hash`, maybe we could respect the new config option there too?

I can add for consistency, but I think it should still work since its optional and defaults to md5

B4nan:
My point is you are introducing a new global config option, but it will have only a limited effect, which is quite unexpected/confusing.

ganeshharihara:
> I guess I wouldn't bother with this place, to support is correctly, we'd have to use env vars most likely (set during `MikroORM.init`), and it might be error prone, since this code path is used from the decorators and those can be executed before the `init` method.
> 
> https://github.com/mikro-orm/mikro-orm/blob/9af0e3815c9e8200ace3e6a09e5ebd23c6f768f8/packages/core/src/metadata/MetadataStorage.ts#L21
> 
> We could just change the algo in v7 to sha256 there, sounds safe.
> 
> But the other places that use the `hash` method all have the config object at hand, so let's respect it there too.
> 
> Maybe we could change the default for the other methods too, it might produce some schema diffs for index names that are too long, but people can opt in to md5 if they want to (or just rename their indexes, the ORM will produce the queries correctly, so its just about running a single migration).

For FIPS deployments, we won't have MD5 at all, so will try to get this to accept the hashAlgorithm setting via a global var in Utils.ts.  LMK if that doesn't work

> My point is you are introducing a new global config option, but it will have only a limited effect, which is quite unexpected/confusing.

Ahh I understand now, sounds good

docs/docs/configuration.md
```diff
@@ -581,6 +581,29 @@ MikroORM.init({
...
+- `'sha256'` - Uses SHA-256 hashing algorithm (FIPS compliant)
+
+This setting affects:
+- Metadata cache file naming and validation
+- File cache adapter hash generation
+- Internal caching mechanisms
+
+
+
 ## Using global Identity Map
```
B4nan:
```suggestion
- Internal caching mechanisms

## Using global Identity Map
```

docs/docs/configuration.md
```diff
@@ -581,6 +581,29 @@ MikroORM.init({
...
+  hashAlgorithm: 'sha256', // Options: 'md5' (default) | 'sha256'
+});
+```
+
+The `hashAlgorithm` option accepts:
+- `'md5'` (default) - Uses MD5 hashing algorithm
+- `'sha256'` - Uses SHA-256 hashing algorithm (FIPS compliant)
+
+This setting affects:
+- Metadata cache file naming and validation
```
B4nan: this has nothing to do with files or validations, it's just a key in memory, we use part of an md5 hash to create a unique key suffix.

i'd skip this line, and change the algo there to sha256 right ahead, i believe it should be safe even outside of major version. i dont really like the approach you picked with the global property - this wont work if entities are imported before the ORM init method is called (since decorators are evaluated on import time, and that code path is mainly used by decorators)

docs/docs/configuration.md
```diff
@@ -581,6 +581,29 @@ MikroORM.init({
 });
 ```
 
+## Setting hashing algorithm (FIPS)
+
+You can configure the hashing algorithm used for caching and metadata operations. This is particularly useful in FIPS-compliant environments where SHA-256 is required instead of the default MD5.
```
B4nan: i wonder how many people know what FIPS mean, your PR is the first time i see it. if you really think we should frame the docs this way, then at least add the full name in parens so people dont have to google this.

docs/docs/configuration.md
```diff
@@ -581,6 +581,29 @@ MikroORM.init({
...
+```
+
+The `hashAlgorithm` option accepts:
+- `'md5'` (default) - Uses MD5 hashing algorithm
+- `'sha256'` - Uses SHA-256 hashing algorithm (FIPS compliant)
+
+This setting affects:
+- Metadata cache file naming and validation
+- File cache adapter hash generation
+- Internal caching mechanisms
```
B4nan: not sure what you are referring to here?

docs/docs/configuration.md
ganeshharihara: removed and just said secured

docs/docs/configuration.md
ganeshharihara: ok that makes sense, I was trying to call out where I made changes but probably too much

docs/docs/configuration.md
ganeshharihara: fixed

packages/core/src/utils/Configuration.ts
```diff
@@ -183,6 +184,8 @@ export class Configuration<D extends IDatabaseDriver = IDatabaseDriver, EM exten
     this.options.baseDir = Utils.absolutePath(this.options.baseDir);
     this.options.preferTs ??= options.tsNode;
 
+    // Set global hash algorithm for static methods that can't access config
+
     if (validate) {
```
B4nan, line 187:
```suggestion
    this.options.preferTs ??= options.tsNode;

    if (validate) {
```

PR #6873, opened 2025-09-23
fix(defineEntity): enhance type of bigint, array, decimal

xcfox:
## Overview
This PR enhance support for `bigint`, `array`, and `decimal` property builders to the `defineEntity` API based on the existing [discussion](https://github.com/mikro-orm/mikro-orm/discussions/6865#discussioncomment-14483413), enhancing type safety and developer experience.

By the way, the automatic inference of composite keys mentioned in the discussion seems much more complex. I'll implement it within a few days.

## Changes

- **BigInt Property**:
  - `p.bigint()` - Returns `bigint` type by default
  - `p.bigint('number')` - Returns `number` type
  - `p.bigint('string')` - Returns `string` type

- **Array Property**:
  - `p.array()` - Default string array
  - `p.array(Number)` or `p.array(i => +i)` - Number array
  - `p.array<T>(toJsValue, toDbValue)` - Custom typed array

- **Decimal Property**:
  - `p.decimal()` - Returns `string` type by default
  - `p.decimal('number')` - Returns `number` type
  - `p.decimal('string')` - Explicitly returns `string` type

## Usage Example

```typescript
const User = defineEntity({
  name: 'User',
  properties: p => ({
    balance: p.bigint(), // bigint type
    safeId: p.bigint('number'), // number type
    tags: p.array(), // string[] type
    scores: p.array(Number), // number[] type
    price: p.decimal(), // string type
    amount: p.decimal('number'), // number type
  }),
});
```

## Backward Compatibility
This change is fully backward compatible and will not affect existing code.

### DIFF
--- a/tests/defineEntity.test.ts
+++ b/tests/defineEntity.test.ts
@@ -693,7 +693,7 @@ describe('PropertyOptionsBuilder', () => {
         settings: { type: types.json, hidden: true },
         bio: { type: types.text, formula: 'concat(first_name, " ", last_name)' },
         status: { enum: true, items: ['active', 'inactive'], nativeEnumName: 'user_status' },
-        tags: { type: new types.array(i => i as string, s => s as string), customOrder: ['tag1', 'tag2', 'tag3'] },
+        tags: { type: new types.array(), customOrder: ['tag1', 'tag2', 'tag3'] },
         extra: { type: types.string, extra: 'VIRTUAL' },
         ignoreChanges: { type: types.string, ignoreSchemaChanges: ['type', 'extra', 'default'] },
       },


tests/defineEntity.test.ts
```diff
@@ -585,7 +693,7 @@ describe('PropertyOptionsBuilder', () => {
         settings: { type: types.json, hidden: true },
         bio: { type: types.text, formula: 'concat(first_name, " ", last_name)' },
         status: { enum: true, items: ['active', 'inactive'], nativeEnumName: 'user_status' },
-        tags: { type: types.array, customOrder: ['tag1', 'tag2', 'tag3'] },
+        tags: { type: new types.array(i => i as string, s => s as string), customOrder: ['tag1', 'tag2', 'tag3'] },
```
B4nan: is this change really needed?

xcfox: Yes, this change is necessary to ensure snapshot matching.  
The modified `p.array()` will return an instance of `ArrayType` instead of the constructor itself, so the values in the snapshot also need to be changed from the constructor to instance values.

Issue #6879, opened 2025-09-24
nativeInsert default value does not account for nullability


adrien2p:
### Describe the bug

When creating a new entity and flushing it, the parameters are generated without respecting the property’s nullability configuration. As a result, if a property is defined as nullable and null is explicitly provided, the ORM instead assigns the default value. This causes a behavioral discrepancy compared to previous versions.

### Reproduction

I am opening a pr to solve this issues which will have the reproduction included

### What driver are you using?

@mikro-orm/postgresql

### MikroORM version

6.5.5

### Node.js version

22+

### Operating system

mac os sequoia

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/mikro-orm/mikro-orm/blob/master/CONTRIBUTING.md).
- [x] Read the [docs](https://mikro-orm.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/mikro-orm/mikro-orm/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mikro-orm/mikro-orm/discussions) or join our [Discord](https://discord.gg/w8bjxFHS7X).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #6880, opened 2025-09-24
fix(): Account for prop nullability upon applying default value

adrien2p:
RESOLVES https://github.com/mikro-orm/mikro-orm/issues/6879

### DIFF
--- a/packages/knex/src/AbstractSqlDriver.ts
+++ b/packages/knex/src/AbstractSqlDriver.ts
@@ -545,7 +545,15 @@ export abstract class AbstractSqlDriver<Connection extends AbstractSqlConnection
     }
 
     const addParams = (prop: EntityProperty<T>, row: Dictionary) => {
-      let value = row[prop.name] ?? prop.default;
+      const rowValue = row[prop.name];
+      const isPropNullable = !!prop.nullable;
+
+      if (isPropNullable && rowValue === null) {
+        params.push(null);
+        return
+      }
+
+      let value = rowValue ?? prop.default;
 
       if (prop.kind === ReferenceKind.EMBEDDED && prop.object) {
         if (prop.array && value) {
--- a/tests/issues/GH6879.test.ts
+++ b/tests/issues/GH6879.test.ts
@@ -1,0 +1,52 @@
+import { Entity, MikroORM, PrimaryKey, Property } from '@mikro-orm/sqlite';
+
+@Entity()
+class User {
+
+  @PrimaryKey()
+  id: number;
+
+  @Property({ type: 'string' })
+  name: string;
+
+  @Property({ type: 'string', nullable: true })
+  email: string | null = null;
+
+  @Property({ type: 'number', nullable: true, default: 0 })
+  age!: number | null;
+
+  constructor(id: number, name: string, email: string) {
+    this.id = id;
+    this.name = name;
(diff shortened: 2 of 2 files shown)


packages/knex/src/AbstractSqlDriver.ts
```diff
@@ -545,7 +545,15 @@ export abstract class AbstractSqlDriver<Connection extends AbstractSqlConnection
     }
 
     const addParams = (prop: EntityProperty<T>, row: Dictionary) => {
-      let value = row[prop.name] ?? prop.default;
+      const rowValue = row[prop.name];
+      const isPropNullable = !!prop.nullable;
+
+      if (isPropNullable && rowValue === null) {
```
B4nan: lets inline this, its already readable enough


```suggestion

      if (prop.nullable && rowValue === null) {
```

adrien2p: let's do that 👍 I see that you applied it already so I ll resolve it

tests/features/fulltext/full-text-search-tsvector.postgres.test.ts
```diff
@@ -154,7 +155,7 @@ describe('full text search tsvector in postgres', () => {
     await orm.em.flush();
 
     expect(mock.mock.calls[0][0]).toMatch(`begin`);
-    expect(mock.mock.calls[1][0]).toMatch(`insert into "book" ("title", "searchable_title", "searchable_title_no_type", "searchable_title_english", "searchable_title_weighted") values ('My Life on The ? Wall, part 1', to_tsvector('simple', 'My Life on The ? Wall, part 1'), to_tsvector('simple', 'My Life on The ? Wall, part 1'), to_tsvector('english', 'My Life on The ? Wall, part 1'), default) returning "id"`);
+    expect(mock.mock.calls[1][0]).toMatch(`insert into "book" ("title", "searchable_title", "searchable_title_no_type", "searchable_title_english", "searchable_title_weighted") values ('My Life on The ? Wall, part 1', to_tsvector('simple', 'My Life on The ? Wall, part 1'), to_tsvector('simple', 'My Life on The ? Wall, part 1'), to_tsvector('english', 'My Life on The ? Wall, part 1'), NULL) returning "id"`);
```
adrien2p, line 157: @B4nan, the resulted payload from empty object after converting to database value is null for the prop, resulting in null here because of the nullability. Do you confirm that it is correct?

tests/features/fulltext/full-text-search-tsvector.postgres.test.ts
```diff
@@ -139,8 +139,9 @@ describe('full text search tsvector in postgres', () => {
     const mock = mockLogger(orm);
     await orm.em.flush();
 
+
     expect(mock.mock.calls[0][0]).toMatch(`begin`);
```
B4nan, line 142:
```suggestion
    await orm.em.flush();

    expect(mock.mock.calls[0][0]).toMatch(`begin`);
```

tests/features/fulltext/full-text-search-tsvector.postgres.test.ts
B4nan: Yeah, that should be fine, at least I hope so 🙃

Issue #6882, opened 2025-09-25
Error with `defineEntity` + `.nullable(true)`: "Exported variable 'UserSchema' has or is using name '__optional' from external module ".../@mikro-orm/core/typings" but cannot be named.ts(4023)"


ialexryan:
### Describe the bug

I'm really excited to adopt the new defineEntity + InferEntity flow from v6.5, but hitting a type error (TS4023) whenever I declare a nullable property.

It seems that when `declaration: true` (or `composite: true` in my case) is enabled in tsconfig.json, using `defineEntity` with any `.nullable(true)` properties fails to typecheck with the following error:

```
Exported variable 'UserSchema' has or is using name '__optional' from external module "@mikro-orm/core/typings" but cannot be named.
```

That error message comes from the attached minimal reproduction repro. Thanks in advance for your time, and for this really cool new cutting-edge feature!

### Reproduction

https://github.com/ialexryan/mikro-orm-4023-error-repro

### What driver are you using?

@mikro-orm/postgresql

### MikroORM version

6.5.5

### Node.js version

node@24.4.0, typescript@5.9.2

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/mikro-orm/mikro-orm/blob/master/CONTRIBUTING.md).
- [x] Read the [docs](https://mikro-orm.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/mikro-orm/mikro-orm/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mikro-orm/mikro-orm/discussions) or join our [Discord](https://discord.gg/w8bjxFHS7X).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

ialexryan:
cc @xcfox – think you might be interested in this and/or have some valuable insights 🙂  (repro repo [here](https://github.com/ialexryan/mikro-orm-4023-error-repro))

B4nan:
Hmm, I am starting to hate typescript for this. The whole point of that symbol is that we don't export it, it's internal. I wonder why its only a problem with the `defineEntity` and not the regular use of the `Opt` type.

xcfox:
A temporary workaround is to place `.nullable()` after `.default()`:

```ts
export const UserSchema = defineEntity({
  name: 'User',
  tableName: 'user',
[...]

PR #6884, opened 2025-09-26
fix(typings): enhance branded type definitions for `Opt`, `Hidden`, `Config` and `RequiredNullable`

xcfox:
This PR will fix https://github.com/mikro-orm/mikro-orm/issues/6882

In the current implementation, for the following entity:
```ts
import { defineEntity, type InferEntity } from '@mikro-orm/core';

export const User = defineEntity({
  name: 'User',
  tableName: 'user',
  properties: (p) => ({
    name: p.string(),
    email: p.string().unique(),
    bio: p.string().nullable(true).default(null),
  }),
});

export type IUser = InferEntity<typeof User>;
```
`IUser` will be inferred as:

```ts
type IUser = {
    id: number;
    name: string;
    email: string;
    bio: string & {
        [__optional]?: 1;
    };
}
```

With the improved brand type, `IUser` will be inferred as:
```ts
type IUser = {
    id: number;
    name: string;
    email: string;
    bio: string & Opt.Brand;
}
```

### DIFF
--- a/packages/core/src/typings.ts
+++ b/packages/core/src/typings.ts
@@ -74,22 +74,46 @@ export const EagerProps = Symbol('EagerProps');
 export const HiddenProps = Symbol('HiddenProps');
 export const Config = Symbol('Config');
 
-// eslint-disable-next-line @typescript-eslint/naming-convention
-declare const __optional: unique symbol;
+export type Opt<T = unknown> = T & Opt.Brand;
+// eslint-disable-next-line @typescript-eslint/no-redeclare
+export declare namespace Opt {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  const __optional: unique symbol;
+  export interface Brand {
+      [__optional]?: 1;
+  }
+}
 
-// eslint-disable-next-line @typescript-eslint/naming-convention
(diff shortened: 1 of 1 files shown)


B4nan (approved):
now that's another level of sorcery, how did you find out about this? :]

xcfox:
If we were to completely redesign these brand types, the best approach would be to use only `interface`:

```ts
export interface Opt {
    [__optional]?: 1;
}

// Use it with intersection:
export class User {
  bio: string & Opt
}
```

Considering compatibility with existing code, the current implementation retains the type alias `type Opt<T>`,  
and names the `Brand` interface as `Opt.Brand`. This ensures that the type alias `Opt` and its `Brand` always appear together.

B4nan:
> If we were to completely redesign these brand types, the best approach would be to use only interface:

Yeah, but then the generic signature wouldn't work, so something to consider for v7, the question is if it's worth it, e.g. for performance reasons. I personally prefer the intersection, so I wouldn't mind dropping the other way if it would mean better perf.

ialexryan:
@xcfox you're my hero! Thank you both for the quick resolution on this, I'll be eagerly awaiting the next release like a kid on Christmas.

PR #6888, opened 2025-09-27
chore: added healthcheck commands for all DB engines

boenrobot:
Use of TCP is explicitly encoded into all commands, ensuring they don't connect successfully until TCP is available, even if shared memory already is.

For MSSQL, the "-b" flag is also added,
which ensures failures are registered as a status code change, not just an error message, which is what happens without it.


B4nan:
Right, I realized that myself when working on the oracle driver some time ago, but forgot to port it to master 🙃 

https://github.com/mikro-orm/mikro-orm/blob/oracledb/.github/workflows/tests.yml#L159

PR #6901, opened 2025-10-02
[RFC] feat(deps): replace eslint with biome

kpervin:
This is a RFC on the viability to migrate from eslint to biome.

Purpose: Biome is quite a bit quicker than eslint, and incorporates formatting instead of relying on a 3rd party (`@stylistic/eslint-plugin`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,6 @@
     "@mikro-orm/mongo-highlighter": "1.0.0",
     "@mikro-orm/sql-highlighter": "1.0.1",
     "@side/jest-runtime": "^1.1.0",
-    "@stylistic/eslint-plugin-ts": "^4.0.0",
     "@types/esprima": "4.0.6",
     "@types/fs-extra": "11.0.4",
     "@types/jest": "29.5.14",
--- a/yarn.lock
+++ b/yarn.lock
@@ -906,17 +906,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@eslint-community/eslint-utils@npm:^4.7.0":
-  version: 4.7.0
-  resolution: "@eslint-community/eslint-utils@npm:4.7.0"
-  dependencies:
-    eslint-visitor-keys: "npm:^3.4.3"
-  peerDependencies:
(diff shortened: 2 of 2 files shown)


B4nan:
If something, I'd like to adopt oxlint at some point, not biome.

kpervin:
> If something, I'd like to adopt oxlint at some point, not biome.

Closing

B4nan:
Otherwise, I appreciate the initiative, I also hate how slow eslint is (especially when combined with type aware rules), not to mention the whole drama with eslint v9 and their flat configs and the amount of things that are implemented without thinking abotu performance (usually third party rules).

B4nan:
> native formatting

Regarding formatters, it's something I keep thinking about. You could say I have a love/hate relationship with them. I like the idea, but I usually hate the implementation, and more importantly, the lack of options (prettier not having anything configurable drives me crazy). On the other hand, when I messed around with dprint, which is highly configurable, it wasn't much better (so many options, trying to find the one I was after was often impossible, or it was non-existent). Lastly, none of the existing formatters allow to control line breaks in a good way, usually they just get rid of them (namely blank lines padding in classes, which we enforce here via eslint). And since they usually are not extensible, you just need to make peace with what their maintainers force you to "like". I eagerly await a brighter future, where I can adopt one too, and it won't make me angry every time I try to commit something.

(I do use formatters at work extensively, can't say it improved much for me personally, rather the opposite, but I do see the value for other collaborators and for PR reviews)

kpervin:
I did not realize I touched on such a sore spot haha. I agree with you on all points. The eslint v9 debacle has our work codebase reluctant to migrate from v8 atm, so I'm constantly on the lookout for something that is faster. I mainly suggested biome due to NextJS' recent inclusion in their `create-next` script, but I agree that oxlint makes more sense due to being maintained by Void0 (Vite). Shame it isn't more feature complete as of now, though.

darkbasic:
> prettier not having anything configurable drives me crazy

I'm glad I'm not the only one.

PR #6910, opened 2025-10-06
fix(defineEntity): update autoincrement method to return Opt<Value>

xcfox:
As mentioned in [this discussion](https://github.com/mikro-orm/mikro-orm/discussions/6865), `somePk: p.bigint().primary().autoincrement()` should be optional when creating an entity.

This PR fixes this issue, so any property marked with `.autoincrement()` will be optional when creating an entity via `em.create(...)`.

### DIFF
--- a/packages/core/src/entity/defineEntity.ts
+++ b/packages/core/src/entity/defineEntity.ts
@@ -110,7 +110,7 @@ export class PropertyOptionsBuilder<Value> {
   /**
    * Explicitly specify the auto increment of the primary key.
    */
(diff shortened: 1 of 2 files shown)


B4nan:
btw one thing that was also mentioned in that discussion - support for the `class` option. right now it breaks things even if the interface is matching the metadata, but since this was never really intended, maybe we should just disallow the `class` option in `defineEntity` completely? and on a similar note, i saw people tried to use the new property helpers with `EntitySchema` directly, which could be pretty nice to support, but then we would likely face the very same type level issues as with the `class` option.

PR #6912, opened 2025-10-07
feat(defineEntity): support for composite keys

xcfox:
This PR implements support for composite primary keys when defining entities using `defineEntity`:

```ts
export const Car = defineEntity({
  name: 'Car',
  properties: p => ({
    name: p.string(),
    year: p.integer(),
  }),
  primaryKeys: ['name', 'year'],
});
```

`Car` will be inferred as:

```ts
type ICar = {
    name: string;
    year: number;
    [PrimaryKeyProp]?: ["name", "year"] | undefined;
}
```

### DIFF
--- a/docs/docs/composite-keys.md
+++ b/docs/docs/composite-keys.md
@@ -3,6 +3,9 @@ title: Composite and Foreign Keys as Primary Key
 sidebar_label: Composite Primary Keys
 ---
 
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
 > Support for composite keys was added in version 3.5
 
 MikroORM supports composite primary keys natively. Composite keys are a very powerful relational database concept, and we took good care to make sure MikroORM supports as many of the composite primary key use-cases. MikroORM supports composite keys of primitive data types as well as foreign keys as primary keys. You can also use your composite key entities in relationships.
@@ -16,6 +19,15 @@ Primary keys need to have their values set before you call `em.persist(entity)`.
 ## Primitive Types only
 
 Suppose you want to create a database of cars and use the model-name and year of production as primary keys:
+
+<Tabs
+groupId="entity-def"
+defaultValue="reflect-metadata"
+values={[
+{label: 'reflect-metadata', value: 'reflect-metadata'},
+{label: 'defineEntity', value: 'define-entity'},
+]}>
+<TabItem value="reflect-metadata">
 
 ```ts
 @Entity()
@@ -37,6 +49,23 @@ export class Car {
 
 }
 ```
+
+</TabItem>
+<TabItem value="define-entity">
+
+```ts
+export const Car = defineEntity({
+  name: 'Car',
+  properties: p => ({
+    name: p.string(),
+    year: p.integer(),
+  }),
+  primaryKeys: ['name', 'year'],
+});
+```
+
+</TabItem>
+</Tabs>
 
 Now you can use this entity:
 
@@ -77,6 +106,15 @@ The semantics of mapping identity through foreign entities are easy:
 
 We keep up the example of an Article with arbitrary attributes, the mapping looks like this:
 
+<Tabs
+groupId="entity-def"
+defaultValue="reflect-metadata"
+values={[
+{label: 'reflect-metadata', value: 'reflect-metadata'},
+{label: 'defineEntity', value: 'define-entity'},
+]}>
+<TabItem value="reflect-metadata">
+
 ```ts
 @Entity()
 export class Article {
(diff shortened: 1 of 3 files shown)


B4nan:
Hmm, why not like this instead?

```ts
export const Car = defineEntity({
  name: 'Car',
  properties: p => ({
    name: p.string().primary(),
    year: p.integer().primary(),
  }),
});
```

I don't think people should use the `primaryKeys` option, it's computed based on the `primary` option on separate properties.

docs/docs/composite-keys.md
```diff
@@ -17,6 +20,15 @@ Primary keys need to have their values set before you call `em.persist(entity)`.
...
 Suppose you want to create a database of cars and use the model-name and year of production as primary keys:
 
+<Tabs
+groupId="entity-def"
+defaultValue="reflect-metadata"
+values={[
+{label: 'reflect-metadata', value: 'reflect-metadata'},
+{label: 'defineEntity', value: 'define-entity'},
+]}>
+<TabItem value="reflect-metadata">
```
B4nan, line 32: if we are adding the tabs here, they should list all the variants, not just those two

xcfox:
**Inferring an ordered key tuple from an object is an enormous challenge.**

Take this code for example:

```ts
export const Car = defineEntity({
  name: 'Car',
  properties: p => ({
    name: p.string().primary(),
    year: p.integer().primary(),
  }),
});
```

Current TypeScript code cannot accurately infer `[PrimaryKeyProp]` as `['name', 'year']`.   
Specifically, this is because the union type obtained from the `keyof` operation is always unordered.   
In current TypeScript, we can barely only infer `[PrimaryKeyProp]` as `('name' | 'year')[]`.


Using `.primary()` directly looks better.  
Considering that `PrimaryKeyProp` is only used in a few cases where tuples are used as `FilterQuery`, a less precise `PrimaryKeyProp` seems acceptable:
```ts
type ICar = {
    name: string;
    year: number;
    [PrimaryKeyProp]?: ('name' | 'year')[] | undefined;
}
```
I'll try to infer `PrimaryKeyProp` directly from `properties` without using `primaryKeys`.

B4nan:
I see, true that, you need an ordered tuple somewhere to be able to use the correct order. Can we make the `primaryKeys` option type safe (so it won't accept anything else than defined properties? even better, only those with `.primary()`?)

> a less precise PrimaryKeyProp seems acceptable:

Yeah, I guess it's better than not inferring this at all.

xcfox:
## Update on October 9th

Inside `defineEntity`, I have reimplemented the type inference part of `PropertyOptionsBuilder`. The current implementation now retains options related to the final property type, such as `{ primary: true }`, `{ ref: true }`, `{ nullable: true }`, etc.

### Automatic Primary Key Inference

`defineEntity` can now infer accurate primary keys:

- Single Primary Key
```ts
const Cat = defineEntity({
  name: 'Cat',
  properties: p => ({
    name: p.text().primary(),
  }),
});
```

The type of `Cat` will be inferred as:
```ts
type ICat = {
    name: string;
    [PrimaryKeyProp]?: "name" | undefined;
}
``` 

- Composite Primary Key
```ts
export const Car = defineEntity({
  name: 'Car',
  properties: p => ({
    name: p.string().primary(),
    year: p.integer().primary(),
    description: p.text(),
  }),
});
```

`Car` will be inferred as:

```ts
type ICar = {
    name: string;
    year: number;
    description: string;
    [PrimaryKeyProp]?: ('name' | 'year')[];
}
```

### Manual Primary Key Declaration

`defineEntity` allows declaring an exact primary key tuple via `primaryKeys`:

```ts
export const Car = defineEntity({
  name: 'Car',
  properties: p => ({
    name: p.string().primary(),
    year: p.integer().primary(),
    description: p.text(),
  }),
  primaryKeys: ['name', 'year'],
});
```

`Car` will be inferred as:
```ts
type ICar = {
    name: string;
    year: number;
    description: string;
    [PrimaryKeyProp]?: ["name", "year"];
}
```

#### `primaryKeys` Type Checking

Only property keys marked with `.primary()` can be used in `primaryKeys: [...]`.

```ts
export const Car = defineEntity({
  name: 'Car',
  properties: p => ({
    name: p.string().primary(),
    year: p.integer().primary(),
    description: p.text(),
  }),
  primaryKeys: ['description'], // ❌ Type "description" is not assignable to type: ...
});
```

### Documentation Update

The usage of composite primary keys across all entity definition methods has been refined in `composite-keys.md`.

B4nan:
Great, thank you! I'll try to review this during the weekend.

One question - were you thinking about the `class` option combined with `defineEntity`? Runtime-wise, it already works, but it breaks the types. And similarly, some people already tried to use `EntitySchema` with the new property builder, and it fails as well, likely the same issue. If we can find a way to make it work with `defineEntity`, maybe we could deprecate EntitySchema as a user facing way in v7, and only use it internally (it would be rather a soft BC, since `defineEntity` returns `EntitySchema`, so it would keep working as usual, I am just thinking about a way to simplify the docs). And if we can't make it work, we should omit it from the available `defineEntity` options.

B4nan:
Also, maybe we should just export the property builder as `p`, just like we export the type map under `t`. I don't like that we suggest doing the `const p = defineEntity.properties` thing currently.

xcfox:
> were you thinking about the class option combined with `defineEntity`?

Using `class` in `defineEntity` is perfectly valid; at the moment, `defineEntity` does not provide good support for the `class` attribute, but this can be easily improved (perhaps with some function overloading).  
I fully agree with removing `EntitySchema` from the documentation, as the `EntitySchema` style offers no advantages over `defineEntity`.  
When working with `class`, using `InferEntity` and `implements` keeps types consistent everywhere, for example:

```ts
const CarProperties = {
  name: p.string().primary(),
  year: p.integer().primary(),
  description: p.string().nullable(),
};

export class Car implements InferEntityFromProperties<typeof CarProperties> {
  name!: string;
  year!: number;
  description: string | null | undefined;

  constructor(name: string, year: number, description?: string) {
    this.name = name;
    this.year = year;
    this.description = description;
  }
}

export const CarSchema = defineEntity({
  name: 'Car',
  properties: CarProperties,
  class: Car,
  serializedPrimaryKey: 'name',
});
```

---

> Also, maybe we should just export the property builder as `p`, just like we export the type map under `t`. I don't like that we suggest doing the `const p = defineEntity.properties` thing currently.

I completely agree.

B4nan:
> export class Car implements InferEntityFromProperties<typeof CarProperties> {

Interesting, that's actually a pretty nice way to go about this. Just wondering if we should pick some fancier name for the `InferEntityFromProperties` type, it's quite a long name :] 

I think the problem I was referring to is the requirement for the `name` option, without it, this fails:

```ts
export const CarSchema = defineEntity({
//  name: 'Car',
  properties: CarProperties,
  class: Car,
  serializedPrimaryKey: 'name',
});
```

image

PR #6919, opened 2025-10-12
fix(defineEntity): enhance `mapToPk()` and remove default `ref: true`

xcfox:
This PR improves the type system for the relation property in `defineEntity`:
- enhance `mapToPk()` type inference
- Remove default `ref: true` from manyToOne and oneToOne relation builders
- Relations now require explicit `.ref()` call to enable reference wrapping

For the following entity:
```ts
const Foo = defineEntity({
  name: 'Foo',
  properties: p => ({
    name: p.string().primary(),
    toOne: () => p.oneToOne(Foo),
    toOneNullable: () => p.oneToOne(Foo).nullable(),
    toOneRef: () => p.oneToOne(Foo).ref(),
    toOneRefNullable: () => p.oneToOne(Foo).ref().nullable(),
    toOnePk: () => p.oneToOne(Foo).mapToPk(),
    toOnePkNullable: () => p.oneToOne(Foo).mapToPk().nullable(),
    scalarRef: p.string().ref(),
    scalarRefNullable: p.string().ref().nullable(),
  }),
});
```

Will be inferred as:
```ts
type IFoo = {
  name: string;
  toOne: IFoo;
  toOneNullable: IFoo | null | undefined;
  toOneRef: Reference<IFoo>;
  toOneRefNullable: Reference<IFoo> | null | undefined;
  toOnePk: string;
  toOnePkNullable: string | null | undefined;
  scalarRef: ScalarReference<string>;
  scalarRefNullable: ScalarReference<string | null | undefined>;
  [PrimaryKeyProp]?: 'name';
}
```

### DIFF
--- a/packages/core/src/entity/defineEntity.ts
+++ b/packages/core/src/entity/defineEntity.ts
@@ -6,7 +6,7 @@ import type { ManyToOneOptions } from '../decorators/ManyToOne';
 import type { OneToManyOptions } from '../decorators/OneToMany';
 import type { OneToOneOptions } from '../decorators/OneToOne';
 import type { ManyToManyOptions } from '../decorators/ManyToMany';
-import type { AnyString, GeneratedColumnCallback, Constructor, CheckCallback, FilterQuery, EntityName, Dictionary, EntityMetadata, PrimaryKeyProp, Hidden, Opt, Primary } from '../typings';
+import type { AnyString, GeneratedColumnCallback, Constructor, CheckCallback, FilterQuery, EntityName, Dictionary, EntityMetadata, PrimaryKeyProp, Hidden, Opt, Primary, EntityClass } from '../typings';
 import type { Reference, ScalarReference } from './Reference';
 import type { SerializeOptions } from '../serialization/EntitySerializer';
 import type { Cascade, DeferMode, LoadStrategy, QueryOrderMap } from '../enums';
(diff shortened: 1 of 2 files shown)


B4nan:
>     profile: p.oneToOne(Profile).mapToPk(), // Type: Reference<Profile>

Where did the `Reference` wrapper come from? I am testing a bit how this works now, and it feels like `profile: p.manyToOne(Profile)` also yields `Reference<Profile>`, that must be dependent on `ref()` call, it's not the default. I am on your branch now, and I see the same behaviour with the fix. Also apparently, nullablity doesnt work correctly, the `parent` option here is required in `em.create`:

```ts
const User = defineEntity({
  name: 'User',
  properties: p => ({
    _id: p.type(ObjectId).primary(),
    parent: () => p.manyToOne(User).nullable(),
  }),
});

const u = orm.em.create(User, {}); // fails because parent is missing
u.parent // this is typed as Reference<User> but it should be just User
```

xcfox:
Currently, `{ ref: true }` is enabled by default in `defineEntity` to achieve optimal type safety.

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/entity/defineEntity.ts#L849
https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/entity/defineEntity.ts#L862

We can use `.ref(false)` to revert to type-unsafe relationship declarations.

---

> Nullability doesn't work correctly.

Yes, I see the problem. I think I should adjust the [order](https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/entity/defineEntity.ts#L967) of `MaybeNullable` and `MaybeRef`.

B4nan:
> Currently, { ref: true } is enabled by default in defineEntity to achieve optimal type safety.

That's pretty bad, we can't have different defaults per entity definition variant, please revert this asap so it's opt-int as with other variants.

PR #6920, opened 2025-10-12
fix(defineEntity): support class-based entity definitions

xcfox:
This PR adds support for class-based entity definitions to the `defineEntity` function, providing an alternative approach to defining entities while maintaining full backward compatibility.

### Code Example

```ts
const bookProperties = {
  id: p.string().serializedPrimaryKey(),
  title: p.string(),
  tags: p.array().$type<string[]>(),
};

class Book implements InferEntityFromProperties<typeof bookProperties> {
  id!: string;
  title!: string;
  tags!: string[];
}

const BookSchema = defineEntity({
  class: Book,
  properties: bookProperties,
});
```

## Related Issues

https://github.com/mikro-orm/mikro-orm/discussions/6865#discussioncomment-14465227

### DIFF
--- a/tests/defineEntity.test.ts
+++ b/tests/defineEntity.test.ts
@@ -53,7 +53,7 @@ describe('defineEntity', () => {
       title!: string;
       tags!: string[];
 
-}
+    }
     const BookSchema = defineEntity({
       class: Book,
       properties: bookProperties,


tests/defineEntity.test.ts
```diff
@@ -40,6 +40,35 @@ describe('defineEntity', () => {
...
+      title: p.string(),
+      tags: p.array().$type<string[]>(),
+    };
+    class Book implements InferEntityFromProperties<typeof bookProperties> {
+
+      id!: string;
+      title!: string;
+      tags!: string[];
+
+}
```
B4nan:
```suggestion
    }
```

PR #6921, opened 2025-10-12
Update metadata-providers.md

Exifers:


B4nan:
this line is actually super dated, nowadays you can just use `Rel` wrapper in case of a dependency cycle on type level

B4nan:
14d2a83288ef61b928748d638b3823154248ca32

also note that what you did was only a change of version snapshot, that's why the PR was sitting here unmerged, such change would get lost on next feature release

PR #6922, opened 2025-10-12
fix(core): add some missing options to `defineEntity` relation builders

B4nan:


B4nan:
@xcfox I've tried to switch all the entities from https://github.com/mikro-orm/mikro-orm/tree/master/tests/entities-schema to the `defineEntity` helper, a few things were missing, which I tried to add on the go:
- `p.datetime()` and `p.time()` accept `length` parameter
- `p.inversedBy` and `p.mappedBy` were missing sometimes and were not type-safe
- `p.formula()`, `p.persist(false)` and `p.version()` makes property marked as `Opt`
- a few other things were missing: `p.orphanRemoval()`
- entity hooks that use inferred entity types cause issues with circular references (see `Author4.ts`)
- `extends` only works with classes, I ended up using the properties export, would be great to have it working the same way as with `EntitySchema`, so `extends: Book4` will have effect on type level too

But the result is hardly usable, running `tsc` fails with `JavaScript heap out of memory` 😅 Just try running `yarn tsc-check-tests`.

If we are to suggest this as the first class citizen in v7, the performance issues will need to be addressed.

xcfox:
performance issues are indeed very important, and we need to spend some time optimizing this issue.

Additionally, in the long run, [ts-go](https://github.com/microsoft/typescript-go) may be the biggest performance improvement.

xcfox:
I have merged `PropertyOptionsBuilder`, `EmbeddedOptionsBuilder`, `ManyToManyOptionsBuilder`, `ManyToOneOptionsBuilder`, `OneToOneOptionsBuilder`, `OneToManyOptionsBuilder` into a single `UniversalPropertyOptionsBuilder`:

- Use the `Options` generic to accumulate the `options` type;
- Use the `IncludeKeys` generic to prevent incorrect calls to non-existent parameters;

I have a simple test locally:

<details>

<summary>Bench Test Code</summary>

```ts
import { bench } from '@arktype/attest';
import {
  type ScalarReference,
  type Reference,
  type PrimaryKeyProp,
  defineEntity,
  EntitySchema,
} from '@mikro-orm/core';

bench('defineEntity with relations', () => {
  const Foo = defineEntity({
    name: 'Foo',
    properties: p => ({
      name: p.string().primary(),
      toOne: () => p.oneToOne(Foo),
      toOneNullable: () => p.oneToOne(Foo).nullable(),
      toOneRef: () => p.oneToOne(Foo).ref(),
      // toOneRefNullable: () => p.oneToOne(Foo).ref().nullable(),
      toOnePk: () => p.oneToOne(Foo).mapToPk(),
      toOnePkNullable: () => p.oneToOne(Foo).mapToPk().nullable(),
      scalarRef: p.string().ref(),
      scalarRefNullable: p.string().ref().nullable(),
    }),
  });
}).types();

bench('defineEntity with ref and nullable', () => {
  const Foo = defineEntity({
    name: 'Foo',
    properties: p => ({
      name: p.string().primary(),
      toOne: () => p.oneToOne(Foo),
      toOneNullable: () => p.oneToOne(Foo).nullable(),
      toOneRef: () => p.oneToOne(Foo).ref(),
      toOneRefNullable: () => p.oneToOne(Foo).ref().nullable(),
      toOnePk: () => p.oneToOne(Foo).mapToPk(),
      toOnePkNullable: () => p.oneToOne(Foo).mapToPk().nullable(),
      scalarRef: p.string().ref(),
      scalarRefNullable: p.string().ref().nullable(),
    }),
  });
}).types();

bench('defineEntity only with ref and nullable', () => {
  const Foo = defineEntity({
    name: 'Foo',
    properties: p => ({
      name: p.string().primary(),
      toOneRefNullable: () => p.oneToOne(Foo).ref().nullable(),
    }),
  });
}).types();

bench('defineEntity only with nullable and ref', () => {
  const Foo = defineEntity({
    name: 'Foo',
    properties: p => ({
      name: p.string().primary(),
      toOneRefNullable: () => p.oneToOne(Foo).nullable().ref(),
    }),
  });
}).types();

bench('EntitySchema', () => {
  interface IFoo {
    name: string;
    toOne: IFoo;
    toOneNullable: IFoo | null | undefined;
    toOneRef: Reference<IFoo>;
    toOneRefNullable: Reference<IFoo> | null | undefined;
    toOnePk: string;
    toOnePkNullable: string | null | undefined;
    scalarRef: ScalarReference<string>;
    scalarRefNullable: ScalarReference<string | null | undefined>;
    [PrimaryKeyProp]?: 'name';
  }

  const FooSchema = new EntitySchema<IFoo>({
    name: 'Foo',
    properties: {
      name: { type: 'string', primary: true },
      toOne: { kind: '1:1', entity: 'Foo' },
      toOneNullable: { kind: '1:1', entity: 'Foo', nullable: true },
      toOneRef: { kind: '1:1', entity: 'Foo', ref: true },
      toOneRefNullable: {
        kind: '1:1',
        entity: 'Foo',
        ref: true,
        nullable: true,
      },
      toOnePk: { kind: '1:1', entity: 'Foo', mapToPk: true },
      toOnePkNullable: {
        kind: '1:1',
        entity: 'Foo',
        mapToPk: true,
        nullable: true,
      },
      scalarRef: { kind: '1:1', entity: 'string', ref: true },
      scalarRefNullable: {
        kind: '1:1',
        entity: 'string',
        ref: true,
        nullable: true,
      },
    },
  });
}).types();
```

</details>

I use ‘@arktype/attest’ to test performance, the results before and after modification are as follows:

| Test Case | Current Result | Before Result | Change | Status |
|-----------|---------------|----------|--------|--------|
| **defineEntity with relations** | 43,683 | 41,059 | +6.39% | ⚠️ Performance degradation |
| **defineEntity with ref and nullable** | 44,029 | 906,231 | -95.14% | ✅ Significant improvement |
| **defineEntity only with ref and nullable** | 42,014 | 903,676 | -95.35% | ✅ Significant improvement |
| **defineEntity only with nullable and ref** | 42,018 | 39,096 | +7.47% | ⚠️ Performance degradation |
| **EntitySchema** | 5,238 | 5,238 | 0.00% | ✅ No change |

B4nan:
I would love to add some arktype tests in here, feel free to add it directly to the codebase.

tests/entities-schema/Author4.ts
```diff
@@ -34,33 +14,40 @@ export class Identity {
...
+    version: p.integer().persist(false),
+    identity: p.embedded(IdentitySchema).object().nullable(),
   },
   hooks: {
     onLoad: [randomHook],
   },
 });
+
+function randomHook(args: EventArgs<any>) {
+  const args1 = args as EventArgs<IAuthor4>;
```
B4nan: @xcfox do you think this is solvable somehow? 

if you change the `EventArgs<any>` to `EventArgs<IAuthor4>`, things fall apart

maybe adding the hooks lazily could help? e.g. `Author4.addHook('onLoad', ...)` (we dont have that API yet)

xcfox: A single `any` can break the circular reference, and I prefer the following style:

```ts
export const Author4 = defineEntity({
  ...
  hooks: {
    onLoad: [randomHook as any],
  },
});

function randomHook(args: EventArgs<IAuthor4>) {
  //
}
```

However, I would prefer that hooks could directly infer the type:

```ts
export const Author4 = defineEntity({
  ...
  hooks: {
    onLoad: [(args) => {
      args.entity // IAuthor4
    }],
  },
});

xcfox:
>  * `extends` only works with classes, would be great to have it working the same way as with `EntitySchema`, so `extends: Book4` will have effect on type level too

One question: **What specific effects does `extends: Book4` have on the type system?**  
In this repository, I almost never see the second parameter of `EntitySchema` being used.

---

I personally use the `class` keyword always.  
However, I believe that when using `defineEntity`, we should favor composition over inheritance.
For OOP enthusiasts, they can certainly use the native `class` and `extends`.   

I personally believe that adding support for `class` and `extends` in `defineEntity` should be seen as a means of maintaining compatibility with legacy code, rather than as the recommended approach. We only need to ensure that existing code using `EntitySchema` can be smoothly migrated into `defineEntity`, rather than adding new features centered around `class` and `extends`.

xcfox:
> If we are to suggest this as the first class citizen in v7

Personally, I subjectively believe that in MikroORM v7, we should prioritize recommending `defineEntity`, followed by `class` + `decorators`. There are many reasons for this:
- A better type system
- Less boilerplate code
- No additional configuration required (`experimentalDecorators`, `reflect-metadata`, `ts-morph`)

I recently came across an article where other ORMs are avoiding the use of decorators: https://joist-orm.io/blog/avoiding-decorators/

B4nan:
> However, I believe that when using defineEntity, we should favor composition over inheritance.

It can be quite tedious to use composition in more complex cases. This is not just about the property definitions, everything is merged based on `extends`: hooks, indexes, uniques, and last but not least, when `extends` holds a reference instead of a string name, we also discover parent entities automatically. And all of this works recursively. I would very much like to preserve how this works here, I dont see a reason why something should work only with decorators and EntitySchema, but not here. 

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/metadata/MetadataDiscovery.ts#L891

The only thing that is relevant on type level is access to the parent properties.

B4nan:
I was rebasing this to v7, and we already have more complex hooks in that entity (https://github.com/mikro-orm/mikro-orm/blob/v7/tests/entities-schema/Author4.ts). It required static props on the class for tests, so I wanted to give actual classes a try. There were some issues with the interop, e.g. many places require `EntitySchema` only for the input (e.g. `p.manyToOne(Author4)` fails when `Author4` is a class), I managed to fix that, but maybe we will need the fix on other places - basically everything that expects an EntitySchema reference needs to support EntityClass too. The `extends` option is not working even for the `EntitySchema` instances, ideally it would for entity classes too.

B4nan:
Here is a backport of the fixes from v7 rebase: 8a03ecabe4a899a2b288b65f62b7e0bb14dc635c

PR #6933, opened 2025-10-19
Change MikroORM CLI installation command

adriancuadrado:
Updated installation command for MikroORM CLI to use --dev flag.


B4nan:
CLI is quite a thin dependency, and Id say many peole use in production, eg to run migrations.

PR #6936, opened 2025-10-20
fix(core): update `extends` type in `defineEntity` and `EntitySchema`

xcfox:
I fixed the type error of the `extends` property in Author4Schema,  
so now the `extends` property of `defineEntity` allows passing a class constructor.


xcfox:
But I notice that in https://github.com/mikro-orm/mikro-orm/commit/10cfd28785b01cebdbf80686debec8567ee1ba30 no handling for class constructors was added, and we may need to complete this logic?

B4nan:
Yeah I realized that later as well, that this wasnt really working before either. Ideally it would work with EntitySchema too, but since we want to make that rather internal thing in v7, it's not a huge deal really.

tests/defineEntity.test.ts
```diff
@@ -187,6 +187,45 @@ describe('defineEntity', () => {
     assert<IsExact<IFoo, { id: number; name: string; createdAt: Opt<Date>; updatedAt: Opt<Date>; [PrimaryKeyProp]?: 'id' }>>(true);
   });
 
+  it('should define entity with class constructor as extends', () => {
+    const p = defineEntity.properties;
```
B4nan: FYI I've already added the top level `p` export.

PR #6939, opened 2025-10-22
feat(sql): infer Database for kysely

xcfox:
Adds `InferKyselyDB` and `InferKyselyTable` type utilities to automatically infer Kysely Database types from MikroORM entities, enabling type-safe Kysely query builders.

### 1. Type Inference Utilities

- **`InferKyselyDB<TEntities, TOptions>`**: Infers a complete Kysely Database type from a collection of entities
- **`InferKyselyTable<TSchema, TNamingStrategy>`**: Infers the corresponding Table type from a single entity

### 2. EntityManager.getKysely() Generic Support

`em.getKysely<TDB>()` supports:
- **Automatic inference**: When no generic is provided, automatically infers the Database type from entities
- **Manual specification**: When a generic is provided, uses the specified Database type

### 3. Type Conversion Rules

- Entity names are converted to snake_case (e.g., `User` → `user`, `UserProfile` → `user_profile`)
- Property names are converted to snake_case (e.g., `fullName` → `full_name`)
- Supports `fieldName` for custom column names
- Automatically handles foreign key columns for relationships (e.g., `author` → `author_name`)
- Correctly handles `autoincrement`, `default`, `nullable` and other properties, generating corresponding `Generated<T>` types

### Basic Usage

```typescript
// Define entities
const User = defineEntity({
  name: 'User',
  properties: {
    name: p.string().primary(),
    email: p.string().nullable(),
  },
});

// Automatically infer Database type
const kysely = orm.em.getKysely();
type DB = InferDBFromKysely<typeof kysely>; // { user: { name: string; email: string | null } }

// Manually specify Database type
interface MyDatabase {
  user: { name: string; email: string | null };
}
const kysely2 = orm.em.getKysely<MyDatabase>();

// Use InferKyselyTable to infer a single table type
type UserTable = InferKyselyTable<typeof User>; // { name: string; email: string | null }
```

### Mixing InferKyselyTable with Manual Type Declarations

```typescript
const User = defineEntity({
  name: 'User',
  properties: {
    name: p.string().primary(),
    email: p.string().nullable(),
  },
});

const Post = defineEntity({
  name: 'Post',
  properties: {
    title: p.string(),
    author: () => p.manyToOne(User),
  },
});

// Manually declare additional table types (e.g., views or native tables)
interface ManualViewTable {
  view_id: number;
  view_count: number;
}

// Mix inferred types with manual types to build a complete Database type
interface MixedDatabase {
  user: InferKyselyTable<typeof User>;           // Use inferred type
  post: InferKyselyTable<typeof Post>;            // Use inferred type
  view_stats: ManualViewTable;        // Use manual type
}

// Create Kysely instance with mixed types
const kysely = orm.em.getKysely<MixedDatabase>();

// Now you can type-safely query all tables
const user = await kysely.selectFrom('user').selectAll().executeTakeFirst();
const post = await kysely.selectFrom('post').selectAll().executeTakeFirst();
const stats = await kysely.selectFrom('view_stats').selectAll().executeTakeFirst();
```

### DIFF
--- a/tests/bench/types/defineEntity.ts
+++ b/tests/bench/types/defineEntity.ts
@@ -1,11 +1,13 @@
+#!/usr/bin/env -S node --import @swc-node/register/esm-register
+
 import { bench } from '@ark/attest';
 import {
   type ScalarReference,
   type Reference,
-  type PrimaryKeyProp,
   defineEntity,
   EntitySchema,
   p,
+  PrimaryKeyProp,
 } from '@mikro-orm/core';
 
 bench('defineEntity with relations', () => {
@@ -22,7 +24,7 @@ bench('defineEntity with relations', () => {
       scalarRefNullable: p.string().ref().nullable(),
     },
   });
-}).types([42446, 'instantiations']);
+}).types([39934, 'instantiations']);
 
 bench('defineEntity with ref and nullable', () => {
   const Foo = defineEntity({
@@ -39,7 +41,7 @@ bench('defineEntity with ref and nullable', () => {
       scalarRefNullable: p.string().ref().nullable(),
     },
   });
-}).types([42792, 'instantiations']);
+}).types([40278, 'instantiations']);
 
 bench('defineEntity only with ref and nullable', () => {
   const Foo = defineEntity({
@@ -49,7 +51,7 @@ bench('defineEntity only with ref and nullable', () => {
       toOneRefNullable: () => p.oneToOne(Foo).ref().nullable(),
     },
   });
-}).types([40763, 'instantiations']);
+}).types([38247, 'instantiations']);
 
 bench('defineEntity only with nullable and ref', () => {
   const Foo = defineEntity({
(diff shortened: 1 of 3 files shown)


B4nan:
Nice, I wanted to experiment with this too, great that you are looking into it.

B4nan:
Do you have a rough estimate when this could be ready? I'd like to rebase the v7 branch in the upcoming week or two.

xcfox:
> Do you have a rough estimate when this could be ready? I'd like to rebase the v7 branch in the upcoming week or two.

The **infer Database for Kysely** feature is now complete.

The **easy-to-use custom Kysely plugin** turned out to be much more complex than I expected, and I estimate it will take another month to finish.

I think I can move the **easy-to-use custom Kysely plugin** to a separate PR, so that **infer Database for Kysely** can be merged first.

B4nan:
Yeah, let's do it that way.

B4nan:
Hmm, I messed up the conflicts. Can you please rebase and resolve those? I was trying to bump vitest to v4, and found quite a lot of weird issues on the way, so ended up with a bump to v3.2 which still required some code changes (v4 randomly fails on my end now).

edit: right, I am dumb, I just need to drop the `type` keyword

packages/core/src/utils/Configuration.ts
```diff
@@ -674,6 +674,14 @@ export interface MikroORMOptions<D extends IDatabaseDriver = IDatabaseDriver, EM
   hashAlgorithm: 'md5' | 'sha256';
 }
 
-export type Options<D extends IDatabaseDriver = IDatabaseDriver, EM extends D[typeof EntityManagerType] & EntityManager = D[typeof EntityManagerType] & EntityManager> =
-  Pick<MikroORMOptions<D, EM>, Exclude<keyof MikroORMOptions<D, EM>, keyof typeof Configuration.DEFAULTS>>
-  & Partial<MikroORMOptions<D, EM>>;
+// export type Options<D extends IDatabaseDriver = IDatabaseDriver, EM extends D[typeof EntityManagerType] & EntityManager = D[typeof EntityManagerType] & EntityManager> =
+//   Pick<MikroORMOptions<D, EM>, Exclude<keyof MikroORMOptions<D, EM>, keyof typeof Configuration.DEFAULTS>>
+//   & Partial<MikroORMOptions<D, EM>>;
```
B4nan: any reason to keep this here?

packages/core/src/utils/Configuration.ts
```diff
@@ -674,6 +674,14 @@ export interface MikroORMOptions<D extends IDatabaseDriver = IDatabaseDriver, EM
...
+
+export interface Options<
+  D extends IDatabaseDriver = IDatabaseDriver,
+  EM extends D[typeof EntityManagerType] & EntityManager = D[typeof EntityManagerType] & EntityManager,
+  Entities extends (string | EntityClass<AnyEntity> | EntityClassGroup<AnyEntity> | EntitySchema)[] = (string | EntityClass<AnyEntity> | EntityClassGroup<AnyEntity> | EntitySchema)[],
+> extends Pick<MikroORMOptions<D, EM>, Exclude<keyof MikroORMOptions, keyof typeof Configuration.DEFAULTS>>, Partial<MikroORMOptions<D, EM>> {
```
B4nan, line 681: this reminds me i wanted to get rid of the computed `Options` type in v7, since the generated docs for that look dumb (its a type alias, so its just printed like [this](https://mikro-orm.io/api/next/core#Options), as opposed to [this](https://mikro-orm.io/api/next/core/interface/MikroORMOptions))

packages/core/src/EntityManager.ts
```diff
@@ -96,6 +96,8 @@ import { TransactionManager } from './utils/TransactionManager.js';
  */
 export class EntityManager<Driver extends IDatabaseDriver = IDatabaseDriver> {
 
+  declare readonly entities?: unknown;
```
B4nan: cant we use generics instead of a declared property? if not, lets at least add `@internal` annotation so it's not publicly documented

tests/bench/types/defineEntity.ts
```diff
@@ -0,0 +1,108 @@
...
+
+bench('defineEntity with relations', () => {
+  const Foo = defineEntity({
+    name: 'Foo',
+    properties: {
+      name: p.string().primary(),
+      toOne: () => p.oneToOne(Foo),
+      toOneNullable: () => p.oneToOne(Foo).nullable(),
+      toOneRef: () => p.oneToOne(Foo).ref(),
+      // toOneRefNullable: () => p.oneToOne(Foo).ref().nullable(),
```
B4nan: why is this one commented out?

tests/features/get-kysely.test.ts
```diff
@@ -0,0 +1,267 @@
...
+        id: p.integer().primary().autoincrement(),
+        title: p.string(),
+        content: p.text(),
+        author: () => p.manyToOne(User),
+        createdAt: p.datetime().onCreate(() => new Date()),
+      },
+    });
+
+    const Comment = defineEntity({
+      name: 'Comment',
```
B4nan, line 293: seeing all the tests here using `name`, will this also work with classes? i guess it should, lets use classes at least in one entity/test to be sure

packages/core/src/EntityManager.ts
xcfox: The purpose of this property is to preserve the original information of the Entity, so that `getKysely()` can correctly retrieve the tables.

We have already allowed users to provide a custom `EntityManager`, which conflicts with using generic type parameters.  
We should mark it as `@internal`. Since it’s intended for internal use, it would be better to name it using the [`~entities`](https://github.com/standard-schema/standard-schema#why-did-you-prefix-the-standard-property-with-) prefix, just like the convention in standard-schema. This way, it won’t interfere with IDE autocomplete suggestions.

tests/bench/types/defineEntity.ts
xcfox: During the previous performance optimization, I discovered that `toOneRefNullable` significantly impacts performance.  

To confirm this, I specifically compared the type inference speed between entities that include this property and those that do not.

tests/features/get-kysely.test.ts
xcfox: Unfortunately, a Class does not carry sufficient option information during column definition. For Classes, we can only perform simple column name transformations, and there is no information available to infer a column like `author_name: string` from a property such as `author: User`.

This is a key reason why we should more strongly recommend the `defineEntity({ name: "" })` syntax in v7.

B4nan: You could still have the EntitySchema instances in the ORM config, I just meant using a class as a backend instead of the `InferEntity` type. That should work, no?

xcfox: You reminded me—yes, that's exactly it! This approach will also deliver better type inference performance!  

I’ll need a bit of time to make the changes to support this feature.

xcfox:
Now we support the `defineEntity + class` syntax, which offers superior type inference performance.

To ensure that the code generated by `entity-generator` can correctly infer the Kysely Database, we need to make `entity-generator` always emit both `tableName` and `PrimaryKeyProp`.

B4nan:
> emit both tableName and PrimaryKeyProp

So with classes you need the symbol, `primaryKeys` option is not enough, since you skip the inference completely if a class definition is present, right?

xcfox:
> since you skip the inference completely if a class definition is present

Yes, that's exactly why using a `class` provides better type inference performance than using `name`.

`defineEntity` directly utilizes the current `class` definition without requiring complex type computations.

tests/features/get-kysely.test.ts
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    class Post {
+
+      id!: number;
+      title!: string;
+      description!: string;
+      author!: User;
+
+      [PrimaryKeyProp]?: 'id';
```
B4nan, line 187: this one shouldn't be needed, right? we need it only for nonstandard PK names (like the `user` above) or when there is a possible collision (e.g. both `id` and `uuid` props are defined)

tests/features/get-kysely.test.ts
```diff
@@ -0,0 +1,383 @@
...
+      tableName: 'posts',
+      properties: {
+        id: p.integer().primary().autoincrement(),
+        title: p.string(),
+        description: p.text(),
+        author: () => p.manyToOne(User),
+      },
+    });
+
+    const orm = MikroORM.initSync({
```
B4nan, line 229: side note, what do you think about using the constructor directly in v7?

```diff
-const orm = MikroORM.initSync({ ... });
+const orm = new MikroORM({ ... });
```

tests/features/get-kysely.test.ts
xcfox: That's right! `type Primary<T>` will make `PrimaryProperty<T>` default to using `id` as the primary key.

tests/features/get-kysely.test.ts
xcfox: I prefer using the constructor directly—it feels more natural and aligns better with OOP conventions.

B4nan:
Rebased the changes from master, and the `extends` no longer accepts class or schema references. I've tried to do my best to fix it, but I can't find the cause.

https://github.com/mikro-orm/mikro-orm/blob/v7/tests/defineEntity.test.ts#L248
https://github.com/mikro-orm/mikro-orm/blob/v7/tests/entities-schema/Author4.ts#L99

PR #6942, opened 2025-10-23
fix(core): support falsy values in scalar references

Webbly25:
Fix issue of ScalarRef throwing on `loadOrFail` when loading a falsey ('', 0, false) value.

Fix logging [Object object] if the primary key is composite.

### DIFF
--- a/packages/core/src/entity/Reference.ts
+++ b/packages/core/src/entity/Reference.ts
@@ -242,7 +242,7 @@ export class ScalarReference<Value> {
       options.failHandler ??= wrapped.__em!.config.get('findOneOrFailHandler');
       const entityName = this.entity!.constructor.name;
       const where = wrapped.getPrimaryKey();
-      const whereString = typeof where === 'object' ? JSON.stringify(where) : where;
+      const whereString = typeof where === 'object' ? inspect(where) : where;
       throw new NotFoundError(`${entityName} (${whereString}) failed to load property '${this.property}'`);
     }
 


packages/core/src/entity/Reference.ts
```diff
@@ -237,12 +237,13 @@ export class ScalarReference<Value> {
   async loadOrFail(options: Omit<LoadReferenceOrFailOptions<any, any>, 'populate' | 'fields' | 'exclude'> = {}): Promise<Value> {
     const ret = await this.load(options);
 
-    if (!ret) {
+    if (ret === undefined || ret === null) {
```
B4nan:
```suggestion
    if (ret == null) {
```

packages/core/src/entity/Reference.ts
```diff
@@ -237,12 +237,13 @@ export class ScalarReference<Value> {
...
       const wrapped = helper(this.entity!);
       options.failHandler ??= wrapped.__em!.config.get('findOneOrFailHandler');
       const entityName = this.entity!.constructor.name;
       const where = wrapped.getPrimaryKey();
-      throw new NotFoundError(`${entityName} (${where}) failed to load property '${this.property}'`);
+      const whereString = typeof where === 'object' ? JSON.stringify(where) : where;
```
B4nan: better use `util.inspect`

```suggestion
      const whereString = typeof where === 'object' ? inspect(where) : where;
```

B4nan (approved):
thanks!

PR #6954, opened 2025-11-01
feat(entity-generator): support generating `defineEntity()` definitions

B4nan:
```ts
await orm.entityGenerator.generate({
  save: true,
  path: './entities',
  entityDefinition: 'defineEntity',
  inferEntityType: true, // by default, a class definition is emitted
});
```


B4nan:
@xcfox when working on this, I realized we are most likely missing support for `EntityRepositoryType`, `EagerProps`, and `Config` symbols with the `defineEntity` + `InferEntity` combo. It's the same kind of thing as `PrimaryKeyProp`.

Also, should we emit the `InferEntity` as a type or interface? I guess an interface should be better for performance, right?

xcfox:
> Also, should we emit the InferEntity as a type or interface? I guess an interface should be better for performance, right?

For types like InferEntity, we can only use type; interface cannot handle such cases.

B4nan:
I went with the interface in the end, since that is what we have in the docs.

PR #6960, opened 2025-11-04
fix(reflection): support of literal props

jeremv42:
Fix an issue with `TsMorphMetadataProvider` when an entity's property name is a literal string or computed name like: `'prop': ... or [`prop`]: ...`


packages/reflection/src/TsMorphMetadataProvider.ts
```diff
@@ -113,7 +113,27 @@ export class TsMorphMetadataProvider extends MetadataProvider {
...
+
+      const nameNode = v.getNameNode();
+
+      if (nameNode instanceof StringLiteral && nameNode.getLiteralText() === prop.name) {
+        return true;
+      }
+
+      if (nameNode instanceof ComputedPropertyName) {
+        const expr = nameNode.getExpression();
+
```
B4nan:
```suggestion

```

PR #6967, opened 2025-11-10
feat(core): support private properties with get/set accessors

### LINKED ISSUE #6932: Virtual Property with `persist: false` hydrated on load from DB
### Describe the bug

Originally reported [in Discord](https://discord.com/channels/1214904142443839538/1215229825090977822/1428867717435232437).

When using [Virtual Properties](https://mikro-orm.io/docs/defining-entities#virtual-properties) with `{ persist: false }`, the virtual property's setter is called when reading from the database. (It seems to be because the field saved in the database has the same name as the virtual property?)

This can have unexpected side effects if your setter modifies the entity. (This marks it as dirty, and will cause it to be re-flushed to the DB, even if you didn't modify it.)

Expected behavior:
* A field that is not persisted should not be read from the database.
* Only the persisted backing field should be read from the DB.

### Reproduction

We have an entity that has a backing field `_foo` which is wrapped by a getter/setter for validation.
```ts
@Property({type: 'jsonb', fieldName: "foo"})
protected _foo: unknown

@Property({ persist: false })
get foo(): Foo {
    // Zod validation, check DB/code types are in sync.
    return Foo.parse(this._settings) 
}

set foo(untrusted: Foo) {
    // Check/trim runtime type before storing:
    this._foo = Foo.parse(untrusted)
}
```

When we read these values in a `transactional()` block, the setter gets called by MikroORM, which causes all records to be marked as dirty:

```ts
await entityManager.transactional(async em => {
    const rows = await em.find(OurEntity, /* WHERE */)

   // Code path that does does not update any records
})
```

This causes all records to be written back to the database unnecessarily. You can see this by enabling debug logging. (Or, in our case, the entity also includes a `version` field, which gets incremented unnecessarily.)

We can validate this double-hydration by updating the setter to something like:
```ts
set foo(untrusted: Foo) {
    if (this._foo !== undefined) { throw new Error(`attempting to re-hydrate value?`) }
    this._foo = Foo.parse(untrusted)
}
```

Here's a stack trace from my non-minimal reproduction showing MikroORM trying to hydrate the virtual field. (My field is called `settings` instead of `foo`.)
```ts
2025-10-17 15:00:10 - [ERROR] - Error: attempting to re-hydrate value?
    at set settings (/Users/codycasterline/code/core-server/apps/merchants/dist/src/quick-actions/entities/segmentation.entity.js:61:19)
    at eval (eval at createFunction (/Users/codycasterline/code/core-server/node_modules/.pnpm/@mikro-orm+core@6.4.7/node_modules/@mikro-orm/core/utils/Utils.js:920:20), <anonymous>:83:21)
    at Utils.callCompiledFunction (/Users/codycasterline/code/core-server/node_modules/.pnpm/@mikro-orm+core@6.4.7/node_modules/@mikro-orm/core/utils/Utils.js:931:20)
    at ObjectHydrator.hydrate (/Users/codycasterline/code/core-server/node_modules/.pnpm/@mikro-orm+core@6.4.7/node_modules/@mikro-orm/core/hydration/ObjectHydrator.js:25:23)
    at EntityFactory.hydrate (/Users/codycasterline/code/core-server/node_modules/.pnpm/@mikro-orm+core@6.4.7/node_modules/@mikro-orm/core/entity/EntityFactory.js:238:27)
    at EntityFactory.create (/Users/codycasterline/code/core-server/node_modules/.pnpm/@mikro-orm+core@6.4.7/node_modules/@mikro-orm/core/entity/EntityFactory.js:84:18)
    at SqlEntityManager.find (/Users/codycasterline/code/core-server/node_modules/.pnpm/@mikro-orm+core@6.4.7/node_modules/@mikro-orm/core/EntityManager.js:152:45)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async /Users/codycasterline/code/core-server/apps/merchants/dist/src/settings/segments/segments.service.js:91:30
    at async fork.getConnection.transactional (/Users/codycasterline/code/core-server/node_modules/.pnpm/@mikro-orm+core@6.4.7/node_modules/@mikro-orm/core/EntityManager.js:968:29)
```

--- 
Attempted workaround:

I noticed there's a [hydrate](https://mikro-orm.io/api/core/interface/PropertyOptions#hydrate) option for properties, but enabling that also disabled the setter from being evaluated during `em.create()` which defeats
 my purpose for using the setter. Plus, the docs say hydrate is "useful for persisted getters" and I don't want to persist this getter!

@mikro-orm/postgresql

### MikroORM version

6.4.7-6.5.8

### Node.js version

23.7.0

### Operating system

macOS 26.0.1

### Validations

B4nan:
When using a private property backed by a public get/set pair, use the `accessor` option to point to the other side.

> The `fieldName` will be inferred based on the accessor name unless specified explicitly.

If the `accessor` option points to something, the ORM will use the backing property directly:

```ts
@Entity()
class User {
  // the ORM will use the backing field directly
  @Property({ accessor: 'email' })
  private _email: string;

  get email() {
    return this._email;
  }

  set email(email: string) {
    return this._email;
  }
}
```

If you want to the ORM to use your accessor internally too, use `accessor: true` on the get/set property instead. This is handy if you want to use a **native private property** for the backing field. 

```ts
@Entity({ forceConstructor: true })
class User {
  #email: string;

  // the ORM will use the get/set field internally
  @Property({ accessor: true })
  get email() {
    return this.#email;
  }

  set email(email: string) {
    return this.#email;
  }
}
```

Closes #6932

### DIFF
--- a/docs/docs/defining-entities.md
+++ b/docs/docs/defining-entities.md
@@ -1538,6 +1538,244 @@ const populatedReport = await em.populate(report, ['reportParameters']);
 console.log(populatedReport.reportParameters.$); 
 ```
 
+## Private property accessors
+
+When using a private property backed by a public get/set pair, use the `accessor` option to point to the other side.
+
+> The `fieldName` will be inferred based on the accessor name unless specified explicitly.
+
+If the `accessor` option points to something, the ORM will use the backing property directly:
+
+<Tabs
+groupId="entity-def"
+defaultValue="reflect-metadata"
+values={[
+{label: 'reflect-metadata', value: 'reflect-metadata'},
+{label: 'ts-morph', value: 'ts-morph'},
+{label: 'defineEntity', value: 'define-entity'},
+{label: 'EntitySchema', value: 'entity-schema'},
+]
+}>
+  <TabItem value="reflect-metadata">
+
+```ts title="./entities/User.ts"
+@Entity()
+export class User {
+  @PrimaryKey()
+  id!: number;
+
+  // the ORM will use the backing field directly
+  @Property({ accessor: 'email' })
+  private _email: string;
+
+  get email() {
+    return this._email;
+  }
+
+  set email(email: string) {
+    return this._email;
+  }
+}
+```
+
+  </TabItem>
+  <TabItem value="ts-morph">
+
+```ts title="./entities/User.ts"
+@Entity()
+export class User {
+  @PrimaryKey()
+  id!: number;
+  
+  // the ORM will use the backing field directly
+  @Property({ accessor: 'email' })
+  private _email: string;
+
+  get email() {
+    return this._email;
+  }
+
+  set email(email: string) {
+    return this._email;
+  }
+}
+```
+
+  </TabItem>
+  <TabItem value="define-entity">
+
+```ts title="./entities/User.ts"
+export class User {
+  id!: number;
+  private _email!: unknown;
+
+  get email(): unknown {
+    return this._email;
+  }
+
+  set email(email: unknown) {
+    this._email = email;
+  }
+}
+
+export const UserSchema = defineEntity({
+  class: User,
+  properties: {
+    id: p.integer().primary(),
+    // the ORM will use the backing field directly
+    email: p.string().accessor('_email'),
+  },
+});
+```
+
+  </TabItem>
+  <TabItem value="entity-schema">
+
+```ts title="./entities/User.ts"
+export class User {
+  id!: string;
+  private _email!: string;
+
+  get email() {
+    return this._email;
+  }
+
+  set email(email: string) {
+    return this._email;
+  }
+}
+
+export const UserSchema = new EntitySchema({
+  class: User,
+  properties: {
+    id: { type: 'integer', primary: true },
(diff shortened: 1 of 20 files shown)


NfNitLoop:
I'm curious how the `persist` and `hydrate` options relate to the new `accessor` option. 

It seems like `accessor: 'email'` has the effect I expected in #6932 from:
* setting `fieldName` on the backing property. (accepting default: `persist: true`)
* setting `persist: false` on the getter/setter prop

And `accessor: true` allows hydration via the property.  (I didn't need this use case but … wouldn't this have happened with just a `@Property()`? I don't understand the change here.)

B4nan:
Why would you want to use those options together? The point of this PR is that you don't decorate/define two properties, just one. It provides a simple interface for this, instead of looking for some weird combination of persist/hydrate on two properties, you just use one option.

> And accessor: true allows hydration via the property. (I didn't need this use case but … wouldn't this have happened with just a @Property()? I don't understand the change here.)

A persisted getter is replaced on hydration with the value.

PR #6972, opened 2025-11-13
Relation wrongly detected as nullable for ManyToOne with cascade

nsams:
Metadata for this field is nullable=true
```
  @ManyToOne({ entity: () => Country, ref: true, cascade: [Cascade.REMOVE] })
  country!: Ref<Country>;
```

Workaround, explicitly set to false:
```
  @ManyToOne({ entity: () => Country, ref: true, cascade: [Cascade.REMOVE], nullable: false })
  country!: Ref<Country>;
```

See test in this PR for reproduction

### DIFF
--- a/tests/issues/GHXXXX.test.ts
+++ b/tests/issues/GHXXXX.test.ts
@@ -1,0 +1,50 @@
+import {
+  Cascade,
+  Entity,
+  ManyToOne,
+  MikroORM,
+  PopulateHint,
+  PrimaryKey,
+  Ref,
+} from '@mikro-orm/sqlite';
+
+@Entity()
+class Location {
+
+  @PrimaryKey()
+  id!: number;
+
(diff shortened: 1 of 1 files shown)


B4nan:
I honestly don't remember why it is like that, the code is super old.

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/metadata/MetadataDiscovery.ts#L491

I'll change it in v7, since it might be seen as breaking.

---

Anyway, do you understand what the `cascade` option is for? In your test case, it means that if you remove the owning entity, the target relation will get removed as well. People usually want the opposite direction, which is `deleteRule: 'cascade'`.

nsams:
Thank you!

---

(I'm not the developer writing this relation code, but I'll forward your comment - makes perfectly sense - thank you)

B4nan:
fixed in v7 via e2b4cba

Issue #6978, opened 2025-11-16
Mongo schema generator doesn't respect property fieldNames on index creation


dzhy89:
### Describe the bug

Mongo schema generator doesn't respect property fieldNames on index creation after recent changes, it works for `Index/Unique` decorators on `Entity` level. I will submit a PR for this issue a bit later. Thanks!

### Reproduction

Let's say we have an entity:

```
@Entity()
@Index({properties: ["name", "email"]})
class User {

  @PrimaryKey()
  _id!: ObjectId;

  @Property({fieldName: "Name"})
  name!: string;

  @Property({fieldName: "Email"})
  email!: string;

}
```

Schema generator is supposed to generate such command:
`db.getCollection('user').createIndex({ Name: 1, Email: 1 }, { unique: false });`

but it generates this:
`db.getCollection('user').createIndex({ name: 1, email: 1 }, { unique: false });`

### What driver are you using?

@mikro-orm/mongodb

### MikroORM version

6.5.9/6.6.0

### Node.js version

20

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/mikro-orm/mikro-orm/blob/master/CONTRIBUTING.md).
- [x] Read the [docs](https://mikro-orm.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/mikro-orm/mikro-orm/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mikro-orm/mikro-orm/discussions) or join our [Discord](https://discord.gg/w8bjxFHS7X).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #6979, opened 2025-11-16
fix(mongo): respect property fieldNames on index creation

dzhy89:
Closes https://github.com/mikro-orm/mikro-orm/issues/6978


B4nan:
I dont get it, how come this behaves any different from the current code? Does the test really fail without your pr?

Anyway, thanks 👍

dzhy89:
> I dont get it, how come this behaves any different from the current code? Does the test really fail without your pr?
> 
> Anyway, thanks 👍

It really fails in the current release, try it. We have faced this issue recently after upgrading from `6.5.8` to `6.5.9`.

PR #6987, opened 2025-11-18
feat(cli): support various TS loaders

B4nan:
Adds native support for various TS loaders for the TS support in the CLI:
- `swc` via `@swc-node/register`, supports metadata reflection
- `tsx`
- `jiti`
- `tsimp`

The default is `auto`, which means it goes through all those options sequentially and picks the first one available in the project dependencies.

To pick a loader explicitly, use the `tsLoader` setting in your package.json:

```json
"mikro-orm": {
  "tsLoader": "jiti"
}
```

Or override it via `MIKRO_ORM_CLI_TS_LOADER` env var.


B4nan:
@octet-stream I was checking #6333 again, and I see you deprecated `alwaysAllowTs` and `preferTs` options. I will drop `alwaysAllowTs`, it's only used to allow TS configs when the TS support is not detected (which you can override with the `preferTs` option anyway, and the detection already respects bun, which was the main reason for this option iirc). 

I am not sure if we can drop `preferTs` though, with the new `tsLoader` option you just say how to approach loading TS files, but `preferTs: false` you can force the ORM to use production builds even if TS support is available (e.g. people keep tsx in their dependencies, or they install dev deps too in production). Those are CLI settings from package.json, they wont be changed for a prod deployment, its not part of the ORM config (as opposed to `preferTs`).

Issue #2857, opened 2022-03-04
BigIntTypes are truncated to 53 bits when retrieved from SQLite database


mcpower:
**Describe the bug**
When retrieving a property with type BigIntType from the database, the number itself is truncated to 53 bits (presumably due to being converted to a JavaScript `number` before being converted to a `string`).

I have observed this with SQLite, but this problem could occur with other databases.

**To Reproduce**
1. Download [bigint.zip](https://github.com/mikro-orm/mikro-orm/files/8184888/bigint.zip), which is a minimal reproducible example using SQLite.
2. Read all five files included in the zip to ensure it is not malicious.
3. Run `npm i && npm run all`.
4. Observe that it prints out `ID from database is 1490404699006197800, real ID is 1490404699006197764`.

**Expected behavior**
The two IDs should be the same.

**Versions**

| Dependency | Version |
| - | - |
| node | 16.14.0 |
| typescript | 4.5.5 |
| mikro-orm | 5.0.4 |

mcpower:
This also occurs for properties which aren't primary keys, which isn't in the minimal reproducible example.

mcpower:
Update - it seems like [better-sqlite3 supports large integers with BigInts](https://github.com/JoshuaWise/better-sqlite3/blob/master/docs/integer.md). sqlite doesn't seem to work with large numbers.

I assume this needs to be fixed upstream in knex, so I've filed a bug (see above).

B4nan:
I don't think this is about MikroORM or knex, this is a limitation of the node-sqlite3 connector. And they apparently don't care about it, seeing this old issue about the same: https://github.com/mapbox/node-sqlite3/issues/922

Closing, this is not really to be fixed here. The issue I linked can probably give you some workarounds, haven't read it fully but apparently you can use blobs to get around it. You can also use better-sqlite driver here, if that supports it out of box (added just recently).

Btw next time please share your reproductions on GH instead of uploading zip files, so they can be easily checked (also on mobile phone).

mcpower:
Thanks for the background! better-sqlite3 does support large integers, but requires a `db.defaultSafeIntegers(true);` call. Currently, knex provides no way of actually enabling that functionality, so by extension MikroORM also can't enable that functionality.

I assume once the knex issue is fixed, adding support for this should be straightforward enough :)

> Btw next time please share your reproductions on GH instead of uploading zip files, so they can be easily checked (also on mobile phone).

[...]

PR #7003, opened 2025-11-27
fix(core): handle BigInt in date hydration

### LINKED ISSUE #7002: Bug: Date hydration fails when using safeIntegers (BigInt)
### Describe the bug

When using `safeIntegers: true` in the driver options (e.g., with `better-sqlite3`), integer columns are returned as `BigInt` types from the driver. However, MikroORM's hydration logic for `Date` properties does not seem to handle `BigInt` values, leading to incorrect parsing or runtime errors when trying to hydrate entities with `Date` fields stored as timestamps.

The `EntityComparator` (or the relevant hydration code generator) checks for `number` or `string` (ISO format), but misses a check for `bigint`, which is what `better-sqlite3` returns when `safeIntegers` is enabled.

### Reproduction

1. Use `@mikro-orm/better-sqlite`.
2. Enable `defaultSafeIntegers: true` in the MikroORM config.
3. Define an entity with a property of type `Date`.
4. Persist an entity with a date value.
5. Fetch the entity from the database.
6. The hydration fails or results in an invalid date because the value coming from the driver is a `BigInt` (e.g., `1732676400000n`), and the hydrator doesn't convert it to `Number` before creating a `Date` object or passing it to `parseDate`.

**Note on Reproduction:**
I’ve submitted a PR to Knex to add support for returning bigints with the better-sqlite3 driver.
- Relation Issues
  - Knex PR: https://github.com/knex/knex/pull/6320
  - history: https://github.com/mikro-orm/mikro-orm/issues/2857

To fully reproduce this behavior, that PR would need to be merged or applied locally.

That said, this issue shows that the Date hydration logic needs proper support for BigInt values, regardless of how the driver is configured.

@mikro-orm/better-sqlite

### MikroORM version

6.x

### Node.js version

20.x

### Operating system

macOS

### Validations

dong-jun-shin:
# Description

Fixes #7002

This PR adds support for `BigInt` values in `Date` hydration. This is particularly useful when using drivers like `better-sqlite3` with `safeIntegers: true` enabled, which returns integers as `BigInt`.

The hydration logic generated by `EntityComparator` has been updated to explicitly handle `typeof value === 'bigint'`. It now converts the `BigInt` value to a `Number` before parsing it as a date, ensuring that timestamp values returned as BigInts are correctly hydrated into Date objects.

## Dependencies

I have submitted a PR to Knex to support the option(`defaultSafeIntegers: true`) for the `better-sqlite3` driver to return bigints (see https://github.com/knex/knex/issues/5050). Testing this change will be possible once that Knex PR is resolved.

## Changes

- Updated `EntityComparator` to include a check for `bigint` type in the generated hydration code for Date properties.
- When a `bigint` is encountered, it is converted to `Number` using `Number(value)` before being processed by `parseDate`.

## Related Issue(s)

- https://github.com/mikro-orm/mikro-orm/issues/2857

## Tests

- [x] Added a test case to verify Date hydration.
- [x] Ran existing tests to ensure no regressions.

### DIFF
--- a/tests/features/embeddables/__snapshots__/GH2391-2.test.ts.snap
+++ b/tests/features/embeddables/__snapshots__/GH2391-2.test.ts.snap
@@ -11,6 +11,8 @@ exports[`onCreate and onUpdate in embeddables (GH 2283 and 2391) result mapper 1
   if (typeof result.foo_audit1_archivedAt !== 'undefined') {
     if (result.foo_audit1_archivedAt == null || result.foo_audit1_archivedAt instanceof Date) {
       ret.foo_audit1_archived = result.foo_audit1_archivedAt;
+    } else if (typeof result.foo_audit1_archivedAt === 'bigint') {
+      ret.foo_audit1_archived = parseDate(Number(result.foo_audit1_archivedAt));
     } else {
       ret.foo_audit1_archived = parseDate(result.foo_audit1_archivedAt);
     }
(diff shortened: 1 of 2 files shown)


B4nan (approved):
thanks!

PR #7010, opened 2025-11-28
perf(core): store subscribers in `Set` to deduplicate them by identity

ruscon:
ref https://github.com/mikro-orm/mikro-orm/discussions/7009

### DIFF
--- a/tests/features/event-manager/EventManager.test.ts
+++ b/tests/features/event-manager/EventManager.test.ts
@@ -1,36 +1,49 @@
-import type { MikroORM } from '@mikro-orm/core';
-import { initORMSqlite } from '../../bootstrap';
-import { Test2Subscriber } from '../../subscribers/Test2Subscriber';
+import { Entity, MikroORM, PrimaryKey } from '@mikro-orm/core';
+import { SqliteDriver } from '@mikro-orm/sqlite';
+
+@Entity()
+class User {
+
+  @PrimaryKey()
+  id!: number;
+
+}
+
+class UserSubscriber {}
 
 describe('EventManager', () => {
-
   let orm: MikroORM;
 
   beforeEach(async () => {
-    orm = await initORMSqlite();
+    orm = await MikroORM.init({
+      entities: [User],
+      subscribers: new Set([new UserSubscriber()]),
+      dbName: `:memory:`,
+      driver: SqliteDriver,
+    });
   });
 
   afterEach(async () => orm.close(true));
 
   test('should register preconfigured subscribers', async () => {
-    expect(orm.em.getEventManager().getSubscribers().size).toEqual(0);
+    expect(orm.em.getEventManager().getSubscribers().size).toEqual(1);
   });
 
-  test('should register a new subscriber', async () => {
-    orm.em.getEventManager().registerSubscriber(new Test2Subscriber());
-    expect(orm.em.getEventManager().getSubscribers().size).toEqual(1);
+  test('should register a new instance of the subscriber', async () => {
+    orm.em.getEventManager().registerSubscriber(new UserSubscriber());
+    expect(orm.em.getEventManager().getSubscribers().size).toEqual(2);
   });
 
   test('should register only one subscriber of the same instance', async () => {
-    const subscriber = new Test2Subscriber();
+    const subscriber = new UserSubscriber();
     orm.em.getEventManager().registerSubscriber(subscriber);
     orm.em.getEventManager().registerSubscriber(subscriber);
     orm.em.getEventManager().registerSubscriber(subscriber);
-    expect(orm.em.getEventManager().getSubscribers().size).toEqual(1);
+    expect(orm.em.getEventManager().getSubscribers().size).toEqual(2);
   });
 
   test('should fork with registered subscribers', async () => {
-    orm.em.getEventManager().registerSubscriber(new Test2Subscriber());
+    orm.em.getEventManager().registerSubscriber(new UserSubscriber());
 
     const em = orm.em.fork();
     expect(orm.em.getEventManager().getSubscribers()).toStrictEqual(em.getEventManager().getSubscribers());
@@ -41,10 +54,9 @@ describe('EventManager', () => {
       freshEventManager: true,
     });
 
-    em.getEventManager().registerSubscriber(new Test2Subscriber());
-    em.getEventManager().registerSubscriber(new Test2Subscriber());
+    em.getEventManager().registerSubscriber(new UserSubscriber());
     expect(em.getEventManager().getSubscribers().size).toEqual(2);
-    expect(orm.em.getEventManager().getSubscribers().size).toEqual(0);
+    expect(orm.em.getEventManager().getSubscribers().size).toEqual(1);
   });
 
   test('should fork with registered subscribers (cloneEventManager: true)', async () => {
@@ -52,9 +64,8 @@ describe('EventManager', () => {
       cloneEventManager: true,
     });
 
-    em.getEventManager().registerSubscriber(new Test2Subscriber());
-    em.getEventManager().registerSubscriber(new Test2Subscriber());
+    em.getEventManager().registerSubscriber(new UserSubscriber());
     expect(em.getEventManager().getSubscribers().size).toEqual(2);
-    expect(orm.em.getEventManager().getSubscribers().size).toEqual(0);
+    expect(orm.em.getEventManager().getSubscribers().size).toEqual(1);
   });
 });


packages/core/src/events/EventManager.ts
```diff
@@ -8,14 +8,18 @@ export class EventManager {
   private readonly listeners: { [K in EventType]?: EventSubscriber[] } = {};
   private readonly entities: Map<EventSubscriber, string[]> = new Map();
   private readonly cache: Map<number, boolean> = new Map();
-  private readonly subscribers: EventSubscriber[] = [];
```
B4nan, line 11: this would only work with class instances, but a subscriber can be a POJO too. we can only check for identity, so a Set would be enough here

ruscon: Сlasses with `Set` will not work, because the class name can be the same, but the instance of it can be different. In this case `Set` will allow 2 instance of the same class.

```ts
var a = new Error()
var b = new Error()
new Set([a, b])
```

result
```
Set(2) {
  Error at <anonymous>:1:9, 
  Error at <anonymous>:1:9
}
```

ruscon: @B4nan, we can try to check via `getSubscribedEntities()` if logically more than 1 subscriber to the same entity should not be allowed.

B4nan: We can't check for a class name for the reasons I noted above - the subscriber might not be a class, there might not be any name.

> if logically more than 1 subscriber to the same entity should not be allowed.

I don't see why we should make such a restriction, you absolutely can have many subscribers for one entity type.

ruscon: And, in theory, the same class instances should be allowed in different `em`'s.

> I don't see why we should make such a restriction, you absolutely can have many subscribers for one entity type.

So, in short, there's no point in doing this at the library level.

B4nan: We can make it a `Set`, then you would just need to make sure they are singletons on your end (or scoped to request or otherwise) if you want to deduplicate them. Otherwise, nope, I don't really see any other way.

ruscon: By the way, should `EventManager`.`entities` also be in `WeakMap`?

ruscon: Did `EventManager.subscribers` as `Set`.

By the way, should `EventManager`.`entities` be as `WeakMap`?

ruscon: > We can make it a Set, then you would just need to make sure they are singletons on your end (or scoped to request or otherwise) if you want to deduplicate them. Otherwise, nope, I don't really see any other way.

Can I add a new method `EventManager.getSubscribers()`
This will allow to check which classes/objects are already registered as subscribers in a particular `EventManager` instance.

B4nan: Sure, a getter wont hurt anything

packages/core/src/events/EventManager.ts
```diff
@@ -8,14 +8,18 @@ export class EventManager {
...
   constructor(subscribers: EventSubscriber[]) {
     subscribers.forEach(subscriber => this.registerSubscriber(subscriber));
   }
 
   registerSubscriber(subscriber: EventSubscriber): void {
-    this.subscribers.push(subscriber);
+    if (this.subscribers.has(subscriber)) {
+      return;
+    }
+
```
B4nan, line 23: this is not necessary, its implicit behaviour of the Set

packages/core/src/events/EventManager.ts
```diff
@@ -8,14 +8,18 @@ export class EventManager {
   private readonly listeners: { [K in EventType]?: EventSubscriber[] } = {};
   private readonly entities: Map<EventSubscriber, string[]> = new Map();
   private readonly cache: Map<number, boolean> = new Map();
-  private readonly subscribers: EventSubscriber[] = [];
+  private readonly subscribers: Set<EventSubscriber> = new Set();
```
B4nan:
```suggestion
  private readonly subscribers = new Set<EventSubscriber>();
```

packages/core/src/events/EventManager.ts
```diff
@@ -85,7 +89,7 @@ export class EventManager {
   }
 
   clone() {
-    return new EventManager(this.subscribers);
+    return new EventManager([...this.subscribers.values()]);
```
B4nan:
```suggestion
    return new EventManager([...this.subscribers]);
```

(the copy isnt really necessary, you could as well just mark the constructor parameter as readonly and accept any `Iterable`)

ruscon: yep, better

packages/core/src/utils/Configuration.ts
```diff
@@ -422,7 +422,7 @@ export class Configuration<D extends IDatabaseDriver = IDatabaseDriver, EM exten
       this.options.autoJoinRefsForFilters ??= false;
     }
 
-    this.options.subscribers = Utils.unique(this.options.subscribers).map(subscriber => {
+    this.options.subscribers = Utils.unique(Array.from(this.options.subscribers)).map(subscriber => {
```
B4nan: we don't need the `Utils.unique` call here now, since we dedupe via the `Set` internally

```suggestion
    this.options.subscribers = [...this.options.subscribers].map(subscriber => {
```

B4nan:
also please adjust some existing test to use this new method

packages/core/src/events/EventManager.ts
ruscon: added this back, because of `constructor(subscribers: Iterable<EventSubscriber>)`.

imo, in v7 better to rewrite to `constructor(subscribers: Set<EventSubscriber>)` and remove the check

packages/core/src/utils/Configuration.ts
ruscon: The absence of `Utils.unique` means that `this.options.subscribers` may contain the same class twice.
But since it is possible to create your own custom drivers and, in theory, a custom `EventManager`, then in order not to do BC now, I added a check back in the`EventManager.registerSubscriber` method.

packages/core/src/events/EventManager.ts
B4nan: this is not needed, the variable where we store things is a Set, the type we accept is not relevant

ruscon: the variable itself yes, but what about this ?

https://github.com/mikro-orm/mikro-orm/pull/7010/files#diff-21653b8eff930a038fb6ba021c51b959da0a183d12d583d6cd9aa7221de52f96L19-L26

```
    this.entities.set(subscriber, this.getSubscribedEntities(subscriber));
    this.cache.clear();
    Utils.keys(EventType)
      .filter(event => event in subscriber)
      .forEach(event => {
        this.listeners[event] ??= [];
        this.listeners[event]!.push(subscriber);
      });
```

ruscon: before my changes, `Utils.unique` did the job

B4nan: You add things to a `Set`. The `Set` itself will never allow duplicates. There is no point in calling `Set.has` before `Set.add` in here.

B4nan: Oh, now I see what you mean, that link didn't really render correctly. You mean this line:

```ts
this.listeners[event]!.push(subscriber);
```

Fair enough. We could also store sets in there to get around it, but I am fine with what you have.

B4nan:
(still waiting for the tests btw)

packages/core/src/events/EventManager.ts
ruscon: I rewrote it so that `Set` would be used for both `listeners` and `entities`, so that everything would be nice.

ruscon:
> (still waiting for the tests btw)

https://github.com/mikro-orm/mikro-orm/pull/7010/files#diff-61c860809f0fef226b133c73f7d2edacab8022d3877a54845141b59303f44c02R113

B4nan:
you added a new method, and you are not testing it anywhere

image

packages/core/src/events/EventManager.ts
```diff
@@ -5,27 +5,37 @@ import { EventType, EventTypeMap, type TransactionEventType } from '../enums';
...
 export class EventManager {
 
-  private readonly listeners: { [K in EventType]?: EventSubscriber[] } = {};
-  private readonly entities: Map<EventSubscriber, string[]> = new Map();
-  private readonly cache: Map<number, boolean> = new Map();
-  private readonly subscribers: EventSubscriber[] = [];
-
-  constructor(subscribers: EventSubscriber[]) {
-    subscribers.forEach(subscriber => this.registerSubscriber(subscriber));
+  private readonly listeners: Partial<Record<EventType, Set<EventSubscriber>>> = {};
```
B4nan:
```suggestion
  private readonly listeners: { [K in EventType]?: Set<EventSubscriber> } = {};
```

tests/features/event-manager/EventManager.test.ts
```diff
@@ -0,0 +1,61 @@
+import type { MikroORM } from '@mikro-orm/core';
+import { initORMMySql } from '../../bootstrap';
```
B4nan: dont use this one, its super heavy, making such a simple test run for way too long. use sqlite with a single inline entity definition

tests/features/event-manager/EventManager.test.ts
```diff
@@ -0,0 +1,60 @@
+import type { MikroORM } from '@mikro-orm/core';
+import { initORMSqlite } from '../../bootstrap';
+import { Test2Subscriber } from '../../subscribers/Test2Subscriber';
+
+describe('EventManager', () => {
+
+  let orm: MikroORM;
+
+  beforeEach(async () => {
+    orm = await initORMSqlite();
```
B4nan: please use a single, simple entity. dont just pick another existing init function. 

i really need to drop those functions, tests need to be simple, those were never supposed to be reused.

tests/features/event-manager/EventManager.test.ts
```diff
@@ -0,0 +1,60 @@
+import type { MikroORM } from '@mikro-orm/core';
+import { initORMSqlite } from '../../bootstrap';
+import { Test2Subscriber } from '../../subscribers/Test2Subscriber';
```
B4nan: and dont use this subscriber either, write one specifically for this test.

B4nan (approved):
thanks!

PR #7034, opened 2025-12-05
Documentation update: Update findOne call to use dynamic ID

connor11528:
In the code provided we wouldn't actually know the ID of the entity so updating to a more realistic scenario


B4nan:
The example is about principle, I think its fine. IMO your change doesn't make it any better, I'd say it actually makes it worse, since `jon.id` doesn't have a value by the time you are using it.

connor11528:
oh i thought since we're querying for it by id = 1 that it had an id. if the create was awaited then would it have an id? i'm just a little unclear on this section. certainly not trying to make the docs worse :)

B4nan:
No, create is a sync method, there is no point in awaiting it. You would need to flush.

PR #7040, opened 2025-12-09
fix(schema): ensure `SchemaGenerator` respects `skipTables` option when diffing a schema

ksemasinghe:
This updates the `prepareSchemaForComparison` method of `SqlSchemaGenerator` to pass down the `skipTables` option from the `schemaGenerator` config.

Prior to this change, this option was only being applied when generating the `toSchema` but not `fromSchema`.

This posed an issue if you were not using snapshots as creating a fresh schema as the target for `fromSchema` would not run through the same "skip" logic that `getTargetSchema` would when generating the `toSchema`.

This led to tables not being properly skipped and sql being generated around these tables in the generated migrations.

Discord convo [ref](https://discord.com/channels/1214904142443839538/1446299371816747018/1446539876802625628)

### DIFF
--- a/tests/features/migrations/Migrator.postgres.test.ts
+++ b/tests/features/migrations/Migrator.postgres.test.ts
@@ -464,20 +464,16 @@ test('ensureTable when the schema does not exist', async () => {
   await orm.close();
 });
 
-// The general idea here is to setup a database, run some manual schema changes and verify that a secondary diff via the
-//  Migrator is able to respect the `skipTables` option.
-// Going with this approach to narrow the scope of the snapshot we generate and use to verify that the changes are correct.
 test('respects the skipTables option when diffing schemas', async () => {
   const baseConfig = {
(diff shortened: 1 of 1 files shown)


tests/features/migrations/Migrator.postgres.test.ts
```diff
@@ -57,6 +57,11 @@ describe('Migrator (postgres)', () => {
       driver: PostgreSqlDriver,
       schema: 'custom',
       logger: () => void 0,
+      schemaGenerator: {
```
ksemasinghe: This adjustment ends up testing this change via the snapshot changes.

This _felt_ appropriate as testing coverage but im open to further suggestions here

ksemasinghe: Ignore, updated tests to isolate this config change as it seemed to cause way too much conflict.

B4nan (approved):
thanks!

PR #7058, opened 2025-12-17
feat(core): allow reusing raw query fragments

Ruby184:
### The Problem
Currently, raw fragments used as keys in object queries are not reusable because they must be cached using a generated string key and cleared after the query is executed. This causes issues when cloning and reusing the same fragment multiple times.

Related discussion: #6052

### The Solution
Use symbols instead of strings to maintain a weak reference to the fragment. When a `RawQueryFragment` is used as a key in an object query, a unique symbol is created for each usage to serve as the reference. This allows us to store the fragment using a weak reference via:

```ts
static #rawQueryReferences = new WeakMap<symbol, RawQueryFragment>();
```

This way, when the symbol is no longer referenced (i.e., no object uses it as a key), the associated `RawQueryFragment` can be automatically garbage collected.

### Downside of the Solution
There is a minor issue with the ordering of keys in an object when combining string and symbol keys (e.g., raw fragments). According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/ownKeys#return_value), when retrieving keys via `Reflect.ownKeys()`, the order is:

1. Non-negative integer indices in ascending order (as strings)
2. Other string keys in the order of property creation
3. Symbol keys in the order of property creation

This causes an issue when using `orderBy` with a mix of raw fragments (symbols) and standard string properties. For example:

```ts
await orm.em.findByCursor(Tag, {}, {
  populate: ['job'],
  first: 3,
  orderBy: {
    [raw(alias => `${alias}.created`)]: 'desc',
    job: {
      [raw(alias => `${alias}.DateCompleted`)]: 'desc',
    },
  },
});
```

The expected order is: `t0.created desc, j1.DateCompleted desc`.  
However, due to the symbol key created by the raw fragment, the actual order becomes: `j1.DateCompleted desc, t0.created desc`.

Since ordering is critical in this context, we need to disallow combining raw fragments and string keys in a single `orderBy` object. Instead, we require the use of an array of order definitions. This is a breaking change.

Under the new rule, the example above would throw an error. The correct usage would be:

```ts
await orm.em.findByCursor(Tag, {}, {
  populate: ['job'],
  first: 3,
  orderBy: [
    { [raw(alias => `${alias}.created`)]: 'desc' },
    { job: { [raw(alias => `${alias}.DateCompleted`)]: 'desc' } },
  ],
});
```

### Open Questions
- In some places, the function `isRaw(val)` is used instead of checking `val instanceof RawQueryFragment`. I assume this was originally intended to also detect `knex.Raw`, not just `RawQueryFragment`. Since we no longer allow using knex queries directly, can we replace these conditions and remove this function, preferring a direct `RawQueryFragment` check?
- Can we make the `raw` function strictly typed (it currently uses `any`) for usage as an object query key? For example, by using a return type like `RawQueryFragment & RawQueryFragmentSymbol` and adding `RawQueryFragmentSymbol` to `ObjectQuery`.

### TODOs
- [ ] Add more tests for raw fragments
- [ ] Add a replacement for `expect(RawQueryFragment.checkCacheSize()).toBe(0);` using `jest-leak-detector`
- [ ] Handle `compareValues` with `RawQueryFragment` in `EntityComparator`
- [x] Move or add a check for combined `orderBy` objects in `QueryBuilder`

### DIFF
--- a/docs/docs/upgrading-v6-to-v7.md
+++ b/docs/docs/upgrading-v6-to-v7.md
@@ -100,6 +100,15 @@ Use `qb.execute()` or `qb.getResult()` instead of awaiting it directly.
 
 This strategy should provide a good compromise between query and hydration performance. It uses joins for to-one relations, while issuing separate queries for to-many relations.
 
+## Mixed `orderBy` with string and raw keys no longer allowed
+
+The implementation of raw SQL fragments changed, they are now represented by a symbol instance rather than a string. This means the previous limitations around reusing of raw fragments are no longer valid. On the other hand, this means mixing of raw and regular string keys is no longer allowed, because there is no way for the ORM to preserve the order of object keys.
+
+```diff
+-em.findAll(User, { orderBy: { email: 'asc', [raw('...')]: 'asc' } });
++em.findAll(User, { orderBy: [{ email: 'asc' }, { [raw('...')]: 'asc' }] });
+```
+
 ## `@CreateRequestContext` requires async function
 
 The decorator converts the function to be async, now it will trigger a validation error if the function it is applied to is not async, since it could result in an ignored promise.
--- a/packages/core/src/utils/QueryHelper.ts
+++ b/packages/core/src/utils/QueryHelper.ts
@@ -226,7 +226,6 @@ export class QueryHelper {
         return this.processJsonCondition<T>(o as FilterQuery<T>, value as EntityValue<T>, [prop.fieldNames[0]] as EntityKey<T>[], platform, aliased);
       }
 
-      // TODO: add test case for customExpression condition
       if (Array.isArray(value) && !Utils.isOperator(key) && !QueryHelper.isSupportedOperator(key as string) && !(customExpression && Raw.getKnownFragment(key)!.params.length > 0) && options.type !== 'orderBy') {
         // comparing single composite key - use $eq instead of $in
         const op = composite && !value.every(v => Array.isArray(v)) ? '$eq' : '$in';
--- a/packages/core/src/utils/RawQueryFragment.ts
+++ b/packages/core/src/utils/RawQueryFragment.ts
@@ -11,12 +11,18 @@ export class RawQueryFragment {
 
   // holds a weak reference to fragments used as object keys
   static #rawQueryReferences = new WeakMap<RawQueryFragmentSymbol, RawQueryFragment>();
-  static #index = 0n;
 
-  readonly #description: string;
+  #key?: RawQueryFragmentSymbol;
 
-  constructor(readonly sql: string, readonly params: unknown[] = []) {
-    this.#description = `[raw]: ${this.sql} (#${RawQueryFragment.#index++})`;
+  constructor(readonly sql: string, readonly params: unknown[] = []) {}
+
+  get key(): RawQueryFragmentSymbol {
+    if (!this.#key) {
+      this.#key = Symbol(this.toJSON()) as RawQueryFragmentSymbol;
+      RawQueryFragment.#rawQueryReferences.set(this.#key, this);
+    }
+
+    return this.#key;
   }
 
   as(alias: string): RawQueryFragment {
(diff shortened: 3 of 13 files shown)


B4nan:
> using jest-leak-detector

Do you think we actually need this? Not sure if there is some vitest alternative, but I wouldn't care about this personally.

>  Add more tests for raw fragments

Technically, all object embeddables are JSON properties, so we have quite a lot of tests.

> In some places, the function isRaw(val) is used instead of checking val instanceof RawQueryFragment. I assume this was originally intended to also detect knex.Raw, not just RawQueryFragment. Since we no longer allow using knex queries directly, can we replace these conditions and remove this function, preferring a direct RawQueryFragment check?

I'd do the opposite, `instanceof` checks are pretty slow in JS, this is sort of a perf optimization, also it prevents weird behaviour in case multiple core packages are present (often the case in monorepos).

> Can we make the raw function strictly typed (it currently uses any) for usage as an object query key? For example, by using a return type like RawQueryFragment & RawQueryFragmentSymbol and adding RawQueryFragmentSymbol to ObjectQuery.

I don't think it's a good idea to pollute the types with `RawQueryFragmentSymbol`. Definitely not something that should be done as part of this PR. Also, I'd argue this is not better than the `any` type. You can use `raw` fragment pretty much everywhere. But feel free to do some PoC and open a draft PR, maybe I am missing something.

B4nan:
> instanceof checks are pretty slow in JS

Maybe this is no longer true, it's been quite some time I measured this. I am trying it now, and it is actually faster to go with `instanceof`, at least in a simple benchmark. In that case, I'd go for using the instanceof check inside `isRaw`, or getting rid of it completely.

Ruby184:
> Do you think we actually need this? Not sure if there is some vitest alternative, but I wouldn't care about this personally.

Essentially, it is not dependent on Jest and can be used with Vitest without any issues. It uses FinalizationRegistry to track object finalization and includes a small hack to expose and trigger gc; that’s all. I’d like to add a test so we can verify that this works correctly when reusing fragments.

> Technically, all object embeddables are JSON properties, so we have quite a lot of tests.

I’ll just try to cover some edge cases to address the missing coverage.

> I'd do the opposite, instanceof checks are pretty slow in JS, this is sort of a perf optimization, also it prevents weird behaviour in case multiple core packages are present (often the case in monorepos). Maybe this is no longer true, it's been quite some time I measured this. I am trying it now, and it is actually faster to go with instanceof, at least in a simple benchmark. In that case, I'd go for using the instanceof check inside isRaw, or getting rid of it completely.

I suppose V8 has optimized this since then, but it could still be slow in cases where we’re checking classes with a long prototype chain. I think we could use [Symbol.hasInstance](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#instanceof_and_symbol.hasinstance)
 with a symbol from the global symbol registry to prevent weird behavior in scenarios where multiple core packages are present.
 
> I don't think it's a good idea to pollute the types with RawQueryFragmentSymbol. Definitely not something that should be done as part of this PR. Also, I'd argue this is not better than the any type. You can use raw fragment pretty much everywhere. But feel free to do some PoC and open a draft PR, maybe I am missing something.

I looked that up, and yes, you might be right. It would be great if TypeScript supported this out of the box - specifically, if a class implementing Symbol.toPrimitive allowed the compiler to infer that its instances can be used as object keys - but this is a long-standing open issue: https://github.com/microsoft/TypeScript/issues/4538. For now, intersecting with `RawQueryFragmentSymbol` should work, but I still need to verify that it behaves as expected. Let’s do this in a separate PR.

B4nan:
Looks like in the realworld case it's still faster without instanceof checks, even in this simple case with no prototype chain, I see a small degradation when applied to the actual codebase.

packages/core/src/utils/EntityComparator.ts
```diff
@@ -729,7 +729,14 @@ export class EntityComparator {
...
         const idx = this.tmpIndex++;
-        context.set(`compareValues_${idx}`, (a: unknown, b: unknown) => prop.customType!.compareValues!(a, b));
+        context.set(`compareValues_${idx}`, (a: unknown, b: unknown) => {
+          if (RawQueryFragment.isKnownFragment(a) || RawQueryFragment.isKnownFragment(b)) {
+            // TODO: improve comparison of raw query fragments
+            return a === b;
+          }
+
+          return prop.customType!.compareValues!(a, b);
+        });
```
B4nan, line 738: why is this needed?

B4nan: ok, now i see. i guess we can live with this. i've removed the todo and used `Raw.getKnownFragment(a) === Raw.getKnownFragment(b)` instead to ensure we compare fragments and not fragment with symbol. otherwise, i think we should compare for identity here. i've removed the clone method as well, since i dont see why we would need it now.

B4nan: i've also reworked how the symbols are created, so instead of creating new symbol every time `toPrimitive` is invoked, its created lazily and cached, so there is always just one symbol for the fragment. this combined with the removal of the clone method should make it play nice with the comparison for identity.

i'll go ahead and merge this, we can polish the rest later, but i want to have this as part of the first RC (which i might just ship later this evening)

Ruby184: > i've also reworked how the symbols are created, so instead of creating new symbol every time `toPrimitive` is invoked, its created lazily and cached, so there is always just one symbol for the fragment. this combined with the removal of the clone method should make it play nice with the comparison for identity.
> 
> i'll go ahead and merge this, we can polish the rest later, but i want to have this as part of the first RC (which i might just ship later this evening)

I’m not sure this is correct - because by self-referencing the symbol on the Raw instance, you’re effectively preventing garbage collection. The symbol is referenced by the fragment, so it stays alive as long as the fragment does. I think this could lead to a memory leak, but I might be wrong - happy to be proven wrong. Maybe will create another PR with the tests for that.

B4nan: I don't think such a simple cycle should be a problem for GC, at least not in node, but I might be wrong as well :] If both the symbol and the fragment instance get unreachable, they should be GCed.

PR #7081, opened 2026-01-11
fix(core): respect orphanRemoval on 1:1 relations when inverse side is a PK

NikitaHerndlhofer:
# fix(core): prevent PK corruption when assigning null to 1:1 relation with orphanRemoval

## Problem

When using `em.assign()` to set a OneToOne relation to `null` inside a transaction, MikroORM throws a validation error instead of deleting the related entity via orphan removal:

```
ValidationError: Value for AttorneyAddress.attorneyProfile is required, 'null' found
```

This occurs when:

- The relation has `orphanRemoval: true`
- The inverse side's property is also the **primary key** (e.g., `primary: true` on the FK)
- The assignment happens inside `em.transactional()`

### Reproduction

```typescript
@Entity()
class AttorneyProfile {
  @PrimaryKey()
  id!: number;

  @OneToOne("AttorneyAddress", {
    mappedBy: "attorneyProfile",
    nullable: true,
    orphanRemoval: true, // Key: should remove address when set to null
  })
  address: Ref<AttorneyAddress> | null = null;
}

@Entity()
class AttorneyAddress {
  @OneToOne(() => AttorneyProfile, {
    inversedBy: "address",
    primary: true, // FK is also the PK
  })
  attorneyProfile!: Ref<AttorneyProfile>;

  [PrimaryKeyProp]?: "attorneyProfile";
}

// Fails inside transaction
await em.transactional(async (txEm) => {
  const profile = await txEm.findOne(AttorneyProfile, id, {
    populate: ["address"],
  });
  txEm.assign(profile!, { address: null });
  await txEm.flush(); // ValidationError thrown here
});
```

## Root Cause

In `EntityHelper.propagateOneToOne()`, when propagating a `null` value to the inverse side, the code was unconditionally setting the inverse property to `null`:

```typescript
if (value == null) {
  entity[prop2.name] = value; // Corrupts entity if prop2 is the PK!
}
```

When the inverse property is a primary key, this corrupts the entity's identity. During flush inside a transaction, the changeset computation encounters this entity with a `null` PK before orphan removal is processed, treating it as a new entity requiring INSERT - which fails validation.

## Solution

Add `!prop2.primary` guards before setting the inverse side to `null`:

```typescript
if (value == null && !prop2.primary) {
  entity[prop2.name] = value;
}
```

The entity will be properly deleted via orphan removal regardless - we just shouldn't corrupt its PK in the meantime.

## Why Only Inside Transactions?

The bug is timing-dependent. Inside transactions, the forked EntityManager's flush processes changesets before orphan removal is fully resolved, exposing the corrupted entity. Outside transactions, the ordering/discovery path differs enough to avoid this race condition.

## Changes

- `packages/core/src/entity/EntityHelper.ts`: Add `!prop2.primary` checks in `propagateOneToOne()` to prevent setting primary key properties to null
- `tests/issues/GH-assign-null-orphan-removal.test.ts`: Add test cases covering the bug scenario

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,11 +13,25 @@ on:
         description: Custom version (only when version=custom)
         required: false
         type: string
+      dist-tag:
+        description: NPM dist-tag (prod or next)
+        required: false
+        type: string
+        default: 'prod'
+      ref:
+        description: Git ref to publish (branch, tag, or commit SHA)
+        type: string
+        default: '6.x'
 
-  workflow_run:
-    workflows: [ 'tests' ]
-    types: [ completed ]
-    branches: [ master ]
+  workflow_call:
+    inputs:
+      ref:
+        description: Git ref to publish (branch, tag, or commit SHA)
+        required: true
+        type: string
+      dist-tag:
+        description: NPM dist-tag (prod or next)
+        type: string
 
 permissions:
   contents: write
@@ -32,7 +46,7 @@ env:
 
 jobs:
   manual-release:
-    if: github.event_name == 'workflow_dispatch'
+    if: github.event_name == 'workflow_dispatch' && inputs.dist-tag == 'prod'
     name: Release ${{ inputs.custom_version || inputs.version }}
     runs-on: ubuntu-latest
 
@@ -40,6 +54,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           fetch-depth: 0
+          ref: ${{ inputs.ref }}
           token: ${{ secrets.GH_TOKEN }}
 
       - uses: actions/setup-node@v6
@@ -81,7 +96,7 @@ jobs:
         run: yarn publish:prod --yes
 
   canary-release:
-    if: github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success'
+    if: inputs.dist-tag == 'next'
     name: Release canary
     runs-on: ubuntu-latest
 
@@ -89,6 +104,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           fetch-depth: 0
+          ref: ${{ inputs.ref }}
 
       - uses: actions/setup-node@v6
         with:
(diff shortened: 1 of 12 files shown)


tests/issues/GH7080.test.ts
```diff
@@ -0,0 +1,123 @@
...
+@Entity()
+class A {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property({ nullable: true })
+  name: string | null = null;
+
+  @OneToOne('B', {
```
B4nan:
```suggestion
  @OneToOne(() => B, {
```

packages/core/src/entity/EntityHelper.ts
```diff
@@ -270,11 +270,17 @@ export class EntityHelper {
...
+    }
+
+    // Skip setting the inverse side to null if it's a primary key - the entity will be removed via orphan removal
+    // Setting a primary key to null would corrupt the entity and cause validation errors
     if (value == null) {
       entity[prop2.name] = value as EntityValue<T>;
     } else if (prop2.mapToPk) {
       entity[prop2.name] = helper(owner).getPrimaryKey() as EntityValue<T>;
-    } else {
+    } else if (value != null) {
```
B4nan:
```suggestion
    } else {
```

packages/core/src/entity/EntityHelper.ts
```diff
@@ -270,6 +270,12 @@ export class EntityHelper {
       }
     }
 
+    if (value == null && prop.orphanRemoval && prop2.primary) {
+      return;
+    }
+
+    // Skip setting the inverse side to null if it's a primary key - the entity will be removed via orphan removal
+    // Setting a primary key to null would corrupt the entity and cause validation errors
```
B4nan:
```suggestion
    // Skip setting the inverse side to null if it's a primary key - the entity will be removed via orphan removal
    // Setting a primary key to null would corrupt the entity and cause validation errors
    if (value == null && prop.orphanRemoval && prop2.primary) {
      return;
    }

```

NikitaHerndlhofer:
Thanks @B4nan!
Lightning fast response, awesome support

That's why love OSS!

NHerndlhofer:
@B4nan Hey, when can I expect this to get released?

PR #7091, opened 2026-01-16
fix(core): normalize populate option in Collection.matching() for M:N relations

lamkashingpaul:
## Problem

When using `Collection.matching()` with a `populate` option on many-to-many relations that use pivot tables, nested relations are not properly populated. The populate option is passed as a raw `string[]` to `loadFromPivotTable()` instead of being normalized into `PopulateOptions[]` format.

### Failing Test Case

The first commit adds a test that demonstrates the issue. Without the fix, the test fails with:
```
tests/features/collection/matching-with-populate.test.ts
  × populate many-to-many relations with matching (20 ms)

  ● populate many-to-many relations with matching

    TypeError: Cannot read properties of undefined (reading 'split')

      1077 |     }
      1078 |
    > 1079 |     const relationsToPopulate = populate.map(({ field }) => field.split(':')[0]);
           |                                                                   ^
      1080 |     const toPopulate: PopulateOptions<T>[] = meta.relations
      1081 |       .filter(prop => prop.kind === ReferenceKind.ONE_TO_ONE && !prop.owner && !prop.lazy && !relationsToPopulate.includes(prop.name))
      1082 |       .filter(prop => fields.length === 0 || fields.some(f => prop.name === f || prop.name.startsWith(`${String(f)}.`)))

      at packages/knex/src/AbstractSqlDriver.ts:1079:67
          at Array.map (<anonymous>)
      at SqliteDriver.autoJoinOneToOneOwner (packages/knex/src/AbstractSqlDriver.ts:1079:42)
      at SqliteDriver.loadFromPivotTable (packages/knex/src/AbstractSqlDriver.ts:1011:27)
      at Collection.matching (packages/core/src/entity/Collection.ts:127:40)
      at Object.<anonymous> (tests/features/collection/matching-with-populate.test.ts:87:17)
```

### What I Found
`Collection.matching()` passes the populate option directly to `loadFromPivotTable()`, which expects `PopulateOptions[]` objects but receives `string[]`. Everywhere else in the codebase (like `EntityManager.find()`, there's a call to `preparePopulate()` that normalizes strings into proper objects. That normalization step is missing here.

### My Workaround
I added `em.preparePopulate()` before calling `loadFromPivotTable()`, but had to make it public (was private).

### Concerns
While I don't think exposing `preparePopulate()` as a public API is a good idea, I'm submitting this PR to show the exact problem with a failing test and provide a working starting point. Happy to revise based on your feedback!

### DIFF
--- a/tests/features/collection/matching-with-populate.test.ts
+++ b/tests/features/collection/matching-with-populate.test.ts
@@ -1,11 +1,4 @@
-import {
-  Collection,
-  Entity,
-  ManyToMany,
-  MikroORM,
-  PrimaryKey,
-  Property,
-} from '@mikro-orm/sqlite';
+import { Collection, Entity, ManyToMany, MikroORM, PrimaryKey, Property } from '@mikro-orm/sqlite';
 
 @Entity()
 class Author {
@@ -45,8 +38,6 @@ beforeAll(async () => {
   orm = await MikroORM.init({
     dbName: ':memory:',
     entities: [Author, Book],
-    debug: ['query', 'query-params'],
-    allowGlobalContext: true, // only for testing
   });
   await orm.schema.refreshDatabase();
 
@@ -92,5 +83,4 @@ test('populate many-to-many relations with matching', async () => {
   expect(books).toHaveLength(1);
   expect(books[0].title).toBe('Book 1');
   expect(books[0].authors).toHaveLength(2);
-
 });
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -2137,6 +2137,7 @@ export class EntityManager<Driver extends IDatabaseDriver = IDatabaseDriver> {
     }, [] as string[]);
   }
 
+  /** @internal */
   async preparePopulate<Entity extends object>(entityName: string, options: Pick<FindOptions<Entity, any, any>, 'populate' | 'strategy' | 'fields' | 'flags' | 'filters'>, validate = true): Promise<PopulateOptions<Entity>[]> {
     if (options.populate === false) {
       return [];


tests/features/collection/matching-with-populate.test.ts
```diff
@@ -0,0 +1,86 @@
...
+afterAll(async () => {
+  await orm.close(true);
+});
+
+test('populate many-to-many relations with matching', async () => {
+  const author1 = await orm.em.findOneOrFail(Author, 1);
+  const books = await author1.books.matching({
+    where: { title: 'Book 1' },
+    populate: ['authors'],
+  });
```
B4nan, line 81: this behaves differently in v7, and i wonder what are you expecting here. `matching` wont update the collection state, it will only give you back the items based on the filter. but you want to populate a circular relation back to the owning entity. thats rather incompatible with `matching`, since you are trying to put the entities in two states at once (`author1.books` shouldnt be initialized this way, but you populate back the author, which means it should). 

i cant say i like the v6 behaviour, since it alters the state of `author1.books` into an inconsistent way (not initialized but has items).

edit: hmm, the populate option doesnt work in v7 for this case at all

B4nan: oh ok this is about the change of default loading strategy, it works the same with joined strategy, but with the balanced one it doesnt. will try to unify this

B4nan: ok, it was about a missing `em.populate` call in the `matching` method: d9b996fff09010c88e54b6e6748e2d98eae9f113

so v7 behaves the same

lamkashingpaul: @B4nan Thanks for keeping the consistency.

To clarify my understanding: I treat `matching()` as a query shortcut. I want filtered related entities with their relations populated, without affecting collection states. I work directly with the returned entities, not by traversing collections.

I've added detailed comments in the test showing my expectations:
- Collections remain uninitialized unless explicitly requested `find()`
- This prevents circular reference issues during serialization
- I acknowledge the inconsistent state (uninitialized but with items) exists, but I don't rely on accessing items from uninitialized collections

Now that v6 and v7 behave consistently, does this approach align with your intended design, or would you prefer different behavior for matching with circular populates?

```typescript
import { Collection, Entity, ManyToMany, MikroORM, PrimaryKey, Property, wrap } from '@mikro-orm/sqlite';
@Entity()
class Author {

  @PrimaryKey()
  id!: number;

  @Property()
  name!: string;

  @ManyToMany(() => Book, book => book.authors)

}

@Entity()
class Book {

  @PrimaryKey()
  id!: number;

  @Property()
  title!: string;

  @ManyToMany({
    entity: () => Author,
    inversedBy: 'books',
    pivotTable: 'book2author',
  })
  authors = new Collection<Author>(this);

}

let orm: MikroORM;

beforeAll(async () => {
  orm = await MikroORM.init({
    dbName: ':memory:',
    entities: [Author, Book],
  });

  const author1 = orm.em.create(Author, {
    id: 1,
    name: 'Author 1',
  });
  const author2 = orm.em.create(Author, {
    id: 2,
    name: 'Author 2',
  });
  const book1 = orm.em.create(Book, {
    id: 1,
    title: 'Book 1',
  });
  const book2 = orm.em.create(Book, {
    id: 2,
    title: 'Book 2',
  });

  author1.books.add(book1, book2);
  author2.books.add(book1, book2);

  await orm.em.flush();
  orm.em.clear();
});

beforeEach(() => {
  orm.em.clear();
});

afterAll(async () => {
  await orm.close(true);
});

test('populate many-to-many relations with matching', async () => {
  const author1 = await orm.em.findOneOrFail(Author, 1);
  const books = await author1.books.matching({
    where: { title: 'Book 1' },
    populate: ['authors'],
  });

  expect(books).toHaveLength(1);
  expect(books[0].title).toBe('Book 1');
  expect(books[0].authors).toHaveLength(2);

  const book1 = books[0];
  const author1FromBook = book1.authors[0];
  expect(author1FromBook.name).toBe('Author 1');

  /**
   * While I have defined a bidirectional many-to-many relationship between Author and Book,
   * I usually work as if it's unidirectional from my starting point (Author → Book → Author).
   *
   * Therefore, when I call matching() on author.books and populate authors inside the resulting books,
   * I don't expect the authors' books collection to be initialized or populated as well,
   * unless I explicitly pass ['authors.books'] to the populate array.
   */
  expect(author1FromBook.books.isInitialized()).toBe(false);

  /**
   * Similarly, I don't expect the original author's books collection to be initialized
   */
  expect(author1.books.isInitialized()).toBe(false);

  /**
   * Even though items exist inside these collections, I never access them like below,
   * since I know I never called author1.books.init() or similar.
   * I expect them to be uninitialized and therefore not safe to access directly.
   *
   * I am not sure if this is due to the many-to-many relationship implementation details,
   * but I acknowledge this is an inconsistent state (uninitialized but with items).
   * I simply don't rely on this behavior in my code.
   */
  expect(author1FromBook.books[0]).toBeDefined();
  expect(author1.books[0]).toBeDefined();

  /**
   * I enjoy the current behavior because I treat matching() as a
   * shortcut to get related entities with filtering and population,
   * without affecting collection states (neither the original author's books
   * nor the authors' books collections inside the resulting books).
   *
   * I work directly with individual entities rather than collections,
   * so I don't mind that those collections remain uninitialized.
   * If I need to work with those collections, I initialize them explicitly.
   */
  const booksWithFind = await orm.em.find(
    Book,
    {
      title: 'Book 1',
      authors: [author1],
    },
    { populate: ['authors'] },
  );
  /**
   * This shows that regular find() queries exhibit the same behavior:
   * populated relations don't have their inverse collections initialized.
   */
  expect(booksWithFind[0].authors[0].books.isInitialized()).toBe(false);

  /**
   * Again, items exist in the uninitialized collection (same inconsistent state),
   * but I don't rely on this. I treat uninitialized collections as off-limits
   * until explicitly initialized.
   */
  expect(booksWithFind[0].authors[0].books[0]).toBeDefined();
  books[0].authors[0].name = 'Changed Author 1';
  await orm.em.flush();

  const refreshedAuthor1 = await orm.em.findOneOrFail(Author, 1);
  expect(refreshedAuthor1.name).toBe('Changed Author 1');

  /**
   * The key benefit: uninitialized collections prevent serialization issues
   * from circular references. This is why I find the current behavior useful.
   */
  expect(wrap(booksWithFind[0]).toObject()).toEqual({
    id: 1,
    title: 'Book 1',
    authors: [
      {
        id: 1,
        name: 'Changed Author 1',
        // books is uninitialized, there is no circular reference here
      },
      {
        id: 2,
        name: 'Author 2',
      },
    ],
  });

  expect(wrap(booksWithFind[0]).toObject()).toEqual(wrap(books[0]).toObject());
});
```

B4nan: It's tricky, I don't see any correct solution (outside of forking, making a properly isolated state), so I'll leave things the way they are. The assertions you provided are passing in both v6 and v7, so at least it behaves the way you expect :]

PR #7128, opened 2026-01-28
perf(core): optimize `AutoPath` and `Loaded` type instantiation costs

B4nan:
Introduce `CollectionShape<T>` and `LoadedCollectionShape<T>` helper types for matching Collection without triggering full interface evaluation. This avoids forcing TypeScript to evaluate Collection's 30+ methods.

The generic parameter allows both matching (`T extends CollectionShape`) and inference (`T extends CollectionShape<infer U>`).

Before → After (type instantiations):

- AutoPath simple property: 133,933 → 1,747 (99% reduction)
- AutoPath with Collection: 134,050 → 1,842 (99% reduction)
- Loaded with Collection: 133,458 → 947 (99% reduction)
- Loaded no populate: 659 → 597 (9% reduction)

Also adds `yarn bench:types` script and CI job to run type performance benchmarks on PRs.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -224,6 +224,38 @@ jobs:
       - name: Test
         run: yarn bench
 
+  bench-types:
+    name: Type Benchmarks
+    if: ${{ !contains(github.event.head_commit.message, '[skip ci]') }}
+    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


B4nan:
@xcfox another _huge_ perf win here

PR #7144, opened 2026-01-31
chore: allow Collection methods to narrow return types

joemckie:
The current implementation doesn't narrow the return type based on the provided predicate.

This change mimics the type used by TypeScript's `Array.filter`:

```ts
filter<S extends T>(predicate: (value: T, index: number, array: T[]) => value is S, thisArg?: any): S[];
```

### DIFF
--- a/packages/core/src/entity/Collection.ts
+++ b/packages/core/src/entity/Collection.ts
@@ -655,7 +655,15 @@ export class Collection<T extends object, O extends object = object> {
    * Returns the first element of this collection that satisfies the predicate.
    */
   find<S extends T>(cb: (item: T, index: number) => item is S): S | undefined;
+
+  /**
+   * Returns the first element of this collection that satisfies the predicate.
+   */
   find(cb: (item: T, index: number) => boolean): T | undefined;
+
+  /**
+   * Returns the first element of this collection that satisfies the predicate.
+   */
   find(cb: (item: T, index: number) => boolean): T | undefined {
     this.checkInitialized();
     let index = 0;
@@ -673,7 +681,15 @@ export class Collection<T extends object, O extends object = object> {
    * Extracts a subset of the collection items.
    */
   filter<S extends T>(cb: (item: T, index: number) => item is S): S[];
+
+  /**
+   * Extracts a subset of the collection items.
+   */
   filter(cb: (item: T, index: number) => boolean): T[];
+
+  /**
+   * Extracts a subset of the collection items.
+   */
   filter(cb: (item: T, index: number) => boolean): T[] {
     this.checkInitialized();
     const items: T[] = [];


B4nan:
thanks! please add some test for this (lets just update some existing ones, a one liner somewhere to see it actually helps)

joemckie:
@B4nan leave it with me, was just a quick off-the-cuff edit via the GitHub UI :sweat_smile:

joemckie:
@B4nan sorted. Not sure if those tests are in the right place, just let me know if you need me to move them :ok_hand:

tests/features/collection/collection-filtering.test.ts
```diff
@@ -0,0 +1,67 @@
+import { Entity, ManyToOne, OneToMany, PrimaryKey, Property, ReflectMetadataProvider } from '@mikro-orm/decorators/legacy';
+import { Collection, MikroORM, type Ref } from '@mikro-orm/sqlite';
+
+@Entity()
```
B4nan: this would have to use STI, right?

joemckie: Possibly, you probably know more than I do there :smile: I can add that in if you need

tests/features/entity-manager/EntityManager.postgre.test.ts
```diff
@@ -1250,8 +1250,7 @@ describe('EntityManagerPostgre', () => {
     expect(book.tags.find(() => false)).toBeUndefined();
 
     // filter
-    expect(book.tags.filter(tag => tag === tagRepository.getReference(tag1.id))).toEqual([tagRepository.getReference(tag1.id)]);
-    expect(book.tags.filter(() => false)).toEqual([]);
```
joemckie, line 1254: Removed this assertion as it's an invalid type now that should be caught by the typechecker

B4nan: what is invalid type? `tagRepository.getReference(tag1.id)` gives you `Tag2` which is what the collection holds

joemckie: `() => false` isn't a type predicate (was referring to the second line)

B4nan: what is invalid about it? 

your changes feel breaking if those lines dont type check.

joemckie: The first line is fine (and has been kept), but `false` isn't a valid value for that collection (it only holds items of type `BookTag2`). Is there a usage I haven't considered here?

B4nan: How come `false` is not a valid return type for a filter callback? Both of those lines return a boolean.

joemckie: I had a brain fart sorry :joy: I'll look into it tomorrow

joemckie: Nvm was a simple fix. Should be good now and I've put the assertion back

tests/features/collection/collection-filtering.test.ts
B4nan: well, its your code, idk what you expect from it. i am trying to understand what you are trying to fix here, and the test is not helping, that 1:m collection wont have anything else than SavedFile entities this way.

joemckie: This test is a weird one, I guess, because it's only asserting the type narrowing, not any database-related logic :smile: 

`Directory.files` contains items with type `SavedFile`, and the test just verifies that when `filter` is ran with a predicate (`item is JSONFile` in this case), the resulting type is `JSONFile[]`

joemckie:
@B4nan after using the lib a bit more, it looks like other collection methods could benefit from the same treatment (`find`, at least, maybe some more). I can add them in; would you like them as separate PRs or added on to this one?

B4nan:
Let's add it here.

packages/core/src/entity/Collection.ts
```diff
@@ -654,6 +654,8 @@ export class Collection<T extends object, O extends object = object> {
   /**
    * Returns the first element of this collection that satisfies the predicate.
    */
+  find<S extends T>(cb: (item: T, index: number) => item is S): S | undefined;
+  find(cb: (item: T, index: number) => boolean): T | undefined;
```
B4nan, line 662: the jsdoc need to be repeated on each overload, unfortunately.

joemckie: Ah ok, was wondering whether that was the case as IntelliSense picks it up correctly

B4nan: typedoc won't unfortunately, we need this for the API docs on the website

joemckie: No probs, good to know :smile: adjusted both methods

B4nan (approved):

PR #7152, opened 2026-02-03
fix: guard prototype.toJSON against being called on prototype itself

### LINKED ISSUE #7151: Logging EntityManager with Pino breaks subsequent queries
## Description

Logging a MikroORM `EntityManager` with Pino causes subsequent `findOne()` calls to fail with:

```
Error: You cannot call 'EntityManager.findOne()' with empty 'where' parameter
```

## Reproduction

```bash
git clone https://github.com/cjroebuck/mikro-orm-pino-bug.git
cd mikro-orm-pino-bug
npm install
npm test
```

One test passes (no logging), one fails (logs EM before query).

**Repository:** https://github.com/cjroebuck/mikro-orm-pino-bug

## Root Cause Analysis

1. MikroORM decorates entity class prototypes with `toJSON()` during discovery
2. Pino uses `safe-stable-stringify`, which calls `toJSON()` on any object it encounters
3. When logging `EntityManager`, the serializer walks the metadata graph and hits entity prototypes
4. Calling `prototype.toJSON()` invokes `EntityTransformer.toObject()` with `this = prototype`
5. This mutates the prototype in a way that affects `Utils.isEntity()` checks
6. Subsequent queries treat plain `{ id: 1 }` objects as entities via prototype chain
7. `QueryHelper.processParams()` collapses the where clause to `{}`, triggering the validation error

## Environment

- MikroORM: 6.5.x
- Pino: 9.x  
- Driver: SQLite
- Node.js: 20+

## Possible Fix

One potential fix would be to make the `toJSON` function more defensive by checking if `this` is actually an entity instance rather than the prototype itself. Alternatively, marking certain metadata properties as non-enumerable could prevent serializers from walking into them.

cjroebuck:
## Summary

When serializers like `safe-stable-stringify` (used by pino) walk an object graph, they may call `toJSON()` on prototype objects found in metadata. This triggers the lazy `__helper` getter on the prototype rather than an entity instance, corrupting it for all future entity instances.

This adds a simple guard to return an empty object if `toJSON` is called with `this === prototype`, preventing the corruption without touching the `__helper` getter.

## Test plan

- [x] Verified fix works with reproduction repo: https://github.com/cjroebuck/mikro-orm-pino-bug

Fixes #7151

### DIFF
--- a/tests/issues/GH7151.test.ts
+++ b/tests/issues/GH7151.test.ts
@@ -1,0 +1,63 @@
+import { MikroORM, defineEntity, p } from '@mikro-orm/sqlite';
+
+const User7151 = defineEntity({
+  name: 'User7151',
+  properties: {
+    id: p.integer().primary(),
+    name: p.string(),
+  },
+});
+
+let orm: MikroORM;
+
+beforeAll(async () => {
+  orm = await MikroORM.init({
+    dbName: ':memory:',
+    entities: [User7151],
+  });
+  await orm.schema.create();
+});
+
+afterAll(async () => {
+  await orm.close(true);
+});
+
+test('calling toJSON on entity prototype should not corrupt subsequent queries (GH7151)', async () => {
+  // Create and persist an entity
+  orm.em.create(User7151, { id: 1, name: 'Test User' });
+  await orm.em.flush();
+  orm.em.clear();
+
+  // Simulate what Pino's safe-stable-stringify does when it walks
+  // the EntityManager's metadata graph and finds entity prototypes.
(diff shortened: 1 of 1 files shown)


B4nan:
Looks good, can you add a test case please?

cjroebuck:
Hi @B4nan, thank you for merging. 

Sorry I was on a train when working on this issue/PR and had to get off before pushing the test and ensuring it was working correctly, looks like you have written almost the same test case, although in a better position in the repo structure (I was adding mine to the entity manager helper test). Thank you for your time on mikro-orm, we really appreciate it.

PR #7159, opened 2026-02-04
refactor: correct persist and oneToMany filtering in InferKyselyDB

xcfox:
This PR fixes two logic errors in the `InferKyselyDB` type inference helper that resulted in incorrect Kysely table definitions.

**Given Entity Definition:**
```typescript
const User = defineEntity({
  name: 'User',
  properties: {
    name: p.string().primary(),
    nonPersistent: p.string().persist(false), // Should be excluded
    posts: () => p.oneToMany(Post),           // Should be excluded (no physical column)
  },
});
```

**Before this fix:**
The inferred type incorrectly included the relation and non-persistent fields, while potentially excluding persistent ones:
```typescript
type UserTable = KyselyDB['users'];
/* 
Result included:
{
  posts: Collection<Post>; // WRONG: 1:m relations don't have columns
  nonPersistent: string;   // WRONG: marked as persist(false)
}
*/
```

**After this fix:**
The inferred type correctly reflects only physical database columns:
```typescript
type UserTable = KyselyDB['users'];
/* 
Result:
{
  name: string; // CORRECT: only physical columns remain
}
*/
```

### DIFF
--- a/packages/sql/src/typings.ts
+++ b/packages/sql/src/typings.ts
@@ -331,13 +331,10 @@ type MaybeJoinKey<TValue, TOptions> = TOptions extends { kind: 'm:1' }
 
 type UnwrapOpt<TValue> = TValue extends Opt<infer OriginalValue> ? OriginalValue : TValue;
 
-type MaybeNever<TValue, TOptions> = TOptions extends { persist: false }
-  ? never
-  : TOptions extends { kind: 'm:n' }
+type MaybeNever<TValue, TOptions> =
+  TOptions extends { persist: false } | { kind: 'm:n' | '1:m' }
     ? never
-    : TOptions extends { kind: '1:m' }
-      ? never
-      : TValue;
+    : TValue;
 
 type ExcludeNever<TMap extends Record<string, any>> = {
   [K in keyof TMap as TMap[K] extends never ? never : K]: TMap[K];


packages/sql/src/typings.ts
```diff
@@ -331,7 +331,13 @@ type MaybeJoinKey<TValue, TOptions> = TOptions extends { kind: 'm:1' }
...
 type UnwrapOpt<TValue> = TValue extends Opt<infer OriginalValue> ? OriginalValue : TValue;
 
-type MaybeNever<TValue, TOptions> = TOptions extends { persist: true } ? never : TOptions extends { kind: 'm:n' } ? never : TValue;
+type MaybeNever<TValue, TOptions> = TOptions extends { persist: false }
+  ? never
+  : TOptions extends { kind: 'm:n' }
+    ? never
+    : TOptions extends { kind: '1:m' }
+      ? never
+      : TValue;
```
B4nan: should be the same as this, right?

```suggestion
type MaybeNever<TValue, TOptions> =
  TOptions extends { persist: false } | { kind: 'm:n' | '1:m' }
    ? never
    : TValue;
```

xcfox: Yes, these two ways of writing are completely equivalent. The new way is more concise.

PR #7192, opened 2026-02-13
refactor: separate driver-specific find options

B4nan:
- Move SQL-only options (groupBy, having, lockMode, lockTableAliases, indexHint, collation, comments, hintComments) out of core FindOptions/CountOptions into SqlFindOptions, SqlCountOptions, etc. in `@mikro-orm/sql`
- Move MongoDB-only options (collation as CollationOptions, indexHint as string | Dictionary, maxTimeMS, allowDiskUse) into MongoFindOptions, MongoCountOptions, etc. in `@mikro-orm/mongodb`
- SqlEntityManager/MongoEntityManager and their repositories override find/count/stream methods to narrow the options type, so users get driver-appropriate autocomplete and type checking
- Replace runtime validation (validateSqlOptions, MongoDB string-collation guard) with compile-time type safety — passing a wrong option shape is now a type error instead of a runtime throw
- Shared base interfaces (SqlQueryExtras, MongoQueryExtras) eliminate property duplication across option variants
- lockMode (pessimistic) and lockTableAliases are SQL-only on find(); FindOneOptions.lockMode still includes LockMode.OPTIMISTIC for both drivers (versioned entity support)

### Breaking changes

- groupBy, having, lockMode, lockTableAliases, indexHint, collation, comments, hintComments removed from core FindOptions — use `SqlFindOptions` (available via `@mikro-orm/sql` and all SQL driver packages)

### DIFF
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -233,7 +233,7 @@ export class EntityManager<Driver extends IDatabaseDriver = IDatabaseDriver> {
     (options as Dictionary)._populateWhere = options.populateWhere ?? this.config.get('populateWhere');
     options.populateWhere = this.createPopulateWhere({ ...where } as ObjectQuery<Entity>, options);
     options.populateFilter = await this.getJoinedFilters(meta, options);
-    const results = await em.driver.find(entityName, where, { ctx: em.transactionContext, em, ...options });
+    const results = await em.driver.find(entityName, where, { ctx: em.transactionContext, em, ...options } as FindOptions<Entity>);
 
     if (results.length === 0) {
       await em.storeCache(options.cache, cached!, []);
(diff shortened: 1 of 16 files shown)


B4nan:
Thoughts about this @xcfox @boenrobot? I am a bit torn here. I like the outcome, but I don't like the implementation very much.

B4nan:
We don't really depend on 3rd party types here, so we could maybe introduce a combined interface in core that would be used internally. 

Realistically, we already do those things a lot, that's what you get with making the public interface more strict. I wouldn't call this a design flaw at all.

xcfox:
I roughly looked at the current implementation. Currently, there are some places inside that actively use `as` operations to bypass type checks.

To be honest, I think for complex functions, it's sufficient to ensure that the function's parameters and return values have the correct types. The variables inside the function are usually straightforward, and too many type annotations will make the internal code more verbose.

That said, perhaps we can use some techniques, such as leaving `groupBy`, `having`, `maxTimeMS` in an internal `interface InternalFindOptions`, and we can define the publicly exposed `FindOptions` as:
```ts
export interface FindOptions extends Omit<InternalFindOptions, 'groupBy' | 'having' | 'maxTimeMS'...>
```
```ts
export interface SqlFindOptions extends Omit<InternalFindOptions, 'maxTimeMS'...>
```
```ts
export interface MongoFindOptions extends Omit<InternalFindOptions, 'groupBy' | 'having'...>
```

This way, we can ensure the precision of options for each driver, while also guaranteeing the internal type checks of MikroORM.

PR #7199, opened 2026-02-16
feat(core): add `EntityName` symbol for decorator-based Kysely type inference

B4nan:
Allow decorator-based entities to opt into automatic Kysely type inference by declaring `[EntityName]?: 'MyEntity'` on the class. This parallels how `defineEntity` entities already get inferred types from `getKysely()`.


docs/docs/kysely.md
```diff
@@ -5,111 +5,238 @@ sidebar_label: Kysely
...
+
+  @Property()
+  lastName!: string; // maps to 'last_name' column
+
+}
+
+const kysely = orm.em.getKysely({
+  tableNamingStrategy: 'entity',
+  columnNamingStrategy: 'property',
+});
```
B4nan, line 56: @xcfox I wonder if we should enable those as defaults, since that is what can easily work for existing users who are likely not using `defineEntity` now

or maybe have a single boolean that would enable all the options at once? or a different method?

xcfox: > that is what can easily work for existing users who are likely not using defineEntity now

I agree that we should provide a better experience for projects using `decorator-based`, and `tableNamingStrategy: 'entity'`, `columnNamingStrategy: 'property'` is a highly recommended configuration.

I thought about this when writing `em.getKysely()`:
Kysely's philosophy is: "What you see is what you get", but `tableNamingStrategy` and `columnNamingStrategy` break this philosophy. These options make the written SQL different from the actually executed SQL. This might confuse developers who are unfamiliar with MikroORM v7 or overly familiar with Kysely. LLMs might generate project code based on the common Kysely style in their training data (`tableNamingStrategy: "table"`).

I believe we need to give developers (especially beginners and developers already familiar with Kysely) enough "ceremony" to let developers and LLMs know that this Kysely instance is different from previous Kysely instances—it is thicker and has more features.

I prefer to add an exported fixed configuration, like:
```ts
export const entityStrategyPreset = {
  tableNamingStrategy: 'entity',
  columnNamingStrategy: 'property',
  processOnCreateHooks: true,
  processOnUpdateHooks: true,
  convertValues: true,
} as const satisfies MikroKyselyPluginOptions
```

This will be very convenient to use, and the IDE will provide the complete configuration of `entityStrategyPreset`:
```ts
const kysely = orm.em.getKysely(entityStrategyPreset)
```

PR #7202, opened 2026-02-17
perf(core): improve `defineEntity` type performance

B4nan:
| Benchmark                               | Master | PR Branch | Change |
|-----------------------------------------|--------|-----------|--------|
| defineEntity with relations              | 13,144 | 6,496     | -51%   |
| defineEntity with ref and nullable       | 13,505 | 6,510     | -52%   |
| defineEntity only ref+nullable           | 11,369 | 1,844     | -84%   |
| defineEntity with relations (class)      | 12,083 | 6,966     | -42%   |
| defineEntity only ref+nullable (class)   | 9,969  | 2,325     | -77%   |
| defineEntity setClass (circular)         | 8,623  | 5,453     | -37%   |
| polymorphic embeddables                  | 3,513  | 3,167     | -10%   |

The main problem was that `<T extends boolean = true>` generics cost ~440 instantiations each. This was always a bit weird and unnecessary, we don't really need to care about using `nullable(false)` when it's the default (and if we really need to, we can think of some other way, but let's not reintroduce this unless we see a good reason).

BREAKING CHANGE:

## Property builder boolean methods no longer accept `true`

The `defineEntity` property builder methods like `nullable()`, `ref()`, `owner()`, `primary()`, `hidden()`, `array()`, `version()`, `lazy()`, and `mapToPk()` no longer accept an explicit `true` argument. Call them without arguments instead — the behavior is identical.

```diff
-name: p.string().nullable(true),
+name: p.string().nullable(),

-tags: () => p.manyToMany(BookTag).owner(true),
+tags: () => p.manyToMany(BookTag).owner(),

-id: p.integer().primary(true).autoincrement(true),
+id: p.integer().primary().autoincrement(),
```

The `persist()` and `autoincrement()` methods still accepts `false` as an argument via overloads:

```ts
// This still works:
id: p.number().autoincrement(false),
token: p.string().persist(false),
```

The `lazy()` method no longer implies `ref()`. If you want a lazy scalar property wrapped in `ScalarReference`, chain `.ref()` explicitly:

```diff
-bio: p.text().lazy(),        // previously wrapped in ScalarReference
+bio: p.text().lazy(),        // now plain lazy scalar
+bio: p.text().lazy().ref(),  // use this for ScalarReference wrapper
```

### DIFF
--- a/docs/docs/upgrading-v6-to-v7.md
+++ b/docs/docs/upgrading-v6-to-v7.md
@@ -617,6 +617,37 @@ const UserSchema2 = new EntitySchema<User>({
 });
 ```
 
+## Property builder boolean methods no longer accept `true`
+
+The `defineEntity` property builder methods like `nullable()`, `ref()`, `owner()`, `primary()`, `hidden()`, `array()`, `version()`, `lazy()`, and `mapToPk()` no longer accept an explicit `true` argument. Call them without arguments instead — the behavior is identical.
+
+```diff
+-name: p.string().nullable(true),
++name: p.string().nullable(),
+
+-tags: () => p.manyToMany(BookTag).owner(true),
++tags: () => p.manyToMany(BookTag).owner(),
+
+-id: p.integer().primary(true).autoincrement(true),
++id: p.integer().primary().autoincrement(),
+```
+
+The `persist()` and `autoincrement()` methods still accept `false` as an argument via overloads:
+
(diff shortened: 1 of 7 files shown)


B4nan:
cc @xcfox

xcfox:
Congratulations! Another huge leap in type performance!

PR #7216, opened 2026-02-26
fix(postgres): fix phantom diffs for complex check constraints

ksemasinghe:
This updates the postgres schema helper to manage some edge cases with check constraints that weren't covered prior.

1. During the schema introspection, the query that is run to get existing check constraints creates a cardinality explosion if there are multiple check constraints with the same name across tables. Adding the join condition of the table name ensures that these dupes get removed. Without this change, the diff operator would constantly detect a diff and would attempt to try fix this by recommending we run some sql to essentially just drop and recreate the check
2. `pg_get_constraintdef` appears to pretty print any contents. This mean that case statements would get formatted to include line breaks as `\n` literals. The lack of `/s` here meant the regex's `.` did not detect this, once again causing a diff to show up when comparing schemas during migrations, triggering an endless cycle of diffs.

Related discord discussion: https://discord.com/channels/1214904142443839538/1214916262308745287/threads/1476342396336996463

_full disclosure_: I asked an LLM to help with the investigation here to figure out why I was constantly getting a drift in my schemas despite applying the suggested SQL. I confirmed its investigation and verified the fix locally before creating this PR.

The changes make sense to me but let me know if there is a potentially better place to get this fix in but the areas it pointed out seem to be at the source of the diffing

### DIFF
--- a/tests/features/schema-generator/check-constraint.postgres.test.ts
+++ b/tests/features/schema-generator/check-constraint.postgres.test.ts
@@ -191,6 +191,7 @@ describe('check constraint [postgres]', () => {
 
     // Create both tables with their constraints
     let diff = await orm.schema.getUpdateSchemaSQL({ wrap: false });
+    expect(diff).not.toBe('');
     await orm.schema.execute(diff);
 
     // After creation the diff must be empty — no phantom constraint changes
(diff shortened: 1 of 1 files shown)


B4nan:
also backported to v6 via #7224

PR #7237, opened 2026-03-03
feat(logger): added support for slow query logging

### LINKED ISSUE #4695: Feature: Log long running queries

sridhar-5:
closes #4695 

- Introduced two slow query options into the Mikro ORM config [Both optional].
  - `slowQueryThreshold` - Threshold in milliseconds for logging slow queries. 
  - `slowQueryLogger` - Factory function to create a custom logger instance for slow queries.
- Introduced a new logger namespace - `slow-query` [To easily identify the slow-query logs. ]
- Added logic in the `executeQuery` method to log slow queries at the warning level if the query time exceeds the slowQuery threshold. 

This is my first contribution to Mikro ORM. I do not have a lot of experience with TypeScript either; I tried to do this while learning it. So please feel free to let me know if I missed some nuances; I'm happy to address the feedback.

### DIFF
--- a/packages/core/src/connections/Connection.ts
+++ b/packages/core/src/connections/Connection.ts
@@ -13,12 +13,14 @@ export abstract class Connection {
   protected platform!: Platform;
   protected readonly options: ConnectionOptions;
   protected readonly logger: Logger;
+  protected readonly slowQueryLogger: Logger;
   protected connected = false;
 
   constructor(protected readonly config: Configuration,
               options?: ConnectionOptions,
               protected readonly type: ConnectionType = 'write') {
     this.logger = this.config.getLogger();
+    this.slowQueryLogger = this.config.getSlowQueryLogger();
     this.platform = this.config.getPlatform();
 
     if (options) {
@@ -150,12 +152,27 @@ export abstract class Connection {
 
     try {
       const res = await cb();
+      const took = Date.now() - now;
+      const threshold = this.config.getSlowQueryThreshold();
+      
       this.logQuery(query, {
         ...context,
-        took: Date.now() - now,
+        took,
         results: Array.isArray(res) ? res.length : undefined,
         affected: Utils.isPlainObject<QueryResult>(res) ? res.affectedRows : undefined,
       });
+
+      if (threshold && took >= threshold) {
+        this.slowQueryLogger.warn('slow-query', `${query} [took ${took} ms]`, {
+          ...context,
+          took,
+          connection: {
+            type: this.type,
+            name: this.options.name || this.config.get('name') || this.options.host,
+          },
+          query,
+        });
+      }
 
       return res;
     } catch (e) {
--- a/packages/core/src/logging/DefaultLogger.ts
+++ b/packages/core/src/logging/DefaultLogger.ts
@@ -67,6 +67,7 @@ export class DefaultLogger implements Logger {
 
   isEnabled(namespace: LoggerNamespace, context?: LogContext) {
     if (context?.enabled !== undefined) { return context.enabled; }
+    if (namespace === 'slow-query') { return true; }
     const debugMode = context?.debugMode ?? this.debugMode;
 
     if (namespace === 'deprecated') {
--- a/packages/core/src/logging/Logger.ts
+++ b/packages/core/src/logging/Logger.ts
@@ -31,7 +31,7 @@ export interface Logger {
 
 }
 
-export type LoggerNamespace = 'query' | 'query-params' | 'schema' | 'discovery' | 'info' | 'deprecated';
+export type LoggerNamespace = 'query' | 'query-params' | 'schema' | 'discovery' | 'info' | 'deprecated' | 'slow-query';
 
 export interface LogContext extends Dictionary {
   query?: string;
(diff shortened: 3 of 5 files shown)


B4nan:
Thanks! Here is what claude found out, all seems valid to me, so let's address it: 

 1. Bug: Date.now() mock in the test breaks the timing logic

  The test mocks Date.now() to return only two values (0 and 5), but the production code in executeQuery calls Date.now() three times: once for now, once inside logQuery (via the took computation — wait, actually took is computed before logQuery in the
  new code, so two calls is enough for the happy path). However, there's a subtle issue: if the test setup's beforeEach (which runs flush() and clear()) triggers queries, those will also call Date.now(). The mock with only two return values will fall
  through to the real implementation for subsequent calls, making the test fragile. The mock should use mockReturnValue for the "before" state and then switch, or use a counter-based approach.

  Also, the test only covers the "slow query IS logged" path. There should be a test that verifies queries below the threshold are not logged to the slow query logger.

  ---
  2. Design: Separate logger instance is over-engineered

  The maintainer's comment says:

  "A slowQueryLogger option (same shape as loggerFactory) to allow routing slow queries elsewhere"

  The PR implements this, but the way it's wired creates unnecessary complexity:

  - A separate Logger instance is stored on both Configuration and Connection, doubling the logger footprint.
  - The slowQueryLogger is eagerly constructed even when slowQueryThreshold is not set, falling back to this.logger. This means every Configuration and every Connection now carries an extra field for a feature most users won't enable.

  Suggestion: Only construct the slow query logger when slowQueryThreshold is actually set. Consider making getSlowQueryLogger() lazy:

	getSlowQueryLogger(): Logger {
	  return this.slowQueryLogger ?? this.logger;
	}

  And only assign this.slowQueryLogger when both slowQueryThreshold and slowQueryLogger are configured. On Connection, don't store a separate field — just call this.config.getSlowQueryLogger() inline at the call site.

  ---
  3. The isEnabled hardcoding in DefaultLogger is wrong

  	if (namespace === 'slow-query') { return true; }

  This is placed in DefaultLogger.isEnabled(), which means any DefaultLogger instance will unconditionally enable the slow-query namespace — even if a user explicitly sets debugMode: ['query'] to filter namespaces. The maintainer said:

  "they always log when the threshold is set"

  The "always" refers to not requiring debug mode, not unconditionally returning true for any logger instance. The correct place for this logic is in the caller — Connection.executeQuery() already knows the threshold was exceeded, so it should pass {
  enabled: true } in the context to bypass debug mode checks, rather than hardcoding namespace behavior in DefaultLogger. This also keeps the DefaultLogger clean for custom logger implementations.

  Fix: Remove the isEnabled change in DefaultLogger, and instead pass enabled: true in the LogContext at the call site in Connection.ts:

	  this.slowQueryLogger.warn('slow-query', `...`, {
	    ...context,
	    enabled: true, // always log when threshold is exceeded
	    took,
	    // ...
	  });

  ---
  4. The slow query log message duplicates information poorly

  The log call constructs its own message string:

  	this.slowQueryLogger.warn('slow-query', `${query} [took ${took} ms]`, { ... });

  But this bypasses the existing logQuery() infrastructure which already handles query highlighting, replica info display, result count formatting, etc. The slow query log will be a plain, unhighlighted, unformatted string while the regular query log gets
   the full treatment. This is inconsistent.

  Suggestion: Use logQuery() on the slow query logger (or reuse the existing logQuery method) with the slow-query namespace context, so slow queries get the same formatting. Or at minimum, route through this.logQuery() style formatting. The maintainer
  didn't specify the exact format, but consistency with existing log output seems clearly intended.

  ---
  5. Missing: slow query logging for failed queries

  The slow query check only happens in the try success branch. A query that takes 10 seconds and then fails is arguably more important to flag as slow than one that succeeds. The catch branch should also check the threshold and log.

  ---
  6. The slowQueryLogger factory receives debugMode and other unrelated options

	  this.slowQueryLogger = slowQueryLoggerFactory({
	    debugMode: this.options.debug,
	    ignoreDeprecations: this.options.ignoreDeprecations,
	    // ...
	  });

  Passing debugMode and ignoreDeprecations to a slow query logger doesn't make sense — the whole point is that slow query logging is independent of debug mode. This should either pass a minimal options set or pass debugMode: true to ensure the logger is
  always active (though the enabled: true approach from point 3 is cleaner).

  ---
  7. Minor: slowQueryLogger option name inconsistency

  The maintainer said it should have the "same shape as loggerFactory". The existing option is named loggerFactory, so the slow query variant should probably be slowQueryLoggerFactory (not slowQueryLogger) for naming consistency. Currently logger in the
  config is the writer function (message: string) => void, while loggerFactory is the factory (options: LoggerOptions) => Logger. Naming this slowQueryLogger when it's actually a factory creates confusion.

  ---
  8. No test for threshold-not-exceeded case

  There's only one test case. At minimum, add a test verifying that when the query is faster than the threshold, the slow query logger is not called.

B4nan:
Do you need help finalizing this? I can chip in, should be pretty simple.

It would make sense to have this in v7, which I will release to stable in a few days (wednesday most likely).

B4nan:
Closing in favor of #7271

PR #7252, opened 2026-03-06
refactor(sql): clone entityMap in MikroKyselyPlugin to prevent race condition in concurrent queries

xcfox:
Fixes a race condition in `MikroKyselyPlugin` where concurrent queries could corrupt the result transformation by sharing the same `entityMap` reference.

## The Bug

When multiple queries execute concurrently, the `transformer.entityMap` is cached by **reference** in `transformQuery()`:

```ts
const entityMap = this.transformer.getOutputEntityMap();
MikroKyselyPlugin.queryNodeCache.set(args.queryId, { entityMap });
```

The transformer is shared across queries, so a subsequent query calling `reset()` clears the same Map instance before the first query's results are transformed. This causes intermittent failures in column name mapping (e.g., `role_ids` not being converted to `roleIds`).

## The Fix

Clone the `entityMap` when caching to break the shared reference:

### DIFF
--- a/packages/sql/src/plugin/index.ts
+++ b/packages/sql/src/plugin/index.ts
@@ -79,7 +79,9 @@ export class MikroKyselyPlugin implements KyselyPlugin {
       UpdateQueryNodeClass.is(args.node) ||
       DeleteQueryNodeClass.is(args.node)
     ) {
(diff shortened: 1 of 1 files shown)


B4nan (approved):
thanks!

PR #7258, opened 2026-03-07
fix(mongo): fix support for deno by handling `TypedArray` when cloning

alexandre-abrioux:
This PR fixes the usage of `@mikro-orm/mongodb` with Deno.

## Issue

Using the newly added reproduction test case in `tests/deno/mongodb.test.ts`, and running it with no additional code change results in the following error:

```
> deno run --allow-read --allow-write --allow-env --allow-net --allow-sys --node-modules-dir=manual tests/deno/mongodb.test.ts
error: Uncaught (in promise) TypeError: this is not a typed array.
    at Uint8Array.values (<anonymous>)
    at Array.from (<anonymous>)
    at Object.toHex (file:///.../mikro-orm/node_modules/bson/lib/bson.cjs:481:22)
    at ObjectId.toHexString (file:///.../mikro-orm/node_modules/bson/lib/bson.cjs:2672:37)
    at eval (eval at createFunction (file:///.../mikro-orm/packages/core/src/utils/Utils.ts:900:14), <anonymous>:13:68)
    at ChangeSetComputer.computePayload (file:///.../mikro-orm/packages/core/src/unit-of-work/ChangeSetComputer.ts:145:20)
    at ChangeSetComputer.computeChangeSet (file:///.../mikro-orm/packages/core/src/unit-of-work/ChangeSetComputer.ts:61:56)
    at UnitOfWork.findNewEntities (file:///.../mikro-orm/packages/core/src/unit-of-work/UnitOfWork.ts:780:46)
    at UnitOfWork.computeChangeSets (file:///.../mikro-orm/packages/core/src/unit-of-work/UnitOfWork.ts:634:12)
    at UnitOfWork.doCommit (file:///.../mikro-orm/packages/core/src/unit-of-work/UnitOfWork.ts:514:12)
```

## Investigation

Why is this bug appearing with Deno? The behavior is different because of [this line of code in the `bson` package](https://github.com/mongodb/js-bson/blob/0712bb15653093766a80bbd4ba104353cd4581e3/src/utils/byte_utils.ts#L69):
- with Node.js, `Buffer.prototype?._isBuffer` is `undefined` ; so `ByteUtils` is using `nodeJsByteUtils`
- with Deno, `Buffer.prototype?._isBuffer` is `true` ; so `ByteUtils` is using `webByteUtils`

The content of the `fromHex` method is different in both implementations:
- With `nodeJsByteUtils`, [it uses `Buffer.from`](https://github.com/mongodb/js-bson/blob/0712bb15653093766a80bbd4ba104353cd4581e3/src/utils/node_byte_utils.ts#L150)
- with `webByteUtils`, [it uses `Uint8Array.from`](https://github.com/mongodb/js-bson/blob/0712bb15653093766a80bbd4ba104353cd4581e3/src/utils/web_byte_utils.ts#L258)

The issue is that our `clone` function in `packages/core/src/utils/clone.ts` does not handle cloning of `Uint8Array` properly.

For instance, whit the following `Uint8Array` array:
```
Uint8Array(12) [
  105, 172, 10
]
```
Cloning it with our util function will result in:
```
Uint8Array {
  "0": 105,
  "1": 172,
  "2": 10
}
```

## Fix

In this PR, we add proper support for cloning [all `TypedArray` subtypes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray#typedarray_objects).

### DIFF
--- a/packages/core/src/utils/clone.ts
+++ b/packages/core/src/utils/clone.ts
@@ -25,6 +25,8 @@ function getPropertyDescriptor<T>(obj: T, prop: keyof T): PropertyDescriptor | n
   return null;
 }
 
+const TypedArray = Object.getPrototypeOf(Uint8Array);
+
 export function clone<T>(parent: T, respectCustomCloneMethod = true): T {
   const allParents: unknown[] = [];
   const allChildren: unknown[] = [];
@@ -81,6 +83,9 @@ export function clone<T>(parent: T, respectCustomCloneMethod = true): T {
     } else if (Buffer.isBuffer(parent)) {
       child = Buffer.allocUnsafe(parent.length);
       parent.copy(child as Buffer);
+      return child;
+    } else if (parent instanceof TypedArray) {
+      child = parent.copyWithin(0);
       return child;
     } else if (parent instanceof Error) {
       child = new (parent as any).constructor(parent.message);
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -351,10 +351,15 @@ jobs:
         with:
           cache: yarn
 
-      - name: Install & build
+      - name: Install & build & pull docker images
         run: |
+          docker compose pull mongo &
           yarn
           yarn build
+          wait
+
+      - name: Init docker
+        run: docker compose up -d mongo --wait
 
       - name: Setup Deno
         uses: denoland/setup-deno@v2
@@ -363,6 +368,9 @@ jobs:
 
       - name: Test (Deno + node:sqlite)
         run: deno run --allow-read --allow-write --allow-env --allow-ffi --node-modules-dir=manual tests/deno/sqlite.test.ts
+
+      - name: Test (Deno + mongodb)
+        run: deno run --allow-read --allow-write --allow-env --allow-net --allow-sys --node-modules-dir=manual tests/deno/mongodb.test.ts
 
   release:
     name: Release
--- a/tests/Utils.win.test.ts
+++ b/tests/Utils.win.test.ts
@@ -175,6 +175,19 @@ describe('Utils', () => {
       expect(copied).toBeInstanceOf(Child);
       expect(copied.name).toBe(expected);
     });
+
+    describe('should copy TypedArray', () => {
+      test('should copy Uint8Array', () => {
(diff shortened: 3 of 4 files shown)


packages/core/src/utils/clone.ts
```diff
@@ -25,6 +25,8 @@ function getPropertyDescriptor<T>(obj: T, prop: keyof T): PropertyDescriptor | n
   return null;
 }
 
+const TypedArray = Object.getPrototypeOf(Uint8Array);
```
alexandre-abrioux, line 28: We can retrieve the `TypedArray` prototype by getting the prototype of any `TypedArray` subtype, like `Uint8Array` ([see MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray#description)).

B4nan:
huh, did I forget to add formatting check to the CI?

please run `yarn format`

edit: and of course, thanks!

alexandre-abrioux:
@B4nan Sorry I missed that! Done ✅

I've also added CI support for the new Deno test. I've tried to make it as light as possible by only pulling and starting the `mongo` service.

Thanks for the review!

packages/core/src/utils/clone.ts
```diff
@@ -82,6 +84,9 @@ export function clone<T>(parent: T, respectCustomCloneMethod = true): T {
       child = Buffer.allocUnsafe(parent.length);
       parent.copy(child as Buffer);
       return child;
+    } else if (parent instanceof TypedArray) {
+      child = parent.copyWithin(0);
```
B4nan, line 88: claude flags this as wrong, since it doesnt actually clone anything. his suggestion is this instead:

```suggestion
      child = parent.slice();
```

i should have paid more attention to this, `copyWithin` mutates, it doesnt create a new copy, any reason you used that? will address this as part of #7256

alexandre-abrioux: You're right, my bad 😬 Sorry, I don't manipulate `TypedArray` often, so I was looking at a method to clone it and misunderstood the docs of `copyWithin`. Indeed, `slice` is more appropriate, thanks for catching it!

PR #7276, opened 2026-03-11
docs: remove string references from entity definitions examples

phxgg:
String references are no longer supported in v7.


B4nan (approved):
thanks! ignore the formatting error, that's from the last release that commits to master without formatting, will fix that tomorrow

PR #7301, opened 2026-03-14
fix(core): apply property `default` values when creating new entity instances

B4nan:
## Summary
- Apply the `default` property option as runtime value during `em.create()` and during flush for manually constructed entities (`new Entity()` + `persist()`)
- Raw SQL defaults (e.g. `raw('current_timestamp')`) are excluded — those remain database-level only
- `onCreate` hooks take precedence over `default` values; user-provided values are never overridden

## Changes
- **EntityFactory.assignDefaultValues()** — after `onCreate` hooks, apply `prop.default` for properties still `null`/`undefined`
- **ChangeSetComputer.processPropertyInitializers()** — same logic during flush for CREATE changesets
- **Tests** — covers `em.create()` immediate application, no-override of user values, flush-time application, and raw SQL exclusion

Closes #7284

## Test plan
- [x] `em.create()` applies scalar `default` values immediately
- [x] `em.create()` does not override user-provided values
- [x] `new A()` + `persist()` + `flush()` applies defaults during flush
- [x] `raw()` SQL defaults are NOT applied in-memory (left to database)
- [x] Existing default-values tests pass across all drivers (sqlite, mysql, postgres)
- [x] EntityFactory and EntityManager.sqlite test suites pass

### DIFF
--- a/tests/features/default-values/default-values.sqlite.test.ts
+++ b/tests/features/default-values/default-values.sqlite.test.ts
@@ -1,10 +1,12 @@
-import { MikroORM, raw } from '@mikro-orm/core';
+import { MikroORM, OptionalProps, raw } from '@mikro-orm/core';
 import { Entity, PrimaryKey, Property, ReflectMetadataProvider } from '@mikro-orm/decorators/legacy';
 import { mockLogger } from '../../helpers.js';
 import { SqliteDriver } from '@mikro-orm/sqlite';
 
 @Entity()
 class A {
+  [OptionalProps]?: 'foo1' | 'foo2' | 'foo3' | 'bar' | 'rawDefault' | 'version';
+
   @PrimaryKey()
   id!: number;
 


packages/core/src/entity/EntityFactory.ts
```diff
@@ -419,8 +420,22 @@ export class EntityFactory {
 
   private assignDefaultValues<T extends object>(entity: T, meta: EntityMetadata<T>): void {
     for (const prop of meta.props) {
+      if (prop.embedded || [ReferenceKind.MANY_TO_ONE, ReferenceKind.ONE_TO_ONE].includes(prop.kind)) {
+        continue;
+      }
+
       if (prop.onCreate) {
         entity[prop.name] ??= prop.onCreate(entity, this.#em);
+      } else if (prop.default != null && !isRaw(prop.default) && entity[prop.name] === undefined) {
```
vaskouk, line 429: Is this also covering when default(null) or this is already covered by .nullable?

B4nan: Why would you use `default(null)`?

vaskouk: Yes doesn't makes sense that's why I am re validating the logic.

B4nan: I don't follow, so you agree your question makes no sense, or what? :D You should not use `p.default(null)`, nullable properties default to null implicitly, it should be a no-op. I recall someone trying to use that approach to fix some type-level issue, which I hope is no longer valid.

vaskouk: Yes I am answering to my non valid question 😂
I also had default(null) which was incorrect, just using nullable is enough.

PR #7312, opened 2026-03-15
fix(decorators): polyfill metadata symbol if not available

DASPRiD:
Polyfill `Symbol.metadata` for runtimes that do not support it natively (e.g. current Node.js).

TypeScript's ES-decorator output uses `Symbol.metadata` as the key to share one metadata object across all decorators in a class:

```js
const _metadata = typeof Symbol === "function" && Symbol.metadata
  ? Object.create(null)
  : void 0;
```

When the symbol is absent, the runtime sets `_metadata = void 0` and every decorator context receives `metadata: undefined`, making it impossible to propagate field-level metadata to the class decorator. Installing the symbol first ensures `_metadata` is always a proper object.

### DIFF
--- a/packages/decorators/src/es/index.ts
+++ b/packages/decorators/src/es/index.ts
@@ -1,17 +1,7 @@
 // Polyfill Symbol.metadata for runtimes that do not support it natively
-// (e.g. current Node.js).
-//
-// TypeScript's ES-decorator output uses `Symbol.metadata` as the key to share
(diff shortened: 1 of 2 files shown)


DASPRiD:
I tried coming up with a test case which could be included, but this would rather require an integration test with already compiled code, so running it purely through NodeJS.

I did verify that this fix properly resolves the issues though in one of my projects.

PR #7314, opened 2026-03-15
feat(core): add `strictNullable()` property builder

B4nan:
## Summary
- Adds `strictNullable()` method to the `defineEntity` property builder that produces `T | null` without `| undefined`
- The existing `nullable()` is unchanged (`T | null | undefined`) for backwards compatibility

## Motivation
When using `defineEntity` with class extension and constructors, properties are always explicitly initialized — they're never `undefined`. Having `| undefined` in the type forces unnecessary null/undefined checks downstream. `strictNullable()` lets users express "this can be null but is always set" ([Discussion #7307](https://github.com/mikro-orm/mikro-orm/discussions/7307)).

### DIFF
--- a/packages/core/src/entity/defineEntity.ts
+++ b/packages/core/src/entity/defineEntity.ts
@@ -66,7 +66,7 @@ export type UniversalPropertyKeys =
   | keyof OneToOneOptions<any, any>
   | keyof ManyToManyOptions<any, any>;
 
-type BuilderExtraKeys = '~options' | '~type' | '$type';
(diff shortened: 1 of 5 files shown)


stoneLeaf:
Just a side note, I have the opposite use case, my config being as follow:
```TypeScript
  ...
  forceUndefined: true,
  ...
```

And thus `T | null | undefined` does not reflect reality, at least in the DB -> JS direction.

B4nan:
`forceUndefined` is an ORM config option, those have no effect on type level.

PR #7322, opened 2026-03-16
docs: install CLI as a dev dependency

adriancuadrado:
Updated installation command for MikroORM CLI to use --dev flag.


B4nan:
I am still not entirely sold on the need for this, the CLI doesn't have to be only dev dep. Andmore importantly, its a thin dependency on its own.

But since you are already the second person suggesting it, lets just merge it, it wont make things worse.

PR #7331, opened 2026-03-17
fix(sql): fix populating TPT relations with embeddables or custom types in child entities

### LINKED ISSUE #7329: Populating a TPT relationship fails if one of the entities has an embeddable
### Describe the bug

When populating a relationship targeting a TPT:
```TypeScript
defineEntity({
  name: "Room",
  properties: {
    id: p.integer().primary(),
    assignedTo: () => p.manyToOne(Person).ref(),
  },
})
```

With one of the entity from the TPT tree having an embeddable:
```TypeScript
defineEntity({
  name: "Employee",
  extends: Person, // Extends the parent TPT entity
  properties: {
    department: p.string(),
    address: p
      .embedded(AddressEmbeddable)
      .onCreate(() => new AddressEmbeddable()),
  },
})
```

Trying to populate the relationship:
```TypeScript
  const room = await this.em.findOne(
    Room,
    { id },
    { populate: ["assignedTo"] },
  )
```

Fails with the follow error: `InvalidFieldNameException: no such column: e3.address`
The query looks as follow:
```SQL
    select `r0`.*,
      `m2`.`team_size` as `m2__team_size`,
      `e3`.`department` as `e3__department`,
      `e3`.`address` as `e3__address`,                  <--- ISSUE HERE
      `e3`.`address_street` as `e3__address_street`,
      ...
```

The embeddable name `address` is injected into the query as a field, which causes the error. In that case, only the prefixed property `address_street` of the embeddable should be selected.

See repro for full details.

### Reproduction

https://github.com/stoneLeaf/mikro-orm-v7-repros/tree/tpt-populate-issue-with-embeddable

@mikro-orm/postgresql

### MikroORM version

7.0.2

### Node.js version

22.22.1

### Operating system

Win

### Validations

B4nan:
## Summary
- Fix `no such column` error when populating a relation targeting a TPT base class where a child entity has an inline embeddable or custom type property
- The old code directly iterated `prop.fieldNames` to build SELECT columns for TPT child tables, which breaks for inline embeddables (they expand into multiple physical columns like `address_street`, `address_city`) and skips `convertToJSValueSQL` wrapping for custom types
- Delegate to `mapPropToFieldNames()` which already handles embeddable expansion, custom type conversion, and formula properties correctly

Closes #7329

### DIFF
--- a/packages/sql/src/AbstractSqlDriver.ts
+++ b/packages/sql/src/AbstractSqlDriver.ts
@@ -2251,13 +2251,12 @@ export abstract class AbstractSqlDriver<
       qb.addPropertyJoin(childMeta.tptInverseProp!, baseAlias, childAlias, JoinType.leftJoin, `[tpt]${meta.className}`);
 
       // Add fields from this child (only ownProps, skip PKs)
-      for (const prop of childMeta.ownProps!.filter(p => !p.primary && this.platform.shouldHaveColumn(p, []))) {
-        for (const fieldName of prop.fieldNames) {
-          const field = `${childAlias}.${fieldName}`;
-          const fieldAlias = `${childAlias}__${fieldName}`;
-          fields.push(raw(`${this.platform.quoteIdentifier(field)} as ${this.platform.quoteIdentifier(fieldAlias)}`));
-        }
-      }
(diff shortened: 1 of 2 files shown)


stoneLeaf:
@B4nan Thanks for the fix!

Issue #7337, opened 2026-03-18
Stop using `async` when generating TypeScript migrations


enduity:
Thanks for your work on this project. It's definitely the coolest ORM I have ever seen 🙂

### Is your feature request related to a problem? Please describe.

When I generate a migration with MikroORM (using the default `TSMigrationGenerator`),
the new file triggers the ESLint rule [`@typescript-eslint/require-await`](https://typescript-eslint.io/rules/require-await/). This rule is included in `tseslint.configs.recommendedTypeChecked`, making it widely used.

While the rule is not catching any real bugs in this case, it's not falsely triggered either – as far as I am aware, there is no situation where [`TSMigrationGenerator`](https://github.com/mikro-orm/mikro-orm/blob/53a84dacfa38177fff34cb09053e1b6ee7450347/packages/migrations/src/TSMigrationGenerator.ts) adds an `await` expression to `up()` or `down()`. Additionally, since e.g. the `up()` method is [typed](https://github.com/mikro-orm/mikro-orm/blob/53a84dacfa38177fff34cb09053e1b6ee7450347/packages/migrations/src/Migration.ts#L24) to return `Promise<void> | void`, the migration runner should therefore accept a synchronous function just fine.

### Describe the solution you'd like

Please consider updating [`TSMigrationGenerator.generateMigrationFile`](https://github.com/mikro-orm/mikro-orm/blob/53a84dacfa38177fff34cb09053e1b6ee7450347/packages/migrations/src/TSMigrationGenerator.ts#L8-L24) to remove the `async` keyword from the generated methods and set their return type to a plain `void`.

### Describe alternatives you've considered

I am aware I can make a custom [`MigrationGenerator`](https://mikro-orm.io/docs/migrations#using-custom-migrationgenerator) for my project. However, I thought that an upstream change could also benefit others. I also wasn't 100% sure that removing `async` wouldn't break anything.

### Additional context

Maybe the keyword is intentional? If yes, perhaps this should just be documented somehow.

I considered that this could be to make the user aware `async` functions are supported by the migrator. In that case, my counterargument is that this can already be inferred from the `Migration` abstract class type definitions; it's not necessary to add this keyword to every migration.

B4nan:
PR welcome, this should be a trivial change 🙃

enduity:
> PR welcome, this should be a trivial change 🙃

[...]

PR #7338, opened 2026-03-18
fix(migrations): stop making up() and down() async unnecessarily

enduity:
Make generated migration `up()` and `down()` methods synchronous. There is no need for them to be `async`, but they trigger [`@typescript-eslint/require-await`](https://typescript-eslint.io/rules/require-await/) and could cause unexpected behaviour in some cases.

Closes https://github.com/mikro-orm/mikro-orm/issues/7337

### DIFF
--- a/tests/features/migrations/__snapshots__/GH7185.test.ts.snap
+++ b/tests/features/migrations/__snapshots__/GH7185.test.ts.snap
@@ -5,11 +5,11 @@ exports[`GH7185 [mariadb] > multiline comments are not split in migration SQL 1`
 
 export class Migration20250101000000 extends Migration {
 
-  override up(): void {
+  override up(): void | Promise<void> {
     this.addSql(\`create table \\\`test_entity\\\` (\\\`id\\\` int unsigned not null auto_increment primary key comment 'line1;\\\\nline2;\\\\nline3', \\\`name\\\` varchar(255) not null comment 'line1;\\\\nline2;\\\\nline3') default character set utf8mb4 engine = InnoDB comment = 'line1;\\\\nline2;\\\\nline3';\`);
   }
 
-  override down(): void {
+  override down(): void | Promise<void> {
     this.addSql(\`drop table if exists \\\`test_entity\\\`;\`);
   }
 
@@ -22,7 +22,7 @@ exports[`GH7185 [mssql] > multiline comments are not split in migration SQL 1`] 
 
 export class Migration20250101000000 extends Migration {
 
-  override up(): void {
+  override up(): void | Promise<void> {
     this.addSql(\`create table [test_entity] ([id] int identity(1,1) not null primary key, [name] nvarchar(255) not null);\`);
     this.addSql(\`if exists(select * from sys.fn_listextendedproperty(N'MS_Description', N'Schema', N'dbo', N'Table', N'test_entity', null, null))
   exec sys.sp_updateextendedproperty N'MS_Description', N'line1;
@@ -50,7 +50,7 @@ line2;
 line3', N'Schema', N'dbo', N'Table', N'test_entity', N'Column', N'name';\`);
   }
 
-  override down(): void {
+  override down(): void | Promise<void> {
     this.addSql(\`if object_id('[test_entity]', 'U') is not null drop table [test_entity];\`);
   }
 
@@ -63,11 +63,11 @@ exports[`GH7185 [mysql] > multiline comments are not split in migration SQL 1`] 
 
 export class Migration20250101000000 extends Migration {
 
-  override up(): void {
+  override up(): void | Promise<void> {
     this.addSql(\`create table \\\`test_entity\\\` (\\\`id\\\` int unsigned not null auto_increment primary key comment 'line1;\\\\nline2;\\\\nline3', \\\`name\\\` varchar(255) not null comment 'line1;\\\\nline2;\\\\nline3') default character set utf8mb4 engine = InnoDB comment = 'line1;\\\\nline2;\\\\nline3';\`);
   }
 
-  override down(): void {
+  override down(): void | Promise<void> {
     this.addSql(\`drop table if exists \\\`test_entity\\\`;\`);
   }
 
@@ -80,7 +80,7 @@ exports[`GH7185 [postgresql] > multiline comments are not split in migration SQL
 
 export class Migration20250101000000 extends Migration {
 
-  override up(): void {
+  override up(): void | Promise<void> {
     this.addSql(\`create table "test_entity" ("id" serial primary key, "name" varchar(255) not null);\`);
     this.addSql(\`comment on table "test_entity" is 'line1;
 line2;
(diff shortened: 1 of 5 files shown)


enduity:
I'm not sure about the type of work, but `fix` at least seemed to fit better than `feature`.

packages/migrations-mongodb/src/TSMigrationGenerator.ts
```diff
@@ -8,14 +8,14 @@ export class TSMigrationGenerator extends MigrationGenerator {
   generateMigrationFile(className: string, diff: { up: string[]; down: string[] }): string {
     let ret = `import { Migration } from '@mikro-orm/migrations-mongodb';\n\n`;
     ret += `export class ${className} extends Migration {\n\n`;
-    ret += `  async up(): Promise<void> {\n`;
+    ret += `  up(): void {\n`;
```
B4nan: what about `void | Promise<void>`, would that work for the linter?

mongo migrations will be always async, so I would rather go that way, so its clear to users they can use async too

B4nan: or actually, since we have a distinct mongo generator, lets just revert this and emit the `async` functions for mongo, only change the SQL variant

enduity: Ah, I didn't notice that the Mongo type is different, sorry. I figured that since there are so many tests, it must be fine 😄 The generated migrations would have actually been broken for Mongo (type errors).

I removed this change from Mongo migrations:
- e8f70ac4 (edit: after force push, accessible only thru https://github.com/enduity/mikro-orm_fork/commit/e8f70ac4b17eaf3541c36a3778dc21766f213201)

I also semi-vibe-coded hacky tests that would've caught this error:
- 1618c741

The tests run type checks on the generated migration files.

enduity: > what about `void | Promise<void>`, would that work for the linter?

Worked for me. However, to me it feels weird to include `Promise<void>` if it doesn't match reality. But it's your call, of course. I think either way is ultimately fine.

tests/features/migrations/Migrator.mongo.test.ts
```diff
@@ -115,6 +117,41 @@ describe('Migrator (mongo)', () => {
...
+
+    try {
+      execSync(`npx tsc --project ${tsConfigPath}`, { stdio: 'pipe' });
+    } catch (e: any) {
+      const output = `${e.stdout?.toString()}\n${e.stderr?.toString()}`;
+      if (output.includes(migration.fileName)) {
+        throw new Error(`TypeScript errors in generated migration:\n${output}`);
+      } else {
+        throw new Error(`Unexpected error during TypeScript compilation:\n${output}`);
+      }
```
B4nan: IMO this is not necessary, not to mention it should have been an assertion that fails, not an error that is thrown. lets just drop it and check the snapshot as other tests do, no need to execute the migration (surely not by creating some dynamic tsconfig and using tsx)

tests/features/migrations/Migrator.test.ts
```diff
@@ -4,6 +4,9 @@ import { Migration, MigrationStorage, Migrator } from '@mikro-orm/migrations';
 import { DatabaseTable, MySqlDriver, DatabaseSchema, SchemaGenerator } from '@mikro-orm/mysql';
 import { rm } from 'node:fs/promises';
 import { initORMMySql, mockLogger } from '../../bootstrap.js';
+import {execSync} from 'node:child_process'
+import {writeFile} from 'node:fs/promises'
+import path from 'node:path'
```
B4nan: also run formatter and linter before you commit, this will fail the CI

packages/migrations-mongodb/src/TSMigrationGenerator.ts
B4nan: Why would you say that? The method can return `Promise<void>`.

tests/features/migrations/Migrator.mongo.test.ts
B4nan: in fact i dont even think we need new tests here, existing tests already use snapshots, to the change will be tested/visible there

tests/features/migrations/Migrator.test.ts
enduity: Sorry, used to commit hooks and editor taking over. And CI running automatically 😅 

**Edit:** I got hooks working now, weird that they weren't automatically configured.

tests/features/migrations/Migrator.mongo.test.ts
enduity: I completely understand, I know this wasn't the best code in the world. Just wanted to try out if that was even doable 🙂 

Removed:
- 38002174
- e739a13b (previously added part of the new tests to the wrong commit somehow...)

Though I would like to mention that while the snapshot can be useful in certain situations, it doesn't help in others. For example, it wouldn't catch the underlying Migration type changing in a way which breaks generated migrations.

In this specific case, it didn't help me because I expected snapshots to require updating anyway. I was not aware mongo had a different Migration type, so the snapshot diffs made sense. Ultimately, of course, it's my mistake. I just wanted to highlight that this is a small gap in the tests.

enduity: > (previously added part of the new tests to the wrong commit somehow...)

I rebased, but kept the history mostly intact (with different commit hashes now, of course). I just wanted to fix the formatting change commit pollution, as the history made no sense. Now the commits flow sensibly.

packages/migrations-mongodb/src/TSMigrationGenerator.ts
enduity: Not in the (non-mongo) generated file itself, as the function is not async and does not explicitly return `Promise<void>`. The inferred type would be `void`.

But I guess `void | Promise<void>` does signal the possibilities better to the user, and readability is most important. Incorporated this change:
- 58a4c763

packages/migrations/src/TSMigrationGenerator.ts
```diff
@@ -8,12 +8,12 @@ export class TSMigrationGenerator extends MigrationGenerator {
   generateMigrationFile(className: string, diff: { up: string[]; down: string[] }): string {
     let ret = `import { Migration } from '@mikro-orm/migrations';\n\n`;
     ret += `export class ${className} extends Migration {\n\n`;
-    ret += `  override up(): void {\n`;
+    ret += `  override up(): void | Promise<void> {\n`;
```
B4nan, line 11: now you need to update the snapshots, this will fail the tests

any reason you dont run the tests locally?

enduity: I was already fixing this 😄 

> any reason you dont run the tests locally?

I do run the tests, I'm just not used to the development flow of this project yet, so I forgot. I'm also used to GitLab automatically setting "Draft" when I push `fixup! ` commits, so I didn't expect you to re-review yet so quickly.

I really don't mean to waste your time, I don't know why I'm making myself seem so amateurish, sorry.

- 6a84546e

Issue #5686, opened 2024-06-07
Virtual properties defined as getters are recognized as required by RequiredEntityData


hartbit:
### Describe the bug

Defining a virtual property using a getter, like in the [documentation](https://mikro-orm.io/docs/defining-entities#virtual-properties) breaks calls to `create` and any other function depending on `RequiredEntityData` as it thinks the property is required.

### Reproduction

Using the following reproduction https://github.com/hartbit/reproduction-virtual-property-getter-required, we get the following TypeScript error:

```ts
orm.em.create(User, { firstName: 'Foo', lastName: 'Bar', email: 'foo' });
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Argument of type '{ firstName: string; lastName: string; email: string; }' is not assignable to parameter of type 'RequiredEntityData<User, never, false>'.
  Property 'fullName' is missing in type '{ firstName: string; lastName: string; email: string; }' but required in type '{ firstName: string; lastName: string; readonly fullName: string; email: string; }'.ts(2345)
```

### What driver are you using?

None

### MikroORM version

6.2.10-dev.28

### Node.js version

20.13.1

### Operating system

macOS

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/mikro-orm/mikro-orm/blob/master/CONTRIBUTING.md).
- [X] Read the [docs](https://mikro-orm.io/docs).
- [X] Check that there isn't [already an issue](https://github.com/mikro-orm/mikro-orm/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mikro-orm/mikro-orm/discussions) or join our [Slack](https://join.slack.com/t/mikroorm/shared_invite/enQtNTM1ODYzMzM4MDk3LWM4ZDExMjU5ZDhmNjA2MmM3MWMwZmExNjhhNDdiYTMwNWM0MGY5ZTE3ZjkyZTMzOWExNDgyYmMzNDE1NDI5NjA).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

B4nan:
You need to mark it as optional on type level, use `OptionalProps` symbol or `Opt` type. 

https://mikro-orm.io/docs/guide/relationships#using-entity-constructor

[...]

PR #7343, opened 2026-03-18
docs: Update Virtual Properties in defining-entities.md

stoneLeaf:
Following issue https://github.com/mikro-orm/mikro-orm/issues/5686 and discussion https://github.com/mikro-orm/mikro-orm/discussions/7336

Updated the docs regarding **virtual properties**:
- added how to declare getters so that they are not marked as required on entity creation (for both `defineEntity()` and decorator based schemas)
- removed the _define entity without class_ tab (as it couldn't apply here)

```TypeScript
  // Opt<T> type is necessary to mark getter properties as non-required on creation
  get fullName2(): Opt<string> {
    return `${this.firstName} ${this.lastName}`;
  }
```

### DIFF
--- a/docs/versioned_docs/version-7.0/defining-entities.md
+++ b/docs/versioned_docs/version-7.0/defining-entities.md
@@ -1914,7 +1914,8 @@ export class User {
     return `${this.firstName} ${this.lastName}`;
   }
 
-  get fullName2() {
+  // Opt return type marks the getter as non-required on entity creation
+  get fullName2(): Opt<string> {
     return `${this.firstName} ${this.lastName}`;
   }
 }
@@ -1947,7 +1948,8 @@ export class User {
     return `${this.firstName} ${this.lastName}`;
   }
 
-  get fullName2() {
+  // Opt return type marks the getter as non-required on entity creation
+  get fullName2(): Opt<string> {
     return `${this.firstName} ${this.lastName}`;
   }
 }
@@ -1984,8 +1986,9 @@ export class User {
     return `${this.firstName} ${this.lastName}`;
   }
 
+  // Opt return type marks the getter as non-required on entity creation
   @Property({ persist: false })
-  get fullName2() {
+  get fullName2(): Opt<string> {
     return `${this.firstName} ${this.lastName}`;
   }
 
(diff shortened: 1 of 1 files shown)


B4nan:
You only updated a snapshot, next docs release would ignore your changes.

https://github.com/mikro-orm/mikro-orm/tree/master/docs/docs

I keep getting docs PRs like this, and I wonder how to fix this.

docs/versioned_docs/version-7.0/defining-entities.md
```diff
@@ -1894,7 +1894,6 @@ groupId="entity-def"
 defaultValue="define-entity-class"
 values={[
 {label: 'defineEntity + class', value: 'define-entity-class'},
-{label: 'defineEntity', value: 'define-entity'},
```
B4nan, line 1897: why are you removing this? we want to stay consistent, every example should have 4 tabs.

docs/versioned_docs/version-7.0/defining-entities.md
```diff
@@ -1903,24 +1902,9 @@ values={[
...
-
-  getFullName() {
-    return `${this.firstName} ${this.lastName}`;
-  }
-
-  get fullName2() {
-    return `${this.firstName} ${this.lastName}`;
-  }
-}
+import { defineEntity, type Opt, p } from '@mikro-orm/core';
```
B4nan: i wouldnt use imports, again, this is inconsistent, we dont want one place with imports while others dont have it.

generally speaking, i would drop all the changes, only add the `Opt` return type, dont try to rewrite stuff like this, not in one example out of hundreds on one page

B4nan:
I see the defineEntity + class examples are mostly wrong on that page, they are supposed to be using `setClass` approach. I'll address that in a separate PR, leaving this one only about the added `Opt` return type.

docs/versioned_docs/version-7.0/defining-entities.md
stoneLeaf: I removed it because it was actually using a class. Maybe a empty script with just a comment saying it can't be done without classes would have been better.

PR #7350, opened 2026-03-20
fix: missing schema name when migration init/create

adlanarifzr:


B4nan:
You should start with a failing test case, so it's clear what you are fixing.

If you are trying to make migrations work with wildcard entities, that's not a fix, that would be a feature. Quite a big one, actually. We'd first need to agree on how that's supposed to work, so far, nobody has suggested anything truly usable.

adlanarifzr:
Hi @B4nan, thanks for the feedback. I have finalized the changes. Would you mind to check?

packages/migrations/src/Migrator.ts
```diff
@@ -272,7 +272,8 @@ export class Migrator extends AbstractMigrator<AbstractSqlDriver> {
     }
 
     const snapshotPath = await this.getSnapshotPath();
-    schema ??= this.#schemaGenerator.getTargetSchema();
+    const _schema = await DatabaseSchema.create(this.em.getConnection(), this.em.getPlatform(), this.config);
```
B4nan, line 275: this is wrong, we dont want to touch database if there is a snapshot file. and getting the whole schema object just to get its name is a complete waste of resources

B4nan (commented):
my previous point still stands, i dont know what you are fixing, no tests, no PR description, and a diff which is highly questionable

B4nan:
Thanks for the contribution, but closing this in favor of #7597 which adds proper runtime-schema support for migrations across all SQL drivers: a `migrations.schema` config option, `migrator.up({ schema })` runtime override, `includeWildcardSchema` for generating unqualified DDL from wildcard entities, a CLI `--schema` flag, plus the full set/reset plumbing (SET search_path on PG, USE on MySQL, ALTER SESSION on Oracle, hard error on MSSQL, no-op on SQLite). Covered by tests across all drivers.

PR #7363, opened 2026-03-20
feat(postgresql): add NativeArrayType for typed native array columns

DASPRiD:
While Mikro ORM has an `ArrayType` in core, it defaults to using `text[]` in Postgres as its primary type. One can extend the `ArrayType` and override that behavior, but that is a lot of unnecessary  work which can be avoided.

This PR introduces `NativeArrayType` in the `postgres` package. It allows wrapping any other type and turning it into an array type. Props are automatically forwarded to the inner type, so the type stays fully configurable.

In theory one can even wrap multiple `NativeArrayType`s to get multi-dimensional arrays in postgres.

## Usage

```ts
// integer[] column
@Property({ type: new NativeArrayType(IntegerType) })
tagIds!: number[];

// decimal(10,2)[] — precision/scale forwarded to inner type
@Property({ type: new NativeArrayType(new DecimalType('number')), precision: 10, scale: 2 })
prices!: number[];

// nullable
@Property({ type: new NativeArrayType(StringType), nullable: true })
labels?: string[] | null;
```

This also works with `defineEntity()`:

```ts
{
  properties: {
    tagIds: p.type(new NativeArrayType(IntegerType)),
  },
}
```

### DIFF
--- a/docs/docs/custom-types.md
+++ b/docs/docs/custom-types.md
@@ -515,6 +515,47 @@ export class CalendarDateArrayType extends ArrayType<CalendarDate> {
 favoriteDays!: CalendarDate[];
 ```
 
+### NativeArrayType (PostgreSQL)
+
+`NativeArrayType` is a PostgreSQL-specific type (from `@mikro-orm/postgresql`) that wraps any other type and produces a native PostgreSQL typed array column. Unlike the core `ArrayType` which always creates a `text[]` column and marshalls values to a delimited string, `NativeArrayType` derives the element column type from the inner type and appends `[]`, so you get columns like `integer[]`, `decimal(10,2)[]`, `boolean[]`, etc.
+
+Per-element conversion is delegated to the inner type, which means features such as `DecimalType`'s precision-aware comparison and custom `convertToJSValue` continue to work on each array element.
+
+The `nullable` option applies to the array column itself (the column may be `NULL`), while all other property options (e.g. `precision`, `scale`, `length`) are forwarded to the inner type when determining the element column type.
+
+```ts
+import { NativeArrayType } from '@mikro-orm/postgresql';
+import { types, DecimalType } from '@mikro-orm/core';
+
+// integer[] column
+@Property({ type: new NativeArrayType(types.integer) })
+ids!: number[];
+
+// decimal(10,2)[] column – precision and scale are forwarded to the inner DecimalType
+@Property({ type: new NativeArrayType(new DecimalType('number')), precision: 10, scale: 2 })
+prices!: number[];
+
+// nullable text[] column
+@Property({ type: new NativeArrayType(types.string), nullable: true })
+tags?: string[] | null;
+```
+
+With `defineEntity` you can use `p.type()` to attach the custom type:
+
+```ts
+import { defineEntity, p, types, DecimalType } from '@mikro-orm/core';
+import { NativeArrayType } from '@mikro-orm/postgresql';
+
+const Product = defineEntity({
+  name: 'Product',
+  properties: {
+    id: p.integer().primary(),
+    tagIds: p.type(new NativeArrayType(types.integer)),
+    prices: p.type(new NativeArrayType(new DecimalType('number'))).precision(10).scale(2),
+  },
+});
+```
+
 ### BigIntType
 
 Since v6, `bigint`s are represented by the native `BigInt` type, and as such, they don't require explicit type in the decorator options:
--- a/packages/postgresql/src/NativeArrayType.ts
+++ b/packages/postgresql/src/NativeArrayType.ts
@@ -1,0 +1,83 @@
+import type { Constructor, EntityProperty, Platform, TransformContext } from '@mikro-orm/core';
+import { Type } from '@mikro-orm/core';
+
+/**
+ * Maps a PostgreSQL native typed array column (e.g. `integer[]`, `decimal(10,2)[]`) to a JS array.
+ *
+ * Unlike the core `ArrayType` which always uses a `text[]` column and marshalls values to/from
+ * a delimited string, `NativeArrayType` derives the element column type from the supplied inner
+ * type and appends `[]`, producing a proper PostgreSQL typed array column. Per-element conversion
+ * is delegated to the inner type, so features like `DecimalType`'s precision-aware comparison
+ * continue to work on each element.
+ *
+ * The `nullable` decorator/defineEntity option applies to the array column itself (i.e. the
+ * column may be `NULL`), while all other property options (e.g. `precision`, `scale`) are
+ * forwarded to the inner type when determining the element column type.
+ *
+ * @example
+ * // integer[] column
+ * @Property({ type: new NativeArrayType(types.integer) })
+ * ids!: number[];
+ *
+ * @example
+ * // decimal(10,2)[] column
+ * @Property({ type: new NativeArrayType(new DecimalType('number')), precision: 10, scale: 2 })
+ * prices!: number[];
+ */
+export class NativeArrayType<InnerJsType, InnerDbType> extends Type<InnerJsType[] | null, InnerDbType[] | null> {
+  readonly #inner: Type<InnerJsType, InnerDbType>;
+
+  constructor(inner: Constructor<Type<InnerJsType, InnerDbType>> | Type<InnerJsType, InnerDbType>) {
+    super();
+    this.#inner = inner instanceof Type ? inner : new inner();
+
+    for (const key of ['platform', 'meta', 'prop'] as const) {
+      Object.defineProperty(this, key, {
+        get: () => this.#inner[key],
+        set: (value) => this.#inner.prop = key === 'prop' ? { ...value, autoincrement: false } : value,
+        enumerable: true,
+        configurable: true,
+      });
+    }
+  }
+
+  override getColumnType(prop: EntityProperty, platform: Platform): string {
+    const innerProp = { ...prop, autoincrement: false };
+    return `${this.#inner.getColumnType(innerProp, platform)}[]`;
+  }
+
+  override convertToDatabaseValue(
+    value: InnerJsType[] | null,
+    platform: Platform,
+    context?: TransformContext,
+  ): InnerDbType[] | null {
+    if (value == null) {
+      return value;
+    }
+
+    return value.map(item => this.#inner.convertToDatabaseValue(item, platform, context));
+  }
+
+  override convertToJSValue(
+    value: InnerDbType[] | null,
+    platform: Platform,
+  ): InnerJsType[] | null {
+    if (value == null) {
+      return value;
+    }
+
+    return value.map(item => this.#inner.convertToJSValue(item, platform));
+  }
+
+  override compareAsType(): string {
+    return 'array';
+  }
+
+  override toJSON(value: InnerJsType[] | null, platform: Platform): InnerJsType[] | InnerDbType[] | null {
+    if (value == null) {
+      return value;
+    }
+
+    return value.map(item => this.#inner.toJSON(item, platform)) as InnerJsType[] | InnerDbType[];
+  }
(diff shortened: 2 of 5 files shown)


B4nan:
>  One can extend the ArrayType and override that behavior, but that is a lot of unnecessary work which can be avoided.

All you need to do is to override the `columnType` option, no need for a custom type here.

Or is this about runtime mapping too?

B4nan:
I can't say I like the implementation, nor the API. Seems way too complicated for what it should do.

I'd rather handle this via `type` option, just like you can use `type: 'number[]'`, we could support other array types.

https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/metadata/MetadataDiscovery.ts#L2142-L2145

Or maybe just let people use the custom type as usual plus add `array: true`. I wouldn't add new type class at all.

DASPRiD:
> Or is this about runtime mapping too?

Yeah, this is also about runtime mapping. E.g. if you have a column `time[]`, this can then automatically be mapped to and from e.g. `Temporal.PlainTime[]`, no need to have a new type for every single array type. Or a `decimal[]` to an array of `Big.js`.

Having `array: true` (or p.something().array()`) would also be a nice API, though of course that would only work properly in postgres as well. I wanted to ensure that people don't get confused by having this in the core.

B4nan:
IMO this should be in core, it's valid for many cases that will work fine with the string concatenated values other dialects use. Think of `bigint[]` or `Date[]`, I actually can't think of a type that wouldn't work this way.

DASPRiD:
`float[]`, `double[]` and `bool[]` are types from the top of my head which wouldn't work with existing types which expect non-string parameters to transform.

But anyway, you also said you were not happy with the implementation either. How approximately would you like to see this implemented? Happy to improve on it.

B4nan:
> float[], double[] and bool[] are types from the top of my head which wouldn't work with existing types which expect non-string parameters to transform.

But why? All of those can be serialized to strings just fine. I don't see why this should be postgres only. Sure, it will work natively on DB level there, but that's not an argument really, I am fine with fallbacks that mostly work the same, I would rather have consistency on API level.

What I don't like is the idea of some new meta type (also its name is not very descriptive), the API is way too verbose, especially because it forces you to use the type implementations that are often no-op. Most of the types like `IntegerType` or `StringType` are there just as internals, they don't do anything really. And even if they do something, the expected usage is `p.decimal()` or `type: 'decimal'` in decorator options, not explicit usage of the type class.

I would lean towards the `array: true` hint (`p.array()` with `defineEntity`). 

But I am also still not sure if we even need this. The current array type supports overriding the `columnType` as well as the mapper functions. Let me try to check how it works now.

B4nan:
Yeah, this really feels unnecessary, especially in postgres, where you get native mapping on the driver level, you don't even need to deal with the custom mapping functions.

Here is how it works currently:

```ts
import { ArrayType, MikroORM } from '@mikro-orm/postgresql';
import { Entity, PrimaryKey, Property, ReflectMetadataProvider } from '@mikro-orm/decorators/legacy';
import { mockLogger } from '../../helpers.js';

@Entity()
class TypedArrayEntity {
  @PrimaryKey()
  id!: number;

  @Property({ type: new ArrayType(i => BigInt(i), i => i.toString()), columnType: 'bigint[]' })
  bigints!: bigint[];

  @Property({ type: new ArrayType(), columnType: 'double precision[]' })
  floats!: number[];

  @Property({ type: new ArrayType(), columnType: 'boolean[]' })
  booleans!: boolean[];

  @Property({ type: new ArrayType(i => new Date(i), i => i.toISOString()), columnType: 'timestamptz[]' })
  dates!: Date[];

  @Property({ type: new ArrayType(), columnType: 'int[]' })
  integers!: number[];

  @Property({ type: new ArrayType(), columnType: 'numeric(10,2)[]', nullable: true })
  decimals?: number[] | null;
}

let orm: MikroORM;

beforeAll(async () => {
  orm = await MikroORM.init({
    metadataProvider: ReflectMetadataProvider,
    entities: [TypedArrayEntity],
    dbName: 'typed_arrays_test',
  });

  await orm.schema.refresh();
});

afterAll(async () => {
  await orm.close(true);
});

test('schema produces correct typed array column types', async () => {
  const sql = await orm.schema.getCreateSchemaSQL();
  expect(sql).toMatch('"bigints" bigint[]');
  expect(sql).toMatch('"floats" double precision[]');
  expect(sql).toMatch('"booleans" boolean[]');
  expect(sql).toMatch('"dates" timestamptz[]');
  expect(sql).toMatch('"integers" int[]');
  expect(sql).toMatch('"decimals" numeric(10,2)[]');
});

test('persist and retrieve bigint[]', async () => {
  const entity = orm.em.create(TypedArrayEntity, {
    bigints: [1n, 9007199254740993n, -42n],
    floats: [1.5],
    booleans: [true],
    dates: [new Date('2025-01-01T00:00:00.000Z')],
    integers: [1],
  });
  await orm.em.persist(entity).flush();
  orm.em.clear();

  const loaded = await orm.em.findOneOrFail(TypedArrayEntity, entity.id);
  expect(loaded.bigints).toEqual([1n, 9007199254740993n, -42n]);
  expect(typeof loaded.bigints[0]).toBe('bigint');
});

test('persist and retrieve float/double precision[]', async () => {
  const entity = orm.em.create(TypedArrayEntity, {
    bigints: [1n],
    floats: [1.1, 2.2, 3.14159, -0.001],
    booleans: [true],
    dates: [new Date('2025-01-01T00:00:00.000Z')],
    integers: [1],
  });
  await orm.em.persist(entity).flush();
  orm.em.clear();

  const loaded = await orm.em.findOneOrFail(TypedArrayEntity, entity.id);
  expect(loaded.floats).toEqual([1.1, 2.2, 3.14159, -0.001]);
  expect(typeof loaded.floats[0]).toBe('number');
});

test('persist and retrieve boolean[]', async () => {
  const entity = orm.em.create(TypedArrayEntity, {
    bigints: [1n],
    floats: [1.0],
    booleans: [true, false, true, false],
    dates: [new Date('2025-01-01T00:00:00.000Z')],
    integers: [1],
  });
  await orm.em.persist(entity).flush();
  orm.em.clear();

  const loaded = await orm.em.findOneOrFail(TypedArrayEntity, entity.id);
  expect(loaded.booleans).toEqual([true, false, true, false]);
  expect(typeof loaded.booleans[0]).toBe('boolean');
});

test('persist and retrieve Date/timestamptz[]', async () => {
  const d1 = new Date('2025-01-01T00:00:00.000Z');
  const d2 = new Date('2000-06-15T12:30:00.000Z');
  const entity = orm.em.create(TypedArrayEntity, {
    bigints: [1n],
    floats: [1.0],
    booleans: [true],
    dates: [d1, d2],
    integers: [1],
  });
  await orm.em.persist(entity).flush();
  orm.em.clear();

  const loaded = await orm.em.findOneOrFail(TypedArrayEntity, entity.id);
  expect(loaded.dates).toHaveLength(2);
  expect(loaded.dates[0]).toBeInstanceOf(Date);
  expect(loaded.dates[0].toISOString()).toBe(d1.toISOString());
  expect(loaded.dates[1].toISOString()).toBe(d2.toISOString());
});

test('persist and retrieve int[]', async () => {
  const entity = orm.em.create(TypedArrayEntity, {
    bigints: [1n],
    floats: [1.0],
    booleans: [true],
    dates: [new Date('2025-01-01T00:00:00.000Z')],
    integers: [10, 20, -30, 0, 2147483647],
  });
  await orm.em.persist(entity).flush();
  orm.em.clear();

  const loaded = await orm.em.findOneOrFail(TypedArrayEntity, entity.id);
  expect(loaded.integers).toEqual([10, 20, -30, 0, 2147483647]);
  expect(typeof loaded.integers[0]).toBe('number');
});

test('persist and retrieve nullable numeric(10,2)[]', async () => {
  const entity = orm.em.create(TypedArrayEntity, {
    bigints: [1n],
    floats: [1.0],
    booleans: [true],
    dates: [new Date('2025-01-01T00:00:00.000Z')],
    integers: [1],
    decimals: [99.99, 0.01, -123.45],
  });
  await orm.em.persist(entity).flush();
  orm.em.clear();

  const loaded = await orm.em.findOneOrFail(TypedArrayEntity, entity.id);
  expect(loaded.decimals).toEqual([99.99, 0.01, -123.45]);

  // test null
  loaded.decimals = null;
  await orm.em.flush();
  orm.em.clear();

  const loaded2 = await orm.em.findOneOrFail(TypedArrayEntity, entity.id);
  expect(loaded2.decimals).toBeNull();
});

test('empty arrays roundtrip correctly', async () => {
  const entity = orm.em.create(TypedArrayEntity, {
    bigints: [],
    floats: [],
    booleans: [],
    dates: [],
    integers: [],
  });
  await orm.em.persist(entity).flush();
  orm.em.clear();

  const loaded = await orm.em.findOneOrFail(TypedArrayEntity, entity.id);
  expect(loaded.bigints).toEqual([]);
  expect(loaded.floats).toEqual([]);
  expect(loaded.booleans).toEqual([]);
  expect(loaded.dates).toEqual([]);
  expect(loaded.integers).toEqual([]);
});

test('no spurious update after loading', async () => {
  const entity = orm.em.create(TypedArrayEntity, {
    bigints: [1n, 2n],
    floats: [1.5, 2.5],
    booleans: [true, false],
    dates: [new Date('2025-01-01T00:00:00.000Z')],
    integers: [10, 20],
    decimals: [1.23, 4.56],
  });
  await orm.em.persist(entity).flush();
  orm.em.clear();

  await orm.em.findOneOrFail(TypedArrayEntity, entity.id);
  const mock = mockLogger(orm);
  await orm.em.flush();

  // no update queries should be issued
  expect(mock.mock.calls).toHaveLength(0);
});
```

Also, if we expose the `array` option (which is now only supported for embeddables, but technically already works for scalars, just not available on type-level), you could just do this (as long as you don't need to adjust the mapping function):

```ts
@Property({ array: true, columnType: 'numeric(10,2)[]', nullable: true })
decimals?: number[] | null;
```

DASPRiD:
```ts
@Property({ array: true, columnType: 'numeric(10,2)[]', nullable: true })
decimals?: number[] | null;
```

I guess this could be enhanced, so that `array` not only takes boolean, but also a tuple for the mapping function? Then one never has to construct `ArrayType` manually. This would also work quite well for `p.array(codec)` In that case, e.g. `mikro-orm-temporal` could expose those codecs.

B4nan:
FWIW, `p.array()` already accepts the mapping functions.

https://github.com/mikro-orm/mikro-orm/blob/master/tests/defineEntity.test.ts#L932-L936

DASPRiD:
Okay, I gave this a few more thoughts.

What do you think about being able to also pass a mikro orm type to `p.array()` and the ArrayType? It can then use those as the codec it already accepts. For temporal for instance, that'd allow to do this:

```ts
p.array(PlainTimeType).columnType("time[]")
```

Or for decorators:

```ts
@Property({ type: new ArrayType(PlainTimeType), columnType: "time[]" })
```

Alternatively to using the constructor, we could have a `ArrayType.fromType(PlainTimeType)` or something like that, to not pollute the constructor.

B4nan:
I still don't see a real benefit over what we already support. This already works:

```ts
@Property({
  type: new ArrayType(v => Temporal.PlainTime.from(v), v => v.toString()),
  columnType: 'time[]',
})
times!: Temporal.PlainTime[];

// Or with defineEntity:
times: p.array(
  v => Temporal.PlainTime.from(v),
  v => v.toString(),
).columnType('time[]')
```

If something, I would go with combination of `p.type()` and `p.array()`:

```ts
times: p.type(PlainTimeType).array()

// or even simpler cases
integers: p.integer().array()
```

Let me ask claude to put this together so we can talk about something concrete.

B4nan:
Let's move the discussion to #7378

PR #7364, opened 2026-03-21
Add mention of dotenvx alongside dotenv.

motdotla:
Adds mention of dotenvx which adds encryption safety as well as default variable expansion out of the box. It's a drop-in replacement from me the creator of dotenv.

### DIFF
--- a/docs/blog/2026-03-11-mikro-orm-7-released.md
+++ b/docs/blog/2026-03-11-mikro-orm-7-released.md
@@ -38,7 +38,7 @@ One of the headline features of v7 is that the `@mikro-orm/core` package now has
 
 This has huge implications for bundle size, cold start times (great news for serverless users!), and overall maintainability. Some of these dependencies are still available as optional peer dependencies - for instance, install `dataloader` if you want the dataloader integration, and `reflect-metadata` if you prefer the reflect-based metadata provider.
 
-> Dotenv support has been removed entirely. If you relied on automatic `.env` loading, you'll need to call `dotenv.config()` yourself before initializing the ORM.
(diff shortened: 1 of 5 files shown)


docs/blog/2026-03-11-mikro-orm-7-released.md
```diff
@@ -38,7 +38,7 @@ One of the headline features of v7 is that the `@mikro-orm/core` package now has
 
 This has huge implications for bundle size, cold start times (great news for serverless users!), and overall maintainability. Some of these dependencies are still available as optional peer dependencies - for instance, install `dataloader` if you want the dataloader integration, and `reflect-metadata` if you prefer the reflect-based metadata provider.
 
-> Dotenv support has been removed entirely. If you relied on automatic `.env` loading, you'll need to call `dotenv.config()` yourself before initializing the ORM.
+> Dotenv support has been removed entirely. If you relied on automatic `.env` loading, you'll need to call `dotenv.config()` yourself before initializing the ORM, or use [`@dotenvx/dotenvx`](https://github.com/dotenvx/dotenvx) (for added encryption–from the creator of dotenv).
```
adriancuadrado, line 41: "from the creator of dotenv" This kinda reads like an ad. We don't need to know who made dotenvx here.

adriancuadrado: Also I'm not even sure if it's ok to promote software like this. If dotenv support has been removed and dotenvx is a drop-in replacement, then it doesn't matter whether you use one or the other to add environment variables because my understanding is that in both cases you still have to call `dotenv.config()`, so I don't think it makes sense to tell users that they can use dotenvx.

B4nan: Yeah, agreed, I don't think this adds value to the readers. We don't link dotenv anywhere, and looking at https://www.dotenv.org/, you already promote dotenvx there, so if this is trying to spread the word, I don't feel it's necessary here. Also, I completely agree that the mention of you as the creator doesn't make sense in our docs. Plus, I really don't like that you are touching my blog post, lol.

PR #7378, opened 2026-03-22
feat(core): support `array` option on scalar properties

B4nan:
## Summary

- Expose `array: true` on `PropertyOptions` so scalar properties can use it via `@Property({ type: IntegerType, array: true })` or `p.integer().array()`
- During discovery, when `array: true` is set on a scalar with a custom type, the type is automatically wrapped in an `ArrayType` that delegates element conversion to the inner type's `convertToJSValue`/`convertToDatabaseValue`
- Column type is inferred as `innerType.getColumnType() + '[]'` (e.g. `int[]`, `text[]`, `numeric(10,2)[]`)

Relates to #7363 — provides a simpler alternative without introducing a new type class.

### Usage

```ts
// defineEntity
integers: p.integer().array()
decimals: p.decimal('number').array().precision(10).scale(2)
custom:   p.type(new PlainTimeType()).array()

// decorators
@Property({ type: IntegerType, array: true })
integers!: number[];
```

### DIFF
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -2114,7 +2114,13 @@ export class MetadataDiscovery {
       });
 
       if (type) {
-        prop.type = type === 'datetime' ? 'Date' : type;
+        if (prop.array) {
+          // built-in type + array: true — force-create instance for ArrayType wrapping
+          prop.customType = new (prop.type as Constructor<Type>)();
+          prop.type = prop.customType.constructor.name;
+        } else {
+          prop.type = type === 'datetime' ? 'Date' : type;
+        }
       } else {
         prop.customType = new (prop.type as Constructor<Type>)();
         prop.type = prop.customType.constructor.name;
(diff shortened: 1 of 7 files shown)


packages/core/src/metadata/MetadataDiscovery.ts
```diff
@@ -2112,6 +2118,27 @@ export class MetadataDiscovery {
...
+      prop.array &&
+      prop.customType &&
+      !(prop.customType instanceof t.array) &&
+      !(prop.customType instanceof t.enumArray) &&
+      prop.kind === ReferenceKind.SCALAR
+    ) {
+      const innerType = prop.customType;
+
+      if (!prop.columnTypes) {
+        prop.columnTypes = [innerType.getColumnType(prop, this.#platform) + '[]'];
```
DASPRiD: This would only work on the Postgres platform though, correct? On other platforms this would create invalid migrations.

B4nan: Yes, it's a draft, no worries, I am already generating tests for other drivers. Other drivers will likely just keep the original column type from the ArrayType, which is usually a simple `text`.

DASPRiD:
LGTM, definitely better than my implementation, I agree. I also like that this additionally supports other database engines this way. Good job! 👍

vaskouk:
Whats the diff than having `prices: p.array(Number)` ?

B4nan:
`p.array(Number)` stores in `text[]` I think? With `p.integer().array()` you get `int[]`.

vaskouk:
> `p.array(Number)` stores in `text[]` I think? With `p.integer().array()` you get `int[]`.

True true. Now i see, nice improvement!

vaskouk:
For custom types, it requires to define it like this:`p.type(new DecimalJsType).array()` or can also stay as `p.type(DecimalJsType).array()` ?

B4nan:
Both should work, will adjust some tests to use that approach too.

vkouk:
> Both should work, will adjust some tests to use that approach too.

@B4nan does it work the way I posted above ? Did you figure out or shall we test and see ?

B4nan:
I dont remember, give it a try :)

PR #7379, opened 2026-03-22
feat(schema): add database trigger support

### LINKED ISSUE #5053: add entity decorators to specify triggers
I'm using a few triggers in my database and I would love to be able to specifiy those in my dbEntities files, hence relying solely on mikroOrm generated migrations to create my db schema.

Very similar to the Index decorator, I would love having a Trigger decorator, along those lines:
``` javascript
@Trigger(
{
  name:'my_trigger'
  insert:true,
  update:true, 
  fn : ` UPDATE x SET t=1 WHERE id=NEW.id; RETURN NEW;`
  }
)
@Entity({ tableName: 'myTable' })
```
For postgres, this would generate the following sql code :

``` sql
CREATE OR REPLACE FUNCTION my_trigger_function()
RETURNS TRIGGER AS $$
BEGIN
  UPDATE x SET t=1 WHERE id=NEW.id, 
  RETURN NEW;
END;
$$ LANGUAGE plpgsql;

CREATE TRIGGER my_trigger
AFTER INSERT OR UPDATE OR DELETE ON myTable
FOR EACH ROW EXECUTE FUNCTION my_trigger_function()
```

Currently I update a migration created by mikro orm to add my trigger creation statement. this works fine.
However, to speed up the db schema creation process (notably for our CI), from time to time we remove all our migrations and restart from a clean slate using `npx mikro-orm migration:create --initial`
doing this will lose the trigger code, so we need to remember to reapply it manually

B4nan:
## Summary

- Adds `@Trigger` decorator (legacy + TC39) and `triggers` option for `defineEntity`/`EntitySchema` to define database triggers managed by the schema generator and migration system
- Triggers are automatically created, diffed (add/change/remove), and dropped during schema updates
- Driver-specific DDL: PostgreSQL (function + trigger), MySQL/MariaDB/SQLite (one per event), MSSQL (multi-event, after/instead of only)
- Schema introspection for all SQL drivers with round-trip idempotency (no spurious diffs)
- MongoDB validation throws when triggers are defined on non-SQL entities

Closes #5053


packages/core/src/metadata/MetadataDiscovery.ts
```diff
@@ -212,6 +212,7 @@ export class MetadataDiscovery {
 
     filtered.forEach(meta => this.initAutoincrement(meta)); // once again after we init custom types
     filtered.forEach(meta => this.initCheckConstraints(meta));
+    filtered.forEach(meta => this.initTriggers(meta));
```
vkouk, line 215: Why don't you loop once to init whatever you want to 😄

B4nan: For this specific line, no reason outside of consistency. For the pattern itself, some properties need to be set on all entities, we can't just iterate them once because of cycles and whatnot.

PR #7382, opened 2026-03-23
fix(core): handle Bun path lookup for native ES decorators

### LINKED ISSUE #7381: Bun 1.3.10 direct TS execution with ES decorators falls back to entity name in `lookupPathFromDecorator`
### Describe the bug

using tools
- ES decorators from `@mikro-orm/decorators/es`
- `TsMorphMetadataProvider`

execute my server with `bun run --watch ./src/app.ts`, MikroORM fails to read entities with error

```
[discovery] - processing entity Task (Task)
163 |       path = path.replace(new RegExp(`^${outDirRelative}`), '');
164 |     }
165 |     path = this.stripRelativePath(path);
166 |     const source = this.sources.find(s => s.getFilePath().endsWith(path));
167 |     if (!source && validate) {
168 |       throw new MetadataError(
                  ^
MetadataError: Source file './Task' not found. Check your 'entitiesTs' option and verify you have 'compilerOptions.declaration' enabled in your 'tsconfig.json'. If you are using webpack, see https://bit.ly/35pPDNn
 entity: undefined,
```

### Reproduction

https://github.com/changchanghwang/mikro-orm-fallback

@mikro-orm/mysql

### MikroORM version

    "@mikro-orm/core": "^7.0.4",     "@mikro-orm/decorators": "^7.0.4",     "@mikro-orm/mysql": "^7.0.4",     "@mikro-orm/reflection": "^7.0.4"

### Node.js version

bun 1.3.10, node 22.22.1

### Operating system

mac os

### Validations

changchanghwang:
## Summary
- fix decorator path lookup on Bun@1.3.10~ ([native Es decorators](https://bun.com/blog/bun-v1.3.10#tc39-standard-es-decorators))
- support stack frames that omit column numbers during path extraction
- test codes for this pr

Closes #7381

### DIFF
--- a/tests/Utils.win.test.ts
+++ b/tests/Utils.win.test.ts
@@ -558,6 +558,32 @@ describe('Utils', () => {
         '    at A (bun:wrap:1:2617)',
       ]),
     ).toBe('Book');
+
+    // bun with transpiled decorator helpers (`__decorate`)
+    expect(
+      lookupPathFromDecorator('Book', [
+        'Error',
+        '    at <anonymous> (/usr/local/var/www/my-project/dist/entities/Book.js:16:54)',
+        '    at __decorate (/usr/local/var/www/my-project/dist/entities/Book.js:4:110)',
+        '    at /usr/local/var/www/my-project/dist/entities/Book.js:22',
+        '    at moduleEvaluation (native:1:11)',
+        '    at moduleEvaluation (native:1:11)',
+        '    at loadAndEvaluateModule (native:2)',
+      ]),
+    ).toBe('/usr/local/var/www/my-project/dist/entities/Book.js');
+
+    // bun with native ES decorators (`bun:wrap`)
+    expect(
+      lookupPathFromDecorator('Book', [
+        'Error',
+        '    at <anonymous> (/opt/app/src/entities/Book.ts:5:47)',
+        '    at se (bun:wrap:1:3694)',
+        '    at /opt/app/src/entities/Book.ts:16',
+        '    at moduleEvaluation (native:1:11)',
+        '    at moduleEvaluation (native:1:11)',
+        '    at loadAndEvaluateModule (native:2)',
+      ]),
+    ).toBe('/opt/app/src/entities/Book.ts');
   });
 
   test('lookup path from decorator on windows', () => {


packages/decorators/src/utils.ts
```diff
@@ -167,7 +171,7 @@ export function lookupPathFromDecorator(name: string, stack?: string[]): string
   }
 
   try {
-    const re = /\(.+\)/i.exec(stack[line]) ? /\((.*):\d+:\d+\)/ : /at\s*(.*):\d+:\d+$/;
+    const re = /\(.+\)/i.exec(stack[line]) ? /\((.*?)(?::\d+){1,2}\)/ : /at\s*(.*?)(?::\d+){1,2}$/;
```
B4nan, line 174: Why limit the size to `{1,2}`? Are you sure it can't be a bigger number?

changchanghwang: The previous regex already assumed at most two numeric segments, since it only matched :line:column. 
`:\d+:\d+`

If you think the current form is a bit less explicit, I can rewrite it to something like `:\d+(?::\d+)?`

B4nan: Oh, this is for groups, I thought its for the digits of the line number.

changchanghwang: Exactly, haha. When you asked me at first, I surprised too :)

changchanghwang:
i don't know where 1 missing line is...

B4nan:
image

https://app.codecov.io/gh/mikro-orm/mikro-orm/pull/7382?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=mikro-orm

Take a look, but don't try too hard, I am fine with a missed line too (or a coverage ignore hint).

changchanghwang:
> Take a look, but don't try too hard, I am fine with a missed line too (or a coverage ignore hint).

It looks like that line became a missing line because the existing stack trace is now covered by bunWrapLine.

since Bun’s stack trace format might still hit that logic, I think it would be better to keep the logic in place. 

As for the test case, it would probably make sense to add it later once we know the exact stack trace that triggers it.

PR #7437, opened 2026-03-30
fix(core): skip phantom delete changesets with null PKs in orphan removal

B4nan:
## Summary
- Entities with composite FK-in-PK that end up in the orphan remove stack without a valid primary key caused `DriverException` during `executeDeletes` because null PKs were passed to the `$in` clause
- Filters null PKs in `ChangeSetPersister.executeDeletes` to prevent the crash
- Skips entities without valid PKs when promoting from `orphanRemoveStack` to `removeStack` in `UnitOfWork.computeChangeSets`
- Regression introduced between 6.4.16 and 6.6.10, manifests with complex entity graphs (multiple models per table, cascade/orphan-removal traversal)

### DIFF
--- a/packages/core/src/unit-of-work/ChangeSetPersister.ts
+++ b/packages/core/src/unit-of-work/ChangeSetPersister.ts
@@ -123,7 +123,12 @@ export class ChangeSetPersister {
 
     for (let i = 0; i < changeSets.length; i += size) {
       const chunk = changeSets.slice(i, i + size);
-      const pks = chunk.map(cs => cs.getPrimaryKey());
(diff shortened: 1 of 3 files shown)


B4nan:
Closing — unable to reproduce the phantom delete issue on either 6.x or master, even with multiple models mapping to the same table (`checkDuplicateTableNames: false`), composite FK-in-PK entities with `orphanRemoval`, transactions, and `em.assign` graph mutations.

The reporter also couldn't produce a minimal reproduction. Without a failing test, we don't want to add defensive null-PK filtering to the core — it would mask the real bug rather than fix it.

If anyone can provide a minimal reproduction, we're happy to revisit.

PR #7494, opened 2026-04-06
fix(core): respect compiledFunctions in driver's EntityComparator

jeffwilde:
### Summary

`DatabaseDriver.setMetadata()` creates its own `EntityComparator` without the `config` argument. This means the driver's comparator — used for `mapResult()` — never sees `compiledFunctions`, so every call to `Utils.createFunction()` falls back to runtime code generation via `new Function()` even when pre-compiled functions are provided. This bypasses the performance benefit of pre-compilation on all runtimes, and causes `EvalError` on eval-free runtimes like Cloudflare Workers.

All read operations are affected (`find`, `findOne`, `stream`, `em.map`, virtual entities, embedded property mapping in joins) because they go through the driver's comparator. Write operations are unaffected — `ChangeSetPersister` and `UnitOfWork` both get their comparator from `config.getComparator()` / `em.getComparator()`, which already has config.

The fix passes `this.config` as the third argument to `EntityComparator` in `setMetadata()`.

### Test

Regression test that exercises the driver's `mapResult()` path with `compiledFunctions` enabled — verifies that `Utils.createFunction` uses the pre-compiled function rather than falling back to `new Function()`. Fails without the fix, passes with it.

### DIFF
--- a/packages/core/src/MikroORM.ts
+++ b/packages/core/src/MikroORM.ts
@@ -239,7 +239,6 @@ export class MikroORM<
     this.em = this.driver.createEntityManager() as EM & { '~entities': Entities };
     (this.em as { global: boolean }).global = true;
     this.#metadata.decorate(this.em);
-    this.driver.setMetadata(this.#metadata);
   }
 
   /**


B4nan:
Looking closer at the code, we can clean it up a bit, the dual `setMetadata` call is apparently not needed nowadays, and we can just use the comparator from the config instance. Will push it to your PR directly.

PR #7520, opened 2026-04-09
feat(core): add a `chunkSize` option to streams

boenrobot:


docs/versioned_docs/version-7.0/streaming.md
```diff
@@ -9,6 +9,7 @@ const stream = em.stream(Book, {
   populate: ['author'],
```
B4nan, line 9: nit: drop the v7 snapshot changes, i'll be shipping this as part of v7.1 with all the other open feature PRs

boenrobot: Done

B4nan:
FYI, I am running reviews throught claude, and this apparently has no effect on MySQL and SQLite. Only PostgreSQL and MSSQL kysely dialects respect that. I guess I would bump this to 101, it indeed sounds like a better default.

> MongoDB's default "batchSize" (which is basically just chunkSize with another name) is 101

From what I read, it's actually only relevant for the first batch. I will unify this to 100 everywhere (in supported dialects).

docs/docs/streaming.md
```diff
@@ -24,6 +25,7 @@ There are several constraints when using streaming:
 - When populating to-many relations, only fully hydrated entities will be returned.
 - You should provide an `orderBy` clause to ensure consistent ordering.
 - With mongodb driver, only root entities can be streamed, `populate` option is ignored.
+- The `chunkSize` option is only honored on PostgreSQL, MSSQL and MongoDB. On MySQL, MariaDB, SQLite and libSQL the underlying driver always streams row-by-row and the option is ignored.
```
boenrobot: What about Oracle?

B4nan: Good point, since we own the dialect, this was simple to handle there.

B4nan (approved):

PR #7526, opened 2026-04-09
fix(core): emit enum check constraints for embedded enum properties

ghempton:
## Description

Enum properties declared on an `@Embeddable()` class and used via `@Embedded()` on a non-embeddable `@Entity()` were silently missing their `check (col in (...))` constraints on SQL drivers that use check constraints for enums (postgres, sqlite, etc.). Directly-declared enum properties on the same entity still worked correctly, so the regression only affected embedded enum columns.

This caused two visible problems:

1. Fresh schema generation omitted enum check constraints for embedded enum columns — no DB-level enforcement of the enum values.
2. For existing v6 databases (which did have those check constraints), the v7 schema comparator emitted `drop constraint …` for every affected check constraint with no corresponding `add` — a silent data-integrity regression on upgrade.

## Root cause

`MetadataDiscovery.initCheckConstraints` iterated `meta.props`, which is the sorted props array populated by `EntitySchema.initProperties()` at schema-init time. By the time `initCheckConstraints` runs in the discovery pipeline, `initEmbeddables` has already added flattened embedded properties like `settings_status` to `meta.properties` (the dictionary), but `meta.sync()` has not yet been called to rebuild `meta.props` from that dictionary. So the enum props coming from embeddables were never visited by the loop.

Instrumenting the call for a minimal entity with one direct enum and one embedded enum showed:

```
initCheckConstraints:
  meta.props.length: 3
  Object.values(meta.properties).length: 5
  enum in meta.props:      [ 'directStatus' ]
  enum in meta.properties: [ 'directStatus', 'settings_status' ]
```

Both enum props are in the dictionary, but only the directly-declared one is in the stale `meta.props` array at that point.

## Fix

Iterate `Object.values(meta.properties)` instead of `meta.props` in the enum check constraint loop. All existing `persist === false` / `nativeEnumName` / string-items guards are preserved, so the only change in behavior is that flattened embedded enum properties are now included.

## Test

Added `tests/features/embeddables/embedded-enum-check-constraints.postgres.test.ts` with a `Demo` entity that has:

- a directly-declared enum (`directStatus`)
- an embedded `Settings` class whose `status` property is an enum

The test asserts that `meta.checks` contains both `demo_embedded_enum_check_direct_status_check` and `demo_embedded_enum_check_settings_status_check`, and that both corresponding `alter table ... add constraint ... check (...)` lines appear in the generated DDL. The test uses `connect: false` so it does not require a live postgres instance.

I verified the test fails without the fix (only `direct_status_check` is produced) and passes with the fix.

### DIFF
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -1928,7 +1928,12 @@ export class MetadataDiscovery {
     }
 
     if (this.#platform.usesEnumCheckConstraints() && !meta.embeddable) {
-      for (const prop of meta.props) {
+      // Iterate `meta.properties` rather than `meta.props` — by this point in
+      // the discovery pipeline `initEmbeddables` has added flattened embedded
+      // properties to `meta.properties`, but `meta.sync()` has not yet been
+      // called, so `meta.props` is still the pre-flatten array and would miss
+      // enum properties that live inside embeddables.
+      for (const prop of Object.values(meta.properties)) {
         if (prop.persist === false || prop.nativeEnumName || !prop.items?.every(item => typeof item === 'string')) {
           continue;
         }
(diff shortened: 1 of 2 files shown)


B4nan (approved):
thanks!

PR #7564, opened 2026-04-14
fix(core): support TPT inheritance targets in polymorphic relations

### LINKED ISSUE #7563: TPT incompatible with polymorphic associations
### Describe the bug

Polymorphic relations assume flat entities (one entity = one table). TPT breaks this because the discriminator resolution doesn't walk the inheritance hierarchy, and the JOIN generation doesn't account for the parent+child table split.

This is manifested in three issues:

1. INSERT: discriminator is null: When assigning a TPT child (e.g. Dog) to a polymorphic ref that lists the abstract parent (Animal), MikroORM resolves the discriminator from the concrete class. Since Dog isn't in the poly union, subject_type is written as null and the reference is silently corrupted.
2. SELECT: child table not joined: When populating the polymorphic relation, only the parent animal table is LEFT JOINed. The child dog table is never joined, so the entity hydrates as a bare Animal missing TPT child properties (e.g. breed).
3. TypeScript: ref(dog) not assignable: Dog extends Animal and Animal is in the union, but Reference<Dog> doesn't satisfy the polymorphic property type. Users must cast to work around it.

### Reproduction

https://github.com/jscheel/mikro-orm-tpt-polymorphic-issue-reproduction

@mikro-orm/sqlite

### MikroORM version

7.0.11-dev.11

### Node.js version

v22.17.0

### Operating system

MacOS

### Validations

jscheel:
When a polymorphic ManyToOne relation (e.g. `@ManyToOne(() => [Animal, Person])`) includes a TPT base class as a target, three bugs prevented it from working:

1. INSERT: The discriminator value (subject_type) was written as NULL because the concrete TPT child class (e.g. Dog) wasn't found in the polymorphic discriminator map, which only contains the root class (Animal). Fixed by walking the prototype chain in `QueryHelper.findDiscriminatorValue()` and the JIT-compiled snapshot code in `EntityComparator`.
2. SELECT: Populating the polymorphic relation didn't JOIN the TPT child tables, so child-specific properties (e.g. breed) were missing and the entity was hydrated as the abstract base class instead of the concrete subclass. Fixed by calling `addTPTPolymorphicJoinsForRelation()` in the polymorphic branch of `getFieldsForJoinedLoad()`, and adding `mapTPTChildFields()` with concrete class resolution in `mapJoinedProps()`.
3. M:N pivot loading: `EntityLoader.findChildrenFromPivotTable` compared `className` exactly, so TPT children (e.g. Dog) didn't match their root target (Animal) and were silently dropped from collection results. Fixed by comparing `root.className` instead.

All fixes use fallback logic that only activates for TPT hierarchies, so existing behavior for plain entities and STI is unaffected. Includes 22 test cases covering INSERT, SELECT (JOINED + SELECT_IN), UPDATE, inverse side, lazy loading, serialization, nested TPT, and mixed-type queries.

Also fixes a pre-existing STI bug in M:N polymorphic pivot table loading. STI children were silently dropped from the population step due to the exact className match. Our root.className comparison fixes both TPT and STI cases.

Closes #7563

### DIFF
--- a/packages/core/src/entity/EntityLoader.ts
+++ b/packages/core/src/entity/EntityLoader.ts
@@ -850,7 +850,12 @@ export class EntityLoader {
     if (prop.polymorphic && prop.polymorphTargets) {
       await Promise.all(
         prop.polymorphTargets.map(async targetMeta => {
-          const targetChildren = unique.filter(child => helper(child).__meta.className === targetMeta.className);
+          const targetChildren = unique.filter(child => {
+            const childMeta = helper(child).__meta;
+            // Use root.className for TPT entities: a Dog (root: Animal) should match
+            // targetMeta Animal. For non-TPT entities, root === self.
+            return childMeta.root.className === targetMeta.root.className;
+          });
           if (targetChildren.length > 0) {
             await populateChildren(targetMeta, targetChildren);
           }
--- a/packages/core/src/utils/EntityComparator.ts
+++ b/packages/core/src/utils/EntityComparator.ts
@@ -817,14 +817,13 @@ export class EntityComparator {
         ret += `      ret${dataKey} = null;\n`;
         ret += `    } else if (typeof entity${entityKey} !== 'undefined') {\n`;
         ret += `      const val${level} = entity${entityKey}${unwrap};\n`;
-        ret += `      let discriminator = ${discriminatorMapKey}.get(val${level}?.constructor);\n`;
-        ret += `      if (discriminator === undefined && val${level}?.constructor) {\n`;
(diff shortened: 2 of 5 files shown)


B4nan:
Looking at those tests, why are you using raw queries all around the place?

jscheel:
@B4nan it just provides tighter isolation for the functionality that is actually being testing in each individual test. Not a hill I would die on though.

B4nan:
I'll handle the M:N part in v7.1 via #7569, since that's hardly a bug fix.

PR #7582, opened 2026-04-16
chore: cap mssql container cpu/memory to prevent idle runaway

B4nan:
## Summary

MSSQL on Apple Silicon runs under x86_64 emulation (Microsoft doesn't publish a native ARM image for full SQL Server). Under emulation, SQL Server's background threads — lazy writer, checkpoint, ghost cleanup, query store cleanup, SQLOS schedulers — can burn several hundred % host CPU even with no workload.

This adds three caps that bring the container from ~800% host CPU at idle to ~2%:

- `cpus: 2` — cgroup CPU cap (ceiling)
- `mem_limit: 3g` — cgroup memory cap, leaving headroom over the buffer-pool limit below
- `MSSQL_MEMORY_LIMIT_MB: 2048` — tells SQL Server's own memory manager to stop growing its buffer pool past 2 GiB, which was the trigger for the thrashing under the previous (unlimited) configuration

On native x86_64 hosts the caps are still a net benefit — 2 GiB / 2 CPUs is plenty for the test workload and keeps the container well-behaved alongside the other 7 DB services.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -81,9 +81,12 @@ services:
       - '1433:1433'
     cap_add:
       - SYS_PTRACE
+    cpus: 2
+    mem_limit: 3g
     environment:
       MSSQL_SA_PASSWORD: Root.Root
       MSSQL_TELEMETRY_ENABLED: 0
+      MSSQL_MEMORY_LIMIT_MB: 2048
       ACCEPT_EULA: 1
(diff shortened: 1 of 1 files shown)


B4nan:
closing while I iterate on the right cap values — the first attempt was too tight and caused ECONNRESETs during the parallel mssql test load

Issue #4286, opened 2023-04-27
Better collation support


stefansundin:
**Is your feature request related to a problem? Please describe.**

I am using mikro-orm with postgresql and I would love the ability to create a collation and then use it in `@Property`.

**Describe the solution you'd like**

Something like this to create the collation:

```typescript
@Collation({
  name: 'case_insensitive',
  provider: 'icu',
  locale: 'und-u-ks-level2',
  deterministic: false,
})
```

Postgres supports the following [`CREATE COLLATION`](https://www.postgresql.org/docs/current/sql-createcollation.html) syntax:

```
CREATE COLLATION [ IF NOT EXISTS ] name (
    [ LOCALE = locale, ]
    [ LC_COLLATE = lc_collate, ]
    [ LC_CTYPE = lc_ctype, ]
    [ PROVIDER = provider, ]
    [ DETERMINISTIC = boolean, ]
    [ VERSION = version ]
)
```

And then to use the collation I want something like the following:

```typescript
  @Property({ collation: 'case_insensitive' })
  email: string;
```

**Describe alternatives you've considered**

I am currently working around this limitation with a manual migration:

```typescript
    this.addSql(
      `create collation if not exists "case_insensitive" (provider = icu, locale = 'und-u-ks-level2', deterministic = false);`,
    );
    this.addSql(
      'alter table "user" alter column "email" set data type text collate "case_insensitive";',
    );
```

It is ok since mikro-orm doesn't try to revert it.

**Additional context**

I think better collation support would be awesome. It is not a fringe feature.

B4nan:
I would rather keep things simple, so the property option sounds good to me, but `@Collation` decorator does not (partially also because that would tie its definition to particular table, while its not table specific) - it would require quite a lot of stuff, you need to implement schema reflection and diffing for that. Even native postgres enums are not supported this way, this feels like much bigger niche than that.

I want to finish the native postgres enums refactor for v6, and this could end up as a very similar feature, so let's see how that goes.

B4nan:
FYI v7.1 will provide native support for column-level collation via #7615

PR #7608, opened 2026-04-23
fix(sql): stabilize migration snapshot serialization

### LINKED ISSUE #7607: Clarify migration snapshot lifecycle and post-apply rewrites
Related to #7234.

## Problem

The current docs explain snapshots mainly as something written by `migration:create` and then reused by future `migration:create` runs instead of diffing against the current database schema.

That is true, but it does not describe the full lifecycle.

In practice, the snapshot file can represent two different states:

- after `migration:create`, it is written from MikroORM's target schema / metadata
- after `migration:up` or `migration:down`, it can be rewritten from the real database schema via introspection

Because the docs currently focus on the first case, it is easy to assume the snapshot is a stable, reviewable artifact of the generated migrations and always represents the expected database schema 1:1.

## Why this is confusing in teams

A common workflow may look like this:

1. Developer A runs `migration:create` and `migration:up` to test changes
1. Developer A runs `migration:down`, deletes it, makes a small change to entities, and only runs `migration:create`
1. Developer A commits the migration and snapshot without also running `migration:up`
1. Developer B pulls the branch and runs `migration:up`
1. Developer B now gets an unexpected snapshot diff in git

That follow-up diff can contain many changes that are unrelated to any newly authored migration, for example:

- table order changes
- type normalization from database introspection
- index / constraint names changing to match what the database actually produced
- additional fields appearing in column metadata

So even if this behavior is technically intended, it is surprising from a workflow perspective and creates repository churn.

## Concrete example

In my case, applying migrations rewrote the snapshot with changes such as:

- table order shuffling
- `varchar(26) collate "C"` becoming `varchar(26)`
- `timestamptz` becoming `timestamptz(6)`
- `int` becoming `int4`
- unique/index/foreign key metadata being renamed or reshaped

The result was a large snapshot diff after `migration:up` even though no new migration had been created.

## Main request

Could the docs make the snapshot lifecycle more explicit?

In particular, it would help to document that:

- snapshots are not only written by `migration:create`
- `migration:up` and `migration:down` can rewrite them to match the actual database schema
- the snapshot is primarily internal migration state used for diffing/conflict detection, not necessarily a stable human-review artifact
- if teams want the committed snapshot to match the real database state, they may need to run migrations before committing

## Small UX improvement

If this behavior is intentional, a small CLI improvement could help prevent accidental commits of the metadata-derived snapshot state.

For interactive runs, `migration:create` could prompt:

```text
The snapshot may be updated to match the real database schema the next time you run migrations.
Would you like to apply pending migrations now? [Y/n]
```

For non-interactive runs, even a warning would help, for example:

```text
Snapshot was generated from entity metadata and may be rewritten after migration:up/down.
Run pending migrations before committing if you want the snapshot to reflect the real database schema.
```

Even if no prompt/warning is added, I think the docs should set expectations here more clearly.

## Longer-term possibilities

These are not the main request for this issue, but they may be useful future directions.

- Prisma's default development workflow (`prisma migrate dev`) creates and applies migrations together, while `--create-only` is the explicit opt-out for manual editing. That makes the provisional state easier to notice.
- Django also supports creating migrations in sequence, but it reconstructs state from the migration history itself rather than relying on a committed snapshot file with two different meanings.

I am not suggesting MikroORM copy either approach directly. I only mention them as examples that this general usability problem can be approached in different ways in the future.

Thanks.

### LINKED ISSUE #7610: fix(migrations): native enums duplicated per-table in migration snapshot
## Bug

Migration snapshot files (`.snapshot-*.json`) include a full copy of all native enum definitions for **every table** in the database, plus once at the schema level. For a database with many tables and many enums, the snapshot file is massively bloated.

## Root cause

In `packages/sql/src/schema/DatabaseSchema.ts`, `addTable()` assigns the schema-level `#nativeEnums` object to each `DatabaseTable` **by reference**:

```ts
// DatabaseSchema.addTable()
table.nativeEnums = this.#nativeEnums; // shared reference to all enums
```

This is intentional for in-memory use (so table-level operations can look up the full enum registry). However, at serialization time two separate `toJSON()` methods both emit the same data:

1. **`DatabaseSchema.toJSON()`** emits `nativeEnums` once at the top level ✓
2. **`DatabaseTable.toJSON()`** emits `nativeEnums` again for **every table** ✗

Since all tables share the same reference (the complete set of enums), a snapshot for a database with 50 tables and 20 enums contains 51 copies of all 20 enum definitions.

## Expected behaviour

Native enums should appear exactly once in the snapshot — at the schema level — where they already live in `DatabaseSchema.toJSON()`.

## Fix

1. Remove `nativeEnums` from `DatabaseTable.toJSON()` — it is redundant since the schema-level copy is canonical.
2. Update `Migrator.getSchemaFromSnapshot()` to wire up the shared reference from schema-level data (`rest.nativeEnums`) rather than from per-table data (`tbl.nativeEnums`), restoring the same shared-reference pattern used during live schema loading.

Old snapshots that still carry per-table `nativeEnums` continue to work because `rest.nativeEnums` (schema level) has always been present and has always contained the same data.

B4nan:
Running `migration:up`/`migration:down` rewrites the snapshot from the real DB via introspection, while `migration:create` writes it from entity metadata — the two serializations diverged enough that applying a migration after `migration:create` produced a large, noisy snapshot diff even though the schemas described the same thing (#7607).

Stabilize `DatabaseSchema.toJSON()` and `DatabaseTable.toJSON()` so both sources emit the same shape:

- Sort tables, views, namespaces, native enums, columns, indexes, checks and foreign keys by name with locale-independent string comparison.
- Pick only the declared fields from `IndexDef` / `ForeignKey` / `CheckDef`; always emit boolean `composite`/`constraint`/`primary`/`unique` flags on indexes; force `constraint: true` on primary indexes; drop the legacy singular `columnName` / `referencedColumnName` aliases that sqlite introspection emits; omit the `no action` default for update/delete rules.
- Normalize the column `type` through the platform's `normalizeColumnType`, so postgres collapses `int`/`int4`/`integer` and `serial`/`bigserial`/`smallserial` to the same canonical form on both paths. Add `double`/`double precision`/`float8` and `real`/`float4` aliases. Recognize multi-word SQL type names (`character varying`, `double precision`, `timestamp with time zone`) in `extractSimpleType` so the lookup tables that already listed them actually fire. Collapse postgres `timestamptz`/`timestamptz(N)` and treat `timetz` / `time with time zone` as `timetz` aliases in the schema diff.
- Drop `precision`/`scale` for integer- and float-family types.
- Derive column-level `primary`/`unique` from index membership, mirroring `DatabaseTable.init()`, so metadata and introspection agree on composite PK columns.
- Skip `unsigned` when the platform does not support it.
- Always emit table-level `comment: null` when unset.
- Treat the string literal `"null"` as an unset default.
- Normalize `length: 0` to `null`.

Sqlite-specific round-trip:

- Recover column length from the declared type (`time(3)` → `length: 3`) in `DatabaseTable.init`, scoped to types with a `getDefaultLength` so display-width artifacts like mysql's `tinyint(1)` don't start showing up as `length: 1` on booleans.
- Override `SqlitePlatform.getDefaultVersionLength()` to `0` — the datetime DDL drops precision and the current-timestamp expression hardcodes ms scaling, so the universal default of 3 was pure snapshot noise.
- Replace the regex-based `getEnumDefinitions` with `extractEnumValuesFromChecks`, walking the already-parsed `CheckDef` list (one fewer DB round-trip).
- Numeric enums stay as plain numeric columns on every dialect (matching the pre-#7607 behaviour and every other major ORM); `addColumnFromProperty` now routes numeric enums through their underlying numeric `mappedType` so the metadata snapshot agrees with introspection without any per-field strip workaround.

Native enum deduplication (#7610): `nativeEnums` lives on the schema by reference and was being serialized for every table — a registry of N enums in a database of M tables landed in the snapshot M+1 times. Drop it from `DatabaseTable.toJSON()` and wire every table back to the shared schema-level reference in `Migrator.getSchemaFromSnapshot()`. Older snapshots with per-table copies still load via the fallback.

Entity-generator: numeric enum values come out of introspection as `string[]` (there's no integer-array on `Column`), so generated code quoted them and the column round-tripped as a string enum. Detect the all-numeric case and emit raw numbers instead.

Also updates the migrations reference and getting-started guide to spell out the two-source lifecycle (`migration:create` from metadata, `migration:up`/`migration:down` from introspection).

Closes #7607
Closes #7610

### DIFF
--- a/docs/docs/guide/03-project-setup.md
+++ b/docs/docs/guide/03-project-setup.md
@@ -655,6 +655,8 @@ npx mikro-orm migration:list
 
 Creating new migration will automatically save the target schema snapshot into the migrations folder. This snapshot will be then used if you try to create a new migration, instead of using the current database schema. This means that if you try to create new migration before you run the pending ones, you still get the right schema diff.
 
+Note that `migration:up` and `migration:down` rewrite the snapshot from the real database schema via introspection, so the committed snapshot may drift from the metadata-derived form produced by `migration:create` until the migrations are actually applied. See the [Snapshots section](../migrations#snapshots) of the migrations reference for details.
+
 > Snapshots should be versioned just like the regular migration files.
 
 Snapshotting can be disabled via `migrations.snapshot: false` in the ORM config.
--- a/docs/docs/guide/03-project-setup.md
+++ b/docs/docs/guide/03-project-setup.md
@@ -655,8 +655,6 @@ npx mikro-orm migration:list
 
 Creating new migration will automatically save the target schema snapshot into the migrations folder. This snapshot will be then used if you try to create a new migration, instead of using the current database schema. This means that if you try to create new migration before you run the pending ones, you still get the right schema diff.
 
-Note that `migration:up` and `migration:down` rewrite the snapshot from the real database schema via introspection, so the committed snapshot may drift from the metadata-derived form produced by `migration:create` until the migrations are actually applied. See the [Snapshots section](../migrations#snapshots) of the migrations reference for details.
-
 > Snapshots should be versioned just like the regular migration files.
 
 Snapshotting can be disabled via `migrations.snapshot: false` in the ORM config.
--- a/docs/docs/migrations.md
+++ b/docs/docs/migrations.md
@@ -90,6 +90,13 @@ This will create the initial migration, containing the schema dump from `schema:
 
 Creating new migration will automatically save the target schema snapshot into migrations folder. This snapshot will be then used if you try to create new migration, instead of using current database schema. This means that if you try to create new migration before you run the pending ones, you still get the right schema diff.
 
+The snapshot file is written from two different sources depending on the command:
+
+- `migration:create` (and `migration:create --initial`) write the snapshot from the target schema derived from your entity metadata.
+- `migration:up` and `migration:down` rewrite the snapshot from the real database schema via introspection, after the migrations have been applied.
+
+The snapshot is primarily internal state used for diffing and conflict detection, not a stable human-review artifact. Running `migration:up`/`migration:down` may rewrite it with details produced by introspection (column type normalization, index and foreign key naming, table ordering, additional column metadata, etc.) that differ from the metadata-derived form written by `migration:create`. If you want the committed snapshot to reflect the real database state, run the pending migrations before committing.
+
 > Snapshots should be versioned just like the regular migration files.
 
 Snapshotting can be disabled via `migrations.snapshot: false` in the ORM config.
--- a/docs/docs/migrations.md
+++ b/docs/docs/migrations.md
@@ -92,10 +92,10 @@ Creating new migration will automatically save the target schema snapshot into m
 
 The snapshot file is written from two different sources depending on the command:
 
-- `migration:create` (and `migration:create --initial`) write the snapshot from the target schema derived from your entity metadata.
-- `migration:up` and `migration:down` rewrite the snapshot from the real database schema via introspection, after the migrations have been applied.
+- `migration:create` (and `migration:create --initial`) write it from the target schema derived from your entity metadata.
+- `migration:up` and `migration:down` rewrite it from the real database schema via introspection, after the migrations have been applied.
 
-The snapshot is primarily internal state used for diffing and conflict detection, not a stable human-review artifact. Running `migration:up`/`migration:down` may rewrite it with details produced by introspection (column type normalization, index and foreign key naming, table ordering, additional column metadata, etc.) that differ from the metadata-derived form written by `migration:create`. If you want the committed snapshot to reflect the real database state, run the pending migrations before committing.
+Both paths produce the same serialized shape for a given schema, so running migrations after `migration:create` does not normally rewrite the snapshot in a meaningful way. Some DB-specific details (e.g. postgres type aliases like `int` vs `int4`) can still produce small diffs.
 
 > Snapshots should be versioned just like the regular migration files.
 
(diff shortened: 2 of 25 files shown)


enduity:
Thanks again for working on this.

I tested the version from this PR against [my repro repo](https://github.com/enduity/mikro-orm-issue-7607-snapshot-reproduction), and the unexpected diffs seem to be gone 🎉

One thing I wanted to point out, though: normalizing `columnType` can affect unsupported custom workflows that rely on it being taken literally in the snapshot/diff path. For example, in the repro repo, I currently [use this workaround](https://github.com/enduity/mikro-orm-issue-7607-snapshot-reproduction/blob/fddd5314b35ceebb73b88dce5ba5de90b7646e13/src/entities/base.entity.ts#L7) for column collations:

```ts
@PrimaryKey({ type: 'varchar', length: 26, columnType: 'varchar(26) collate "C"' })
```

What’s good is that this still gets emitted correctly when the column is first created, so it is not silently stripped from the initial migration.

However, a limitation shows up later, when an existing column is modified. If that custom `columnType` is added or removed after the initial metadata-based migrations have already been created and applied, `migration:create` no longer sees it as a schema change, so it cannot generate a migration for the collation change. (To be fair, even if it did generate one, I'm not sure it would work as intended)

I’m not suggesting this should block the PR, since this is an unsupported workaround and it only affects migration generation, not runtime behavior. Still, it may be another argument for eventually adding native column collation support:
- https://github.com/mikro-orm/mikro-orm/issues/4286

Also, there may be other `columnType`-based workarounds with similar edge cases that we have not considered.

B4nan:
#7615 🙃

enduity:
Good work, thank you for implementing this!

PR #7612, opened 2026-04-24
fix(core): TPT metadata drop when subclass registered before abstract

### LINKED ISSUE #7609: TPT schema generator silently drops subclass when registered before its abstract parent
### Describe the bug

When using Table-Per-Type (TPT) inheritance, a concrete subclass is silently excluded from the generated schema (and from the final `MetadataStorage`) if it appears before its abstract parent in the `entities` array.

This happens in practice whenever entity discovery relies on a file glob such as `entities: ['dist/**/*.entity.js']`, because file-system glob order is alphabetical 
— e.g. `comment-like.entity.js` is loaded before `like.entity.js`. The bug is order-dependent and has no warning/error: `discovery finished, found N entities` still reports the correct count, but the generated CREATE SCHEMA SQL is missing one of the child tables and `MetadataStorage` never links it as a TPT child (`inheritanceType: undefined`, `tptParent: undefined`).

Swapping the same entities so that the abstract parent comes first makes everything work. Registration order should not affect correctness of schema generation.

If this is confirmed as a bug, I am happy to submit a PR — would appreciate a pointer to where TPT parent-child resolution happens in MetadataDiscovery (I suspect `initTPTRelationships` / `finalizeTPTInheritance`).

### Reproduction

https://github.com/ParkGiraffe/mikro-orm-tpt-ordering-repro

None

### MikroORM version

@mikro-orm/core 7.0.13-dev.2

### Node.js version

Node.js : 24.11.1, typescript : 5.9.3

### Operating system

macOS (Darwin 25.3.0, arm64). Also reproduced on Ubuntu (GitHub Actions CI).

### Validations

ParkGiraffe:
When using @Entity({ abstract: true, inheritance: 'tpt' }), the @Entity decorator did not set meta.name (it was skipped for abstract entities). As a result the TPT root entity had name=undefined, causing the MetadataDiscovery filter (meta.root.name) to silently drop the entire TPT hierarchy — root and all concrete subclasses.

Two fixes:
- @Entity decorator (legacy + ES): set meta.name even for abstract entities when inheritance: 'tpt' is present, since TPT root entities always have their own table.
- MetadataDiscovery.initTPTRelationships: mirror the STI pattern by checking the raw `inheritance` decorator option (always available) alongside inheritanceType (set later during discovery), so child entities are handled correctly regardless of registration order. Also use `??=` for tptChildren to avoid clearing already-registered children.

Closes #7609

### DIFF
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -1718,10 +1718,10 @@ export class MetadataDiscovery {
 
     if (inheritance === 'tpt' && meta.root === meta) {
       meta.inheritanceType = 'tpt';
-      meta.tptChildren = [];
+      meta.tptChildren ??= [];
     }
 
-    if (meta.root.inheritanceType !== 'tpt') {
+    if (meta.root.inheritanceType !== 'tpt' && (meta.root as Dictionary).inheritance !== 'tpt') {
       return;
     }
 
--- a/packages/decorators/src/es/Entity.ts
+++ b/packages/decorators/src/es/Entity.ts
@@ -11,7 +11,7 @@ export function Entity<Owner extends EntityClass<unknown> & EntityCtor>(
     Utils.mergeConfig(meta, metadata, options);
     meta.class = target as unknown as Constructor<Owner>;
 
-    if (!options.abstract || meta.discriminatorColumn) {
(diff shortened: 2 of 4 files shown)


ParkGiraffe:
Closing — I wasn't aware @B4nan had already fixed this. Thanks for the quick fix!

PR #7636, opened 2026-04-28
fix(core): avoid crash on STI subclass with 1:M override

### LINKED ISSUE #7635: [BUG] STI subclass @OneToMany override crashes metadata discovery
When two STI subclasses both declare a `@OneToMany` with the same property name but different target entities, `MikroORM.init` crashes:

```
TypeError: rootProp.fieldNames is not iterable
    at MetadataDiscovery.js:1339:49
    at MetadataDiscovery.initSingleTableInheritance (MetadataDiscovery.js:1311)
```

Looks like the inverse-side counterpart of #7598. The fix in #7599 added `sameRelationTargetRoot()` to skip the rename branch for narrowed owning-side overrides, but it bails out for anything that isn't `MANY_TO_ONE` or `ONE_TO_ONE`:

```ts
if (prop.kind !== ReferenceKind.MANY_TO_ONE && prop.kind !== ReferenceKind.ONE_TO_ONE) {
  return false;
}
```

So OneToMany falls through into the rename branch, hits `initFieldName(rootProp, ...)` which doesn't set `fieldNames` for inverse-side props, and then `[...rootProp.fieldNames]` blows up.

The targets in the repro aren't narrowing the same hierarchy either — they're entirely separate entities — so the rename branch shouldn't run for them at all.

### Expected behavior

Discovery should succeed. OneToMany on STI siblings has no column to merge, so it should just inherit.

### Reproduction

`npm i @mikro-orm/core@7.0.13 @mikro-orm/postgresql@7.0.13` and run:

```js
import { MikroORM, defineEntity } from '@mikro-orm/core';

const p = defineEntity.properties;

const Animal = defineEntity({
  name: 'Animal',
  abstract: true,
  discriminatorColumn: 'type',
  properties: {
    id: p.string().primary(),
    type: p.string(),
    name: p.string(),
  },
});

const DogToy = defineEntity({
  name: 'DogToy',
  properties: {
    id: p.string().primary(),
    name: p.string(),
    owner: () => p.manyToOne(Dog),
  },
});

const CatToy = defineEntity({
  name: 'CatToy',
  properties: {
    id: p.string().primary(),
    name: p.string(),
    owner: () => p.manyToOne(Cat),
  },
});

const Dog = defineEntity({
  name: 'Dog',
  extends: Animal,
  discriminatorValue: 'DOG',
  properties: {
    toys: () => p.oneToMany(DogToy).mappedBy('owner'),
  },
});

const Cat = defineEntity({
  name: 'Cat',
  extends: Animal,
  discriminatorValue: 'CAT',
  properties: {
    toys: () => p.oneToMany(CatToy).mappedBy('owner'),
  },
});

const { PostgreSqlDriver } = await import('@mikro-orm/postgresql');
await MikroORM.init({
  driver: PostgreSqlDriver,
  entities: [Animal, Dog, Cat, DogToy, CatToy],
  dbName: 'never_connects',
  connect: false,
});
```

@mikro-orm/postgresql

### MikroORM version

7.0.13

### Node.js version

24.15.0

### Operating system

Windows 11

### Validations

belmeopmenieuwesim:
STI children that declare an `@OneToMany` with a different target than the root crashed during metadata discovery with `TypeError: rootProp.fieldNames is not iterable`. The rename branch in `initSingleTableInheritance`, which exists to disambiguate physical FK columns across STI children, was being entered for inverse-side relations — but inverse-side properties never populate `fieldNames`, so spreading them blew up.

This handles both narrowing within the same hierarchy (e.g. `Dog extends Pet` narrows `toys: Toy` to `DogToy`) and disjoint targets (e.g. `Dog.items: DogItem` / `Cat.items: CatItem` with no shared base). `sameRelationTargetRoot` now recognizes 1:m and inverse m:n so narrowed overrides flow through the existing `narrowedRelationOverride` path and preserve the root's broad target; for disjoint overrides, each child keeps its own local typed property rather than letting the last-processed child clobber the root's declaration.

Closes #7635

### DIFF
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -1560,7 +1560,12 @@ export class MetadataDiscovery {
       return false;
     }
 
-    if (prop.kind !== ReferenceKind.MANY_TO_ONE && prop.kind !== ReferenceKind.ONE_TO_ONE) {
+    if (
+      prop.kind !== ReferenceKind.MANY_TO_ONE &&
+      prop.kind !== ReferenceKind.ONE_TO_ONE &&
+      prop.kind !== ReferenceKind.ONE_TO_MANY &&
+      prop.kind !== ReferenceKind.MANY_TO_MANY
+    ) {
       return false;
     }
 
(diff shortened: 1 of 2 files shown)


B4nan (approved):
thanks!

PR #7642, opened 2026-04-28
refactor: preserve JsonType for json/jsonb properties with array TS type

### LINKED ISSUE #7641: @Property({ type: 'json' }) with T[] runtime type writes Postgres-array literals instead of JSON since #7378
JSON-typed properties whose TypeScript type is an array (e.g. `string[]`, `Foo[]`) are mis-mapped to `ArrayType` instead of `JsonType` when `TsMorphMetadataProvider` is used. Inserts then go through `Platform.marshallArray()` instead of `JsonType.convertToDatabaseValue` (`JSON.stringify`).

This affects **all of these variants** equally:

- `@Property({ type: 'json' }) foo!: T[]`
- `@Property({ type: 'jsonb' }) foo!: T[]`
- `@Property({ columnType: 'json' }) foo!: T[]`
- `@Property({ columnType: 'jsonb' }) foo!: T[]`
- `@Property({ type: JsonType }) foo!: T[]`
- `T` may be a primitive (`string`, `number`) or an object — both are affected.

The bug is **not Postgres-specific** — it's generic at the metadata layer. Symptom severity differs by driver:

| Driver | What `Platform.marshallArray` writes | Symptom |
|---|---|---|
| PostgreSQL | `{a,b}` (PG array literal) into `json`/`jsonb` | Hard error: `invalid input syntax for type json — Expected ":", but found ","` (non-empty); silent corruption to `{}` (empty array becomes object) |
| MySQL / MariaDB | `a,b` into `JSON` column | Hard error from MySQL JSON validator |
| SQLite / MSSQL | `a,b` into TEXT-backed column declared as `json` | Silent: ArrayType handles read+write so the entity round-trips, but the stored value is not valid JSON — breaks any external consumer or `JSON_*` SQL function |

Regression introduced in v7.1.0-dev by #7378 ("feat(core): support `array` option on scalar properties"). Last known good: v7.0.13.

### Reproduction

```ts
import { Entity, PrimaryKey, Property } from '@mikro-orm/core';
import { MikroORM, PostgreSqlDriver } from '@mikro-orm/postgresql';
import { TsMorphMetadataProvider } from '@mikro-orm/reflection';

@Entity()
class Item {
  @PrimaryKey() id!: number;
  @Property({ type: 'json' }) features!: string[];
}

  driver: PostgreSqlDriver,
  entities: [Item],
  metadataProvider: TsMorphMetadataProvider,
  dbName: 'repro',
});
await orm.schema.refresh();

const em = orm.em.fork();
em.create(Item, { features: ['a', 'b'] });
await em.flush();
// → DriverException: invalid input syntax for type json
//   Detail: Expected ":", but found ","

em.create(Item, { features: [] });
await em.flush();
// → INSERT succeeds, but the column round-trips as `{}` (object), not `[]` (array).
```

@mikro-orm/postgresql

### MikroORM version

7.1.0-dev.23

### Node.js version

22.x

### Operating system

Windows 11

### Validations

belmeopmenieuwesim:
The `array: true` wrap pass added in #7378 re-wraps any already-set
customType in ArrayType. With TsMorphMetadataProvider, reflection sets
`prop.array = true` from the TS array type, so a `@Property({ type: 'json' })`
on `string[]` had its JsonType silently re-wrapped — inserts then went
through `Platform.marshallArray()` and produced invalid JSON in the
column (hard SQL error on Postgres/MySQL, silent corruption on
SQLite/MSSQL).

Closes #7641

### DIFF
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -2333,6 +2333,7 @@ export class MetadataDiscovery {
       prop.customType &&
       !(prop.customType instanceof t.array) &&
       !(prop.customType instanceof t.enumArray) &&
+      !(prop.customType instanceof t.json) &&
       prop.kind === ReferenceKind.SCALAR
     ) {
       const innerType = prop.customType;
--- a/tests/issues/GH7641.test.ts
+++ b/tests/issues/GH7641.test.ts
@@ -1,0 +1,35 @@
+import { JsonType, MikroORM } from '@mikro-orm/sqlite';
+import { TsMorphMetadataProvider } from '@mikro-orm/reflection';
+import { Entity, PrimaryKey, Property } from '@mikro-orm/decorators/legacy';
+
+@Entity()
+class A {
+  @PrimaryKey()
+  id!: number;
+
+  @Property({ type: 'json' })
+  features!: string[];
+}
+
+@Entity()
+class B {
+  @PrimaryKey()
+  id!: number;
+
+  @Property({ type: 'jsonb' })
+  features!: string[];
+}
+
+test('GH issue 7641: json/jsonb properties with array TS type stay mapped to JsonType under TsMorph', async () => {
+  const orm = await MikroORM.init({
+    metadataProvider: TsMorphMetadataProvider,
+    metadataCache: { enabled: false },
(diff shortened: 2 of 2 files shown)


B4nan (approved):
thanks!

PR #7650, opened 2026-05-03
docs(sqlite): add Cloudflare Durable Objects section

jeffwilde:
Adds a `## Using Cloudflare Durable Objects` subsection to `docs/docs/usage-with-sqlite.md`, parallel to the existing D1 / Bun SQLite / Turso subsections.

Cloudflare Durable Objects provide per-instance SQLite storage via `ctx.storage.sql`. The generic `SqliteDriver` from `@mikro-orm/sql` works with [`kysely-durable-objects`](https://github.com/jeffwilde/kysely-durable-objects) (a Kysely dialect for the DO storage runtime) to make MikroORM work end-to-end inside a Durable Object.

The new section covers:

- Constructor pattern with lazy ORM init via `??=`
- Why `dbName` is required by the API but ignored at runtime (no connection string — the database is the one tied to this DO instance)
- The two MikroORM-config requirements: `implicitTransactions: false` (the dialect raises on `BEGIN`) and `compiledFunctions` (workerd blocks `new Function()` at request time)
- A short Limitations subsection. Notably, unlike D1, query streaming **does** work — DO storage is in-process SQLite with a real cursor.

Marked **Experimental** to match the convention used for D1.

No source changes. Pure docs.

### DIFF
--- a/docs/docs/usage-with-sqlite.md
+++ b/docs/docs/usage-with-sqlite.md
@@ -121,6 +121,55 @@ D1 has significant limitations compared to regular SQLite:
 
 See the [D1 SQL documentation](https://developers.cloudflare.com/d1/sql-api/sql-statements/) for more details on supported SQL statements.
 
+## Using Cloudflare Durable Objects
+
+> **Experimental:** Durable Objects SQLite support is experimental. Use with caution.
+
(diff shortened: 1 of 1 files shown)


B4nan (approved):
thanks!

PR #7684, opened 2026-05-08
fix(core): apply filters to inverse 1:1 populate joins instead of duplicating them

### LINKED ISSUE #7680: Filter on abstract TPT base entity causes eager OneToOne relation to not be populated
### Describe the bug

_See repro_: https://github.com/stoneLeaf/mikro-orm-v7-repros/tree/tpt-base-filter-breaks-eager-onetoone-population

### Context

Abstract TPT base entity with a soft-delete filter, a "part" entity with inverse OneToOne relations back to two concrete subtypes, and those subtypes each holding an eager owner-side OneToOne to the part entity:

```TypeScript
const PersonSchema = defineEntity({
  name: "Person",
  abstract: true,
  inheritance: "tpt",
  filters: {
    excludeDeleted: { name: "excludeDeleted", cond: { deletedAt: null }, default: true },
  },
  properties: {
    id: p.integer().primary(),
    name: p.string(),
    deletedAt: p.datetime().nullable(),
  },
});

const BadgeSchema = defineEntity({
  name: "Badge",
  properties: {
    id: p.integer().primary(),
    code: p.string(),
    parentEmployee: () => p.oneToOne(Employee).mappedBy("badge").ref().nullable(),
    parentContractor: () => p.oneToOne(Contractor).mappedBy("temporaryBadge").ref().nullable(),
  },
});

// Concrete subtype with an eager owner-side OneToOne to Badge
const EmployeeSchema = defineEntity({
  name: "Employee",
  extends: Person,
  properties: {
    id: p.integer().primary(),
    badge: () => p.oneToOne(Badge).owner().eager(),
  },
});
```

### Issue

Fetching an `Employee` no longer populates the eager `badge` relation:

```TypeScript
const found = await em.findOneOrFail(Employee, { id: 1 }, { populate: ["*"] });

found.badge; // null — should be a Badge instance
```

The debug log reveals the cause. Without the filter the second query (loading `Badge`) produces clean left joins for the inverse back-references:

```sql
-- Without filter (filters: false) — badge IS populated ✓
select `b0`.*,
  `p2`.`id` as `p1__id`, `p2`.`name` as `p1__name`, `p2`.`deleted_at` as `p1__deleted_at`,
  `p1`.`badge_id` as `p1__badge_id`,
  `p4`.`id` as `p3__id`, `p4`.`name` as `p3__name`, `p4`.`deleted_at` as `p3__deleted_at`,
  `p3`.`temporary_badge_id` as `p3__temporary_badge_id`
from `badge` as `b0`
left join (`person_employee` as `p1` inner join `person` as `p2` on `p1`.`id` = `p2`.`id`)
  on `b0`.`id` = `p1`.`badge_id`
left join (`person_contractor` as `p3` inner join `person` as `p4` on `p3`.`id` = `p4`.`id`)
  on `b0`.`id` = `p3`.`temporary_badge_id`
where `b0`.`id` in (1)
```

With the filter active, MikroORM emits **duplicate JOINs** for the same two tables — one filtered pair (`p1`/`p3`, with `deleted_at is null` on the JOIN condition) and one unfiltered pair (`p5`/`p7`):

```sql
-- With filter (default: true) — badge is NOT populated ✗
select `b0`.*,
  `p2`.`id` as `p1__id`, `p2`.`name` as `p1__name`, `p2`.`deleted_at` as `p1__deleted_at`,
  `p1`.`badge_id` as `p1__badge_id`,
  `p4`.`id` as `p3__id`, `p4`.`name` as `p3__name`, `p4`.`deleted_at` as `p3__deleted_at`,
  `p3`.`temporary_badge_id` as `p3__temporary_badge_id`,
  `p6`.`id` as `p5__id`,   -- duplicate, unfiltered — only selects id
  `p8`.`id` as `p7__id`    -- duplicate, unfiltered — only selects id
from `badge` as `b0`
left join (`person_employee` as `p1` inner join `person` as `p2` on `p1`.`id` = `p2`.`id`)
  on `b0`.`id` = `p1`.`badge_id` and `p2`.`deleted_at` is null        -- filtered
left join (`person_contractor` as `p3` inner join `person` as `p4` on `p3`.`id` = `p4`.`id`)
  on `b0`.`id` = `p3`.`temporary_badge_id` and `p4`.`deleted_at` is null  -- filtered
left join (`person_employee` as `p5` inner join `person` as `p6` on `p5`.`id` = `p6`.`id`)
  on `b0`.`id` = `p5`.`badge_id`                                       -- duplicate, unfiltered
left join (`person_contractor` as `p7` inner join `person` as `p8` on `p7`.`id` = `p8`.`id`)
  on `b0`.`id` = `p7`.`temporary_badge_id`                             -- duplicate, unfiltered
where `b0`.`id` in (1)
```

The unfiltered duplicate joins (`p5`/`p7`) select only `id` — no FK columns. They appear to be used by MikroORM's hydration for relationship detection, and cause MikroORM to hydrate the `badge` property on `Employee` as `null`.

### Expectation

The `badge` relation should be populated regardless of whether a filter is applied to the base TPT entity. The two queries should be identical except for the `deleted_at is null` condition on the first query's `WHERE` clause.

### Reproduction

https://github.com/stoneLeaf/mikro-orm-v7-repros/tree/tpt-base-filter-breaks-eager-onetoone-population

@mikro-orm/postgresql

### MikroORM version

7.0.14

### Node.js version

22.22.1

### Operating system

Win

### Validations

B4nan:
## Summary

When a filter was active on the target of an inverse OneToOne relation, the existing populate hint produced a JOIN (forced by `joinedProps()`) but the filter was applied via a separate `:ref` hint that created a redundant JOIN. With TPT inheritance the duplicate joins confused hydration, so the eager relation was hydrated as `null`.

The fix recognises that inverse OneToOne always produces a JOIN regardless of the load strategy, so the filter can be applied to the existing JOIN without needing a duplicate `:ref` join.

Closes #7680

### DIFF
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -553,7 +553,9 @@ export class EntityManager<Driver extends IDatabaseDriver = IDatabaseDriver> {
         prop.strategy || hint.strategy || options.strategy || this.config.get('loadStrategy'),
         prop.kind,
       );
-      const joined = strategy === LoadStrategy.JOINED && prop.kind !== ReferenceKind.SCALAR;
+      const joined =
+        (strategy === LoadStrategy.JOINED || (prop.kind === ReferenceKind.ONE_TO_ONE && !prop.owner)) &&
+        prop.kind !== ReferenceKind.SCALAR;
 
       if (!joined && !hint.filter) {
         continue;
@@ -645,16 +647,18 @@ export class EntityManager<Driver extends IDatabaseDriver = IDatabaseDriver> {
         let found = false;
 
         for (const hint of populated) {
-          if (!hint.all) {
-            hint.filter = true;
-          }
-
           const strategy = getLoadingStrategy(
             prop.strategy || hint.strategy || options.strategy || this.config.get('loadStrategy'),
             prop.kind,
           );
+          // inverse 1:1 always produces a JOIN (forced by `joinedProps()`), regardless of strategy
+          const willJoin = strategy === LoadStrategy.JOINED || (prop.kind === ReferenceKind.ONE_TO_ONE && !prop.owner);
 
-          if (hint.field === `${prop.name}:ref` || (hint.filter && strategy === LoadStrategy.JOINED)) {
+          if (!hint.all || willJoin) {
+            hint.filter = true;
+          }
+
+          if (hint.field === `${prop.name}:ref` || (hint.filter && willJoin)) {
             found = true;
           }
         }
--- a/tests/issues/GH7680.test.ts
+++ b/tests/issues/GH7680.test.ts
@@ -1,0 +1,65 @@
+import { defineEntity, MikroORM, p } from '@mikro-orm/sqlite';
+
+const PersonSchema = defineEntity({
+  name: 'Person',
+  abstract: true,
+  inheritance: 'tpt',
+  filters: {
+    excludeDeleted: { name: 'excludeDeleted', cond: { deletedAt: null }, default: true },
+  },
+  properties: {
+    id: p.integer().primary(),
+    name: p.string(),
+    deletedAt: p.datetime().nullable(),
+  },
+});
+
+const BadgeSchema = defineEntity({
+  name: 'Badge',
+  properties: {
+    id: p.integer().primary(),
+    code: p.string(),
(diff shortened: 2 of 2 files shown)


stoneLeaf:
@B4nan Confirmed working, thanks again!

PR #7693, opened 2026-05-10
feat(core): support stored procedures and functions

### LINKED ISSUE #5253: Represent stored routes and provide type safe way to execute them

In some cases, it may be appropriate to define functionality in stored procedures or functions, particularly in otherwise chatty series of queries that don't need additional input beyond the initial one. In PostgreSQL, there's also the fact that stored functions are what's needed to be attached to triggers.

I'm thinking something along the lines of

```typescript
@Routine<AddRecord>({
  type: 'procedure',// required. type: 'procedure' | 'function'
  name: 'add_record',//optional. Respects naming strategy as for entities when omitted
  comment: 'Adds a record, and returns the newly added record',//optional
  security: 'invoker',//optional
  definer: 'root',//optional

  deterministic: true,//optional. Explicitly sets the (NOT) DETERMINISTIC flag if provided

//optional. Sets the data access (for MySQL, that's CONTAINS SQL, NO SQL, READS SQL DATA, or MODIFIES SQL DATA).
  dataAccess: 'MODIFIES SQL DATA',

// Required for functions, and defines the returned type.
// For procedures, if present, defines the type of result set
// (i.e. output is always an array of this type, even if guaranteed to be a single record by the query).
// Multiple result sets can be defined as an array where each member is the type of that result set.
// Whenever an entity type is specified, lazy scalars always get resolved if returned by the procedure, and references are not loaded.
// Returned entities themselves load/refresh the identity map and result caches, just like a regular find() call with an empty populate.
// Users wishing to have procedures return certain shapes without an associated entity can define virtual entities for use here.
  returns: () => RecordEntity,

//optional. Default is driver specific. Mostly useful in PostgreSQL, where one can define custom references to C libs
  language: 'SQL',

// Optional for routines in language other than SQL.
// can be a string that is the procedure body,
// or a function that takes the name of the inputs of the class itself, similarly to generated column decorator,
// and the entity manager that is creating the procedure (not the caller! this function is not evaluated at call time).
// This is mostly to enable the use of entity aware query builder.
// Returns the body of the procedure to be created.
  body:  (params, em) => `BEGIN
SET ${params.hash} = SHA1(CONCAT(${params.name}, ${params.age}, 'secret salt'));
INSERT INTO record_entity SET hash = ${params.hash}, name = ${params.name}, age = ${params.age};
SELECT * FROM record_entity WHERE hash = ${params.hash};
END`
});
export class AddRecord{
  @Property({columnType: 'tinyint'});
  age: number;

  @Property({columnType: 'varchar(255)'});
  name: string;

  // Special decorator, only for routines. If out is set to true, this param becomes out param.
  // Without this decorator, params are in params.
  // Params that are both in and out can be defined with explicit `@Param({in: true, out: true})`
  // Params that are out or in/out params need to be wrapped with Ref, to enable their post call fill.
  // Params that are out are also implicitly optional.
  // Params that are in or in/out params are only optional if they have default values defined.
  @Param({out: true});
  @Property({columnType: 'char(40)'});
  hash: Ref<string>;
}
```

and then when calling
```typescript
// prepare holder for out param.
const hash = new Ref<string>();
// Type of newRecord inferred from returns, potentially void if the procedure does not have a returns defined.
const newRecord = await orm.em.callProcedure(
  AddRecord,
  /*type hints from class.*/
  {name: 'Jon Snow', years: 30, hash}
);

/* resolves to a set of SQL statements like

START TRANSACTION; -- Only if not already in a transaction, or if nested transactions are supported.
SET SESSION `hash` = NULL; -- Set to initial value of reference, or NULL if not provided.
CALL add_record('John Snow', 30, @`hash`);
SELECT @`hash`; -- Only if ref is provided in callProcedure
SET SESSION `hash` = NULL; -- Always set to NULL in SQL, to prevent data leaks in other statements.
COMMIT;  -- Only if not already in a transaction, or if nested transactions are supported.

*/

// do something else with newRecord entity, including f.e. populating its relations.

// Now that we called the procedure, this will not be undefined.
// In this particular example, we could've gotten the same value out of newRecord instead, but for the sake of example...
const loadedHash = hash.unwrap();
```

And of course, it goes without saying that eventually, the schema generator and entity generator should be able to handle those definitions, so that these can be versioned along with the rest of the DB.

The way to use those right now is to define them manually, perhaps in separately maintained and versioned SQL files, and use raw queries to manually map the results back. Extremely inconvenient, which means most people would rather never use them, even if appropriate.

**Additional context**
Related to #5053

B4nan:
## Summary

First-class support for declaring, managing, and invoking stored procedures and functions. Closes #5253.

Shipping as **experimental** — the metadata shapes may evolve in a patch release while we gather real-world feedback.

### Declaration

Routines are declared with the `Routine` class from `@mikro-orm/core` and registered via a dedicated `routines` config option, separate from `entities` (the same way `subscribers` is its own slot):

```ts
const HashUser = new Routine({
  name: 'hash_user',
  type: 'function',
  params: { name: { type: 'varchar(255)' }, salt: { type: 'varchar(255)' } },
  returns: { runtimeType: 'string', columnType: 'char(40)' },
  body: 'SELECT SHA1(CONCAT(name, salt))',
});

await MikroORM.init({ entities: [User], routines: [HashUser] });

// args typed as `{ name: string; salt: string }`, result typed as `string`:
const hash = await em.callRoutine(HashUser, { name: 'jon', salt: 'pepper' });
```

No decorator form — routines are callables, not entities; the class body would always be empty, and a decorator would invite users to put `@Property()` fields on it expecting them to mean something. The single `Routine` class keeps the surface honest and carries the same arg/return type generics.

`TArgs` and `TReturn` are inferred from the literal config so `em.callRoutine` is fully typed without threading generics through the call site. For cases where inference is too loose (typically `runtimeType: 'object'` returns that should be a concrete shape rather than `Dictionary`), `Routine.create<TArgs, TReturn>(config)` accepts explicit override generics.

### TS type inference

The argument shape inferred for `em.callRoutine(R, args)` resolves in this order, per param:

1. explicit `runtimeType` (`'string' | 'number' | 'boolean' | 'bigint' | 'Buffer' | 'Date' | 'object' | 'any'`) — wins when set
2. `type` set to a `Type` class or instance — JSType extracted from the `Type<JSType, DBType>` generic
3. `type` set to a SQL string — mapped via `SqlTypeToTs<S>` (`varchar(255)`, `text`, `uuid`, `decimal`, `numeric`, `money`, `bigint` → `string`; `int`, `smallint`, `real`, `double precision` → `number`; `boolean`, `bool`, `bit`, `tinyint(1)` → `boolean`; `timestamp`, `date`, `time` → `Date`; `json`, `jsonb` → `Dictionary`; `bytea`, `blob`, `binary` → `Buffer`). `decimal`/`numeric`/`money`/`bigint` default to `string` because drivers typically return them as strings to preserve precision — pass `runtimeType: 'number'` or `runtimeType: 'bigint'` to opt into a different type when the value range is safe.
4. fallback `any`

`ref: true` wraps the value in `ScalarReference`; `nullable: true` adds `| null`.

Type bench in `tests/bench/types/routines.ts` locks instantiation counts for the realistic call-site usage (12-param wide proc, mixed primitives, Type class at `type`, `ref:` for OUT params).

### Schema management

Per-dialect `SchemaHelper.createRoutine` / `dropRoutine` / `getAllRoutines`:

- **PostgreSQL** + **PGlite** (inherits): `pg_proc` + `pg_get_functiondef`
- **MySQL** / **MariaDB**: `information_schema.routines` + `parameters`
- **MSSQL**: `sys.sql_modules` + `sys.parameters` (return type from the `parameter_id = 0` row); T-SQL `@`-prefixed body params; routine DDL in its own batch
- **Oracle**: `USER_PROCEDURES` + `USER_SOURCE` + `USER_ARGUMENTS` (return type from the `POSITION = 0` row, `PACKAGE_NAME is null` filter); unconstrained PL/SQL formal param types; `drop … if exists` for idempotent refresh
- **SQLite**: silent-skip schema-side — no DDL emitted, `schema:diff` produces no churn
- **libSQL**: silent-skip schema-side; runtime calls throw (the libsql client doesn't implement UDF registration)

`SchemaComparator` detects added/changed/removed routines with platform-aware type normalisation (`int` ↔ `integer`, `varchar` ↔ `character varying`, etc.) and body normalisation that strips outer `BEGIN ... END` wrappers and trailing semicolons. MSSQL folds metadata-side `direction: 'out'` to `'inout'` via the new `SchemaHelper.normaliseRoutineParamDirection` hook — T-SQL has no distinct OUT-only direction at the catalog level, and otherwise pure-OUT declarations would churn on every diff. `stripStatementNewlines` keeps line comments and multi-line string literals in routine bodies intact while still preventing the schema-generator's `;\n` statement splitter from breaking the DDL apart.

### Runtime — `em.callRoutine(routine, args)`

- **PostgreSQL**: `CALL proc(...)` returned row → ScalarReference for OUT/INOUT
- **MySQL / MariaDB**: session-variable bridge (`SET @v := ?; CALL proc(?, @v); SELECT @v`). When the procedure has OUT/INOUT params and the caller is not already inside a `ctx`, the three statements are pinned to a single pool connection via an implicit transaction — MySQL `@vars` are connection-scoped, so without this they could split across pool connections and silently read NULL.
- **MSSQL**: batched `DECLARE @v ...; SET @v = ?; EXEC dbo.proc ?, @v OUTPUT; SELECT @v`
- **Oracle**: anonymous PL/SQL block via `oracledb` with `BIND_INOUT` / `BIND_OUT`. Throws if invoked inside `em.transactional` — Oracle's routine path needs its own pool connection with `autoCommit: true` for refcursors and OUT binds, so it can't share the EM transaction. `dataAccess: 'reads-sql-data'` / `'no-sql'` functions are allowed through.
- **SQLite**: functions bridged via `better-sqlite3`'s `db.function()` UDF when a `bodyJs` JS fallback is provided; procedures throw
- **libSQL**: throws (`libsql.Database.function()` is in the type defs but not implemented at runtime)

The `select <qualified>(?, ?…)` scalar-function path and the OUT-`ScalarReference.set()` loop live as shared `callRoutineFunction` / `applyRoutineOutParams` helpers on the base `Connection` class, used by every SQL driver. `convertRoutineInbound` / `convertRoutineOutbound` are sibling protected methods marshalling values through `Type.convertToDatabaseValue` / `convertToJSValue` when a `customType` (or `Type` at `type`) is declared.

### Custom types on params and returns

The canonical way to attach a `Type` to a routine param or scalar return is the `type` field — a `Type` class or instance drives all three sides at once: SQL column type (via `Type.getColumnType`), TS type (extracted from the `Type<JSType, DBType>` generic), and marshalling.

```ts
const CountItems = new Routine({
  name: 'count_items',
  type: 'function',
  params: { data: { type: JsonType } },
  returns: { type: StringType },
  body: '...',
});
```

The longer `runtimeType + columnType + customType` triple is still supported when you genuinely need to split column type, runtime type, and marshaller — e.g. a `Type<string, string>` on top of a non-default column type.

### Multi-result-set procedures (no metadata declaration required)

The drivers auto-detect end-of-stream so users don't have to count result sets upfront. When a procedure emits any result sets, `em.callRoutine` returns `Dictionary[][]` (one row array per set); otherwise it returns `void`.

- **MySQL / MariaDB**: native — proc body has N `SELECT`s; mysql2's response array carries each set, terminated by an OkPacket the connection filters out
- **PostgreSQL**: declare `refcursor` OUT params and the connection auto-FETCHes each cursor (callers must wrap in `em.transactional` so the cursors stay alive)
- **Oracle**: declare `sys_refcursor` OUT params (oracledb `DB_TYPE_CURSOR` binds)
- **MSSQL**: not surfaced — tedious flattens recordsets through the kysely wrapper without per-set boundaries. Use `em.getConnection().execute()` for raw access.
- **SQLite** / **libSQL** / **Mongo**: no stored-procedure concept, throws

### Reverse-engineering via `generate-entities`

The entity generator emits routine source code alongside entities. `EntityGenerator` calls `schema.loadRoutines()` explicitly so introspected routines reach the emitter, and `RoutineSourceFile` always emits `export const X = new Routine({...})` regardless of `entityDefinition` mode — there is no decorator form. Routine bodies are escaped properly: backslashes and `${...}` template-literal interpolations in user-supplied SQL no longer corrupt the generated source.

### Tests

Live-database integration tests for every supported dialect under `tests/features/stored-routines/` plus metadata, comparator, helper, validator, and entity-generator suites. Type-level inference tests under `tests/types.test.ts` cover `SqlTypeToTs`, `RoutineArgs<typeof R>`, and the `Type`-at-`type` extraction.

Also fixes a pre-existing regex bug in PG's `parseRoutineParams` where a parameter name starting with `in`/`out`/`inout`/`variadic` (e.g. `input`) was mis-parsed as a direction keyword followed by the rest of the name.


docs/docs/stored-routines.md
```diff
@@ -0,0 +1,161 @@
...
+- Schema generator silently skips them — no DDL is emitted, `schema:diff` produces no changes.
+- `em.callRoutine` for `type: 'function'` routines that declare `bodyJs` registers the JS function as a UDF via the underlying driver (better-sqlite3's `db.function()`) on first call, then dispatches `SELECT routine_name(?, ?, ...)`.
+- `em.callRoutine` for procedures (or functions without `bodyJs`) throws clearly. SQLite has no analog for stored procedures, so they cannot be bridged.
+
+This lets you run the same entities and routine declarations against PostgreSQL/MySQL/MSSQL/Oracle in production while using SQLite in tests, provided your tests only call functions that have a `bodyJs` fallback.
+
+## Limitations
+
+- The default body diff strips trailing semicolons, normalises whitespace, and compares the result. If the database canonicalises your body more aggressively than that, set `ignoreSchemaChanges: ['body']` on the affected routines.
+- Multi-result-set procedures are currently only supported on MySQL.
```
boenrobot: > Multi-result-set procedures are currently only supported on MySQL.

How are the result types declared? Neither the docs or tests show.

Also, based on this article:

https://www.codestudy.net/blog/postgresql-function-returning-multiple-result-sets/

MSSQL should be able to support it in the same fashion as MySQL (just doing EXEC should return all the result sets, with EXEC being the equivalent of CALL in MySQL).

With PGSQL, it appears result sets are only returned in special typed (refcursor) OUT params.

B4nan: Good catch — that line was misleading. The high-level `em.callRoutine` API only ever surfaces a single scalar (for functions) or `void` + mutated OUT/INOUT refs (for procedures), regardless of driver. There's no metadata for declaring multiple result sets and no code path that walks them, so claiming MySQL-only support was just docs drift.

Reworded the limitation to reflect reality: multi-result-set procedures aren't surfaced through the high-level API on any driver — callers need `em.getConnection().execute()` for raw access. Adding proper result-set declarations (e.g. `refcursor` OUT params on PG, or batch-result handling on MSSQL/MySQL) would be a follow-up feature, not in scope for this PR.

Fixed in 7e2f19861.

B4nan: Btw I asked claude to implement the follow-ups as well, let's see what it ends up with. Since it's a completely new thing, I wouldn't mind documenting this as experimental and waiting for some real-world feedback. I'd like to ship this in v7.1 later this week, so if you see something else that doesn't match your initial expectations, let me know (sooner the better 🙃). I haven't reviewed this myself yet fully, so I hope it won't be too crazy.

B4nan:
Quick follow-up to @boenrobot's review comment about multi-result-set support — pulled in three things rather than punting them to follow-up issues (fd59d5e16):

**1. Multi-result-set procs are now real.** `resultSets: N` on the routine declaration; per-driver plumbing:

- **MySQL / MariaDB**: native — proc body has N SELECTs.
- **PostgreSQL**: refcursor OUT params, auto-FETCHed inside the active transaction. So the codestudy.net article's pattern works — declare `refcursor` OUT params with `direction: 'out', ref: true` and call inside `em.transactional`.
- **Oracle**: `sys_refcursor` OUT params (oracledb `DB_TYPE_CURSOR`).
- **MSSQL**: throws with a clear "use em.getConnection().execute()" message. Tedious flattens recordsets through our kysely wrapper, so wiring it through the high-level API would mean bypassing kysely — left for later.
- SQLite/Mongo: throws (no procedure concept).

`em.callRoutine` returns `Dictionary[][]` when `resultSets` is set.

**2. Custom Type on params/returns.** `customType` on a param marshals values through `convertToDatabaseValue`/`convertToJSValue` for IN, OUT, INOUT and scalar function returns. So `JsonType`, `ArrayType`, user types — all work at the routine boundary now.

**3. `generate-entities` emits routines.** Reverse-engineering from an existing DB produces `@Routine` decorator source, `defineRoutine()` consts, or `new RoutineSchema()` — same `entityDefinition` switch as for entities.

Also fixed a regex bug in PG's param parser that surfaced once I had a real-world IN param named `input` (the alternation `IN|OUT|INOUT|VARIADIC` was greedy-matching `IN` against the prefix).

Tests added: integration coverage for PG/MySQL/Oracle multi-rs round-trips, MSSQL throw, validator rejections, customType marshalling on PG and SQLite, and unit + integration tests for routine emission in entity-generator.

docs/docs/stored-routines.md
boenrobot: Definitely needs some "experimental; may change in even a patch release" warning in the docs. I foresee the multi result set being hard to get right and "bolt on", since even _specifying_ it is kind of hard, and yet, most of the use cases for stored procedures involve multiple result sets (well... I mean, most of the use cases that aren't also covered by a stored function; that, and out params are the only things not doable in stored functions).

_Technically_, a procedure can execute different SELECT based on condition, and MySQL and MSSQL will happily emit whatever branch was taken, and only that branch. Which means you could have a procedure with known body and input... and output type that can only be determined after the call.

I think the only way to support hydration of result sets from such procedures would be to have a callback that takes the "fields" and "index" from the returned result set, and returns the type the rows from that result set should be hydrated as.

idk about MSSQL, but with MySQL's client, there's
```ts
query.on('fields', function(fields, index) {
    // the fields for the result rows that follow
})
```
So if `fields` and `index` are both passed to a user provided callback e.g.
```ts
(fields: Record<string, unknown>, index: number) => {
  if (index % 2 === 0) {
    // Application knows that at every even position, we have either authors or publishers.
    if (typeof fields.author_id !== 'undefined') {
      return Author;
    }
    return Publisher;
  }
  // ... and at every odd position, we have books related to the author/publisher from before.
  return Book;
}
```

MikroORM can then react accordingly in `query.on('result', ...)`.

But I'm aware this approach is not exactly TS friendly. MikroORM needs to take this, infer the return type of `callRoutine()` as `(Author[] | Publisher[] | Book[])[]`... And then the caller can check with instanceof what they're getting, if they care (or just blindly JSON if they don't...).

And then you separately from the "branching" cases, you have non-branching ones that can safely and more accurately be defined as a tuple, e.g. `[Author, Publisher, Book]` to declare that the return from `callRoutine()` is `[Author[], Publisher[], Book[]]`.

B4nan:
> Definitely needs some "experimental; may change in even a patch release" warning in the docs.

Added — top-of-page admonition + a more specific "likely to evolve" note at the multi-result-set section in 69b114409.

On the hydration callback idea: agreed, the dynamic `(fields, index) => EntityClass` shape is the right escape hatch for procs that branch on input. The MySQL `query.on('fields', ...)` hook makes it natural there, and tedious has the equivalent `column metadata` event for MSSQL. The bigger lift is wiring up the entity hydration pipeline (which currently runs through QueryBuilder result mapping) for arbitrary `fields → entity meta` matches, since the procedure body's columns don't have to mirror the entity's column names — would probably need a `nameMap`/`fields` mapping on the callback return.

For the static tuple form `[Author, Book]`, the `RoutineReturns` union already has a slot for `(() => EntityName<any>)[]` — it's defined but not yet wired into the hydration path. So that one is mostly a runtime plumbing task on top of the existing types.

I'm leaving both as follow-ups since they're additive on top of the experimental v1 (no behavior change for users who don't opt in), and we can iterate on the callback signature once we see real procs with this shape in the wild. The current `resultSets: N` raw-rows mode at least unblocks people who just want the result sets without entity hydration.

B4nan:
> Or in more practical terms, if the returned result is not an array, it marks the end of the CALL query. So no need to annotate number of result sets.

Good catch — pushed in 279538437. Dropped `resultSets: N` from the metadata entirely; the drivers already have the signal natively:

- **MySQL / MariaDB**: `result.filter(Array.isArray)` after CALL gives exactly the result-set list. The OkPacket is the end-of-stream marker, just like you said.
- **PostgreSQL**: refcursor OUT params are detected by `type: 'refcursor'`; we FETCH each cursor that's declared. Non-refcursor OUT params still behave as scalar refs.
- **Oracle**: sys_refcursor OUT params are detected by type regex and bound as `DB_TYPE_CURSOR`.
- **MSSQL**: still no surfacing of multi-rs through `em.callRoutine` — tedious flattens the recordsets through our kysely wrapper without per-set boundaries, so users still need `em.getConnection().execute()` for raw access.

The validator rules for `resultSets` are gone, and all `resultSets: N` declarations in tests have been removed. The return shape is purely runtime-detected now: scalar for functions, `void` for procedures with no result sets, `Dictionary[][]` for procedures that emit any. The "likely to evolve" note on the docs now points only at entity-typed hydration (static `returns: [Author, Book]` tuple form + dynamic `(fields, index) => EntityClass` callback) as the next iteration, which is where the typed-hydration discussion from earlier in the thread lands naturally.

B4nan:
I'll remove the decorators completely from here, in the current implementation, they are just noise.

PR #7706, opened 2026-05-11
fix(schema): centralize identifier-length truncation across platforms

nuzayets:
## What

Auto-derived check-constraint names that exceed a platform's identifier limit are now routed through `Platform.getIndexName(..., 'check')`, the same path FK/index/unique names already take. PG/MySQL/SQLite overrides have their type union widened so the existing hash-truncation pattern covers check constraints. Oracle and MSSQL get matching overrides at the 128-char boundary.

## Why

Before this change, check-constraint names were generated directly via `NamingStrategy.indexName(..., 'check')`, bypassing per-platform truncation. The result:

- **Postgres** (NAMEDATALEN=63): the engine silently truncates over-long identifiers on `ADD CONSTRAINT` (NOTICE-only, not an error), so the entity-derived name in `toSchema` never matched the introspected name in `fromSchema`. `migration:check` and `schema:update --dump` reported a phantom drop+add of the same constraint on every run.
- **MySQL / MariaDB** (64): the engine rejected the statement outright with an identifier-too-long error, so `migration:create` produced SQL that couldn't be applied.
- **Oracle / MSSQL** (128): never had any truncation override, so the same shape of bug applied with a much higher threshold.

Trivially reproducible via long table name + embedded prefix + enum property — the natural `*_check` name exceeds 63 chars.

## Approach

`Platform.getIndexName` now accepts `'check'` in its type union. The default implementation still delegates to `NamingStrategy.indexName`. PG/MySQL/SQLite overrides only had to widen their signatures; their existing hash-truncation logic now applies to check names automatically. Oracle and MSSQL get the same pattern at 128 chars.

`MetadataDiscovery.initCheckConstraints` and `EntityMetadata.sync` now go through `Platform.getIndexName(..., 'check')` instead of `NamingStrategy.indexName` directly.

## Scope notes

- **User-supplied identifier names** (`@Check({ name: '...' })`, explicit `foreignKeyName`, etc.) still pass through verbatim. The same pre-existing exposure for over-long *explicit* names remains, consistent with how FK / index / unique already work today. Worth a separate follow-up if desired.
- **No comparator changes.** A defensive normalization layer in `SchemaComparator` was considered but proved redundant once metadata-side truncation is in place.
- **One-time rename for pre-fix DBs.** Users currently running with check names that the engine silently truncated will see a one-time `drop + add` diff on the next `migration:check` as the constraint is renamed to the hash-truncated form. But, the current behaviour also emits a `drop + add` diff that never resolves anyway. This time, the diff is stable after rename.

## Tests

`tests/issues/GHx47.test.ts` covers all six SQL dialects:

- **Postgres / MySQL / MariaDB**: 73-char natural check name; assert emitted name fits the platform limit and `getUpdateSchemaSQL` is empty after `refresh()`.
- **SQLite**: 131-char natural name; assert it passes through unchanged (no limit) and round-trips clean.
- **Oracle / MSSQL**: 131-char natural name; assert emitted name fits 128 chars and round-trips clean.

### DIFF
--- a/packages/core/src/metadata/MetadataDiscovery.ts
+++ b/packages/core/src/metadata/MetadataDiscovery.ts
@@ -1975,7 +1975,7 @@ export class MetadataDiscovery {
 
     for (const check of meta.checks) {
       const fieldNames = check.property ? meta.properties[check.property].fieldNames : [];
-      check.name ??= this.#namingStrategy.indexName(meta.tableName, fieldNames, 'check');
+      check.name ??= this.#platform.getIndexName(meta.tableName, fieldNames, 'check');
 
       if (check.expression instanceof Function) {
         check.expression = check.expression(columns, table);
@@ -2004,7 +2004,7 @@ export class MetadataDiscovery {
 
         if (expression) {
           meta.checks.push({
-            name: this.#namingStrategy.indexName(meta.tableName, prop.fieldNames, 'check'),
+            name: this.#platform.getIndexName(meta.tableName, prop.fieldNames, 'check'),
             property: prop.name,
             expression,
           });
--- a/packages/core/src/platforms/Platform.ts
+++ b/packages/core/src/platforms/Platform.ts
@@ -813,8 +813,7 @@ export abstract class Platform {
   }
 
   /**
-   * Returns the default name of index for the given columns. Per-platform overrides may apply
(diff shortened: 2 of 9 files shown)


B4nan (approved):
thanks!

PR #7711, opened 2026-05-12
feat(postgres): support `driverOptions.onPoolCreated` hook for `pg.Pool`

nuzayets:
The `pg` driver re-emits errors from idle pool connections as `pool.on('error', ...)`. Without a Pool-level listener, errors from idle connections the server closes (failover, maintenance) surface as uncaught exceptions and crash the process. `onCreateConnection` can't solve this — it hands you a checked-out `Client`, not the `Pool`, and the pool re-emits regardless of what listeners exist on the client.

Today the only workaround is constructing a `PostgresDialect` manually and passing it via `driverOptions`, which forces callers to duplicate the `TypeOverrides` setup `PostgreSqlConnection.mapOptions` already does.

This adds an optional `driverOptions.onPoolCreated(pool)` callback, invoked once with the `pg.Pool` right after construction:

```ts
import { MikroORM, PostgreSqlDriver } from '@mikro-orm/postgresql';
import type { Pool } from 'pg';

MikroORM.init({
  driver: PostgreSqlDriver,
  driverOptions: {
    onPoolCreated: (pool: Pool) => {
      pool.on('error', err => logger.warn('idle pg pool error', err));
    },
  },
});
```

Follows the same convention as MySQL's `driverOptions.maxIdle` (`packages/mysql/src/MySqlConnection.ts`): an undocumented-in-types driver-specific key that the connection class extracts before merging the rest into the underlying pool config. No changes to `@mikro-orm/core` or to the exported `PostgreSqlOptions` surface.

### DIFF
--- a/packages/postgresql/src/PostgreSqlConnection.ts
+++ b/packages/postgresql/src/PostgreSqlConnection.ts
@@ -7,9 +7,14 @@ import { AbstractSqlConnection, Utils } from '@mikro-orm/sql';
 /** PostgreSQL database connection using the `pg` driver. */
 export class PostgreSqlConnection extends AbstractSqlConnection {
   override createKyselyDialect(overrides: PoolConfig): PostgresDialect {
-    const options = this.mapOptions(overrides);
+    const { onPoolCreated, ...poolOverrides } = (overrides ?? {}) as PoolConfig & {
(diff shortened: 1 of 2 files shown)


B4nan (approved):
thanks!

PR #7719, opened 2026-05-14
fix(core): resolve polymorphic TPT-child filter columns to parent table

### LINKED ISSUE #7717: Polymorphic `manyToOne` targeting TPT sub-class entities resolves default filter conditions against the child sub-table alias instead of the parent table
### Describe the bug

_Repro_: https://github.com/stoneLeaf/mikro-orm-v7-repros/tree/polymorphic-tpt-target-default-filter-wrong-table-alias

## Related issues

- #7317 — Polymorphic relation with default filter on one target creates an incorrect INNER JOIN, fixed in v7.0.3 by #7332
- #7563 — Polymorphic relation with TPT sub-class targets (INSERT/SELECT bugs), fixed in v7.0.11 by #7564

Both fixes are present in v7.0.15. The remaining case described here involves **all three features simultaneously**: polymorphic `manyToOne`, TPT sub-class targets, and any default filter whose condition references a column that exists only on the TPT parent table (not on child sub-tables). The reproduction below uses a soft-delete filter as a concrete example, but the same crash occurs with any filter in this position.

## Problem

When a `Part` entity holds a polymorphic `manyToOne([DeviceA, DeviceB])` where both targets are TPT sub-classes of an abstract `Device` parent, and the parent carries a default filter whose condition references a parent-table column (here: an embedded audit property resolved to `audit_deleted_at`), populating the inverse collection (`DeviceA.parts`) throws:

```
InvalidFieldNameException: column d0.audit_deleted_at does not exist
```

In TPT, the embedded `audit_deleted_at` column lives on the parent table (`device`). The child sub-tables (`device_a`, `device_b`) do not have it. MikroORM's polymorphic join code path applies the filter condition directly to the child sub-table alias, bypassing the parent table join that would ordinarily carry the column.

## Entity setup

```typescript
const AuditEmbeddableSchema = defineEntity({
  name: "AuditEmbeddable",
  embeddable: true,
  properties: {
    deletedAt: p.datetime().nullable(),
  },
});
class AuditEmbeddable extends AuditEmbeddableSchema.class {}
AuditEmbeddableSchema.setClass(AuditEmbeddable);

// Any default filter whose cond references a parent-table-only column triggers
// the bug. Soft-delete is used here as a concrete example.
const PARENT_COLUMN_FILTER = {
  excludeSoftDeleted: {
    name: "excludeSoftDeleted",
    cond: { audit: { deletedAt: null } },  // resolves to `audit_deleted_at`
    default: true,
  },
};

// Abstract parent — `audit_deleted_at` lives in THIS table only
const DeviceSchema = defineEntity({
  name: "Device",
  tableName: "device",
  abstract: true,
  inheritance: "tpt",
  filters: { ...PARENT_COLUMN_FILTER },
  properties: {
    id: p.integer().primary(),
    name: p.string(),
    audit: () => p.embedded(AuditEmbeddable).onCreate(() => new AuditEmbeddable()),
  },
});
class Device extends DeviceSchema.class {}
DeviceSchema.setClass(Device);

// Child TPT entities — sub-tables hold `id` only, NOT `audit_deleted_at`
const DeviceASchema = defineEntity({
  name: "DeviceA",
  tableName: "device_a",
  extends: Device,
  properties: {
    id: p.integer().primary(),
    parts: () => p.oneToMany(Part).mappedBy("parentDevice"),
  },
});
class DeviceA extends DeviceASchema.class {}
DeviceASchema.setClass(DeviceA);

const DeviceBSchema = defineEntity({
  name: "DeviceB",
  tableName: "device_b",
  extends: Device,
  properties: { id: p.integer().primary() },
});
class DeviceB extends DeviceBSchema.class {}
DeviceBSchema.setClass(DeviceB);

// Polymorphic owning side
const PartSchema = defineEntity({
  name: "Part",
  tableName: "part",
  properties: {
    id: p.integer().primary(),
    parentDevice: () => p.manyToOne([DeviceA, DeviceB]).ref(),
  },
});
class Part extends PartSchema.class {}
PartSchema.setClass(Part);
```

## Steps to reproduce

```typescript
await orm.em.find(DeviceA, {}, { populate: ["parts"] });
```

`EntityLoader.findChildren` issues the child-find query for `Part`. Because `Part.parentDevice` is polymorphic, MikroORM JOINs both target sub-tables and applies their inherited default filter to each join. The generated SQL reads (PostgreSQL):

```sql
SELECT "p0".*
FROM "part" AS "p0"
LEFT JOIN "device_a" AS "d0"
  ON "p0"."parent_device_id" = "d0"."id"
  AND "p0"."parent_device_type" = 'device_a'
  AND "d0"."audit_deleted_at" IS NULL      -- ← does not exist on device_a
LEFT JOIN "device_b" AS "d1"
  ON "p0"."parent_device_id" = "d1"."id"
  AND "p0"."parent_device_type" = 'device_b'
  AND "d1"."audit_deleted_at" IS NULL      -- ← does not exist on device_b
WHERE "p0"."parent_device_id" IN (1)
  AND "p0"."parent_device_type" = 'device_a'
```

## Expected behaviour

Any column referenced in a default filter on the TPT parent entity belongs to the parent table (`device`). The filter condition should be resolved against the parent table, not against the child sub-table alias. For a regular TPT query MikroORM correctly joins the parent table and uses its alias; the polymorphic join code path does not, causing any filter column that lives only on the parent table to be unresolvable.

## Actual behaviour

The query references `device_a.audit_deleted_at` (and `device_b.audit_deleted_at`). Those columns do not exist. PostgreSQL throws `InvalidFieldNameException: column d0.audit_deleted_at does not exist`; SQLite throws `no such column: d0.audit_deleted_at`.

### Reproduction

https://github.com/stoneLeaf/mikro-orm-v7-repros/tree/polymorphic-tpt-target-default-filter-wrong-table-alias

@mikro-orm/postgresql

### MikroORM version

7.0.15

### Node.js version

22.22.1

### Operating system

Win

### Validations

B4nan:
## Summary

When a polymorphic `manyToOne` targeted a TPT sub-class whose default filter referenced a column on the TPT parent table (e.g. an embedded audit field), populating the inverse collection threw `column <alias>.audit_deleted_at does not exist`.

Two paths contributed:
- The per-target polymorphic LEFT JOINs did not INNER JOIN the parent table, so column resolution had no parent alias to rewrite to.
- `getJoinedFilters` also auto-joined the first polymorphic target via `populateFilter`, redundantly applying the filter against the child alias (and silently dropping rows pointing to other targets when the relation was not nullable).

## Fix

- Reuse the existing TPT-parent walk for polymorphic TPT-child targets so the parent table is INNER JOINed and registered in `tptAlias`; `processNestedJoins` nests it inside the polymorphic LEFT JOIN so rows pointing to other polymorphic targets are not dropped.
- Skip emitting populate-filter entries for polymorphic to-one relations; they are already handled per-target via the `:ref filter` hint with discriminator-scoped WHERE checks.

Closes #7717

### DIFF
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -561,6 +561,14 @@ export class EntityManager<Driver extends IDatabaseDriver = IDatabaseDriver> {
         continue;
       }
 
+      // Polymorphic to-one relations are already filtered via per-target LEFT JOINs created
+      // for the `:ref filter` hint; emitting a populate-filter entry here would auto-join only
+      // the first target meta and either drop rows pointing to other targets or reference
+      // parent-table-only columns on the child sub-table alias for TPT children.
+      if (prop.polymorphic && [ReferenceKind.MANY_TO_ONE, ReferenceKind.ONE_TO_ONE].includes(prop.kind)) {
+        continue;
+      }
+
       const filters = QueryHelper.mergePropertyFilters(prop.filters, options.filters);
       const where = await this.applyFilters<Entity>(prop.targetMeta!.class, {}, filters, 'read', {
         ...options,
--- a/packages/sql/src/AbstractSqlDriver.ts
+++ b/packages/sql/src/AbstractSqlDriver.ts
@@ -2132,6 +2132,14 @@ export abstract class AbstractSqlDriver<
             schema,
           );
 
+          // For polymorphic targets that are TPT child entities, INNER JOIN parent tables so that
+          // filter conditions referencing parent-table columns resolve to the correct alias. The
+          // INNER JOINs get nested inside the polymorphic LEFT JOIN by processNestedJoins, which
+          // keeps the resulting query valid for rows pointing to other polymorphic targets.
+          if (targetMeta.inheritanceType === 'tpt' && targetMeta.tptParent) {
+            this.addTPTParentJoinsForRelation(qb, targetMeta as EntityMetadata<T>, tableAlias, targetPath);
+          }
+
           // For polymorphic targets that are TPT base classes, also LEFT JOIN
           // all descendant tables so child-specific fields can be selected.
           if (targetMeta.inheritanceType === 'tpt' && targetMeta.tptChildren?.length && !ref) {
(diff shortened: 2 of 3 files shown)


stoneLeaf:
Confirmed working, thanks @B4nan

PR #7789, opened 2026-05-27
feat(postgresql): support `EXCLUDE` constraints via `@Check`

### LINKED ISSUE #7787: Add PostgreSQL EXCLUDE constraint support
## Is your feature request related to a problem? Please describe.

MikroORM supports most PostgreSQL constraint types (`NOT NULL`, `UNIQUE`, `PRIMARY KEY`, `FOREIGN KEY`, `CHECK`), but lacks support for [`EXCLUDE`](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-EXCLUSION) constraints.

This is typically used to prevent overlapping reservations for a resource directly at the database level:

```sql
CREATE EXTENSION IF NOT EXISTS btree_gist;

ALTER TABLE room_booking
ADD CONSTRAINT room_booking_no_overlap
EXCLUDE USING gist (
	room_id WITH =,
	tstzrange(starts_at, ends_at, '[)') WITH &&
);
```

## Describe the solution you'd like

I propose adding an entity-level decorator like `@ExcludeConstraint()`. Like `@Check()`, this should tie into schema generation, migrations, introspection, and diffing.

```ts
@Entity()
@ExcludeConstraint({
	name: "room_booking_no_overlap",
	using: "gist",
	expression: (columns) =>
		`${columns.room} WITH =, tstzrange(${columns.startsAt}, ${columns.endsAt}, '[)') WITH &&`,
})
export class RoomBooking {
	// ...
}
```

I am not attached to the exact decorator/API name, but the important part is that MikroORM treats this as a constraint rather than an index.

## Describe alternatives you've considered

- **Abusing `@Index({ expression: "..." })`**: You can use this to emit the `ALTER TABLE ... ADD CONSTRAINT` statement, but it breaks schema diffing. Postgres stores the object as a constraint, while MikroORM tracks it as an index. Lifecycle operations fail because MikroORM tries to drop or update an index, but Postgres expects constraint semantics.
- **`@Check()`**: Only evaluates the current row, so it can't compare against other rows.
- **Application-level validation**: Prone to race conditions.

## Additional context

I realize `EXCLUDE` constraints are a somewhat obscure and **PostgreSQL-specific** feature, but they are the canonical Postgres solution for scheduling and reservation overlap rules.

Thank you for the continued good work on MikroORM.

Related docs:

- PostgreSQL [`EXCLUDE` constraints](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-EXCLUSION)
- PostgreSQL [`btree_gist`](https://www.postgresql.org/docs/current/btree-gist.html)
- MikroORM [`@Index()`](https://mikro-orm.io/docs/indexes)
- MikroORM [`@Check()`](https://mikro-orm.io/docs/decorators#check)

B4nan:
The `@Check` decorator already covers the lifecycle EXCLUDE constraints need (ALTER/DROP CONSTRAINT, name-based matching, opaque expression diff), so this teaches the existing pipeline to pass the body through verbatim when the expression itself is a full table-constraint body — no new decorator, no new option.

- `SchemaHelper.createCheck` skips the `check (…)` wrap when the expression starts with `EXCLUDE`.
- `PostgreSqlSchemaHelper.getChecksSQL` UNIONs in `contype = 'x'` rows; since EXCLUDE doesn't appear in `information_schema.constraint_column_usage`, that branch resolves table/schema directly from `pg_class` / `pg_namespace`.
- `getAllChecks` keeps the raw `pg_get_constraintdef()` text when it isn't a `CHECK ((…))` wrap, so round-trip comparison through `diffExpression` is symmetric.
- The index introspection query filters out `contype = 'x'` rows so the index pg auto-creates to back the constraint doesn't leak through as a stray index needing a `drop index`.

```ts
@Entity()
@Check({
  name: 'room_booking_no_overlap',
  expression: 'exclude using gist (room_id with =, during with &&)',
})
class RoomBooking { /* … */ }
```

Closes #7787

### DIFF
--- a/docs/docs/defining-entities.md
+++ b/docs/docs/defining-entities.md
@@ -1511,6 +1511,32 @@ export class Book {
   </TabItem>
 </Tabs>
 
+### Exclusion constraints (PostgreSQL)
+
+PostgreSQL's [`EXCLUDE`](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-EXCLUSION) constraints share the same lifecycle as `CHECK` (table-level constraint, dropped via `ALTER TABLE … DROP CONSTRAINT`), so `@Check()` accepts a full constraint body as the `expression` and emits it verbatim when it starts with `EXCLUDE`:
+
+```ts
+@Entity()
+@Check({
+  name: 'room_booking_no_overlap',
+  expression: 'exclude using gist (room_id with =, during with &&)',
+})
+export class RoomBooking {
+
+  @PrimaryKey()
+  id!: number;
+
+  @Property()
+  room_id!: number;
+
+  @Property({ columnType: 'tstzrange' })
+  during!: string;
+
+}
+```
+
+This requires the `btree_gist` extension when the constraint uses `=` on a non-range type. The schema generator round-trips EXCLUDE constraints through introspection just like check constraints — no extra diffing churn.
+
 ## Database Triggers
 
 You can define database triggers via the `@Trigger()` decorator or the `triggers` option in `defineEntity`/`EntitySchema`. Triggers are managed by the schema generator and migration system — they are created, updated, and removed automatically.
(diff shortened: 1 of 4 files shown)


enduity:
Thanks! 😄

PR #7795, opened 2026-05-28
fix(cli): allow `discovery:export` to generate an empty list of entities

boenrobot:
That is, if discovery.warnWhenNoEntities is set to false.

This enables users to still get a barrel file present, and with a typed EM. Notably important during prototyping, when one hasn't authored entities yet, but they still need to hook the ORM into their application.


B4nan (approved):
thanks!

PR #7814, opened 2026-05-30
fix(postgres): only diff partitioning for entities that declare `partitionBy`

B4nan:
The schema generator introspects partitioning for every table, so any mismatch between the introspected definition and the metadata made `schema:update`/`migration:create` throw `partition definitions … not supported automatically`. A table partitioned in the database but mapped by an entity without `partitionBy` failed schema diffing as a result.

Partitioning is now diffed only when the entity declares `partitionBy`; otherwise the existing partitioning is left untouched (no diff, no throw). Adding partitioning to an existing table or changing an existing definition still can't be applied in place, so it throws on a normal run to avoid silently losing the change — unless safe mode is enabled (`schema:update --safe` or `migrations: { safe: true }`), in which case it is skipped. Note `migrations.safe` defaults to `false`, so `migration:create` throws unless safe mode is turned on.


B4nan:
Folded into #7816 — the rebuild work supersedes the interim throw-gating here, and keeping it as one PR gives a single changelog entry for the whole partitioning-on-existing-tables change.

PR #7834, opened 2026-06-01
feat(cli): added searchConfiguration() to analyze config file for context

boenrobot:
The logic is the same pre-existing one from CLIHelper.getConfiguration(),
but now exposed separately.

This feature is particularly important for projects where the config file is imported statically.
It being in core also ensures this logic is available for use at runtime, outside of the CLI.


B4nan:
Does this have to be in core? Also, do you really need to use the CLI for the use case you mentioned? Maybe it would be better to provide some programatic API for the discovery export command?

B4nan:
> CLI is generally not a runtime dependency (only a dev one) and it doesn't seem right to require it as one. And it's only this searchConfiguration() operation is needed at runtime.

But you said you want this so you can use `discovery:export`, so how is this a runtime concern? The method is not used anywhere in the core, it's really just for the CLI, so I would rather keep it there. The core is not aware of any config loading in v7, a config is an object, not a path to a file.

B4nan:
> import configExports from './../../../../mikro-orm.config';

But this is your file, you should know what it does and act based on it here (e.g. use a default import). You don't need a generic ORM helper that reflects how the CLI deals with it - you are not using the CLI here.

Or is that would you'd like to do inside the nuxt adapter? If so, if feels a bit wrong that you'd mess with dynamic config resolution on that level. That is really just a concern of the ORM CLI, no other part of the codebase does this nowadays. The core expects the user to provide the right options, I'd expect the same from the nuxt adapter, similarly to how the nest adapter expects an explicit way to deal with multiple contexts.

> with a dedicated export (not part of index)

I can't say I like that idea, this dummy helper doesn't really deserve its own export path, it has no dependencies.

B4nan (approved):

PR #7842, opened 2026-06-02
fix(core): use owner targetKey for inverse collection loads

jaronpate:
Populate, find, and Collection.load/count for inverse OneToMany relations now resolve identity from the owning side's targetKey (e.g. uuid) instead of the parent's primary key. Child hydration and FK id collection in EntityLoader follow the same rule.

Extract RelationIdentity helpers for entity/collection keys and OneToMany child wiring. Collection reuses getInverseCollectionOwnerValue for load and count conditions.

Add non-pk-relation-target tests for inverse populate, find+populate, and collection.loadItems.

Full disclosure AI helped me write this. There is no way I could absorb enough info about the codebase to make this change. Please review with a close eye.

### DIFF
--- a/tests/features/non-pk-relation-target/non-pk-relation-target.postgres.test.ts
+++ b/tests/features/non-pk-relation-target/non-pk-relation-target.postgres.test.ts
@@ -9,8 +9,7 @@ import {
   ReflectMetadataProvider,
 } from '@mikro-orm/decorators/legacy';
 
-// Entity in custom schema
-@Entity({ schema: 'custom_schema' })
+@Entity()
 class Author {
   @PrimaryKey()
   id!: number;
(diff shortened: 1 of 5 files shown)


tests/features/non-pk-relation-target/non-pk-relation-target.postgres.test.ts
```diff
@@ -22,11 +21,11 @@ class Author {
   @Property()
   name!: string;
 
-  @OneToMany(() => Book, b => b.author)
+  @OneToMany(() => Book, book => book.author)
   books = new Collection<Book>(this);
 }
 
-@Entity({ schema: 'custom_schema' })
```
boenrobot, line 29: This is changing a pre-existing test to not support a feature (custom schema on an entity definition level).

If your change breaks this test, then the implementation needs changing I think.

Unless maybe the test was wrong in the first place, but I don't see why it would be.

B4nan: Yeah, also the changes don't even compile. I'll most likely reuse the tests from it and try to fix it myself, without touching the existing tests.

jaronpate:
Thanks for the quick fix. Apologies the original patch was not more helpful.

PR #7849, opened 2026-06-04
fix(core): avoid stack overflow in IdentityMap.keys()/values() for large maps

ianmihura:
## What

`IdentityMap.keys()` and `IdentityMap.values()` build their result with `ret.push(...arr)`. When a single entity class holds more than ~125k entries, the spread exceeds V8's maximum argument count and throws `RangeError: Maximum call stack size exceeded`, crashing the process.

## Why

`TransactionManager.mergeEntitiesToParent` calls `getIdentityMap().keys()` to test emptiness after a nested transaction merges into a parent EM. A long-lived parent identity map with >125k entities of one class triggers the overflow.

## Fix

Replace the argument spreads with `for...of` push loops, mirroring the existing `[Symbol.iterator]()` implementation in the same class. No API or behavior change other than no longer crashing on large maps.

## Test

Added `tests/features/unit-of-work/identity-map.test.ts`: populates one class store with 200k entries and asserts `keys()`/`values()` return all of them. Fails with the RangeError before the fix, passes after.


B4nan:
I don't mind the fix, but please target master branch first, then we can cherry pick to 6.x

tests/features/unit-of-work/identity-map.test.ts
```diff
@@ -0,0 +1,18 @@
+import { IdentityMap } from '@mikro-orm/core';
+
+describe('IdentityMap large maps', () => {
```
B4nan: also this doesnt really deserve its own file. lets put it inside https://github.com/mikro-orm/mikro-orm/blob/3010820b9c06d51839b931200d9386264a3a56c2/tests/features/unit-of-work/UnitOfWork.test.ts#L10

B4nan (approved):

ianmihura:
@B4nan I see you already merged to master. Thanks!

PR #7873, opened 2026-06-12
fix(cli): discovery:export casts the DriverEntityManager to the right type

boenrobot:
Specifically, `Constructor<EntityManager>`, where `EntityManager` is the grafted type. Without this cast, the type import is correct,
but the type can't be correctly inferred out of the value, as the value type carries the type without the "~entities" graft.


B4nan (approved):
thanks!

PR #7876, opened 2026-06-12
fix(mssql): return result sets from callRoutine without OUT params

### LINKED ISSUE #7872: Calling a routine without output variables always return undefined
### Describe the bug

When you call a routine using `callRoutine` without using any output variables in the procedure itself, it always returns undefined despite the query returning one or several results.

The result is available here https://github.com/mikro-orm/mikro-orm/blob/51ba1ab879d5f3107bc704feb2a6c44ba96061de/packages/mssql/src/MsSqlConnection.ts#L168 but is ignored because `outVars` length is 0.

I suggest checking `outVars` length at this line https://github.com/mikro-orm/mikro-orm/blob/51ba1ab879d5f3107bc704feb2a6c44ba96061de/packages/mssql/src/MsSqlConnection.ts#L174 and always returning the result.

So we can still use output variables with references and get the raw result if needed.

In my case, I have to use a stored procedure I don't own. It's a stored procedure from an external database I need to use.
The new Routine feature is neat because I use it to mock the result locally (during dev) and ensure parameters and return type safety.

### Reproduction

https://github.com/SteadEXE/mikro-orm-repro/tree/routines-without-out-vars

`npm run test`

@mikro-orm/mssql

### MikroORM version

mikro-orm-repro@0.0.1 /mikro-orm-repro ├── @mikro-orm/cli@7.1.4 ├── @mikro-orm/core@7.1.4 ├── @mikro-orm/decorators@7.1.4 ├── @mikro-orm/migrations@7.1.4 ├── @mikro-orm/mssql@7.1.4 ├── @swc-node/register@1.11.1 ├── @swc/cli@0.8.1 ├── @swc/core@1.15.41 ├── @types/node@24.13.2 ├── oxfmt@0.54.0 ├── reflect-metadata@0.2.2 ├── source-map-support@0.5.21 ├── typescript@6.0.3 ├── unplugin-swc@1.5.9 └── vitest@4.1.8

### Node.js version

24.16.0

### Operating system

Fedora 44 (Linux)

### Validations

SteadEXE:
## Summary
- Fix `callRoutine` in MSSQL driver to return result sets from stored procedures without OUT/INOUT parameters
- Previously, routines without output variables always returned `undefined` despite the query returning results
- Now returns the actual result from the execute call, whether there are output variables or not

## Changes
- `packages/mssql/src/MsSqlConnection.ts`: Return `result` instead of `undefined` in both branches (with and without output variables)
- `tests/features/stored-routines/mssql.test.ts`: Added test case to verify result sets are returned from procedures without OUT/INOUT params

## Verification
- Added test case that creates a procedure with SELECT statement and verifies the result is returned
- Test covers the scenario described in issue #7872

Fixes #7872

### DIFF
--- a/packages/mssql/src/MsSqlConnection.ts
+++ b/packages/mssql/src/MsSqlConnection.ts
@@ -168,7 +168,7 @@ export class MsSqlConnection extends AbstractSqlConnection {
     const result = await this.execute(batch.join('; '), allValues, 'all', ctx);
 
     if (outVars.length === 0) {
-      return undefined as T;
+      return result as T;
     }
 
     const rows = result as Dictionary[];
@@ -178,7 +178,7 @@ export class MsSqlConnection extends AbstractSqlConnection {
       args,
     );
 
-    return undefined as T;
+    return result as T;
   }
 
   override async commit(
--- a/tests/features/stored-routines/mssql.test.ts
+++ b/tests/features/stored-routines/mssql.test.ts
@@ -112,4 +112,35 @@ describe('stored routines — MSSQL', () => {
 
     await orm2.close(true);
   });
+
+  it('em.callRoutine returns result sets from procedures without OUT/INOUT params', async () => {
+    const GetRecords = Routine.create<{ p_name: string }, RecordEntity[]>({
+      name: 'get_records',
+      type: 'procedure',
+      params: {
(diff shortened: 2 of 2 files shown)


B4nan:
Closing in favor of #7879, thanks anyway!

PR #7886, opened 2026-06-15
feat(cli): add a "-q" param, to suppress informational output

boenrobot:
This lets the CLI more seamlessly interact with larger scripts, particularly when called as a JS module (where the output buffer is shared with the caller).

Errors are still outputted.


B4nan:
I would rather see this as a global flag that would work on the `dump` helper level. Feels weird to add it only to this one command.

B4nan:
Fair enough, this one is the least important out of the 3 docs-wise. Happy to merge once you rebase it.

B4nan (approved):

PR #7920, opened 2026-06-22
fix(core): enable RequestContext to use almost all fork options

boenrobot:
Notably, this enables per request abort signal passing. Also flush mode and event manager override (in case one wants to override them for some routes), while still using the globals for most.

Explicitly excludes the options that do not make sense when using RequestContext.


B4nan (approved):
thanks!

PR #7925, opened 2026-06-23
feat(core): allow specifying a callback for the RequestContext options

boenrobot:
This enables setting different options per instance, which is particularly useful with multiple instances present.


B4nan:
Can you please retarget all the 3 PRs to `next` branch?

B4nan:
Let's document this, same for the other two PRs. Nothing huge, a short paragraph or a simple example is enough.

B4nan (approved):
thanks!

PR #7931, opened 2026-06-24
feat(mongo): add `$all` query operator support

lblblong:
## Summary

- Add `$all` to `QueryOperator` and `OperatorMap` for MongoDB array queries (match documents whose array field contains all specified values)
- Allow `$all` inside embeddable query conditions
- Reject `$all` in SQL drivers with a descriptive error pointing at `$contains`, instead of building invalid SQL
- Add mongo integration tests, a sqlite test for the unsupported case, and a QueryHelper unit test ensuring `$all` is not rewritten to `$in`

## Motivation

MongoDB supports the native `$all` operator for array fields, but MikroORM did not register it as a query operator. Queries like `{ tags: { $all: [id1, id2] } }` were incorrectly transformed to `{ tags: { $all: { $in: [...] } } }`.

Made with [Cursor](https://cursor.com)


B4nan (approved):
thanks

Issue #7714, opened 2026-05-12
Expression indexes on parent TPT entities are propagated to child TPT tables


stoneLeaf:
### Describe the bug

_See repro_: https://github.com/stoneLeaf/mikro-orm-v7-repros/tree/parent-tpt-expression-index-wrongly-propagated-to-tpt-children

## Context

- Abstract parent TPT entity (`Animal`) with `indexes: [{ expression: fn }]` where `fn` reads `columns['name']`
- `name` is a column of the **parent** table only
- Child TPT entity (`Dog`) whose table holds only `id` and `breed`

## Issue

When a parent TPT entity defines an `indexes` entry whose `expression` is a function that looks up columns by property name, MikroORM propagates that index to every child TPT entity table. The `columns` Record passed to the expression function for child tables contains only the child's own columns — not the parent's. Accessing a parent-only property (e.g. `columns['name']`) returns `undefined`, resulting SQL to have an empty column list:

```sql
-- Parent table — correct
CREATE UNIQUE INDEX "animal_name_not_deleted_unique" ON "animal" (name) WHERE deleted_at IS NULL;

-- Child table — invalid: empty column list crashes the database
CREATE UNIQUE INDEX "dog__not_deleted_unique" ON "dog" () WHERE deleted_at IS NULL;
```

Executing the schema creation then fails:

```
SyntaxErrorException: syntax error at or near ")"
```

## Expectation

Parent TPT entity `indexes` definitions should **not be propagated to child entity tables**.

### Reproduction

https://github.com/stoneLeaf/mikro-orm-v7-repros/tree/parent-tpt-expression-index-wrongly-propagated-to-tpt-children

### What driver are you using?

@mikro-orm/postgresql

### MikroORM version

7.0.15

### Node.js version

22.22.1

### Operating system

win

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/mikro-orm/mikro-orm/blob/master/CONTRIBUTING.md).
- [x] Read the [docs](https://mikro-orm.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/mikro-orm/mikro-orm/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mikro-orm/mikro-orm/discussions) or join our [Discord](https://discord.gg/w8bjxFHS7X).
[...]

PR #7944, opened 2026-06-30
fix(sql): do not propagate TPT parent check constraints to child tables

### LINKED ISSUE #7935: TPT: base enum CHECK constraint duplicated onto child subtype tables
### Describe the bug

For TPT entities with an enum on the base table, the generated schema duplicates the enum's CHECK constraint onto the child tables.

Below is a verbose AI description:

---

  With `inheritance: 'tpt'`, schema generation emits a **base** entity's enum `CHECK` constraint onto the **child** subtype table as well. In TPT the column lives only on the parent table, so the child-table CHECK references a non-existent column and the generated DDL is invalid — `schema.create()` / `migration:up` fails.

  - **sqlite:** `no such column: status`
  - **mssql:** error **207**, `Invalid column name 'status'`

  This is **driver-independent** — the wrong-table constraint is produced in the shared schema layer (`DatabaseSchema.fromMetadata`), so it reproduces on sqlite, mssql, postgres and mysql. The reproduction uses in-memory sqlite per the template guidance.

  ## Entities (defineEntity; same happens with the decorator API)

  ```ts
  enum Status { ACTIVE = 'active', INACTIVE = 'inactive' }

  const PersonSchema = defineEntity({
    name: 'Person', abstract: true, inheritance: 'tpt',
    properties: { id: p.integer().primary(), name: p.string(), status: p.enum(() => Status) },
  });
  class Person extends PersonSchema.class {}
  PersonSchema.setClass(Person);

  const EmployeeSchema = defineEntity({
    name: 'Employee', extends: PersonSchema,
    properties: { department: p.string() }, // no `status` here
  });
  class Employee extends EmployeeSchema.class {}
  EmployeeSchema.setClass(Employee);
  ```

  ## Expected

  The `status` CHECK is emitted on the `person` table only.

  ## Actual

  `getCreateSchemaSQL()` also emits a CHECK on `employee`, whose table has no `status` column:

  ```sql
  create table `person` (`id` integer not null primary key autoincrement, `name` text not null, `status` text check (`status` in ('active', 'inactive')) not null);
  create table `employee` (`department` text not null, `id` integer not null primary key, constraint `employee_id_foreign` foreign key (`id`) references `person` (`id`) on update cascade on delete cascade,
  constraint `employee_status_check` check (`status` in ('active', 'inactive')));
  -- the `employee` table has no `status` column
  ```

  `await orm.schema.create()` then throws `no such column: status`.

  ## Notes

 - The runtime persistence layer already handles this correctly — inserting an `Employee` writes `status` to `person` and only `department`/PK to `employee`.
  - It's only DDL generation that's wrong: `DatabaseSchema.fromMetadata` already restricts TPT child **columns** to `ownProps`, but the **checks** loop iterates `meta.checks` (including inherited entries) without that filter, so inherited enum CHECKs leak onto child tables.

### Reproduction

https://github.com/Subtletree/mikro-orm-tpt-check-repro

@mikro-orm/mssql

### MikroORM version

7.1.4 also tested on 7.1.5-dev.26

### Node.js version

22.20.0

### Operating system

Ubuntu

### Validations

NoahStarkenburg:
In table-per-type inheritance an inherited column lives only on the parent table, but its check constraint (for example the one generated for an enum) was still emitted onto each child table. The child-table check then references a column that does not exist there, so the generated DDL is invalid and `schema.create()` fails.

Checks whose column is not present on the child table are now skipped, the same way parent indexes are handled (#7714).

Closes #7935

### DIFF
--- a/packages/sql/src/schema/DatabaseSchema.ts
+++ b/packages/sql/src/schema/DatabaseSchema.ts
@@ -344,6 +344,14 @@ export class DatabaseSchema {
 
       for (const check of meta.checks) {
         const columnName = check.property ? meta.properties[check.property].fieldNames[0] : undefined;
+
+        // In TPT, inherited columns live on the parent table only, so a check
+        // that references such a column must not be emitted onto the child
+        // table, where that column does not exist.
+        if (meta.inheritanceType === 'tpt' && meta.tptParent && columnName && !table.hasColumn(columnName)) {
+          continue;
+        }
+
         const expression = isRaw(check.expression)
           ? platform.formatQuery(check.expression.sql, check.expression.params)
           : (check.expression as string);
--- a/tests/issues/GH7935.test.ts
+++ b/tests/issues/GH7935.test.ts
@@ -1,0 +1,44 @@
+import { Entity, Enum, PrimaryKey, Property, ReflectMetadataProvider } from '@mikro-orm/decorators/legacy';
+import { MikroORM } from '@mikro-orm/sqlite';
+
+// GH #7935 - An enum CHECK constraint defined on an abstract TPT parent must not
+// be propagated to child TPT tables. In TPT the enum column lives only on the
+// parent table, so a copy of the check on the child references a column that
+// does not exist there and produces invalid DDL (e.g. sqlite `no such column`).
+
+@Entity({ inheritance: 'tpt' })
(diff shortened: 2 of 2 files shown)


B4nan (approved):
thanks!

PR #7951, opened 2026-07-04
test(sqlite): add findOne empty where validation coverage

Ayush7614:
## Summary
- Adds SQLite coverage for `findOne` / `findOneOrFail` rejecting empty, null, and undefined `where` clauses
- Verifies primary-key shorthand and non-empty filters still work as expected

## Test plan
- [x] `yarn test tests/features/entity-manager/find-one-empty-where.sqlite.test.ts`
- [x] `yarn lint`


B4nan:
same here

PR #7953, opened 2026-07-04
test(sqlite): add merge without identifier validation coverage

Ayush7614:
## Summary
- Adds SQLite coverage for `em.merge()` rejecting entity instances and data objects without a primary key
- Verifies merge succeeds when a valid identifier is provided

## Test plan
- [x] `yarn test tests/features/entity-manager/merge-without-identifier.sqlite.test.ts`
- [x] `yarn lint`


B4nan:
we have this covered in tests already

PR #7956, opened 2026-07-06
fix(mysql): treat (current_date) and curdate() date defaults as equal

kalu06:
A date column with a (current_date) default churned a no-op migration on every diff: MySQL/MariaDB report the default back as curdate(), while the entity keeps (current_date), so hasSameDefaultValue() never saw them as equal.

Instead of special-casing DateType in the cross-dialect SchemaComparator, this normalizes the introspected curdate() back to (current_date) via a single MySqlSchemaHelper.DEFAULT_VALUES entry. Direction matters: norm[0] is also the value emitted into the entity's CREATE TABLE DDL, and a bare curdate() there is invalid (a date expression default must be parenthesized) — so the entity side is left untouched and only the introspected side is normalized.

Both dialects report curdate() identically, so the one entry (shared via MariaDbSchemaHelper extends MySqlSchemaHelper) fixes both. Tests cover both.

### DIFF
--- a/packages/sql/src/dialects/mysql/MySqlSchemaHelper.ts
+++ b/packages/sql/src/dialects/mysql/MySqlSchemaHelper.ts
@@ -20,6 +20,7 @@ export class MySqlSchemaHelper extends SchemaHelper {
   static readonly DEFAULT_VALUES = {
     'now()': ['now()', 'current_timestamp'],
     'current_timestamp(?)': ['current_timestamp(?)'],
+    'curdate()': ['(current_date)'],
     '0': ['0', 'false'],
   };
 
--- a/packages/sql/src/dialects/mysql/MySqlSchemaHelper.ts
+++ b/packages/sql/src/dialects/mysql/MySqlSchemaHelper.ts
@@ -20,7 +20,7 @@ export class MySqlSchemaHelper extends SchemaHelper {
   static readonly DEFAULT_VALUES = {
     'now()': ['now()', 'current_timestamp'],
     'current_timestamp(?)': ['current_timestamp(?)'],
-    'curdate()': ['curdate()', '(current_date)'],
+    'curdate()': ['(current_date)', 'curdate()'],
     '0': ['0', 'false'],
   };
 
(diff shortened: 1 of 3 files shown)


packages/sql/src/dialects/mysql/MySqlSchemaHelper.ts
```diff
@@ -20,6 +20,7 @@ export class MySqlSchemaHelper extends SchemaHelper {
   static readonly DEFAULT_VALUES = {
     'now()': ['now()', 'current_timestamp'],
     'current_timestamp(?)': ['current_timestamp(?)'],
+    'curdate()': ['(current_date)'],
```
B4nan:
```suggestion
    'curdate()': ['curdate()', '(current_date)'],
```

B4nan (approved):
thanks!

packages/sql/src/dialects/mysql/MySqlSchemaHelper.ts
```diff
@@ -20,6 +20,7 @@ export class MySqlSchemaHelper extends SchemaHelper {
   static readonly DEFAULT_VALUES = {
     'now()': ['now()', 'current_timestamp'],
     'current_timestamp(?)': ['current_timestamp(?)'],
+    'curdate()': ['curdate()', '(current_date)'],
```
B4nan:
```suggestion
    'curdate()': ['(current_date)', 'curdate()'],
```

packages/sql/src/dialects/mysql/MySqlSchemaHelper.ts
```diff
@@ -20,6 +20,7 @@ export class MySqlSchemaHelper extends SchemaHelper {
   static readonly DEFAULT_VALUES = {
     'now()': ['now()', 'current_timestamp'],
     'current_timestamp(?)': ['current_timestamp(?)'],
+    'curdate()': ['(current_date)', 'curdate()'],
```
B4nan, line 23: i feel like this was supposed to look like this, the key should be the canonic form. lets see if the tests pass or not, but the original way was IMO wrong.


```suggestion
    '(current_date)': ['(current_date)', 'curdate()'],
```

kalu06: Alright, thanks 👍

PR #7959, opened 2026-07-07
fix(mssql): map FK, CHECK, PK/unique, deadlock and lock-timeout errors to driver exceptions

Subtletree:
The MSSQL exception converter only recognised 7 error numbers; several common constraint violations fell through to the generic DriverException. Map them to the appropriate typed exceptions, matching the other dialects and Doctrine DBAL:

- 547 -> ForeignKeyConstraintViolationException, or CheckConstraintViolationException when the message names a CHECK constraint (547 is shared by both; the FOREIGN KEY/REFERENCE/CHECK keyword stays English even on localized servers, so it is matched on the bare token)
- 4712 (TRUNCATE blocked by an FK reference) -> ForeignKeyConstraintViolationException
- 2627 (PK/unique constraint) -> UniqueConstraintViolationException, alongside the existing 2601 (unique index)
- 1205 -> DeadlockException, 1222 -> LockWaitTimeoutException

### DIFF
--- a/packages/mssql/src/MsSqlExceptionConverter.ts
+++ b/packages/mssql/src/MsSqlExceptionConverter.ts
@@ -1,6 +1,10 @@
 import {
+  CheckConstraintViolationException,
+  DeadlockException,
   ExceptionConverter,
+  ForeignKeyConstraintViolationException,
   InvalidFieldNameException,
+  LockWaitTimeoutException,
   NonUniqueFieldNameException,
   NotNullConstraintViolationException,
   SyntaxErrorException,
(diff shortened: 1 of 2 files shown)


packages/mssql/src/MsSqlExceptionConverter.ts
```diff
@@ -14,8 +18,8 @@ import {
 /** Converts MSSQL native errors into typed MikroORM driver exceptions. */
 export class MsSqlExceptionConverter extends ExceptionConverter {
   /**
-   * @see https://docs.microsoft.com/en-us/sql/relational-databases/errors-events/mssqlserver-511-database-engine-error?view=sql-server-ver15
-   * @see https://github.com/doctrine/dbal/blob/master/src/Driver/AbstractPostgreSQLDriver.php
+   * @see https://learn.microsoft.com/en-us/sql/relational-databases/errors-events/database-engine-events-and-errors
+   * @see https://github.com/doctrine/dbal/blob/master/src/Driver/AbstractSQLServerDriver.php
```
B4nan: this link doesnt exist

Subtletree:
Fixed lint error

packages/mssql/src/MsSqlExceptionConverter.ts
Subtletree: Ah dam, have updated with a correct and more relevant link. AI force-pushed without asking sorry, would have preferred a new commit.

B4nan (approved):
thanks!

PR #7961, opened 2026-07-08
fix(core): infer `string` for `time` type properties instead of `any`

kalu06:
`time`-typed properties defined via `defineEntity` currently infer as `any`:

```ts
const Foo = defineEntity({
  name: 'Foo',
  properties: p => ({
    t1: p.time(),        // any
    t2: p.type('time'),  // any
  }),
});
```

TimeType extended Type without type arguments (relying on the JSType = string default) and overrode convertToDatabaseValue(value: any, …). The defineEntity inference helper InferJSType derives the runtime type from every position where Type's first type parameter appears — including that value: any parameter — so the any poisoned the inference and collapsed the result to any. DateType avoids this because its value: any sits in convertToJSValue (the DBType slot, already erased in the matcher).

Fix: declare the type parameters explicitly (Type<string | null | undefined, …>, mirroring DateType) and type the override's parameter/return so the inference resolves to string. Runtime behavior is unchanged.

### DIFF
--- a/packages/core/src/types/TimeType.ts
+++ b/packages/core/src/types/TimeType.ts
@@ -4,9 +4,9 @@ import type { EntityProperty } from '../typings.js';
 import { ValidationError } from '../errors.js';
 
 /** Maps a database TIME column to a JS `string` in HH:MM:SS format. */
-export class TimeType extends Type {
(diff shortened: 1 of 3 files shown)


B4nan (approved):
thanks!

PR #7968, opened 2026-07-09
fix: Custom Type desyncs __originalEntityData causing bad changesets

### LINKED ISSUE #7966: Custom Type desyncs __originalEntityData causing bad changesets
### Describe the bug

When a to-one relation is assigned a raw scalar PK (via `em.create()`'s data, or equivalently `em.getReference(Entity, id)`), and that PK's `Type` maps between two **different string shapes** in JS-form vs DB-form (not just casing/formatting — e.g. ULID ↔ UUID), MikroORM's change-tracking snapshot (`__originalEntityData`) ends up stored in JS-form, while `EntityComparator.prepareEntity()` always derives DB-form on every subsequent flush. The two never compare equal, so MikroORM believes the PK changed and emits `UPDATE <table> SET id = <db-form> WHERE id = <js-form>` on every flush touching that entity — even though it was never modified by application code.

On loosely-typed PK columns this fails silently (0 rows affected). On Postgres with a native `uuid` column it throws `invalid input syntax for type uuid`.

## Root cause

- `EntityComparator.prepareEntity()` unconditionally runs any scalar prop with a `customType` through `convertToDatabaseValue()`.
- `UnitOfWork.register()` seeds `__originalEntityData` directly from the raw hydration input the first time an entity is registered, with no equivalent conversion.
- Whether that raw input gets converted to DB-form first depends on `ObjectHydrator`'s `hydrateScalar`, which only writes the DB-form value back when **both** `convertCustomTypes` and `customType.ensureComparable()` are `true`.
- `convertCustomTypes` defaults to `false` for this exact path (`EntityManager.getReference()`'s documented default, and the same default in `EntityFactory.extractConstructorParams()` for relations assigned a raw scalar in `em.create()`) — correct for setting `entity.id` itself, but it means the DB-form write-back never runs, so the snapshot is seeded straight from the un-converted JS-form value.

Doesn't surface with built-in `UuidType`/`ObjectId` since their JS/DB forms are close enough to identical. Surfaces for any custom `Type` where the two forms are genuinely different strings.

---
Root-caused with help from Claude Code, which traced the mechanism through the hydration/change-tracking source and reduced it to the driver-agnostic repro linked above and helped me prepare this Bug report.

### Reproduction

https://gist.github.com/Kalos-S/d34963fa0e2448ee2f82599ffb466f91

None

### MikroORM version

7.1.5

### Node.js version

26.0.0

### Operating system

_No response_

### Validations

weknowyourgame:
fixes #7966

### DIFF
--- a/packages/core/src/entity/EntityFactory.ts
+++ b/packages/core/src/entity/EntityFactory.ts
@@ -202,7 +202,8 @@ export class EntityFactory {
         loaded: options.initialized,
       });
 
-      if (options.recomputeSnapshot) {
+      // `data` may still hold un-converted customType values here, so recompute from the entity itself (GH #7966).
+      if (!options.newEntity && (options.recomputeSnapshot || meta.forceConstructor || !options.convertCustomTypes)) {
         wrapped.__originalEntityData = this.#comparator.prepareEntity(entity);
       }
     }
--- a/tests/issues/GH7966.test.ts
+++ b/tests/issues/GH7966.test.ts
@@ -1,0 +1,106 @@
+import { MikroORM, Type, helper } from '@mikro-orm/sqlite';
+import { Entity, ManyToOne, PrimaryKey, Property, ReflectMetadataProvider } from '@mikro-orm/decorators/legacy';
+import { mockLogger } from '../helpers.js';
+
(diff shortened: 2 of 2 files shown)


B4nan:
This is not passing the full test suite, closing in favor of #7973

PR #7977, opened 2026-07-11
fix(sql): coerce to.default to string in SchemaComparator.hasSameDefaultValue

### LINKED ISSUE #7976: SchemaComparator.hasSameDefaultValue throws "to.default.toLowerCase is not a function" on non-string (numeric/boolean) column defaults
## Describe the bug

`SchemaComparator.hasSameDefaultValue(from, to)` handles `from.default` defensively (`from.default.toString().toLowerCase()`) but calls `.toLowerCase()` **directly** on `to.default` at the two `=== 'null'` guards:

```ts
hasSameDefaultValue(from: Column, to: Column): boolean {
  if (
    from.default == null ||
    from.default.toString().toLowerCase() === 'null' ||   // <- .toString() (safe)
    from.default.toString().startsWith('nextval(')
  ) {
    return to.default == null || to.default.toLowerCase() === 'null';   // <- no .toString()  💥
  }

  if (to.default == null || to.default.toLowerCase() === 'null') {      // <- no .toString()  💥
    return false;
  }
  // ...
}
```

When `to.default` is a non-string value (a **number** or **boolean** — e.g. a column whose entity default is `@Property({ default: 0 })` or `@Property({ default: true })`), `to.default.toLowerCase()` throws:

```
TypeError: to.default.toLowerCase is not a function
```

This aborts schema diffing / migration generation. The `from` side already anticipates non-string defaults (hence its `.toString()`), so the `to` side is simply missing the same coercion.

## Reproduction

No database connection required — the crash is in the pure comparison method:

```js
import { MikroORM, EntitySchema, PostgreSqlDriver } from '@mikro-orm/postgresql';
import { SchemaComparator } from '@mikro-orm/knex';

const Foo = new EntitySchema({ name: 'Foo', properties: { id: { type: 'integer', primary: true } } });
const orm = await MikroORM.init({ driver: PostgreSqlDriver, entities: [Foo], connect: false });
const comparator = new SchemaComparator(orm.em.getPlatform());

// from.default == null hits the first guard; to.default is a number → crash
comparator.hasSameDefaultValue({ default: null }, { default: 0 });
// TypeError: to.default.toLowerCase is not a function
// (same with { default: false })

await orm.close(true);
```

Encountered in practice when running migration generation (`migrator.create` / schema diff) over a set of entities that include numeric/boolean column defaults.

## Expected behavior

`hasSameDefaultValue` should treat `to.default` as defensively as `from.default` and not crash on a non-string default — e.g. `hasSameDefaultValue({ default: null }, { default: 0 })` should return `false` (a null DB default and a `0` entity default are not equal).

## Suggested fix

Mirror the `from` side — coerce `to.default` with `.toString()` on both guards:

```ts
return to.default == null || to.default.toString().toLowerCase() === 'null';
// ...
if (to.default == null || to.default.toString().toLowerCase() === 'null') {
```

## Versions

| Dependency | Version |
| --- | --- |
| `@mikro-orm/knex` / `@mikro-orm/sql` | 7.1.4 (verified present on `master` and in the published 7.1.5) |
| node | 24 |
| Database | PostgreSQL (bug is driver-agnostic — lives in `@mikro-orm/sql`) |

I have a fix + a DB-free regression test ready and will open a PR.

SiphoChris:
## What's this PR do?

Fixes a crash in `SchemaComparator.hasSameDefaultValue` when a column's `to.default` is a non-string value.

The method coerces `from.default` defensively (`from.default.toString().toLowerCase()`) but calls `.toLowerCase()` **directly** on `to.default` at the two `=== 'null'` guards. When `to.default` is a **number** or **boolean** (a column whose entity default is `@Property({ default: 0 })` or `@Property({ default: true })`), it throws:

```
TypeError: to.default.toLowerCase is not a function
```

…which aborts schema diffing / migration generation. This PR mirrors the `from` side and coerces `to.default` with `.toString()` on both guards.

Closes #7976.

## The change

```diff
-      return to.default == null || to.default.toLowerCase() === 'null';
+      return to.default == null || to.default.toString().toLowerCase() === 'null';
     }

-    if (to.default == null || to.default.toLowerCase() === 'null') {
+    if (to.default == null || to.default.toString().toLowerCase() === 'null') {
```

## Tests

Added a DB-free unit test (`tests/features/schema-generator/GH7976.test.ts`) that constructs a `SchemaComparator` from an in-memory ORM and asserts `hasSameDefaultValue` no longer throws on numeric/boolean `to.default`, and still returns the correct result:

- `hasSameDefaultValue({ default: null }, { default: 0 })` → `false` (was: `TypeError`)
- `hasSameDefaultValue({ default: null }, { default: false })` → `false` (was: `TypeError`)
- `hasSameDefaultValue({ default: null }, { default: null })` → `true` (unchanged)

Verified downstream against the published `@mikro-orm/sql@7.1.4`: before the change `migrator.create()` over entities with numeric/boolean column defaults crashed with the above `TypeError`; after the change it completes. The bug is present on `master` and in the published `7.1.5`.

## Checklist

- [x] `fix` conventional-commit message
- [x] DCO `Signed-off-by`
- [x] Added a regression test
- [x] No public API change (behaviour-only fix)

### DIFF
--- a/packages/sql/src/schema/SchemaComparator.ts
+++ b/packages/sql/src/schema/SchemaComparator.ts
@@ -1247,10 +1247,10 @@ export class SchemaComparator {
       from.default.toString().toLowerCase() === 'null' ||
       from.default.toString().startsWith('nextval(')
     ) {
-      return to.default == null || to.default.toLowerCase() === 'null';
+      return to.default == null || to.default.toString().toLowerCase() === 'null';
     }
 
-    if (to.default == null || to.default.toLowerCase() === 'null') {
+    if (to.default == null || to.default.toString().toLowerCase() === 'null') {
       return false;
     }
 
--- a/tests/features/schema-generator/GH7976.test.ts
+++ b/tests/features/schema-generator/GH7976.test.ts
@@ -1,0 +1,39 @@
+import { MikroORM, SchemaComparator } from '@mikro-orm/sqlite';
+import type { Column } from '@mikro-orm/sql';
+
+// GH #7976 — SchemaComparator.hasSameDefaultValue threw `to.default.toLowerCase is not a function`
+// when `to.default` was a non-string (a number or boolean — e.g. a column whose entity default is
+// `@Property({ default: 0 })` or `@Property({ default: true })`). The `from` side was already coerced
+// with `.toString()`; the `to` side was not.
+describe('GH 7976', () => {
+  let orm: MikroORM;
+  let comparator: SchemaComparator;
+
+  const col = (def: unknown): Column => ({ default: def } as unknown as Column);
+
+  beforeAll(async () => {
+    orm = await MikroORM.init({ dbName: ':memory:', entities: [], discovery: { warnWhenNoEntities: false } });
+    comparator = new SchemaComparator(orm.em.getPlatform());
+  });
+
+  afterAll(() => orm.close(true));
+
+  it('does not throw on a numeric `to.default` (from.default null)', () => {
+    expect(() => comparator.hasSameDefaultValue(col(null), col(0))).not.toThrow();
+    expect(comparator.hasSameDefaultValue(col(null), col(0))).toBe(false);
+  });
+
+  it('does not throw on a boolean `to.default` (from.default null)', () => {
+    expect(() => comparator.hasSameDefaultValue(col(null), col(false))).not.toThrow();
+    expect(() => comparator.hasSameDefaultValue(col(null), col(true))).not.toThrow();
+    expect(comparator.hasSameDefaultValue(col(null), col(false))).toBe(false);
+  });
+
+  it('does not throw on a numeric `to.default` (from.default a real value)', () => {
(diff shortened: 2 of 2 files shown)


SiphoChris:
Closing — this turned out to be on our side, not MikroORM's.

Root cause was a corrupted migration snapshot: one column's default had been recorded as the raw JSON boolean `false` instead of the string `"false"` that the schema generator normally emits. Every other boolean column in our codebase was `"false"`/`"true"` (string); that single hand-edited entry was the only non-string default in the whole schema, and it's what fed a boolean into `hasSameDefaultValue` (crashing `to.default.toLowerCase()`). Fixing the snapshot to the string form the generator itself produces resolves it completely, with no framework change and no patch.

Thanks for the fast response and for pushing back — you were right that column defaults are string-normalized, so a trivial repro never reproduced. If you'd still like the defensive `.toString()` coercion at the two `=== 'null'` guards as hardening (the `BooleanType` branch already does `'' + to.default`), I'm happy to leave the PR open; otherwise I'll close both. Closing for now.

PR #7984, opened 2026-07-14
feat(core): add native support for row level security

B4nan:
Adds first-class PostgreSQL row level security support, as discussed in #6137 (and #6413, #6493).

## What's included

- **Policies as entity metadata** — `policies` (which implies RLS enabled) and `rowLevelSecurity: boolean | 'force'` (with `false` for staging policies without enforcement) options on decorators, `defineEntity` and `EntitySchema`, with `using`/`check` expressions as strings, `raw()` or typed callbacks receiving the property→column mapping. Full schema generator lifecycle: DDL emission, `pg_policies` introspection, diffing (changed policies are dropped and recreated) and migration snapshots. Dropping/recreating the schema now preserves declared policies (#6493).
- **Session context at runtime** — `em.fork({ session: { variables, role } })`, `em.setSessionContext()` / `clearSessionContext()`. Under the default `sessionContext: 'transaction'` strategy the ORM emits `select set_config(..., true)` and `set local role` right after `begin`, and wraps out-of-transaction operations (including raw `em.execute()`, native ops, upsert reloads and pivot loads) in a short implicit transaction — only when a session context is set, zero overhead otherwise. The `'connection'` strategy applies the context on every pool acquire via a composed `onReserveConnection` instead. Write violations map to `RowLevelSecurityViolationException`; result cache keys include the session context.
- **Filter bridge** — `@Filter({ ..., rls: true })` compiles the filter condition into a database policy at schema time (casts derived from property types, session variable names `mikro.<filter>.<arg>`, overridable via `rls: { setting }`), and `em.setFilterParams()` stages the matching session variables — one declaration drives both the app-level `where` and the database-level policy. Non-compilable conditions fail loudly at schema time.
- **Entity generator** round-trips policies and RLS flags in all output modes.
- **Docs**: new `row-level-security.md` guide including the operational caveats (owner/superuser bypass vs `force`, schema role vs app role grants, fail-closed behavior and the pooled-connection empty-string GUC subtlety, pgBouncer modes).

PostgreSQL only (`postgresql` + `pglite`; pglite supports the `'transaction'` strategy only, as there is no pool to hook); other drivers reject RLS metadata at discovery. The e2e suite runs against a non-owner database role so the isolation guarantees are tested for real, and schema tests assert live round-trips (apply diff → next diff is empty) for every policy mutation class.


rinode:
Is there an ETA on when this will be available? We are very interested in this native support!

B4nan:
I didn't really find the time to read the diff just yet 🙃 But it will land in v7.2 for sure, and once merged, it will be available as a canary build on npm. I hope I will get to this in the upcoming days finally.

B4nan:
cc @RWOverdijk, I remember you were trying to hack this yourself some time ago

PR #7989, opened 2026-07-16
ci: pin third-party actions to full commit SHAs

kobihikri:
CI supply-chain hardening: every third-party action in all four workflow files (`tests.yml`, `release.yml`, `docs.yml`, `codeql-analysis.yml`) is now pinned to a full commit SHA, with the resolved version kept as a trailing comment. A tag or branch can move after review; a SHA can't. Renovate understands the `sha # vX.Y.Z` convention and bumps the digest and comment together.

The original PR pinned the three mutable-ref usages (`crate-ci/typos@master` twice, `apify/workflows/execute-workflow@main` in the release-dispatch job). The maintainer follow-up commits extend the same treatment to the remaining actions (checkout, setup-node, codeql-action, setup-deno, codecov, cpu-cores, the Pages actions) and bump the pins to current: typos v1.49.0 and the latest `apify/workflows` main. All other pinned SHAs already match each action's latest release, so behavior is unchanged.

Thanks @kobihikri for kicking this off.


B4nan (approved):
thanks! i've updated the other places as well so its consistent

PR #8004, opened 2026-07-18
fix(core): fix infinite recursion when flushing a `mapToPk` FK with value `0`

### LINKED ISSUE #8003: Flushing an entity with a `mapToPk` m:1 FK set to `0` hangs in infinite recursion (RangeError: Maximum call stack size exceeded)
**Describe the bug**
When an entity has a `ManyToOne` relation with `mapToPk: true` and the FK value is `0` (a legal PK for integer keys, e.g. legacy schemas), `em.flush()` crashes with `RangeError: Maximum call stack size exceeded`.

**To reproduce**
```ts
@Entity()
class Author {
  @PrimaryKey()
  id!: number;
}

@Entity()
class Book {
  @PrimaryKey()
  id!: number;

  @ManyToOne(() => Author, { mapToPk: true })
  author!: number;
}

const em = orm.em.fork();
em.create(Book, { id: 1, author: 0 });
await em.flush(); // RangeError: Maximum call stack size exceeded
```
Deterministic on current master with better-sqlite; any driver should reproduce since the loop is driver-independent.

**Root cause**
`ChangeSetComputer.processProperty` guards with `if (!target)` — a raw FK of `0` is falsy, so it's treated as "no target" and the unwrap branch re-enters itself forever. `Utils.unwrapProperty` only yields non-null values and `processToOne` early-returns for `mapToPk`, so relaxing the guard to `target == null` is behavior-safe for `0`/`''`/`false`.

Related: #8000 covers the read side of the same falsy-FK class (1:m collections not populated when the parent PK is `0`). This one is the write side and is a hard crash.

I have a fix with a regression test ready and will open a PR right after this issue.

xianjianlf2:
Closes #8003

`ChangeSetComputer.processProperty` guarded the unwrap branch with `if (!target)`, so a raw `mapToPk` FK of `0` was treated as "no target" and the branch re-entered itself until the stack overflowed — `em.flush()` on such an entity crashed with `RangeError: Maximum call stack size exceeded`, deterministically.

Fix relaxes the guard to `target == null`. This is behavior-safe for the other falsy values too: `Utils.unwrapProperty` only yields non-null values, and `processToOne` early-returns for `mapToPk`, so `0`/`''`/`false` FKs flow through the same path a truthy PK does.

Companion to #8002 — that PR fixes the read side of the falsy-FK class (1:m population), this one fixes the write side (flush crash).

Verified: `GH8003.test.ts` red on master (stack overflow) → green after; unit-of-work sqlite suites pass; `yarn build` / format / lint clean.

### DIFF
--- a/packages/core/src/unit-of-work/ChangeSetComputer.ts
+++ b/packages/core/src/unit-of-work/ChangeSetComputer.ts
@@ -185,7 +185,8 @@ export class ChangeSetComputer {
   }
 
   private processProperty<T extends object>(changeSet: ChangeSet<T>, prop: EntityProperty<T>, target?: unknown): void {
-    if (!target) {
+    // check for `null`/`undefined` explicitly, the target can be a falsy raw PK value like `0` (e.g. with `mapToPk`)
+    if (target == null) {
       const targets = Utils.unwrapProperty(changeSet.entity, changeSet.meta, prop);
       targets.forEach(([t]) => this.processProperty(changeSet, prop, t));
 
--- a/tests/issues/GH8003.test.ts
+++ b/tests/issues/GH8003.test.ts
@@ -1,0 +1,53 @@
+import { Collection } from '@mikro-orm/core';
+import {
+  Entity,
+  ManyToOne,
+  OneToMany,
+  PrimaryKey,
+  Property,
+  ReflectMetadataProvider,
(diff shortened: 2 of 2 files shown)


B4nan (approved):
thanks

PR #8005, opened 2026-07-19
fix(sql): preserve query builder field types across joins and CTEs

takaram:
Preserve selected field types in `QueryBuilder` return types when:

- Adding joins, including lateral joins, after `select()`
- Switching to a CTE with `from()`

This prevents selected-field information from being lost during query builder chaining.


B4nan (approved):
next time please run the tests to ensure you didnt break something 🙃 

thanks anyway!

PR #8028, opened 2026-07-26
fix(sql): avoid invalid join when nested populate mixes joined and select-in loads

Nahid-NHB:
Nested populate of two `OneToMany` collections that share the same inverse side, where one has a `where` filter, could generate invalid SQL when the load strategy splits them across a JOIN and a separate SELECT_IN query.

`buildPopulateWhere` recursed into every child hint unconditionally. A child that gets loaded via SELECT_IN already applies its own `where` in that query, so folding the same condition into the join tree here either attached it to the wrong join or forced an extra join just to express it.

With a TPT entity that extra join ended up referencing a parent-table alias from an out-of-scope nested join:

- sqlite: `no such column: b1.archived_at`
- postgres: `invalid reference to FROM-clause entry for table "b1"`

Fix restricts the recursion to children that actually land in the joined-load tree at that nesting level, using `joinedProps` with the resolved strategy.

Repro in `tests/issues/GHx60.test.ts`.

### DIFF
--- a/packages/sql/src/AbstractSqlDriver.ts
+++ b/packages/sql/src/AbstractSqlDriver.ts
@@ -3105,7 +3105,7 @@ export abstract class AbstractSqlDriver<
   protected buildPopulateWhere<T extends object>(
     meta: EntityMetadata<T>,
     joinedProps: PopulateOptions<T>[],
-    options: Pick<FindOptions<any>, 'populateWhere'>,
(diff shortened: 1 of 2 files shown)


B4nan:
Closing as already fixed by #8030, next time please reference the issue you are trying to solve, I'd prefer merging your PR if I knew it was about the same thing :]

PR #8041, opened 2026-07-29
fix(oracle): resolve the default schema name from the connection user

### LINKED ISSUE #8040: [Oracle] SchemaGenerator fails to discover existing tables when database user !== dbName
### Describe the bug

When running a schema update (schema:update or programmatically via orm.schema.update()) on an Oracle database, MikroORM fails to detect existing tables if the database user is different from the dbName specified in the connection configuration.

Because the existing schema metadata is not properly resolved or filtered under these conditions, the Schema Generator assumes the tables do not exist and generates redundant CREATE TABLE statements. Executing these statements results in a TableExistsException: ORA-00955: name is already used by an existing object error.

If the user and dbName match identically, the ORM correctly discovers the existing tables, the dump is empty as expected, and the update runs smoothly.

### Reproduction

https://github.com/miesgre/mikro-orm-repro-oracle-update

@mikro-orm/oracledb

### MikroORM version

7.1.8

### Node.js version

24

### Operating system

Linux, OSX

### Validations

lazerg:
On Oracle the connection opens as the configured `user` and only falls back to `dbName` when no user is set, so tables end up owned by that user's schema. `getDefaultSchemaName()` returned `dbName` regardless, so when the two differ the schema introspection filtered `all_tables` by the wrong owner, never saw the existing tables, and re-emitted `CREATE TABLE`, which fails with `ORA-00955: name is already used by an existing object`.

This resolves the default schema from the connection user the same way `OracleConnection` does, and keeps the `dbName` fallback so setups where the user is not set behave exactly as before.

Fixes #8040

### DIFF
--- a/packages/oracledb/src/OraclePlatform.ts
+++ b/packages/oracledb/src/OraclePlatform.ts
@@ -187,7 +187,7 @@ export class OraclePlatform extends AbstractSqlPlatform {
   }
 
   override getDefaultSchemaName(): string | undefined {
-    return this.config.get('dbName');
+    return this.config.get('user', this.config.get('dbName'));
   }
 
   override getVarcharTypeDeclarationSQL(column: { length?: number }): string {
--- a/tests/features/schema-generator/default-schema-name.oracle.test.ts
+++ b/tests/features/schema-generator/default-schema-name.oracle.test.ts
@@ -1,0 +1,29 @@
+import { MikroORM, OracleDriver, OraclePlatform, type Options } from '@mikro-orm/oracledb';
+
+async function init(options: Partial<Options>) {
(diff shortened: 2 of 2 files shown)


lazerg:
Closing this. Deriving the default schema from the configured `user` looks right in isolation but regresses the Oracle schema generator: during setup it reconnects as the management user and leaves that value on the config, so introspection then filters `all_tables` by the wrong owner. The `SchemaGenerator.oracle3` tests catch it.

A correct fix needs the owner resolved from the live connection rather than the mutable config, validated against a real Oracle instance (including identifier casing). I would rather withdraw this than leave a half-right change on record.

PR #8047, opened 2026-07-29
fix(core): keep a nulled embedded array in the entity snapshot

### LINKED ISSUE #8045: Setting a nullable `array: true` embedded to `null` is silently dropped when passing an entity instance to `qb().update()`
### Describe the bug

We have observed behaviour where when an entity instance (rather than a plain object) is passed to `qb().update()`, a nullable embedded array property that has been set to `null` is silently removed from the update payload. The column is omitted from the `SET` clause entirely, so the previous (stale) value remains in the database with no error or warning. Is this expected behaviour?

Passing a plain object with the same `null` value does set the value to null when passed to the QueryBuilder update, it only seems to occur when entity instances are passed through the QueryBuilder.

This previously was not an issue for us when using v5.2 but the upgrade to v7 has both changed the handling of undefined values, where previously they were converted to NULL but are now just omitted from the 'SET'. Also when we explicitly set an undefined value to NULL all are included in the `SET` with a NULL value, but the only exception is the nullable embedded array property 

## Potential cause

`QueryBuilder` converts entity instances to data via the comparator's snapshot generator:

```js
// QueryBuilder.js
if (Utils.isEntity(data)) {
    data = this.em?.getComparator().prepareEntity(data) ?? serialize(data);
}
```

The compiled snapshot code generated by `EntityComparator` guards `array: true` embeddables with `Array.isArray` and has no branch for `null`:

```js
// generated snapshot code
if (Array.isArray(entity.formats)) {
    ret.formats = [];
    // ...
    ret.formats = cloneEmbeddable(ret.formats);
}
// no else branch — null falls through, key never added to snapshot

//Actual code in the repo
getEmbeddedArrayPropertySnapshot(meta, prop, context, level, path, dataKey) { 
	const entityKey = path.map(k => this.wrap(k)).join(''); 
	const ret = []; 
	const padding = ' '.repeat(level * 2); 
	const idx = this.#tmpIndex++; 
	
	ret.push(`${padding}if (Array.isArray(entity${entityKey})) {`); // ← the guard 
	ret.push(`${padding} ret${dataKey} = [];`); 
	ret.push(`${padding} entity${entityKey}.forEach((_, idx_${idx}) => {`); 
	ret.push(this.getEmbeddedPropertySnapshot(...)); 
	ret.push(`${padding} });`); 
	if (this.shouldSerialize(prop, dataKey)) { 
		ret.push(`${padding} ret${dataKey} = cloneEmbeddable(ret${dataKey});`); 
	} 
	ret.push(`${padding}}`); // ← closes the if; no else 
	return ret.join('\n'); 
}
```

`Array.isArray(null)` is `false`, so the key never lands in the prepared data. Compare with scalar props, which are guarded with `typeof entity.prop !== 'undefined'` — that passes for `null`, which is why nulling scalar columns through the same path works. Non-array (`object: true`) embeddeds also handle `null` correctly; only the `array: true` shape drops it.

The following code replicates the issue

## Expected behaviour

If this is an issue would the expected behaviour be when setting a nullable `array: true` embedded property to `null` on an entity instance and passing it to `qb().update()` should produce `formats = null` in the `SET` clause, consistent with:

- passing a plain object payload with `formats: null`,
- scalar and `object: true` embedded properties through the same entity-instance path.

A possible fix is an `else if (entity.formats === null) ret.formats = null;` branch in the generated snapshot code (possibly only when the QB is the consumer), or having the QueryBuilder use a proper serializer instead of the comparator snapshot for update payloads.

### Reproduction

```ts
import { MikroORM, EntitySchema } from '@mikro-orm/sqlite';

type Format = {
   name: string;
};

type Doc = {
  id: number;
  formats: Format[] | null;
};

const FormatSchema = new EntitySchema<Format>({
  name: 'Format',
  embeddable: true,
  properties: { name: { type: 'string' } },
});

const DocSchema = new EntitySchema<Doc>({
  name: 'Doc',
  properties: {
  id: { type: 'number', primary: true },
  formats: { kind: 'embedded', entity: () => FormatSchema, array: true, nullable: true },
 },
});

const orm = await MikroORM.init({ dbName: ':memory:', entities: [FormatSchema, DocSchema],  allowGlobalContext: true });
await orm.schema.refresh();
await orm.em.createQueryBuilder(DocSchema).insert({ id: 1, formats: [{ name: 'a' }] }).execute();

const entity = orm.em.create(DocSchema, { id: 1, formats: [{ name: 'x' }] }, { persist: false });
entity.formats = null;

// 1. prepareEntity drops the null embedded array entirely
console.log(orm.em.getComparator().prepareEntity(entity));

// actual: { id: 1 } <- formats key missing
// expected: { id: 1, formats: null }

// 2. so qb().update(entity) omits the column from the SET clause

const qb = orm.em.createQueryBuilder(DocSchema).update(entity).where({ id: 1 });
console.log(qb.getFormattedQuery());

// actual: update `doc` set `id` = 1 where `id` = 1
// expected: update `doc` set `id` = 1, `formats` = null where `id` = 1

await qb.execute();
console.log(await orm.em.getConnection().execute('select * from doc'));
// actual: [ { id: 1, formats: '[{"name":"a"}]' } ] <- stale value silently kept

// 3. plain object payload works correctly (inconsistent with entity path)

const qb2 = orm.em.createQueryBuilder(DocSchema).update({ formats: null }).where({ id: 1 });
console.log(qb2.getFormattedQuery());

// update `doc` set `formats` = null where `id` = 1

await orm.close(true);
```

@mikro-orm/postgresql

### MikroORM version

7.1.4

### Node.js version

22

### Operating system

_No response_

### Validations

lazerg:
Passing an entity instance to `qb().update()` routes it through `EntityComparator.prepareEntity`. The snapshot generated for a nullable `array: true` embedded property only covered the `Array.isArray` case, so setting the property to `null` fell through without any branch. The key never landed in the prepared data, the column dropped out of the `SET` clause, and the stale value stayed in the database.

Scalar and `object: true` embedded properties already keep their `null`, and a plain object payload with `formats: null` writes it correctly, so only the entity-instance array path was affected. The array branch now handles `null` the same way, recording `null` for the key.

Closes #8045

### DIFF
--- a/packages/core/src/utils/EntityComparator.ts
+++ b/packages/core/src/utils/EntityComparator.ts
@@ -626,6 +626,8 @@ export class EntityComparator {
       ret.push(`${padding}  ret${dataKey} = cloneEmbeddable(ret${dataKey});`);
     }
 
+    ret.push(`${padding}} else if (entity${entityKey} === null) {`);
+    ret.push(`${padding}  ret${dataKey} = null;`);
     ret.push(`${padding}}`);
 
     return ret.join('\n');
--- a/tests/features/embeddables/__snapshots__/entities-in-embeddables.mongo.test.ts.snap
+++ b/tests/features/embeddables/__snapshots__/entities-in-embeddables.mongo.test.ts.snap
@@ -103,6 +103,8 @@ exports[`embedded entities in mongo > diffing 1`] = `
                 
                 }
               });
+            } else if (entity.profile1.identity.links[idx_0].metas === null) {
+              ret.profile1_identity_links[idx_0].metas = null;
             }
 
             if (typeof entity.profile1 !== 'undefined' && typeof entity.profile1.identity !== 'undefined' && typeof entity.profile1.identity.links !== 'undefined') {
@@ -118,6 +120,8 @@ exports[`embedded entities in mongo > diffing 1`] = `
           }
         });
         ret.profile1_identity_links = cloneEmbeddable(ret.profile1_identity_links);
+      } else if (entity.profile1.identity.links === null) {
+        ret.profile1_identity_links = null;
       }
 
       if (typeof entity.profile1 !== 'undefined' && typeof entity.profile1.identity !== 'undefined') {
@@ -219,6 +223,8 @@ exports[`embedded entities in mongo > diffing 1`] = `
                 
                 }
               });
+            } else if (entity.profile2.identity.links[idx_2].metas === null) {
+              ret.profile2.identity.links[idx_2].metas = null;
             }
 
             if (typeof entity.profile2 !== 'undefined' && typeof entity.profile2.identity !== 'undefined' && typeof entity.profile2.identity.links !== 'undefined') {
@@ -233,6 +239,8 @@ exports[`embedded entities in mongo > diffing 1`] = `
           
           }
         });
+      } else if (entity.profile2.identity.links === null) {
+        ret.profile2.identity.links = null;
       }
 
       if (typeof entity.profile2 !== 'undefined' && typeof entity.profile2.identity !== 'undefined') {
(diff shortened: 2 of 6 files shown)


lazerg:
Closing this as a duplicate. I missed that #8046 had already landed and closed the issue while I was working on the same fix.

PR #8051, opened 2026-07-29
fix(migrations): keep migration-owned triggers out of the snapshot

### LINKED ISSUE #8049: migration:up poisons snapshot with migration-owned external triggers
### Describe the bug

`migrator.up()` rewrites the migration snapshot from live introspection. If a
migration creates a PostgreSQL trigger with custom SQL, that trigger is copied
into the entity-metadata snapshot even though it is not declared in entity
metadata.

The next `migrator.checkSchema()` returns `true` and `migration:check` asks for a
new migration that drops the trigger. This makes migration-owned PostgreSQL
artifacts and snapshots incompatible.

### Reproduction

Minimal pnpm-only reproduction:

https://gist.github.com/vkartaviy/3607e3da5111d8f659a50f84df6394f0

It uses one `EntitySchema` table and one external statement trigger. The script
creates and drops its own uniquely named PostgreSQL database.

```text
{
  "after": {
    "hash": "46488bed6cae2d389ac7a82ca7c8ffe31c214d36f83be406091862c76d5bfdb2",
    "triggerCount": 1
  },
  "before": {
    "hash": "42bc441236a7512dd41b9b78d738d9de835bf5201578a17f918f25db1d9794e5",
    "triggerCount": 0
  },
  "migrationNeeded": true
}
```

### Expected behavior

Migration-owned triggers that are absent from target entity metadata should
remain in the live database but should not be copied into the entity-metadata
snapshot. ORM-managed triggers declared through `EntitySchema.triggers` must
continue to round-trip normally.

### Additional context

The behavior comes from `Migrator.runMigrations()`: after `up`/`down`, it
introspects the live schema and compares/stores it without distinguishing
metadata-managed triggers from migration-owned external triggers.

A local proof-of-concept filters introspected triggers against
`schemaGenerator.getTargetSchema()` before snapshot comparison/storage. It
keeps the external-trigger snapshot byte-identical with
`checkSchema() === false`, while a separate `EntitySchema.triggers` control also
stays byte-identical and retains its managed trigger.

Versions used:

- MikroORM 7.1.8
- PostgreSQL 17.9
- Node.js 24.14.1
- pnpm 11.13.0

B4nan:
`migration:up` and `migration:down` rewrite the snapshot from live introspection, which pulled in triggers created by the migrations themselves. Those are not declared in entity metadata, so the next `checkSchema()` reported drift and `migration:check` asked for another migration to drop them.

Introspected triggers are now filtered against the target schema before the snapshot is compared and stored. Migration-owned triggers stay in the database but never reach the snapshot, while triggers declared via metadata keep round-tripping.

Closes #8049

### DIFF
--- a/packages/migrations/src/Migrator.ts
+++ b/packages/migrations/src/Migrator.ts
@@ -219,6 +219,20 @@ export class Migrator extends AbstractMigrator<AbstractSqlDriver> {
         ctx,
       );
 
+      // triggers created by the migrations themselves are absent from entity metadata, so
+      // snapshotting them would make `checkSchema()` ask for a migration that drops them again
+      const tablesWithTriggers = schema.getTables().filter(table => table.getTriggers().length > 0);
+
+      if (tablesWithTriggers.length > 0) {
+        // wildcard-schema entities declare their triggers in metadata too, so they need to be part of the target
+        const target = this.#schemaGenerator.getTargetSchema(undefined, true);
+
+        for (const table of tablesWithTriggers) {
+          const declared = target.getTable(table.getShortestName(false));
+          table.setTriggers(table.getTriggers().filter(trigger => declared?.hasTrigger(trigger.name)));
+        }
+      }
+
       // keep the snapshot authored by `migration:create` when the migrated DB still matches it
       // semantically — rewriting it from introspection only churns cosmetic serialization noise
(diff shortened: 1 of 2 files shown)


B4nan:
Closing this. `ignoreTriggers: true` is the supported answer for migration-owned triggers, see #8049 for the details.

PR #8062, opened 2026-07-30
fix(migrations): keep dollar-quoted blocks intact when splitting statements

### LINKED ISSUE #8061: PostgreSQL dollar-quoted trigger bodies are split at internal semicolons
## Describe the bug

PostgreSQL trigger functions generated from entity metadata are dollar-quoted
PL/pgSQL blocks. When the trigger body contains more than one statement,
internal semicolons followed by newlines are treated as migration/schema
statement boundaries.

This affects two public paths:

1. `migrator.createInitial()` returns a split `diff.up`, so the generated
   migration contains separate `addSql()` calls for fragments of one function.
2. `orm.schema.refresh()` executes the same fragments separately and fails with
   `syntax error at or near ";"`.

The generated schema SQL before statement splitting is valid. The split result
from the reproduction is:

```text
create table "items" ("id" serial primary key, "value" int not null default 0);
create or replace function "items_items_increment_after_insert_fn"() returns trigger as $$ begin update items set value = value + 1 where id = NEW.id;
return NEW;; end; $$ language plpgsql;
create trigger "items_increment_after_insert" AFTER INSERT on "items" for each ROW execute function "items_items_increment_after_insert_fn"();
```

But `createInitial().diff.up` contains the function as two entries:

```json
[
  "create or replace function ... where id = NEW.id;",
  "return NEW;; end; $$ language plpgsql;"
]
```

The likely common boundary is the SQL statement splitting that recognizes
single-quoted strings but not PostgreSQL dollar-quoted blocks.

Expected: semicolons/newlines inside `$tag$ ... $tag$` remain part of one SQL
statement for both migration generation and schema execution.

## Reproduction

https://gist.github.com/vkartaviy/bd393135dc77bd8f34fa6cd718e185ac

The gist uses only public `EntitySchema`, `Migrator`, and schema-generator APIs,
pins all MikroORM packages to `7.1.8`, uses pnpm, and includes a PostgreSQL 17
Docker command. `pnpm repro` prints the split migration diff and then reproduces
the `schema.refresh()` syntax error.

## Environment

- Driver: `@mikro-orm/postgresql`
- MikroORM: `7.1.8` (current latest stable)
- Node.js: `v24.14.1`
- pnpm: `11.0.0` in the standalone reproduction
- PostgreSQL: `17`
- OS: macOS `26.5.2`

## Validations

- Read the contributing guidelines and docs.
- Searched open and closed issues for trigger-body, dollar-quoted, PL/pgSQL,
  migration-splitting, and schema-refresh reports.
- Reproduced from a clean standalone pnpm install and dedicated PostgreSQL
  database.

lazerg:
The migration statement splitter breaks a SQL dump on `;\n` boundaries and re-merges the pieces that fall inside a single-quoted string. It has no notion of PostgreSQL dollar quoting, so a PL/pgSQL trigger body with more than one statement gets cut at the first internal `;\n`. `createInitial()` then emits the function as two `addSql()` fragments, and neither fragment is valid on its own.

Dollar-quoted blocks (`$$ ... $$` and the tagged `$tag$ ... $tag$` form) are now tracked the same way as string literals, so their contents stay in one statement.

Fixes #8061

### DIFF
--- a/packages/migrations/src/Migrator.ts
+++ b/packages/migrations/src/Migrator.ts
@@ -379,16 +379,54 @@ export class Migrator extends AbstractMigrator<AbstractSqlDriver> {
     // Split SQL by statement boundaries (semicolons followed by newline) rather than
     // just newlines, to preserve multiline statements like view definitions.
     // Blank lines (from double newlines) are preserved as empty strings for grouping.
-    // Splits inside single-quoted string literals are re-merged (GH #7185).
+    // Splits inside single-quoted string literals (GH #7185) or dollar-quoted blocks
+    // like PL/pgSQL function bodies (GH #8061) are re-merged.
     const splitStatements = (sql: string) => {
       const result: string[] = [];
       let buf = '';
 
+      // an unterminated single-quoted string or dollar-quoted block ($$ or $tag$) means the boundary is inside a literal
+      const insideLiteral = (s: string) => {
+        let inString = false;
+        let dollarTag: string | null = null;
+
+        for (let i = 0; i < s.length; i++) {
+          if (dollarTag) {
+            if (s.startsWith(dollarTag, i)) {
+              i += dollarTag.length - 1;
+              dollarTag = null;
+            }
+
+            continue;
+          }
+
+          if (inString) {
+            inString = s[i] !== `'`;
+            continue;
+          }
+
+          if (s[i] === `'`) {
+            inString = true;
+            continue;
+          }
+
+          if (s[i] === '$') {
+            const match = /^\$(?:[a-z_]\w*)?\$/i.exec(s.slice(i));
+
+            if (match) {
+              dollarTag = match[0];
+              i += dollarTag.length - 1;
+            }
+          }
+        }
+
+        return inString || dollarTag !== null;
+      };
+
       for (const chunk of sql.split(/;\n/)) {
         buf += (buf ? ';\n' : '') + chunk;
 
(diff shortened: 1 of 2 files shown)


B4nan:
Thanks for digging into this, and sorry to close it. I went a different way in #8063.

The `;\n` sequence is our statement boundary marker, and generated DDL is not supposed to contain it in the first place; that is what `stripStatementNewlines` is for, and every `createRoutine` already calls it. `createTrigger` just never got it. Teaching the splitter to parse around the marker fixes one of the two consumers: `SqlSchemaGenerator.execute` still splits on `;\n` whenever `supportsMultipleStatements()` is false, so `schema.create/refresh` stays broken on pglite and on mysql without `multipleStatements`. And since dollar quoting is postgres-only, mysql/sqlite/mssql triggers were torn apart the same way and would stay that way.

There is also a second bug behind the report that the splitter side cannot reach: `createTrigger` appends `;` unconditionally, so a body ending with `;` emits `return NEW;; end`. That is the actual source of the `syntax error at or near ";"` in the issue. Plain postgres runs the whole group as one multi-statement query, so the splitting alone never broke `refresh()` there.

One note on the test: `filter(sql => sql.includes('create or replace function'))` matches only the first fragment when the statement is torn, so `toHaveLength(1)` passes on master too. Only the `language plpgsql` assertion actually failed before the fix.

Appreciate the effort, and please do keep them coming.

PR #8094, opened 2026-08-02
fix(core): duck-type `Routine.is()` to survive the CJS/ESM dual-package hazard

### LINKED ISSUE #8092: Routine.is() uses bare instanceof — breaks under the CJS/ESM dual-package hazard that #7289 fixed for EntitySchema
> _Heads up: this report was drafted with the help of an AI coding assistant. I've reviewed it and tried to verify the details against the source, but apologies in advance if anything is inconsistent or wrong — please point it out and I'll gladly correct it._

## Describe the bug

`Routine.is()` still uses a bare `instanceof` check, so it breaks under the **CJS/ESM dual-package hazard** — the same problem #7289 fixed for `EntitySchema.is()`, but `Routine.is()` was not given the equivalent duck-typing fallback.

```ts
// packages/core/src/metadata/Routine.ts  (master)
static is(item: unknown): item is Routine {
  return item instanceof Routine;
}
```

vs the hardened version:

```ts
// packages/core/src/metadata/EntitySchema.ts  (after #7289)
static is(item: unknown): item is EntitySchema {
  if (item instanceof EntitySchema) {
    return true;
  }
  return item != null && typeof item === 'object'
      && item.constructor?.name === 'EntitySchema' && 'meta' in item;
}
```

When a TypeScript config loader (e.g. `tsx`) loads the ORM config in a different module realm than the CLI, the config's `Routine` instances are not `instanceof` the CLI's `Routine` class. `Configuration.normaliseRoutines()` then throws:

```
Error: 'routines' entry is not a stored routine declaration. Use a Routine class instance.
    at Configuration.normaliseRoutines (.../@mikro-orm/core/utils/Configuration.js)
```

So configs that declare `routines: [...]` still fail under exactly the loaders #7289 was meant to unblock. `EntitySchema`-only configs work; the moment a stored routine is registered, the CLI breaks.

## Reproduction

1. MikroORM v7 (`@mikro-orm/core@7.1.9`), a `"type": "commonjs"` project.
2. A config declaring at least one stored routine:
   ```ts
   export default defineConfig({
     // ...
     routines: [myRoutine], // a `new Routine(...)` instance
   });
   ```
3. Run any CLI command through `tsx`, e.g. `MIKRO_ORM_CLI_TS_LOADER=tsx mikro-orm migration:list` (or run the CLI under `tsx`).

→ throws `'routines' entry is not a stored routine declaration. Use a Routine class instance.`

The same config + command with `MIKRO_ORM_CLI_TS_LOADER=jiti` works, because jiti loads `@mikro-orm/core` as a single instance — confirming it's the dual-package hazard, not an invalid routine declaration.

## Expected behavior

`Routine.is()` should tolerate cross-realm instances the way `EntitySchema.is()` does (post-#7289), so `routines`-using configs load under `tsx`/`swc` like the rest of the config after #7289.

## Suggested fix

Mirror the `EntitySchema.is()` fallback — `Routine` instances carry a `type` discriminator:

```ts
static is(item: unknown): item is Routine {
  if (item instanceof Routine) {
    return true;
  }
  return item != null && typeof item === 'object'
      && item.constructor?.name === 'Routine' && 'type' in item;
}
```

## Versions

| Dependency | Version |
| --- | --- |
| `@mikro-orm/core` | 7.1.9 (also reproduced against `master`) |
| Node.js | 22.16 |
| loader | `tsx` 4.21 |
| TypeScript | 7.0.2 |

Related: #7289 (the `EntitySchema.is()` fix that established the pattern).

lazerg:
`Routine.is()` used a bare `instanceof`, so a `Routine` coming from a second copy of `@mikro-orm/core` (the CJS/ESM dual-package hazard, hit under tsx/swc) failed the check and `MikroORM.init` rejected a valid `routines` config. This applies the same duck-typing fallback #7289 added to `EntitySchema.is()`, keyed on the `type` discriminator.

Fixes #8092

### DIFF
--- a/packages/core/src/metadata/Routine.ts
+++ b/packages/core/src/metadata/Routine.ts
@@ -173,6 +173,10 @@ export class Routine<
   }
 
   static is(item: unknown): item is Routine {
-    return item instanceof Routine;
+    if (item instanceof Routine) {
+      return true;
+    }
+
+    return item != null && typeof item === 'object' && item.constructor?.name === 'Routine' && 'type' in item;
   }
 }
--- a/tests/features/stored-routines/validator.test.ts
+++ b/tests/features/stored-routines/validator.test.ts
@@ -184,4 +184,27 @@ describe('stored routines — Routine + validator', () => {
 
     expect(r.returnCustomType).toBe(stringType);
   });
+
+  it('Routine.is', () => {
+    const routine = new Routine({
+      name: 'is_routine_test',
+      type: 'function',
+      returns: { runtimeType: 'number', columnType: 'int' },
+      body: 'select 1',
+    });
+    // real instances pass via instanceof
+    expect(Routine.is(routine)).toBe(true);
+    // duck-type fallback: object with matching constructor name and `type` property
+    const fake = Object.create({ constructor: { name: 'Routine' } });
+    Object.defineProperty(fake, 'type', { value: 'function', enumerable: true });
+    expect(Routine.is(fake)).toBe(true);
+    // non-matching values
+    expect(Routine.is(null)).toBe(false);
+    expect(Routine.is(undefined)).toBe(false);
+    expect(Routine.is({})).toBe(false);
+    expect(Routine.is('string')).toBe(false);
+    expect(Routine.is(42)).toBe(false);
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    expect(Routine.is(() => {})).toBe(false);
+  });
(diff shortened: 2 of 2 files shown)


lazerg:
The one red shard (Tests node 26, 4/4) failed while pulling the Docker images, not in the tests. All the other database shards passed. A re-run should clear it.

B4nan (approved):

PR #8097, opened 2026-08-02
fix(core): map cursor offsets based on property types

DASPRiD:
## Summary

`Cursor.decode` restored types by sniffing the string shape: any timestamp-looking string became a JS `Date`, regardless of the target property. Custom types received a `Date` their converters reject (found with a Temporal `Instant` wrapper type, where every paged request failed on postgres), sub-millisecond precision was silently truncated, and plain string properties holding timestamp-shaped values were coerced into dates, returning wrong pages. On top of that, the cursor condition is built at the driver level, after the EM has already run custom type conversion on the user `where`, so cursor offsets never went through `convertToDatabaseValue` at all.

`Cursor.decode` now returns raw JSON values, and the offsets are mapped per property inside the existing `createCursorCondition` recursion:

- ISO strings become `Date` instances only when the property maps to a datetime column (checked via the platform's mapped type), never based on the string shape.
- Custom types are restored via `convertToJSValue`. On the string-cursor path, a type is expected to accept its own `toJSON` output there, same as the serialization round trip already requires for `Date` values.
- The finished condition then runs through `QueryHelper.processWhere` with custom type conversion enabled, so it is processed exactly like a user-provided `where` (nested relations, embeddables and operators all go through the standard pipeline).
- Values compared against JSON documents (object-mode embeddables, `JsonType` properties) keep their serialized form, with custom types restored first so distinct DB/JSON forms convert correctly.
- POJO `after`/`before` values skip the JSON round trip. Entity instances, `Reference`/`ScalarReference` wrappers and embeddable class instances are accepted as cursor values, and an entity or reference can be passed as the cursor argument itself.

Behavior notes:

- Cursor offsets a custom type cannot convert now raise the type's own error. Previously such values were coerced or passed through and silently returned wrong pages.
- Some offsets now bind in their `convertToDatabaseValue` form, e.g. bigint in number mode binds `'1003'` instead of `1003`; the returned rows are identical. On mssql, string offsets now bind as `N'...'` unicode literals like any other where condition.
- `Cursor.decode` is public API and no longer revives ISO-shaped strings into `Date` instances; callers relying on that get the raw JSON values back.
- The cursor wire format is unchanged, existing cursors keep working.

### DIFF
--- a/packages/core/src/drivers/DatabaseDriver.ts
+++ b/packages/core/src/drivers/DatabaseDriver.ts
@@ -41,8 +41,11 @@ import { EntityManager } from '../EntityManager.js';
 import { CursorError, ValidationError } from '../errors.js';
 import { DriverException } from '../exceptions.js';
 import { helper } from '../entity/wrap.js';
+import { Reference } from '../entity/Reference.js';
 import { PolymorphicRef } from '../entity/PolymorphicRef.js';
 import { JsonType } from '../types/JsonType.js';
+import { DateTimeType } from '../types/DateTimeType.js';
+import { QueryHelper } from '../utils/QueryHelper.js';
 import { MikroORM } from '../MikroORM.js';
 
 /** Abstract base class for all database drivers, implementing common driver logic. */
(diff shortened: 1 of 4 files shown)


B4nan:
Now that's a crazy amount of new code for a bug fix (and it's all in a single inflated method), I'll see if I can find a more native solution.

Issue #8090, opened 2026-08-01
applyJoinedFilters with autoJoinRefsForFilters: false references a join alias before it is defined (errorMissingRTE)


jonasnilsberg:
## Bug: `applyJoinedFilters` with `autoJoinRefsForFilters: false` references a join alias before it is defined (`errorMissingRTE`)

When `autoJoinRefsForFilters: false` is set and a query auto-joins an entity whose `@Filter` condition references a relation, the generated SQL references a table alias in a join ON clause **before** that alias is defined in the FROM clause, causing `missing FROM-clause entry for table "l2"` (`errorMissingRTE`).

### Reproduction

https://github.com/jonasnilsberg/mikro-orm-v7-applyjoinedfilters-bug

```bash
pnpm install
createdb mikro_orm_v7_bug_repro
pnpm test
```

### Setup

- `Product` has M:1 → `Company`
- `Company` has a `@Filter` with `{ $or: [{ id: company }, { locations: locations }] }` where `locations` is a O:M relation
- ORM config has `autoJoinRefsForFilters: false`
- Query: `em.find(Product, { company: { code: 'ACME' } }, { filters: { auth: true } })`

### Generated SQL (buggy)

```sql
select "p0".* from "product" as "p0"
inner join "company" as "c1" on "p0"."company_id" = "c1"."id"
  and ("c1"."id" in ('company-1') or "l2"."id" in ('loc-1'))  -- ❌ l2 not yet defined
left join "location" as "l2" on "c1"."id" = "l2"."company_id"  -- defined too late
where "c1"."code" = 'ACME'
```

The `l2` alias (Location, created by the filter's `{ locations: [...] }` condition) is referenced in the Company join ON clause, but the Location join is added **after** the Company join.

### Expected behavior

With `autoJoinRefsForFilters: false`, the filter's `{ locations: [...] }` leg (alias `l2`) should be emitted so that no join ON clause references an alias before it is defined in FROM — either by placing each leg in its own join's ON clause (as the `true` path does) or by ordering the `l2` join before the `c1` join that references it.

Note this is **not** a FROM-ordering difference between the two settings. With `autoJoinRefsForFilters: true` (default), `l2` is *still* emitted after `c1` in the FROM clause — the query works because the `true` path splits the filter's `$or` legs into each join's own ON clause, so the company join never forward-references `l2`:

```sql
inner join "company" as "c1" on "p0"."company_id" = "c1"."id"
[...]

PR #8100, opened 2026-08-03
fix(sql): prevent forward join alias references from earlier joined paths

### LINKED ISSUE #8099: applyJoinedFilters forward-references an alias created by an earlier joined path (errorMissingRTE)
`QueryBuilder.applyJoinedFilters` applies a joined entity's own `@Filter` to its join's `ON` clause. When that filter references a relation whose alias was already created by an **earlier** joined path, the condition is emitted into an `ON` clause that references the alias **before** its join is added → `missing FROM-clause entry for table "l2"` on PostgreSQL.

Sibling of #8090; **not** fixed by #8091 (reproduces with it applied and unapplied). Independent of the `autoJoinRefsForFilters` setting (default, `true`, or `false`) — the joins here come from the filter conditions themselves, not from ref auto-joining.

### Reproduction

https://github.com/jonasnilsberg/mikro-orm-v7-nested-filter-forward-ref

```bash
pnpm install && createdb mikro_orm_v7_nested_filter_repro && pnpm test
```

Three entities, each with a `default: true` `@Filter` named `auth`:
- `Grant` → `User`: `{ user: { location: { company } } }`
- `User` → `Location`: `{ location: { company } }`
- `Location` → `Company`: `{ company }`

Query: `em.findOne(Grant, {}, { fields: ['id'] })`.

### Generated SQL (buggy)

```sql
select "g0"."id" from "grant" as "g0"
inner join "user" as "u1"
  on "g0"."user_id" = "u1"."id" and "l2"."company_id" = 1   -- ❌ l2 not yet defined
inner join "location" as "l2"
  on "u1"."location_id" = "l2"."id" and "l2"."company_id" = 1
where "l2"."company_id" = 1
```

The `Grant` filter builds the `user` (`u1`) and `location` (`l2`) joins. `applyJoinedFilters` then applies `User`'s own filter to the `u1` join, resolving to `l2.company` — and emits it into `u1`'s `ON`, referencing `l2`, which is only joined on the next line.

### Root cause

`applyJoinedFilters` has a fold (added in #8091 / #7681) that nests joins created **during the current join's pass** (`joinCountBefore …`) to avoid exactly this forward reference. But `l2` was created earlier by the `Grant` filter's path, so it's before `joinCountBefore` and the fold skips it.

### Expected

`l2` must be defined before any `ON` references it (order the `location` join first, or route the condition to `WHERE` as the `Grant` filter's equivalent already does).

### Versions

- `@mikro-orm/core` / `@mikro-orm/postgresql`: 7.1.7
- Node.js: 22.x · PostgreSQL: 14+

lazerg:
When a joined entity's `@Filter` resolves to a relation whose alias was already created by an earlier joined path, `applyJoinedFilters` puts the condition in that join's `on` clause even though the referenced alias is joined on a later line, so PostgreSQL reports `missing FROM-clause entry for table "l2"`.

This is the sibling of #8090. The fold added in #8091 only nests joins created during the current pass, so an alias created earlier is skipped. Nest those referenced joins into the filtered join as well, so they render inside its parenthesized group and share the scope of the outer `on` clause.

Closes #8099

### DIFF
--- a/packages/sql/src/query/QueryBuilder.ts
+++ b/packages/sql/src/query/QueryBuilder.ts
@@ -1530,6 +1530,8 @@ export class QueryBuilder<
           join.cond = { ...cond };
         }
 
+        this.nestReferencedJoins(join);
+
         // For polymorphic LEFT JOIN filters, add a WHERE condition to enforce the filter
         // only for rows matching this target's discriminator value. This ensures rows pointing
         // to other polymorphic targets are not excluded.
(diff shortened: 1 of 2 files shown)


B4nan:
superseded by #8101

PR #8119, opened 2026-08-06
fix(entity-generator): keep index column order when columns map to relations

### LINKED ISSUE #8116: @mikro-orm/entity-generator: Index column order is lost when columns are mapped to relation properties
### Describe the bug

Hello,

When generating entities from an existing database, the column order of indexes is not preserved if those columns are part of composite foreign keys mapped to relation properties.

**The Root Cause**
In `packages/sql/src/schema/DatabaseTable.ts`, the `getIndexProperties()` method collects property names into a `Set` and then returns them based on insertion order:

```ts
// packages/sql/src/schema/DatabaseTable.ts
const propBaseNames = new Set<string>();
…
return Array.from(propBaseNames).map(…);   // ~Line 764

```

When physical columns are collapsed into relation properties, the resulting array reflects the order in which the properties were first encountered, not the actual index column order. If a table shares a column across multiple composite foreign keys, several relations might be added to the Set at once. When those columns are encountered again later in the index definition, they are already in the Set and their positions are not updated to match the index.

For example, on our schema, this results in the following mismatch:

```text
uq_role_grants_anchored
  Database actual: (tenant_id, user_id, role_id, org_node_id, coverage)
  Generated order: (org_node_id, role_id, user_id, tenant_id, coverage)

```

*(Another index, `uq_org_nodes_tenant_parent_name`, is reordered in a similar way).*

**Impact**
While this reordering does not break uniqueness constraints, it has two significant side effects:

1. **Performance:** Leading-column selectivity is altered, which dictates whether the query planner can actually use the index efficiently.
2. **Migrations:** Any schema-diff tooling comparing the database `indexdef` against the generated metadata will flag a difference, which blocks the generator's output from being used as a reliable migration source.

### Reproduction

You can reproduce this by creating a table with two composite foreign keys that share their first column, alongside a unique index where the column order differs from the order in which the relations are declared.

```sql
-- Minimal reproduction schema shape:
CREATE UNIQUE INDEX uq_probe_anchored
  ON role_grants (tenant_id, user_id, role_id, org_node_id, coverage)
  WHERE org_node_id IS NOT NULL;

```

**Expected:** The generated entity lists the index properties in the exact order defined in the database (`tenant_id, user_id, role_id, org_node_id, coverage`).
**Actual:** The relation properties are emitted in the order they were first collected by the `Set`.

@mikro-orm/postgresql

### MikroORM version

7.1.10 (@mikro-orm/core, @mikro-orm/postgresql, @mikro-orm/sql)

### Node.js version

v24.18.0, TypeScript 6.0.3

### Operating system

macOS 27.0 (arm64)

### Validations

lazerg:
`getIndexProperties()` collected property names of a composite index into a `Set`, so a property covering several columns landed at the position of the first index column it touches. With `scalarPropertiesForRelations: 'always'`, all foreign keys of the reported table start with the shared `tenant_id`, so every one of them got collected while processing the leading column and the result came out in foreign key discovery order instead of the index order. A unique index on `(tenant_id, user_id, role_id, org_node_id, coverage)` was generated as `['orgNodeId', 'roleId', 'tenantId', 'userId', 'coverage']`.

The collected names now remember the first and last index position they were seen at, and ties on the first position are broken by the last one, so the generated order follows the columns again.

Fixes #8116

### DIFF
--- a/packages/sql/src/schema/DatabaseTable.ts
+++ b/packages/sql/src/schema/DatabaseTable.ts
@@ -706,13 +706,24 @@ export class DatabaseTable {
     fksOnStandaloneProps: Map<string, { fkIndex?: IndexDef; currentFk: ForeignKey }>,
     namingStrategy: NamingStrategy,
   ) {
-    const propBaseNames = new Set<string>();
+    const propBaseNames = new Map<string, { first: number; last: number }>();
     const columnNames = index.columnNames;
     const l = columnNames.length;
 
     if (columnNames.some(col => !col)) {
       return;
     }
+
+    const addPropBaseName = (baseName: string, position: number) => {
+      const positions = propBaseNames.get(baseName);
+
+      if (positions) {
+        positions.last = position;
+        return;
+      }
+
+      propBaseNames.set(baseName, { first: position, last: position });
+    };
 
     for (let i = 0; i < l; ++i) {
       const columnName = columnNames[i];
@@ -725,7 +736,7 @@ export class DatabaseTable {
         }
         // It has a prop named after it.
         // Add it and move on.
-        propBaseNames.add(columnName);
+        addPropBaseName(columnName, i);
         continue;
       }
 
@@ -733,7 +744,7 @@ export class DatabaseTable {
       // include this prop and move on.
       const columnPropFk = fksOnColumnProps.get(columnName);
       if (columnPropFk && !columnPropFk.columnNames.some(fkColumnName => !columnNames.includes(fkColumnName))) {
-        propBaseNames.add(columnName);
+        addPropBaseName(columnName, i);
         continue;
       }
 
@@ -747,7 +758,7 @@ export class DatabaseTable {
         }
 
         if (!fk.columnNames.some(fkColumnName => !columnNames.includes(fkColumnName))) {
-          propBaseNames.add(propName);
+          addPropBaseName(propName, i);
           propAdded = true;
         }
       }
@@ -761,9 +772,10 @@ export class DatabaseTable {
       return;
     }
 
-    return Array.from(propBaseNames).map(baseName =>
-      this.getPropertyName(namingStrategy, baseName, fksOnColumnProps.get(baseName)),
-    );
+    // Props sharing their first column would otherwise follow FK discovery order, so break ties on the last one.
+    return Array.from(propBaseNames)
+      .sort(([, a], [, b]) => a.first - b.first || a.last - b.last)
+      .map(([baseName]) => this.getPropertyName(namingStrategy, baseName, fksOnColumnProps.get(baseName)));
   }
 
   private getSafeBaseNameForFkProp(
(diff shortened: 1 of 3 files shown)


packages/sql/src/schema/DatabaseTable.ts
```diff
@@ -761,9 +772,10 @@ export class DatabaseTable {
       return;
     }
 
-    return Array.from(propBaseNames).map(baseName =>
-      this.getPropertyName(namingStrategy, baseName, fksOnColumnProps.get(baseName)),
-    );
+    // Props sharing their first column would otherwise follow FK discovery order, so break ties on the last one.
+    return Array.from(propBaseNames)
+      .sort(([, a], [, b]) => a.first - b.first || a.last - b.last)
```
boenrobot, line 777: I'm not sure I understand this...

OK, a Set is ~unordered~ ordered by insertion into the Set rather than index prop order, so you may get the props in a wrong order. I get that part. A Map with the prop as a key fixes this.

But why store first AND last occurrence in the Map and sort by both? Why not store just the ~first (a single number) by not adding to the map if present~ last occurrence, sort by that, and output the keys in that order?

lazerg: The `first` values tie, so sorting on them alone is not enough.

The loop over `fksOnStandaloneProps` runs inside a single iteration of `i`, and it can add more than one new entry in that one pass, because one column can belong to several standalone FKs whose columns are all covered by the index.

The test in this PR hits exactly that. The index is `(tenant_id, user_id, role_id, org_node_id, coverage)`, and `tenant_id` belongs to all four FKs: `fk_tenant`, `fk_user`, `fk_role`, `fk_org`. At `i = 0` all four props are added, so all four get `first = 0`. Sorting on `first` alone then leaves them in Map insertion order, which is FK discovery order from the database, alphabetical by constraint name in this case. I checked by dropping `|| a.last - b.last` and rerunning the test: the output becomes `['orgNodeId', 'roleId', 'tenantId', 'userId', 'coverage']` instead of `['tenantId', 'userId', 'roleId', 'orgNodeId', 'coverage']`.

`last` is what separates them. `fk_tenant` ends at position 0, `fk_user` at 1, `fk_role` at 2, `fk_org` at 3, so the props come out in the order the index actually covers their columns.

boenrobot: Yeah, I checked that too, and you're right... but then leaving only `a.last - b.last` and removing `a.first - b.first` works too, which would in turn mean that a simple `propBaseNames.set(propName, i)` instead of `addPropBaseName(propName, i)` (plus sorting by the values) should also work... Or is there a different case that is just not covered by existing tests (including this new one)?

lazerg: Sorting by `last` alone breaks when one relation spans the columns of another. Index `(p_a, q_a, q_b, filler, p_b)`, FKs `(p_a, p_b)` and `(q_a, q_b)`:

- by `first`: `['p1', 'p2', 'filler']`
- by `last`: `['p2', 'filler', 'p1']`

Master already emits the first one, so `first` keeps current output and `last` only breaks ties. Added that case as a test in 4d873de.

PR #8127, opened 2026-08-06
fix(postgres): emit the index access method when creating indexes

### LINKED ISSUE #8124: @mikro-orm/sql: PostgreSQL index introspection drops the access method (GIST, GIN, etc.), regenerating them as standard B-trees
### Describe the bug

Hello,

When introspecting a PostgreSQL database, the index access method (e.g., `USING gist`, `USING gin`, `USING brin`) is silently dropped unless the index contains functional columns. When the ORM recreates the index, it defaults back to a standard B-tree.

**The Root Cause**
In `packages/sql/src/dialects/postgresql/PostgreSqlSchemaHelper.ts`, `getAllIndexes()` reads the full `pg_get_indexdef()` string. However, it only retains that raw expression if the index contains functional columns:

```ts
// packages/sql/src/dialects/postgresql/PostgreSqlSchemaHelper.ts
const hasFunctionalColumns = index.index_def.some(col => PostgreSqlSchemaHelper.FUNCTIONAL_COL_RE.exec(col))
...
if (hasFunctionalColumns) {
  // Functional-column expression can't be diffed structurally — keep the whole CREATE
  indexDef.expression = index.expression
}

```

For an index over plain columns, the resulting `IndexDef` only retains `columnNames`, `unique`, and `where`. The access method is discarded. Consequently, `getCreateIndexSQL()` emits the default B-tree syntax:

```sql
-- What is actually in the database:
CREATE INDEX idx_org_nodes_path_gist ON public.org_nodes USING gist (path)

-- What is read back and re-emitted by the generator:
CREATE INDEX "idx_org_nodes_path_gist" ON "org_nodes" ("path")

```

**Impact**
This is a silent failure because PostgreSQL happily accepts the re-emitted DDL. It builds a B-tree over the column (e.g., an `ltree` column). However, a B-tree cannot answer the `<@` ancestry queries that the GIST index was explicitly created for, causing every subtree query to silently degrade into a sequential scan.

The exact same regression applies to GIN, BRIN, Hash, and SP-GiST indexes over plain columns. Any workflow that uses `schema:update` or `migration:create` against a database holding these indexes will inadvertently drop and recreate them as B-trees.

**Expected behaviour**

An introspected index should carry enough metadata to be recreated accurately. The ORM should either record the access method on `IndexDef` and emit `USING <am>`, or fall back to keeping the entire `pg_get_indexdef()` string (the same way functional indexes currently do).

**Suggested fix**

The least intrusive fix might be to treat a non-default access method the same way a functional column is treated, preserving the raw expression:

```diff
-const hasFunctionalColumns = index.index_def.some(col => PostgreSqlSchemaHelper.FUNCTIONAL_COL_RE.exec(col))
+const accessMethod = /\busing\s+(\w+)/i.exec(index.expression ?? '')
+const hasFunctionalColumns =
+  index.index_def.some(col => PostgreSqlSchemaHelper.FUNCTIONAL_COL_RE.exec(col)) ||
+  (!!accessMethod && accessMethod[1].toLowerCase() !== 'btree')

```

### Reproduction

You can reproduce this by creating a table with an `ltree` column and a GIST index:

```sql
CREATE EXTENSION IF NOT EXISTS ltree;
CREATE TABLE t (path ltree NOT NULL);
CREATE INDEX idx_t_path ON t USING gist (path);

```

Then run the following introspection:

```ts
const schema = await DatabaseSchema.create(orm.em.getConnection(), orm.em.getPlatform(), orm.config)
console.log(schema.getTable('t')!.getIndexes())

```

**Expected:** The index definition retains `USING gist`.
**Actual:** The access method is absent, and the generated SQL will recreate it as a standard B-tree.

@mikro-orm/postgresql

### MikroORM version

7.1.11

### Node.js version

v24.18.0, TypeScript 6.0.3

### Operating system

macOS 27.0 (arm64)

### Validations

lazerg:
Introspection already records the access method of a PostgreSQL index on `IndexDef.type`, but `getCreateIndexSQL()` never emitted it, so every index came back out as a B-tree. An entity declaring `@Index({ properties: ['payload'], type: 'gin' })` generated `create index "..." on "..." ("payload")`, and the same applies to the `type` that the entity generator writes out for an introspected gist, gin or brin index. PostgreSQL accepts the DDL without complaint, which is what makes it easy to miss: the index exists, it just cannot answer the queries it was created for.

`getCreateIndexSQL()` now asks a `getIndexTypeClause()` hook for the access method and places it after the table name. The hook returns an empty string everywhere except PostgreSQL, where it emits `using <type>`. Full-text indexes keep going through `getFullTextIndexExpression()`, which already spells out `using gin`.

Closes #8124

### DIFF
--- a/packages/sql/src/dialects/postgresql/PostgreSqlSchemaHelper.ts
+++ b/packages/sql/src/dialects/postgresql/PostgreSqlSchemaHelper.ts
@@ -1557,6 +1557,18 @@ export class PostgreSqlSchemaHelper extends SchemaHelper {
     return super.getCreateIndexSuffix(index);
   }
 
+  /**
+   * PostgreSQL index access method, e.g. `gist`, `gin` or `brin`. Full-text indexes are emitted
+   * via `getFullTextIndexExpression()`, which already spells out the access method.
+   */
+  protected override getIndexTypeClause(index: IndexDef): string {
+    if (typeof index.type === 'string' && index.type !== 'fulltext') {
+      return ` using ${index.type}`;
+    }
+
+    return super.getIndexTypeClause(index);
+  }
+
   private getIndexesSQL(tables: Table[]): string {
     return `select indrelid::regclass as table_name, ns.nspname as schema_name, relname as constraint_name, idx.indisunique as unique, idx.indisprimary as primary, contype, condeferrable, condeferred,
       array(
--- a/packages/sql/src/schema/SchemaHelper.ts
+++ b/packages/sql/src/schema/SchemaHelper.ts
@@ -255,7 +255,8 @@ export abstract class SchemaHelper {
     tableName = this.quote(tableName);
     const keyName = this.quote(index.keyName);
     const defer = index.deferMode ? ` deferrable initially ${index.deferMode}` : '';
-    let sql = `create ${index.unique ? 'unique ' : ''}index ${keyName} on ${tableName}`;
+    const type = this.getIndexTypeClause(index);
+    let sql = `create ${index.unique ? 'unique ' : ''}index ${keyName} on ${tableName}${type}`;
 
     if (index.unique && index.constraint) {
       sql = `alter table ${tableName} add constraint ${keyName} unique`;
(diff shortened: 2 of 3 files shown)


lazerg:
#8126 landed the same change while this was still in CI, and it covers the comparator side too. Closing this one.

PR #8132, opened 2026-08-08
fix(sql): compare quoted decimal defaults numerically

### LINKED ISSUE #8131: Schema never up to date when a decimal property has a string default containing a decimal point
### Describe the bug

`getUpdateSchemaSQL()` emits an `alter column ... set default` immediately after the schema was
created from that same metadata, and keeps emitting it on every subsequent run.

Reproduction: <REPRO LINK>

```ts
const Product = defineEntity({
  name: 'Product',
  properties: p => ({
    id: p.integer().primary(),
    price: p.decimal().precision(12).scale(2).default('0.00'),
  }),
});

await orm.schema.refresh();
await orm.schema.getUpdateSchemaSQL();
// expected: ''
// actual:   alter table "product" alter column "price" set default '0.00';
```

`getCreateSchemaSQL()` for the same entity produces:

```sql
create table "product" ("id" serial primary key, "price" numeric(12,2) not null default '0.00');
```

so the comparator rejects the DDL it just generated. Applying the `alter` does not help; the next
diff is identical.

## Cause

`MetadataDiscovery.initDefaultValue` quotes string defaults, so metadata holds `'0.00'`. Postgres
reports the column default as `0.00`. In `SchemaComparator.hasSameDefaultValue`:

```ts
if (to.mappedType instanceof DecimalType && Number.isFinite(+from.default) && Number.isFinite(+to.default)) {
  return this.platform.formatDecimal(from.default, to.scale) === this.platform.formatDecimal(to.default, to.scale);
}
```

`+"'0.00'"` is `NaN`, so the branch is skipped and the fallthrough compares `0.00` against `'0.00'`
as strings.

The operand that fails the guard is `to.default` (metadata), not `from`. `mappedType` is
`DecimalType`, so the branch is reached rather than bypassed by a type mismatch.

## Related cases

Same entity, four properties, all `numeric(12,2)`:

| default | metadata | postgres | |
| --- | --- | --- | --- |
| `'0.00'` | `'0.00'` | `0.00` | mismatch |
| `'0'` | `'0'` | `'0'::numeric` | ok, `normalizeDefaultValue` strips the cast |
| `0` | `0` | `0` | ok, never quoted |
| `defaultRaw: '0.00'` | `0.00` | `0.00` | ok, skips `initDefaultValue` |

\#7797 is the numeric-literal case, which the `DecimalType` branch handles. This one does not reach
it.

### Reproduction

https://github.com/yayza/mikroorm-reproduction

@mikro-orm/postgresql

### MikroORM version

7.1.12-dev.4

### Node.js version

Node 24

### Operating system

Windows / WSL

### Validations

lazerg:
`SchemaComparator.hasSameDefaultValue` compares decimal defaults numerically, but metadata keeps string defaults quoted, so `+"'0.00'"` is `NaN` and that branch is skipped. The comparison falls back to strings, where the `0.00` postgres reports never matches the quoted `'0.00'`, and `getUpdateSchemaSQL()` keeps emitting the same `alter column ... set default` for a schema it just created from that metadata.

Both values are now unquoted before the numeric check, so the quoted literal compares equal to what the database reports.

The snapshot serialization in `DatabaseTable.toJSON` had the same finite-number guard, so `migration:create` (metadata side) and `migration:up` (introspection side) kept rewriting the snapshot file with `'0.00'` vs `0.00`. That spot now strips the quotes too before collapsing to the canonical numeric form.

Closes #8131

### DIFF
--- a/packages/sql/src/schema/DatabaseTable.ts
+++ b/packages/sql/src/schema/DatabaseTable.ts
@@ -1401,8 +1401,13 @@ export class DatabaseTable {
       // mysql stores decimal defaults padded to scale (`0` → `0.00`); collapse to canonical numeric form
       // so the metadata-side (`0`) and introspection-side (`0.00`) snapshots agree
       let defaultValue: string | null = c.default ?? null;
-      if (defaultValue != null && c.mappedType instanceof DecimalType && Number.isFinite(+defaultValue)) {
-        defaultValue = this.#platform.formatDecimal(defaultValue, c.scale).toString();
+      if (defaultValue != null && c.mappedType instanceof DecimalType) {
+        // string defaults like `default: '0.00'` are quoted in metadata, so strip the quotes first
+        const unquoted = defaultValue.replace(/^'(.*)'$/, '$1');
+
+        if (Number.isFinite(+unquoted)) {
+          defaultValue = this.#platform.formatDecimal(unquoted, c.scale).toString();
+        }
       }
       const normalized: Dictionary = {
         name: c.name,
--- a/packages/sql/src/schema/SchemaComparator.ts
+++ b/packages/sql/src/schema/SchemaComparator.ts
@@ -1259,6 +1259,12 @@ export class SchemaComparator {
     return parseJsonSafe(val);
   }
 
+  private parseDecimalDefault(defaultValue: string | number): number | null {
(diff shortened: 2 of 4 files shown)


B4nan (commented):
thanks!

PR #8144, opened 2026-08-14
fix(core): do not mutate the options object provided by the caller

### LINKED ISSUE #8143: EntityManager find and stream methods mutate caller-provided options, allowing query state to leak across calls
EntityManager `find*` and `stream` methods mutate the options object supplied by the caller. In particular, they add a normalized `populate` value and EntityManager-specific context to the original object.

This becomes dangerous when an options object is intentionally reused as a base configuration. A query without population can add `populate: []` to that object, affecting later queries and potentially leaking state across requests in long-running applications.

The relevant flow in `EntityManager.findOne()` is approximately:

```ts
em.prepareOptions(options);

// ...

options.populate = await em.preparePopulate(entityName, options);

// The object is cloned only after the mutations above.
options = { ...options };
```

`prepareOptions()` also assigns values such as `schema`, `signal`, logging context, and abort strategy directly onto the provided object.

### Expected behavior

Query options should be treated as caller-owned input and should remain unchanged after the query:

```ts
const options = { filters: false };
const before = structuredClone(options);

await em.findOne(User, { id: userId }, options);

expect(options).toEqual(before);
```

If mutation is an intentional API contract, it should be explicitly documented. However, immutable handling would be much safer and would match the usual expectation for an options argument.

### Real-world impact

We used a module-level object for a reusable filter scope:

```ts
const BASE_OPTIONS = { filters: false } as const;
```

One unpopulated query passed it directly:

```ts
await repository.findOne(where, BASE_OPTIONS);
```

That query caused MikroORM to persist `populate: []` on `BASE_OPTIONS`.

A later query composed its options from that base:

```ts
await repository.findOne(where, {
  populate: ['organization', 'connection'],
  ...BASE_OPTIONS,
});
```

After the first query, the spread unexpectedly overwrote the requested relationships:

```ts
{
  populate: [],
  filters: false,
}
```

The resulting behavior persisted across HTTP requests because the mutated object lived at module scope. Restarting the application temporarily resolved it by recreating the object.

### Suggested fix

Clone options at the public method boundary before performing any normalization:

```ts
async findOne(entityName, where, options = {}) {
  options = { ...options };  // or a nested clone

  em.prepareOptions(options);
  // ...
}
```

Any nested arrays or objects that are modified during normalization should also be cloned, or the preparation methods could return new options instead of modifying their input.

### Reproduction

I have not had a chance to create a reproduction repo, but I think this is clear enough

```ts
const options = { filters: false } as const;

console.log(options);
// { filters: false }

await em.findOne(User, { id: userId }, options);

console.log(options);
// Approximately:
// {
//   filters: false,
//   populate: [],
//   logging: { id: ... },
//   loggerContext: { id: ... },
// }
```

A stricter reproduction is to freeze the input:

```ts
const options = Object.freeze({ filters: false });

await em.findOne(User, { id: userId }, options);
```

This should throw when MikroORM attempts to assign fields such as `options.populate`.

Ensure the entity is not already resolved from the identity map when reproducing, for example by calling `em.clear()` before `findOne()`.

_No response_

### MikroORM version

7.1.11

### Node.js version

v24.16.0

### Operating system

MacOS

### Validations

lazerg:
`prepareOptions()` assigned `schema`, `signal`, the abort strategy and the logger context straight onto the options object it received, and `em.find()`, `em.findOne()` and `em.stream()` then wrote the normalized `populate` hint onto it too. All of that happens before the local `options = { ...options }` copy, so the object the caller owns comes back changed. A module level base config like `const BASE_OPTIONS = { filters: false }` picks up `populate: []` from the first query that runs without a populate hint, and every later query spreading that base loses its own hint. Nothing resets it, so in a long running app the leak survives across requests.

`prepareOptions()` now returns a copy instead of writing into its argument, and every caller reassigns the result, which covers `upsert()`, `insert()`, `nativeUpdate()`, `nativeDelete()`, `lock()` and `em.populate()` as well. `findAndCount()` and `findByCursor()` set `flushMode` and `overfetch` before that call, so they copy first too. The manual copies in `em.count()` and in both `countBy()` implementations are no longer needed.

Fixes #8143

### DIFF
--- a/packages/core/src/EntityManager.ts
+++ b/packages/core/src/EntityManager.ts
@@ -233,7 +233,7 @@ export class EntityManager<Driver extends IDatabaseDriver = IDatabaseDriver> {
     }
 
     const em = this.getContext();
-    em.prepareOptions(options);
+    options = em.prepareOptions(options);
     const meta = this.metadata.get<Entity>(entityName);
     em.validateIndexUsage(meta, where, options);
     await em.tryFlush(entityName, options);
@@ -346,7 +346,7 @@ export class EntityManager<Driver extends IDatabaseDriver = IDatabaseDriver> {
     > = {} as any,
   ): AsyncIterableIterator<Loaded<Entity, Hint, Fields, Excludes>> {
     const em = this.getContext();
-    em.prepareOptions(options);
+    options = em.prepareOptions(options);
     (options as Dictionary).strategy = 'joined';
     await em.tryFlush(entityName, options);
     const where = (await em.processWhere(entityName, options.where ?? {}, options, 'read')) as FilterQuery<Entity>;
@@ -840,7 +840,7 @@ export class EntityManager<Driver extends IDatabaseDriver = IDatabaseDriver> {
   ): Promise<[Loaded<Entity, Hint, Fields, Excludes>[], number]> {
     const em = this.getContext(false);
     await em.tryFlush(entityName, options);
-    options.flushMode = 'commit'; // do not try to auto flush again
+    options = { ...options, flushMode: 'commit' }; // do not try to auto flush again
 
     return Promise.all([
       em.find(entityName, where as any, options),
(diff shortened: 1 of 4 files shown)


lazerg:
The node 24 3/4 job stopped in its Init docker step with `mssql did not stabilise within the startup window`, so no test ran on that shard. The missing coverage upload from it is also what puts codecov/project at -2.42%, while codecov/patch reports 100% of the diff hit. The same shard passed on node 22 and node 26, so a rerun should clear both.

B4nan (approved):

PR #8198, opened 2026-08-22
fix(postgres): strip array-typed casts cleanly when recovering check constraints

### LINKED ISSUE #8197: `@mikro-orm/sql`: PostgreSQL check constraint recovery fails on array-typed casts, producing invalid SQL (`ARRAY[...][]`)
### Describe the bug

Hello,

When introspecting PostgreSQL check constraints that involve array-typed casts (such as `::text[]`), the cast-stripping regex fails to match the array bracket suffix. This leaves behind an orphaned `[]` in the recovered expression, resulting in invalid SQL that fails with a syntax error when re-emitted as DDL.

**The Root Cause**
In `packages/sql/src/dialects/postgresql/PostgreSqlSchemaHelper.ts`, `getAllChecks()` uses a regex to strip PostgreSQL-added `(expr)::type` casts back to `expr`:

```ts
// packages/sql/src/dialects/postgresql/PostgreSqlSchemaHelper.ts
const def = m ? m[1].replace(/\(([^()]*)\)::\w+/g, '$1') : single ? single[1] : check.expression;

```

Because `\w+` does not match the trailing `[]` of an array type name (such as `text[]`), the replacement consumes `(ARRAY[...])::text` but leaves the `[]` intact.

For example, PostgreSQL stores an `IN` list check on a `varchar` column as:

```sql
CHECK (((mode IS NULL) OR ((mode)::text = ANY ((ARRAY['a'::character varying, 'b'::character varying])::text[]))))

```

During introspection, the expression is stripped down to:

```sql
(mode IS NULL) OR (mode = ANY (ARRAY['a'::character varying, 'b'::character varying][]))

```

**Why This Is Easily Missed**
A simple, standalone `IN` list check on a column is usually intercepted by `getEnumDefinitions()`, which rebuilds the check expression from parsed enum values and discards the string recovery result. However, if the check contains any additional logic (such as an `IS NULL` disjunct or multi-term predicates), it bypasses enum conversion and surfaces this broken expression.

**Impact**
Any workflow that reads existing check constraints and writes them back (such as `schema:update` or `migration:create`) will produce invalid DDL that fails with `42601 syntax error at or near "["` at execution time.

**Suggested Fix**
Allow the regex to optionally match the array bracket suffix `(?:\[\])?`:

```diff
-const def = m ? m[1].replace(/\(([^()]*)\)::\w+/g, '$1') : single ? single[1] : check.expression;
+const def = m ? m[1].replace(/\(([^()]*)\)::\w+(?:\[\])?/g, '$1') : single ? single[1] : check.expression;

```

I would be glad to open a PR for this containing the fix and a regression test under `tests/features/schema-generator/check-constraint.postgres.test.ts`.

### Reproduction

Create a table with a check constraint containing a disjunct:

```sql
CREATE TABLE t (
  mode varchar(16), 
  CONSTRAINT chk_mode CHECK (mode IS NULL OR mode IN ('a', 'b'))
);

```

Introspect the table:

```ts
const schema = await DatabaseSchema.create(orm.em.getConnection(), orm.em.getPlatform(), orm.config);
console.log(schema.getTable('t')!.getChecks().find(c => c.name === 'chk_mode')!.expression);

```

**Expected:** `(mode IS NULL) OR (mode = ANY (ARRAY['a'::character varying, 'b'::character varying]))`

**Actual:** `(mode IS NULL) OR (mode = ANY (ARRAY['a'::character varying, 'b'::character varying][]))` *(Note the trailing `[]`)*

Executing DDL with the recovered expression fails with `42601 syntax error at or near "["`.

@mikro-orm/postgresql

### MikroORM version

7.1.13

### Node.js version

v24.18.0, TypeScript 7.0.2

### Operating system

macOS 27.0 (arm64)

### Validations

hprogq:
Postgres stores an `IN` list on a varchar column with the whole array cast, e.g. `(ARRAY['a', 'b'])::text[]`. The cast strip in `getAllChecks()` matched the type with `::\w+`, which cannot cover an array type name, so it consumed `(ARRAY[...])::text` and left the trailing `[]` orphaned. The recovered expression is no longer valid SQL, and re-adding it as DDL fails with a syntax error:

```
recovered: (mode IS NULL) OR (mode = ANY (ARRAY['a'::character varying, 'b'::character varying][]))
expected:  (mode IS NULL) OR (mode = ANY (ARRAY['a'::character varying, 'b'::character varying]))
```

Bare `IN`-list checks never surface this: the enum-as-check detection rebuilds their expression from the parsed items. It bites on any check not shaped like a pure enum membership, e.g. one with an `is null` disjunct.

The regression test sits next to the existing check-recovery tests and asserts both the recovered expression and that re-adding the constraint from the recovered text executes.

Closes #8197

### DIFF
--- a/packages/sql/src/dialects/postgresql/PostgreSqlSchemaHelper.ts
+++ b/packages/sql/src/dialects/postgresql/PostgreSqlSchemaHelper.ts
@@ -699,7 +699,7 @@ export class PostgreSqlSchemaHelper extends SchemaHelper {
       // SchemaHelper.createCheck).
       const m = /^check \(\((.*)\)\)$/is.exec(check.expression);
       const single = m ? null : /^check \((.*)\)$/is.exec(check.expression);
-      const def = m ? m[1].replace(/\(([^()]*)\)::\w+/g, '$1') : single ? single[1] : check.expression;
+      const def = m ? m[1].replace(/\(([^()]*)\)::\w+(?:\[\])?/g, '$1') : single ? single[1] : check.expression;
       ret[key].push({
         name: check.name,
         columnName: check.column_name,
--- a/tests/features/schema-generator/check-constraint.postgres.test.ts
+++ b/tests/features/schema-generator/check-constraint.postgres.test.ts
@@ -452,4 +452,48 @@ describe('check constraint [postgres]', () => {
     await orm.schema.dropDatabase();
     await orm.close();
   });
+
+  test('array-typed cast in a non-enum check keeps brackets balanced on introspection [postgres]', async () => {
+    const orm = await initORMPostgreSql();
+    const meta = orm.getMetadata();
+    await orm.schema.update();
+
+    // pg stores `mode in ('a', 'b')` on a varchar column as
+    // `CHECK (((mode IS NULL) OR ((mode)::text = ANY ((ARRAY['a'::character varying, 'b'::character varying])::text[]))))`
+    // — the array literal is cast as a whole, so the cast strip must consume the array-type suffix
+    // too, otherwise the recovered expression keeps an orphaned `[]` and re-emitting it as DDL is a
+    // syntax error. The `is null` disjunct keeps the check out of the enum-as-check detection, which
+    // would otherwise rebuild the expression from the parsed items and mask the recovery.
+    const newTableMeta = new EntitySchema({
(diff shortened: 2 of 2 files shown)


B4nan (approved):
thanks

Issue #7677, opened 2026-05-07
Discriminator/narrowing issue with embedded inheritance: type check on union does not narrow type for embedded discriminated union


motopods:
### Describe the bug

When using embedded inheritance and discriminated unions for embeddables (e.g. a pet property being Cat | Dog with discriminatorColumn 'type'), TypeScript type narrowing does not occur after a conditional check on the discriminator (e.g. if (pet.type === AnimalType.CAT) { pet.canMeow }). This results in a type error: "Property 'canMeow' does not exist on type 'Cat | Dog'." Expected: The check on `type` should properly narrow the type for downstream usage, like for stand-alone entities. See sample code and error below:

It appears that checking the discriminator field fails to trigger TypeScript's control flow analysis for type narrowing with embeddables. Is this a bug or design limitation? This type narrowing works with entity inheritance but not for embeddables/unions. Please advise if workaround is possible, or if MikroORM could emit additional type hints for proper discrimination.

### Reproduction

```ts
import { MikroORM } from '@mikro-orm/sqlite';
import { defineEntity, p } from '@mikro-orm/core';

export enum AnimalType {
  CAT,
  DOG,
}

const AnimalSchema = defineEntity({
  name: 'Animal',
  embeddable: true,
  abstract: true,
  discriminatorColumn: 'type',
  properties: {
    type: p.enum(() => AnimalType),
    name: p.string(),
  },
});

export class Animal extends AnimalSchema.class {}
AnimalSchema.setClass(Animal);

const CatSchema = defineEntity({
  name: 'Cat',
  embeddable: true,
  extends: Animal,
  discriminatorValue: AnimalType.CAT,
  properties: {
    canMeow: p.boolean().nullable(),
  },
});

export class Cat extends CatSchema.class {}
CatSchema.setClass(Cat);

const DogSchema = defineEntity({
  name: 'Dog',
  embeddable: true,
  extends: Animal,
  discriminatorValue: AnimalType.DOG,
  properties: {
    canBark: p.boolean().nullable(),
  },
});

export class Dog extends DogSchema.class {}
DogSchema.setClass(Dog);

const OwnerSchema = defineEntity({
  name: 'Owner',
  properties: {
    id: p.integer().primary(),
    name: p.string(),
    pet: () => p.embedded([Cat, Dog]),
  },
});

export class Owner extends OwnerSchema.class {}
OwnerSchema.setClass(Owner);

let orm: MikroORM;

beforeAll(async () => {
[...]

PR #8201, opened 2026-08-25
fix(core): don't require the discriminator value when creating an STI entity

### LINKED ISSUE #8200: Type requires discriminator value to be set when creating STI entity
### Describe the bug

Let's say I have an STI entity set up like this:

```
const BasePerson = defineEntity({
  name: "BasePerson",
  abstract: true,
  discriminatorColumn: "type",
  properties: {
    id: p.integer().primary(),
    type: p.text(),
  },
});

const Person = defineEntity({
  name: "Person",
  extends: BasePerson,
  discriminatorValue: "person",
  properties: {
    name: p.text(),
  },
});
```

If I `orm.em.create` a Person, the value of `type` should be `person`. MikroORM seems to understand this, because this code runs and flushes successfully. But the types don't understand this, and demand that you manually specify type:

Image

### Reproduction

Runnable version of what I described above: https://github.com/nkalupahana/mikroorm-reproductions/tree/sti-type

@mikro-orm/sqlite

### MikroORM version

7.1.13

### Node.js version

v22.22.3

### Operating system

26.5.2 (25F84)

### Validations

lazerg:
Creating an STI entity defined via `defineEntity()` required passing the discriminator value explicitly, even though the ORM already sets it from the schema on insert. The inherited discriminator property was narrowed to the child's `discriminatorValue` but stayed required, so `em.create()` rejected data without it.

It is now declared via `[OptionalProps]` instead. Branding it with `Opt` would have been shorter, but it breaks the discriminated union added in #7677, as an intersection type no longer works as a discriminant. `em.create()` also assigns the value to the entity itself now, which used to stay undefined until the entity was reloaded.

Fixes #8200

### DIFF
--- a/packages/core/src/entity/EntityFactory.ts
+++ b/packages/core/src/entity/EntityFactory.ts
@@ -149,6 +149,15 @@ export class EntityFactory {
     }
 
     data = { ...data };
+
+    if (options.newEntity && meta2.root.inheritanceType === 'sti' && meta2.discriminatorValue != null) {
+      const prop = meta2.properties[meta2.root.discriminatorColumn as EntityKey<T>];
+
+      if (prop && prop.userDefined !== false) {
+        data[prop.name] ??= meta2.discriminatorValue as EntityValue<T>;
+      }
+    }
+
     const entity = exists ?? this.createEntity<T>(data, meta2, options);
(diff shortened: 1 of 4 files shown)


B4nan (approved):
thanks

PR #8202, opened 2026-08-25
fix(sql): keep row-value tuples in `$in` on a `raw()` key

lamkashingpaul:
## Problem

A `raw()` key can hold a column tuple, and `$in` with an array of tuples rendered a real row-value comparison in v6, where knex did the rendering. The native query builder in v7 flattens it one level:

```ts
qb.where({
  [raw('(`tool`.`class_id`, `tool`.`role_id`)')]: { $in: [[-1, -1], [1, 1], [2, 2]] },
});
```

```sql
-- v6
where (`tool`.`class_id`, `tool`.`role_id`) in ( values (-1, -1), (1, 1), (2, 2))
-- v7
where (`tool`.`class_id`, `tool`.`role_id`) in (-1, -1, 1, 1, 2, 2)
```

Not cosmetic — the driver rejects the flattened SQL. sqlite reports `IN(...) element has 1 term - expected 2`, mysql `Operand should contain 2 column(s)`. And the pairs are meaningful together: `classId in (1, 2) and roleId in (1, 2)` is a wider, different condition.

## Cause

`QueryBuilderHelper` already renders row values, but gates it on `fields.length > 1`. A raw key is opaque, so `fields` is just `[key]` and the payload falls through to the scalar branch, which binds each tuple as a single param; `Platform.formatQuery` then flattens each array into `-1, -1`.

## Fix

The arity of a raw key can't be inferred from the key, so the row-value form is detected from the payload instead — an `$in`/`$nin` list where every item is an array — and reuses the existing tuple rendering.

The output is the portable `((-1, -1), (1, 1), (2, 2))` form rather than knex's `values` spelling. That is what v7 already emits for composite non-raw keys, and unlike `values` it also works on MySQL before 8.0.19.

Deliberately unchanged:

- A flat array on a raw key keeps its scalar meaning, so `[[1, 1]]` is how to ask for a single tuple.
- `$in: []` still renders `1 = 0`.
- MSSQL and Oracle both report `allowsComparingTuples() === false`, so a *composite key* is decomposed into `(a = ? and b = ?) or ...` there. A `raw()` fragment can't be decomposed -- the column names aren't recoverable from it -- so it gets the row-value form on every platform. Oracle accepts that and raw tuple keys work; SQL Server has no row value constructor and rejects it, exactly as it did in v6, where knex rendered the same `(a, b) in ((1, 1), (2, 2))` shape for a raw key on every dialect. So no regression on either, and Oracle gains a working case.

The first commit is the failing test, the second is the fix. Coverage is sqlite plus the two platforms that report `allowsComparingTuples() === false`; seven of the eleven cases fail on the first commit.

Reproduction: https://github.com/lamkashingpaul/reproduction/tree/raw-tuple-in

### DIFF
--- a/packages/sql/src/query/QueryBuilderHelper.ts
+++ b/packages/sql/src/query/QueryBuilderHelper.ts
@@ -782,6 +782,13 @@ export class QueryBuilderHelper {
     const replacement = this.getOperatorReplacement(op, value);
     const rawField = Raw.isKnownFragmentSymbol(key);
     const fields = rawField ? [key as unknown as string] : Utils.splitPrimaryKeys(key);
+    // a raw key can hold a column tuple like `(a, b)`, but its arity is opaque, so we detect the
+    // row-value form from the payload instead: an `$in`/`$nin` list where every item is an array
+    const rowValues =
+      rawField &&
+      ['$in', '$nin'].includes(op) &&
+      Array.isArray(value[op]) &&
+      value[op].every((v: unknown) => Array.isArray(v));
 
     if (fields.length > 1 && Array.isArray(value[op])) {
       const singleTuple = !value[op].every((v: unknown) => Array.isArray(v));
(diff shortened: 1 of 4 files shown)


B4nan (approved):
thanks!

PR #8237, opened 2026-09-03
fix(postgres): forward `pool.min` to the `pg` pool

### LINKED ISSUE #8234: `pool.min` is silently ignored by the PostgreSQL driver
### Describe the bug

`PoolConfig.min` is still part of the public `Options` type and documented as "Minimum number of connections to keep in the pool", but `PostgreSqlConnection.mapOptions()` only forwards `pool.max` and `pool.idleTimeoutMillis` to `pg.Pool`. `pool.min` is never read anywhere in `@mikro-orm/*`, so the created `pg.Pool` always runs with `min = 0`, and every idle connection is closed after pg-pool's default `idleTimeoutMillis` of 10 s, regardless of what `pool.min` says.

`pg-pool` itself supports `min` (3.14.0 in my install), so the value could simply be forwarded the same way `max` is.

Relevant code: `packages/postgresql/src/PostgreSqlConnection.ts`, `mapOptions()`:

```ts
Utils.defaultValue(ret, 'max', pool?.max);
Utils.defaultValue(ret, 'idleTimeoutMillis', pool?.idleTimeoutMillis);
// no mapping for pool?.min
```

### Reproduction

https://github.com/takaram/mikro-orm-reproduction/tree/postgres-pool-min

@mikro-orm/postgresql

### MikroORM version

7.1.15-dev.9

### Node.js version

v24.20.0

### Operating system

Ubuntu

### Validations

lazerg:
`PostgreSqlConnection.mapOptions()` forwards `pool.max` and `pool.idleTimeoutMillis` to `pg`, but not `pool.min`, so the pool always ran with `min = 0` and closed every idle connection once pg-pool's idle timeout elapsed, no matter what `pool.min` was set to.

`pg-pool` supports `min`, so this maps it the same way `max` is already mapped. The other drivers map it to their own pool option already.

Closes #8234

### DIFF
--- a/packages/postgresql/src/PostgreSqlConnection.ts
+++ b/packages/postgresql/src/PostgreSqlConnection.ts
@@ -43,6 +43,7 @@ export class PostgreSqlConnection extends AbstractSqlConnection {
   mapOptions(overrides: PoolConfig): PoolConfig {
     const ret = { ...this.getConnectionOptions() } as PoolConfig;
     const pool = this.config.get('pool');
+    Utils.defaultValue(ret, 'min', pool?.min);
     Utils.defaultValue(ret, 'max', pool?.max);
     Utils.defaultValue(ret, 'idleTimeoutMillis', pool?.idleTimeoutMillis);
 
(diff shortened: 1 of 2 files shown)


B4nan (approved):
thanks!

PR #8239, opened 2026-09-03
feat(core): add `through` option for read-only to-one relations resolved via subquery

### LINKED ISSUE #8238: Add support for HasOneThrough / HasManyThrough relationships

We use a shared pivot entity to model typed graph-like relationships between entities. Please do not that I gave a "dummy" example, not the actual one from our proprietary code:

```ts
@Entity()
class Edge {
  @PrimaryKey()
  key!: string;

  @ManyToOne(() => Customer)
  start!: Customer;

  @ManyToOne(() => Transaction)
  end!: Account;

  @Property()
  state!: string;
}
```

We want to expose filtered relationships through this entity, such as an `Customer.account` relation where the matching edge has a specific `type` and `state`.

`ManyToMany` with a filtered pivot can approximate to-many relationships, but there is no native way to expose a scalar to-one relation through a pivot. Custom loaders also lose native `populate`, query, typing, serialization, and UnitOfWork behavior.

Native through-relation decorators or equivalent schema APIs:

```ts
@HasOneThrough({
  entity: () => Account,
  through: () => Edge,
  ownerProperty: 'start',
  targetProperty: 'end',
  where: {
    state: 'cleared',
  },
})
account!: Ref<Account>;
```

A corresponding `HasManyThrough` relation would expose `Collection<T>`.

Ideally these relations would support:

- `populate`
- joined/select-in loading strategies
- nested query criteria and ordering
- `Loaded<>` type inference
- serialization
- polymorphic and inherited/TPT targets
- configurable read-only or writable behavior
- cardinality validation for `HasOneThrough`

For writable relations, the API would need a way to initialize additional pivot properties or delegate pivot creation to a factory.

- Adding direct foreign keys, which duplicates relationship state already stored in the pivot.
- Using a hidden `ManyToMany` collection plus a scalar getter.
- Writing custom DataLoaders or repository methods.
- Creating application-level decorators that only store metadata.
- Forking MikroORM to add another relation kind.

These approaches either duplicate persistence state or lose native ORM behavior.

**Additional context**

We are using MikroORM 7.1.x with PostgreSQL and TPT inheritance.

A read-only `HasOneThrough` implementation would already solve the main use case. It could potentially reuse the existing many-to-many join planner while hydrating the result as a scalar reference and rejecting multiple matching rows.

B4nan:
Adds a `through` option to `ManyToOne` and `OneToOne` relations. The relation is resolved via a correlated subquery on another entity instead of a FK column. This covers a pivot entity carrying the link with extra columns to filter on, and also picking a single item out of a to-many relation (e.g. the latest one) by pointing `through` at the target entity itself.

```ts
@ManyToOne(() => Account, { through: () => Edge, where: { state: 'cleared' }, ref: true, nullable: true })
account?: Ref<Account>;

@OneToOne(() => Edge, { through: () => Edge, orderBy: { id: 'desc' }, ref: true, nullable: true })
latestEdge?: Ref<Edge>;
```

`where` and `orderBy` apply to the `through` entity and the first matching row is used. Under the hood this is the existing non-persistent formula relation, so populate with both strategies, `Loaded<>` inference, serialization, `mapToPk` and conditions on the relation work without new code paths. The SQL platform generates the subquery with the query builder, so values go through custom types and platform quoting. The relation is read-only, writes go through the pivot entity.

Discovery checks that the `through` entity has M:1 properties to both sides and rejects composite PK targets. Conditions that would need a join inside the subquery are rejected at query time, because a join alias could shadow the outer owner alias. The new docs section also covers the manual `formula` approach for cases the option does not handle.

Closes #8238

### DIFF
--- a/docs/docs/relationships.md
+++ b/docs/docs/relationships.md
@@ -625,7 +625,7 @@ const CustomerSchema = defineEntity({
     edges: () => p.oneToMany(Edge).mappedBy('start'),
     // the account linked via an edge in the `cleared` state
     account: () => p.manyToOne(Account).through(() => Edge).where({ state: 'cleared' }).ref().nullable(),
-    // the most recent edge of this customer
+    // the newest edge of this customer, picked out of the `edges` collection
     latestEdge: () => p.oneToOne(Edge).through(() => Edge).orderBy({ id: 'desc' }).ref().nullable(),
   },
 });
@@ -661,7 +661,7 @@ export const Customer = defineEntity({
     edges: () => p.oneToMany(Edge).mappedBy('start'),
     // the account linked via an edge in the `cleared` state
     account: () => p.manyToOne(Account).through(() => Edge).where({ state: 'cleared' }).ref().nullable(),
-    // the most recent edge of this customer
+    // the newest edge of this customer, picked out of the `edges` collection
     latestEdge: () => p.oneToOne(Edge).through(() => Edge).orderBy({ id: 'desc' }).ref().nullable(),
   },
 });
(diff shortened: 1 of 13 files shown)


vkouk:
@B4nan Can you elaborate in which cases this is useful / needed to use?

B4nan:
When you want a to-one relation but you don't have a foreign key.

PR #8241, opened 2026-09-04
docs: updated schema-first-guide.md reference to the hash option to the latest version name

boenrobot:
Because this is a change in argon2, not in MikroORM, and the guide never makes mention of a version,
all versions of the docs are updated to assume latest argon2.


B4nan (approved):
Thanks

PR #8253, opened 2026-09-05
feat(cli): add Nub TypeScript loader option

colinhacks:
Adds `@nubjs/loader` as a CLI TypeScript loader.

It runs last in the automatic selection, and only when the project `tsconfig.json` is used, since Nub discovers the tsconfig on its own and cannot be pointed at a custom path. Selecting it explicitly together with a custom tsconfig path throws.

### DIFF
--- a/docs/docs/upgrading-v6-to-v7.md
+++ b/docs/docs/upgrading-v6-to-v7.md
@@ -136,11 +136,11 @@ TypeScript support was previously provided by `ts-node`. In v7, the CLI supports
 - `oxc` via `@oxc-node/core`, supports metadata reflection
 - `swc` via `@swc-node/register`, supports metadata reflection
 - `tsx`
-- `nub` via `@nubjs/loader`, selected explicitly and using the project `tsconfig.json` (available from v7.2)
+- `nub` via `@nubjs/loader`, supports metadata reflection (available from v7.2)
 - `jiti`
 - `tsimp`
 
-The default is `auto`, which means it goes through all those options sequentially and picks the first one available in the project dependencies.
+The default is `auto`, which means it goes through all those options sequentially and picks the first one available in the project dependencies. `nub` is skipped there, it is always opt-in.
 
 To pick a loader explicitly, use the `tsLoader` setting in your `package.json`:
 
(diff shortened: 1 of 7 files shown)


B4nan (changes requested):
thanks! just in time for the next feature release, please retarget this to `next` branch and drop the changes in the versioned docs, i will be shipping v7.2 on monday including a new docs snapshot, the old ones should stay intact since older versions dont support nub

B4nan:
> The loader is not considered by automatic selection

What's the reason for omitting it there?

colinhacks:
@nubjs/loader is opt-in so installing it for another tool does not change the CLI’s existing automatic loader selection.

B4nan:
But why not? That's my question. How is it different from the other options that auto-register?

docs/docs/upgrading-v6-to-v7.md
```diff
@@ -136,6 +136,7 @@ TypeScript support was previously provided by `ts-node`. In v7, the CLI supports
 - `oxc` via `@oxc-node/core`, supports metadata reflection
 - `swc` via `@swc-node/register`, supports metadata reflection
 - `tsx`
+- `nub` via `@nubjs/loader`, selected explicitly and using the project `tsconfig.json`
```
B4nan:
```suggestion
- `nub` via `@nubjs/loader`, selected explicitly and using the project `tsconfig.json` (available from v7.2)
```

B4nan:
Thanks! Pushed a few tweaks on top (2bd3b9fb8):

- The `detects a preloaded Nub loader` test was passing no matter what. `detectTypeScriptSupport()` short-circuits on `process.env.VITEST` a few lines before the `execArgv` scan, so the new `@nubjs/loader` check had no coverage at all. It now stubs `VITEST`/`TS_JEST`/`MIKRO_ORM_CLI_ALWAYS_ALLOW_TS` empty and asserts a negative case too.
- Dropped the `normalizePath()` wrapping in the tsconfig guard, `absolutePath()` already normalizes on every return path.
- Docs: nub does support metadata reflection (it maps `experimentalDecorators` to oxc's legacy mode together with `emitDecoratorMetadata`), which is usually what people pick a loader on, so the bullet says so now. Also noted that it resolves the nearest tsconfig per importing directory rather than "the project tsconfig.json", that TC39 decorators are not supported yet, and that `auto` skips it.
- Small comment on the `continue` in the loader loop plus a JSDoc note, so the opt-in-only behavior is visible from the source.

docs/docs/upgrading-v6-to-v7.md
```diff
@@ -152,7 +152,7 @@ To pick a loader explicitly, use the `tsLoader` setting in your `package.json`:
 
 Or override it via `MIKRO_ORM_CLI_TS_LOADER` env var.
 
-To use Nub, install `@nubjs/loader` and select it explicitly:
+To select Nub explicitly, install `@nubjs/loader`.
```
B4nan: all TS loaders should be installed explicitly, I'd drop this note

colinhacks:
There is no fundamental registration difference. It now runs after every existing automatic loader and is skipped only when `tsConfigPath` is non-default, because the loader cannot accept that setting.

colinhacks:
The updated branch is waiting for approval of the tests workflow. Could that run be approved?

colinhacks:
The naming hold is lifted. Continuing with the published `@nubjs/loader` package.

B4nan (approved):

PR #8259, opened 2026-09-07
docs: fix wrong default propagation claim for em.transactional in upgrade guide

### LINKED ISSUE #8258: Upgrade guide says em.transactional defaults to REQUIRES_NEW, Transactions page says NESTED
Two docs pages disagree with each other on the default propagation for `em.transactional`, and I think one of them is just wrong.

Transactions page (`transactions.md#transaction-propagation`):

> By default, MikroORM uses `NESTED` propagation for explicit transactions via `em.transactional`. When you call `em.transactional()` within another transaction, it automatically creates a savepoint (nested transaction) rather than joining the existing transaction.

Upgrade guide (`upgrading-v6-to-v7.md`, "Default propagation in `@Transactional` is `REQUIRED`"):

> The previous default was `REQUIRES_NEW`, which always started a new transaction. `REQUIRES_NEW` remains the default for the `em.transactional` method.

Before opening a PR I went and checked which one matches reality. `em.transactional()` calls `TransactionManager.handle()`, which does `options.propagation ??= TransactionPropagation.NESTED` (`packages/core/src/utils/TransactionManager.ts:23`) — that's the only code path, there's no branch where it falls back to `REQUIRES_NEW`. The JSDoc on `EntityManager.transactional()` says the same thing in plain English ("a new savepoint (nested transaction) will be created by default"), and `tests/features/transactions/transaction-propagation.sqlite.test.ts` has a test specifically checking that nesting produces a `SAVEPOINT`. So the Transactions page is right and the upgrade guide sentence is wrong.

I also checked whether the upgrade guide used to be correct and just went stale. It didn't, as far as I can tell — the `options.propagation ??=` line hasn't changed since it was introduced in #6802, and it was already `NESTED` there. The wrong sentence got added later, in the same commit that changed the *decorator's* default to `REQUIRED`, but that commit didn't touch `em.transactional`'s default at all (only a type signature widening). So it looks like the sentence was wrong from the moment it was written, probably a mix-up between "the decorator used to default to REQUIRES_NEW" and "em.transactional still defaults to REQUIRES_NEW."

Same sentence shows up in both `docs/docs/upgrading-v6-to-v7.md` and the `version-7.2` snapshot, word for word.

Proposed fix, matching how the Transactions page already phrases it:

> `NESTED` remains the default for the `em.transactional` method, which creates a savepoint when a transaction is already active.

Happy to send a PR for this — just didn't want to guess which of your own two pages was the intended source of truth before submitting the diff.

aahmedsamy:
Closes #8258

The upgrade guide's "Default propagation in `@Transactional` is `REQUIRED`" section says:

> `REQUIRES_NEW` remains the default for the `em.transactional` method.

That's wrong. The Transactions page says `NESTED` is the default for `em.transactional`, and the code backs that up:

- `em.transactional()` calls `TransactionManager.handle()`, which does `options.propagation ??= TransactionPropagation.NESTED` (`packages/core/src/utils/TransactionManager.ts:23`) — there's no other path.
- `EntityManager.transactional()`'s JSDoc already says a savepoint is created by default when nesting.
- `transaction-propagation.sqlite.test.ts` has a test asserting that nesting produces a `SAVEPOINT`.

Same wrong sentence appears in both `docs/docs/upgrading-v6-to-v7.md` and the `version-7.2` snapshot, so both are fixed here.

New wording, matching how the Transactions page already phrases it:

> `NESTED` remains the default for the `em.transactional` method, which creates a savepoint when a transaction is already active.

### DIFF
--- a/docs/docs/upgrading-v6-to-v7.md
+++ b/docs/docs/upgrading-v6-to-v7.md
@@ -434,7 +434,7 @@ The mechanism for processing serialized primary keys in MongoDB driver has chang
 
 ## Default propagation in `@Transactional` is `REQUIRED`
 
-The default [propagation mode](./transactions.md#transaction-propagation) of the `@Transactional` decorator is now `REQUIRED`, which means that if there is an ongoing transaction, the decorated method will join it; otherwise, a new transaction will be started. The previous default was `REQUIRES_NEW`, which always started a new transaction. `REQUIRES_NEW` remains the default for the `em.transactional` method.
+The default [propagation mode](./transactions.md#transaction-propagation) of the `@Transactional` decorator is now `REQUIRED`, which means that if there is an ongoing transaction, the decorated method will join it; otherwise, a new transaction will be started. The previous default was `REQUIRES_NEW`, which always started a new transaction. `NESTED` remains the default for the `em.transactional` method, which creates a savepoint when a transaction is already active.
 
 ## `dataloader` dependency
 
--- a/docs/versioned_docs/version-7.0/upgrading-v6-to-v7.md
+++ b/docs/versioned_docs/version-7.0/upgrading-v6-to-v7.md
@@ -452,7 +452,7 @@ The mechanism for processing serialized primary keys in MongoDB driver has chang
 
 ## Default propagation in `@Transactional` is `REQUIRED`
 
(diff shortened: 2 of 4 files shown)


B4nan (approved):
thanks!

PR #8277, opened 2026-09-15
fix(sql): respect property returning hints on inserts

nikzanda:
## Summary

Properties explicitly marked `returning: true` can be omitted from INSERT results, leaving managed entities out of sync with their persisted values after `flush()`.

For example, given `name: p.string().trim().returning()`, inserting `' Example '` stores `'Example'` in the database but leaves `entity.name` as `' Example '`. With this fix, the INSERT returns the normalized value and `entity.name` becomes `'Example'` before `flush()` resolves. The same applies to uppercase and lowercase normalization when the property is marked for returning.

Honor property returning hints in `nativeInsertMany()` for both `RETURNING` and SQL Server's `OUTPUT`, including properties whose values are supplied in the insert payload. Apply the same rule to QueryBuilder INSERTs. The Unit of Work hydrates the returned values through its existing mapping, without an additional SELECT.

This fixes the existing property-level `returning` option for INSERTs. Properties without that option retain their current normalization behavior, and explicit QueryBuilder `.returning(...)` lists still take precedence.

Related discussion: [Property returning hints during flush (#7269)](https://github.com/mikro-orm/mikro-orm/discussions/7269).


B4nan (approved):
thanks!

PR #8281, opened 2026-09-16
fix(query-builder): do not inline raw query parameters

fortizflexidao:
Catched a bug where:

```typescript
.where(raw('field = ?', ['some?value']))
```

triggered an error in knex. The issue is that knex would keep expecting and additional parameter because the provided argument has a question  mark. The fix is to pass the `raw` arguments untouched to knex.


B4nan:
Please check if this is valid on master, if so, PR master first.

fortizflexidao:
The bug is not reproducible in the 7.x branch. Do you want me to submit a pull request with the test to prevent potential regression?

B4nan:
Thanks for checking it. Yeah, let's do that, more tests won't hurt :]

B4nan (approved):

fortizflexidao:
Here is there PR. Regards! https://github.com/mikro-orm/mikro-orm/pull/8286

PR #8289, opened 2026-09-17
fix(core): escape the remaining property names in generated code

Dev-next-gen:
Following #8221 I went looking for the positions it did not cover, and there are a few more places where the compiled hydrator and comparator interpolate a property name as raw code.

In `ObjectHydrator`:

- `key: '<targetKey>'` in the `createReference()` options
- the child names of an inline embeddable, used as object literal keys
- the `value?.['<inverse>']` probe in the collection item mapper
- the 1:1 back-reference assignment, which went through `safeKey()`. That folds `\W` to `_` instead of escaping, so for an inverse side named `owner'quote` the hydrator wrote the back-reference to `owner_quote` and left the declared property undefined — no syntax error, just a silently wrong object. It is the only place `safeKey()` was used for a real property access rather than for a generated variable name, so it now goes through `wrap()` like every other access.

In `EntityComparator`:

- the composite primary key object literal keys and the matching `cond.<pk>` reads, in both `getPkGetter()` and `getPkGetterConverted()`
- `val?.<targetKey>` in the snapshot generator

Every one of these renders identically for a plain identifier — `objectKey('id')` is `id`, `wrap('id')` is `.id`, `quote('id')` is `'id'` — so the generated code is byte-identical for names that already worked, and no snapshot moved. `objectKey()` is just the inline identifier check that was already sitting at the embeddable site, lifted next to `quote()` in both classes so the pk getters can use it too; both helpers stay private.

I added four cases to `tests/features/hydration/special-property-names.test.ts`. Before the change three of them fail with a `SyntaxError` coming out of `new Function()` — from `getEntityHydrator`, `getPkGetter` and `getSnapshotGenerator` respectively — and the 1:1 one fails on the mangled key.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.


B4nan (approved):

B4nan:
thanks!

PR #8291, opened 2026-09-21
fix(core): do not mutate the where condition and data passed to the EntityManager

### LINKED ISSUE #8290: em.find() / em.findOne() mutate the caller's where object (entity values are replaced by their primary keys)
### Describe the bug

When the `where` condition passed to `em.find()` / `em.findOne()` contains an entity instance as a value
(e.g. `{ author }`), MikroORM rewrites that property **on the caller's object** to the entity's primary key.
After the call, `where.author` is `1` instead of the `Author` instance that was passed in.

`where` is an input, so this breaks any code that reuses the same object after the query — for example a
find-or-create helper that passes `params` to `findOne()` and then reads `params.author` to create the row.

On 7.1.14 entities inside operator arrays (`{ author: { $in: [author] } }`) were left alone, which made this
easy to miss; on `next` (7.2.0-dev.8) the array elements are replaced as well.

### Reproduction

https://github.com/daemyung/reproduction/tree/where-mutation

```ts
const author = await em.findOneOrFail(Author, { name: 'Jon' });
const where = { author };

await em.find(Book, where);

expect(where.author).toBe(author); // AssertionError: expected 1 to be Author { id: 1, name: 'Jon' }
```

### Expected behavior

The normalized condition is built into a new object and the caller's where is left untouched.

@mikro-orm/postgresql

### MikroORM version

7.1.14, 7.2.0-dev.8

### Node.js version

24

### Operating system

_No response_

### Validations

lazerg:
`QueryHelper.processWhere()` normalizes the condition in place: entity values become primary keys, PK-only objects get inlined, group operators get lifted and, with `ignoreUndefinedInQuery`, undefined keys get dropped. `em.find()`/`em.findOne()` hand their `where` argument to this code untouched, so after `em.find(Book, where)` with `where = { author }`, `where.author` was `1` instead of the `Author` instance. That breaks any code that reuses the object once the query resolves (e.g. a find-or-create helper).

`processWhere()` now copies the plain object and array structure of the condition first, keeping leaf values such as entities by reference, and normalizes the copy.

The data passed to the write methods had the same problem, since `QueryHelper.processObjectParams()` replaced entities with primary keys in the caller's object. It builds a new object now, which covers `em.create()`, `em.insert()`, `em.insertMany()`, `em.nativeUpdate()`, `em.upsert()`, `em.upsertMany()` and the query builder. `em.upsertMany()` also stopped replacing the rows in the caller's array.

Fixes #8290


B4nan (approved):
thanks!

PR #8293, opened 2026-09-21
fix(sql): resolve inline embeddable columns that shadow the property name

### LINKED ISSUE #8292: Joined populate throws when embedded property has the same name as its column
### Describe the bug

After upgrading from v6 to v7, populating a relation with the joined strategy started throwing for one of our entities.

The entity has an embeddable with a single property, and that property's column has the same name as the embedded property itself (`User.password` -> `Password.value` stored in the `password` column, `prefix: ''`). Loading `User` directly works, `select-in` works, only the joined populate fails:

```
DriverException: Cannot use 'as u1__password' alias on embedded property 'u1.password' because it expands to multiple columns. Alias individual fields instead (e.g. 'u1.password.propertyName as u1__password').
```

The error message was confusing to me since the embeddable only has one column and I'm not aliasing anything myself, the alias comes from the populate. My guess is `u1.password` gets resolved to the embedded property instead of the `password` column.

v6 had a problem with the same setup too, just a different one: the joined load works, but if you then change another entity and flush, it tries to update `user` with nothing in it:

```
Empty .update() call detected! Update data does not contain any values to update. This will result in a faulty query. Table: `user`. Columns: .
```

Again only after a joined load, select-in is fine.

Renaming the embedded property (we went with `hashedPassword`) makes both go away, so we're fine for now, but it took a while to figure out what was going on.

Might be related to #8271 / #8272, but it still fails on 7.2.1 and on the current `next`, which should already have that fix.

### Reproduction

https://github.com/JH8459/reproduction

`src/example.test.ts` has 4 tests. direct load and select-in pass, joined populate and the flush one fail.

- 7.2.1-dev.24 (next): joined populate + flush fail
- 7.2.1, 7.2.0: same
- 6.6.16: joined populate passes, flush fails with the empty update

I also tried it on mariadb 11.4, same results.

@mikro-orm/sqlite (also reproduced with @mikro-orm/mariadb)

### MikroORM version

7.2.1-dev.24

### Node.js version

v24.11.0

### Operating system

macOS

### Validations

lazerg:
A joined populate selects each property of the joined table as `alias.prop as alias__col`. When an inline embeddable has a column named after the embedded property itself (`prefix: ''` plus a matching `fieldName`), that select resolves to the embedded property rather than to the column, and the query builder rejected the alias because an embeddable "expands to multiple columns".

An aliased select whose name matches a column of the entity now resolves to that column, whether the column belongs to the embeddable, to a nested one, or to an unrelated scalar property. Without an alias the name still means the embedded property and expands to all of its columns.

This also fixes two smaller problems in the same branch. A `persist: false` child of an inline embeddable crashed the expansion. The expanded columns were never qualified with the alias they were requested on, so selecting an embeddable through a join alias read it from the main table.

Fixes #8292


B4nan (approved):
thanks!

PR #8299, opened 2026-09-21
fix(sql): respect property returning hints on updates and upserts

nikzanda:
Properties marked `returning: true` stay stale on managed entities after `flush()` updates them. Assigning `' Customer '` to a trimmed property persists `'Customer'`, but the entity keeps the spaces. #8277 fixed this for inserts. This does the same for single and batched updates, QueryBuilder updates and upserts.

- `returning` is normalized once in metadata discovery to physical, hydratable columns. Formulas, inverse sides and props with `persist: false` or `hydrate: false` lose the hint there, so every existing `prop.returning` check keeps working as is.
- Single updates add the marked props and raw values to the explicit returning list. Batched updates share one returning list between the base driver, Oracle and MSSQL, with per-column `convertToJSValueSQL` for custom types and composite relations.
- MSSQL batch updates emit `OUTPUT`, with `OUTPUT INTO` on tables that have triggers. MSSQL upserts skip the redundant reload SELECT.
- The QueryBuilder update path goes through `helper.mapper()` like an explicit `.returning()` list already does. TPT tables only return their own columns and mixed STI batches resolve child columns through the root.
- Upserts include the marked props in the returning list even when the payload supplies them. Merge, exclude and ignore keep their current semantics.

MySQL and MariaDB are unchanged. Native and QueryBuilder writes still do not synchronize managed entities.

Related discussion: [Property returning hints during flush (#7269)](https://github.com/mikro-orm/mikro-orm/discussions/7269).


B4nan (approved):
thanks!

PR #8300, opened 2026-09-22
fix(sql): respect property returning hints on updates and upserts

B4nan:
Properties marked `returning: true` stay stale on managed entities after `flush()` updates them. Assigning `' Customer '` to a trimmed property persists `'Customer'`, but the entity keeps the spaces. #8277 fixed this for inserts. This does the same for single and batched updates, QueryBuilder updates and upserts.

Alternative to #8299 with the same tests and a smaller source change. The tests are copied verbatim, except one regex that only passed without colored query logs.

- `returning` is normalized once in metadata discovery to physical, hydratable columns. The existing `prop.returning` checks in the insert paths stay as they are and there is no new `isReturningProperty()` helper to call at every site.
- The QueryBuilder update path goes through `helper.mapper()`, the same way an explicit `.returning()` list already does, so the hand-rolled custom type conversion is gone.
- The base driver, the Oracle driver and the new MSSQL trigger override share the batch update returning list via `getUpdateReturningProperties()`, which replaces the copy the Oracle driver kept.

Behaviour is the same as in #8299. MSSQL batch updates now emit `OUTPUT`, with `OUTPUT INTO` on tables that have triggers, and MSSQL upserts skip the redundant reload SELECT. Composite relations return every column with per-column SQL conversions, TPT tables only return their own columns, and mixed STI batches resolve child columns through the root. MySQL and MariaDB are unchanged.


B4nan:
Folded into #8299 instead.

